PR #1399, opened 2020-08-05
feat(ci): Upload test coverage to codecov

nojnhuh:


.github/workflows/main.yml
```diff
@@ -161,6 +161,9 @@ jobs:
           go-version: 1.14
       - name: Test
         run: make go-test-coverage
+      - name: Upload Coverage
+        if: ${{ success() }}
+        run: bash <(curl -s https://codecov.io/bash)
```
bridgetkromhout, line 166: curlbash, fun :)

michelleN:
@nojnhuh what does that line do?

nojnhuh:
This is how they describe it in their docs: https://docs.codecov.io/docs/about-the-codecov-bash-uploader

Looks like there may be a native GitHub Actions solution. I'll look into that.

nojnhuh:
Looks like their GitHub action does basically the same thing. https://github.com/codecov/codecov-action/blob/master/index.js#L127

That looks easier to customize though, so we can start using that once we need it. By itself it doesn't look like we gain anything over the plain script.

michelleN:
ah so we've already exported the CODECOV token?
`export CODECOV_TOKEN="token"`

nojnhuh:
That's only required for private repos.

PR #1425, opened 2020-08-06
docs(patterns/certificates): update relative paths for wiki links

ross-mcdermott:
Update relative paths for wiki links


michelleN:
Thank you @ross-mcdermott !

PR #1430, opened 2020-08-06
scripts/demo: remove dependency on GNU parallel

shashankram:
Run the commands in background and wait on them instead.

### DIFF
--- a/demo/reset-counters.sh
+++ b/demo/reset-counters.sh
@@ -7,9 +7,10 @@ source .env
 
 # This script assumes that port forwarding has already been established.
 # See ./scripts/port-forward-all.sh to enable fort forwarding for Bookstore demo.
-# Requires GNU Parallel
-parallel curl -I -X GET ::: \
-         http://localhost:8080/reset \
-         http://localhost:8081/reset \
-         http://localhost:8082/reset \
-         http://localhost:8083/reset
+curl -I -X GET http://localhost:8080/reset &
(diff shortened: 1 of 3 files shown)


nojnhuh:
Can you update the docs to remove that as a pre-req? Or is that being handled in a separate PR?

demo/reset-counters.sh
```diff
@@ -7,9 +7,10 @@ source .env
 
 # This script assumes that port forwarding has already been established.
 # See ./scripts/port-forward-all.sh to enable fort forwarding for Bookstore demo.
-# Requires GNU Parallel
-parallel curl -I -X GET ::: \
-         http://localhost:8080/reset \
-         http://localhost:8081/reset \
-         http://localhost:8082/reset \
-         http://localhost:8083/reset
+curl -I -X GET http://localhost:8080/reset &
```
eduser25, line 10: Should we add some arbitrary timeout (`--connect-timeout`)? Can get stuck for some time otherwise (order of minutes from what I'm [reading](https://unix.stackexchange.com/questions/94604/does-curl-have-a-timeout))

shashankram: Would make it a separate change if the problem exists prior to this change.

shashankram:
> Can you update the docs to remove that as a pre-req? Or is that being handled in a separate PR?

done, thanks.

PR #1445, opened 2020-08-07
demo: cleanup stale namespace before recreating

shashankram:
Rerunning the demo in specific environments results in an error because the existing
namespace is not deleted prior to recreation:
Error from server (AlreadyExists): namespaces "osm-system" already exists


eduser25:
Merging on behalf

shashankram:
> Merging on behalf

Thanks Edu! (I asked Edu to merge on my behalf)

PR #1456, opened 2020-08-07
Log error in catalog/debugger if listing namespaces fails

elenaspasovaspasova:
Logging an error as was suggested in review of #1311 

- [ ] New Functionality
- [ ] Documentation
- [ ] Install
- [ ] Control Plane
- [ ] CLI Tool
- [ ] Certificate Management
- [ ] Networking
- [ ] Metrics
- [ ] SMI Policy
- [ ] Security
- [ ] Tests / CI System
- [X] Other

### DIFF
--- a/pkg/catalog/debugger.go
+++ b/pkg/catalog/debugger.go
@@ -81,6 +81,7 @@ func (mc *MeshCatalog) ListMonitoredNamespaces() []string {
 	namespaces, err := mc.namespaceController.ListMonitoredNamespaces()
 
 	if err != nil {
+		log.Error().Err(errListingNamespaces).Msg("Failed to list monitored namespaces")
 		return nil
 	}
 
--- a/pkg/catalog/errors.go
+++ b/pkg/catalog/errors.go
@@ -13,4 +13,6 @@ var (
 	errDomainNotFoundForService              = errors.New("no host/domain found to configure service")
 	errNoTrafficSpecFoundForTrafficPolicy    = errors.New("no traffic spec found for the traffic policy")
 	errServiceNotFound                       = errors.New("service not found")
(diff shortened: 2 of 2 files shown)


pkg/catalog/errors.go
```diff
@@ -13,4 +13,6 @@ var (
 	errDomainNotFoundForService              = errors.New("no host/domain found to configure service")
 	errNoTrafficSpecFoundForTrafficPolicy    = errors.New("no traffic spec found for the traffic policy")
 	errServiceNotFound                       = errors.New("service not found")
+	errListingNamespaces                     = errors.New("Failed to list monitored namespaces")
```
shashankram, line 16:
```suggestion
	errListingNamespaces                     = errors.New("error listing namespaces)
```

pkg/catalog/debugger.go
```diff
@@ -81,6 +81,7 @@ func (mc *MeshCatalog) ListMonitoredNamespaces() []string {
 	namespaces, err := mc.namespaceController.ListMonitoredNamespaces()
 
 	if err != nil {
+		log.Error().Err(errListingNamespaces).Msg("Failed to list monitored namespaces")
```
shashankram, line 84: If both the error object's message (`.Err()`) and the log message (`.Msg()`) are the same, it seems unnecessary, and the error log would duplicate this error message which isn't useful.
If the `namespaceController.ListMonitoredNamespaces()` is already logging the error, this log is unnecessary.

elenaspasovaspasova: Good call out, thank you @shashankram! Closing this issue in favor of the error already being logged in `namespaceController.ListMonitoredNamespaces()`

PR #1457, opened 2020-08-07
demo: disable egress by default

shashankram:
Since the install command has egress disabled for default
for the purpose of simplifying manual demos, disable egress
requests in the demo apps as well. An `ENABLE_EGRESS` env
variable is used to turn on/off egress in the demo and
in the client demo apps.

- [ ] New Functionality
- [ ] Documentation
- [ ] Install
- [ ] Control Plane
- [ ] CLI Tool
- [ ] Certificate Management
- [ ] Networking
- [ ] Metrics
- [ ] SMI Policy
- [ ] Security
- [X] Tests / CI System
- [ ] Other

- Does this change contain code from or inspired by another project? `No`


demo/cmd/common/books.go
```diff
@@ -102,10 +103,13 @@ func GetBooks(participantName string, meshExpectedResponseCode int, egressExpect
...
-		buyBook:        false,
-		httpEgressURL:  false,
-		httpsEgressURL: false,
+		booksBought: false,
+		buyBook:     false,
+	}
+
+	if enableEgress {
+		urlSuccessMap[httpEgressURL] = false
+		urlSuccessMap[httpsEgressURL] = false
```
eduser25, line 112: not going to request a change, but since you parametrized the urlSuccessMap I'd have said to parametrized urlExpectedRespCode too

shashankram: It shouldn't matter since response code checks for egress only apply if the egress URLs aren't going to be used. I found it cleaner to avoid if/else unless necessary. The purpose of the change is not to wrap every egress related code in conditionals, but rather avoid egress requests unless enabled. WDYT?

eduser25: i'm fine with that

PR #1463, opened 2020-08-08
Removed redundant namespace creation from manual demo script

NillsF:

- [ ] New Functionality
- [X] Documentation
- [ ] Install
- [ ] Control Plane
- [ ] CLI Tool
- [ ] Certificate Management
- [ ] Networking
- [ ] Metrics
- [ ] SMI Policy
- [ ] Security
- [ ] Tests / CI System
- [X] Other

**No**

## Impact of this PR
Remove the namespace objects from the following files:

- docs/example/manifests/bookbuyer.yaml
- docs/example/manifests/bookstore-v1.yaml
- docs/example/manifests/bookthief.yaml
- docs/example/manifests/bookwarehouse.yaml

## Issue this solves
When running through the manual demo script, it appears that namespaces are created twice. 

In the current version of the demo script, namespaces are first created using:
```bash
for i in bookstore bookbuyer bookthief bookwarehouse; do kubectl create ns $i; done
```

And are afterwards created again with the following command:
```bash
kubectl create -f docs/example/manifests/apps/
```
As the namespaces are part of the files in that directory. ([example for bookbuyer](https://github.com/openservicemesh/osm/blob/main/docs/example/manifests/apps/bookbuyer.yaml)) 
This results in a message that the resource already exists when doing the `kubectl create`.
## Impact on automated demo
I _believe_ there should be no impact on removing the namespaces from `docs/example/manifests/apps/` because the automated demo script uses inline YAML to create the applications. For example [deploy-bookwarehouse.sh](https://github.com/openservicemesh/osm/blob/main/demo/deploy-bookstore.sh)


phillipgibson:
The manifest files are needed for the automated demo, so even though they reference creating the same namespace it doesn't affect the deployment. Kubernetes will just ignore it the attempt to create a new existing namespace.

shashankram:
Maybe keeping the namespace resource in the manifests file isn't a bad idea, because it ensures the individual manifests files can be deployed independently without having to manually create the namespace. In the context of the instructions, its redundant, but the manifests alone being enough to deploy an app when the namespace hasn't been manually created seems beneficial to me to demo other scenarios.

nojnhuh:
@shashankram If the manifests were used to create the namespaces and deployments at the same time, when does the step to label the namespaces occur?

phillipgibson:
Yes @shashankram let's keep the manifest intact. I am creating a PR to update the manual demo and will take out the instructions for creating the namespaces, since this is handled by the bookstore app deploy.

shashankram:
> @shashankram If the manifests were used to create the namespaces and deployments at the same time, when does the step to label the namespaces occur?

The namespace could be labeled later as well, but with the caveat that if the namespace is labeled post application and OSM deployment, then the application pods would need to be restarted for the sidecars to be injected. Its possible for the users to label namespace whenever they want, with the condition that the application pods should be deployed post namespace labeling for automatic sidecar injection. If this cannot be satisfied, a rolling restart on the application pod can be performed.

nojnhuh:
Maybe the best compromise is not to change this at all then? That way the manifests are still self-contained and we don't have to add a step to do a rolling restart of all the pods.

shashankram:
> Maybe the best compromise is not to change this at all then? That way the manifests are still self-contained and we don't have to add a step to do a rolling restart of all the pods.

That would be my suggestion, to keep the manifests as is.

shashankram:
> > Maybe the best compromise is not to change this at all then? That way the manifests are still self-contained and we don't have to add a step to do a rolling restart of all the pods.
> 
> That would be my suggestion, to keep the manifests as is.

@NillsF are you fine with this?

NillsF:
I'm fine with keeping as is if that is best for the demo itself. 

Just thinking out load, an option to avoid the "error" message when creating the resources, we could change the current `kubectl create` to a `kubectl apply` in the instruction markdown.
Thoughts?

shashankram:
> I'm fine with keeping as is if that is best for the demo itself.
> 
> Just thinking out load, an option to avoid the "error" message when creating the resources, we could change the current `kubectl create` to a `kubectl apply` in the instruction markdown.
> Thoughts?

That seems great. Would you mind sending a PR for this?

NillsF:
Will create one later today, and will close this one.

PR #1465, opened 2020-08-08
docs: Remove colloquialism from dev doc (#1460)

phillipgibson:


README.md
```diff
@@ -16,14 +16,13 @@ The OSM project builds on the ideas and implementations of many cloud native eco
   - [Features](#features)
   - [SMI Specification Support](#smi-specification-support)
 - [OSM Design](#osm-design)
-- [Getting Started](#getting-started)
+- [Install](#install)
```
shashankram, line 19: `Getting Started` seems reasonable and easier to jump onto than `Install`. Would suggest to keep this as is, with a section on install.

README.md
```diff
@@ -62,20 +61,14 @@ OSM runs an Envoy based control plane on Kubernetes, can be configured with SMI
 
 Read more about [OSM's high level goals, design, and architecture](DESIGN.md).
 
-## Getting Started
-
-Below are quick getting started instructions. For a more detailed example usage guide and demo walkthrough, see the [OSM Example Usage Guide](/docs/example/README.md).
+## Install
```
shashankram, line 64: We should retain Getting Started section. Getting Started doesn't mean install, could also mean just reading the docs, design etc.

ksubrmnn: Agreed

phillipgibson:
Resubmitting as branch got orphaned.

PR #1474, opened 2020-08-09
WIP: Create update command for osm controlplane update

nitishm:
Signed-off-by: Nitish Malhotra <nitish.malhotra@gmail.com>

- [ ] New Functionality
- [ ] Documentation
- [ ] Install
- [ ] Control Plane
- [X] CLI Tool
- [ ] Certificate Management
- [ ] Networking
- [ ] Metrics
- [ ] SMI Policy
- [ ] Security
- [ ] Tests / CI System
- [ ] Other

No


cmd/cli/update.go
```diff
@@ -0,0 +1,142 @@
...
+	if err != nil {
+		return err
+	}
+
+	values, err := u.cfg.resolveValues()
+	if err != nil {
+		return err
+	}
+
+	// TODO (nitishm): Should we check if the controlplane deployment exists ?
```
nitishm, line 155: @michelleN Do we want to check for existing deployments to perform the upgrade. My hunch is that it is irrelevant but I would rather discuss this first.

nitishm:
@draychev @shashankram I am trying to write unit tests for the update command.
The `install` command installs a new object, into an empty `store`
The `delete` command just stores the release name in the helm `store`

Using the below `release.Mock`,
```go
rel := release.Mock(&release.MockReleaseOptions{Name: meshName})
store.Create(rel)
```
, how can I populate the store with an installed object that I would want to update.

I am not savvy with `Ginkgo` and Behavior testing in general. If I try calling `installCmd.run()` and then do a `updateCmd.run()` the test seems to fail.

cmd/cli/update.go
```diff
@@ -0,0 +1,170 @@
...
+The default Kubernetes namespace that gets updated is called osm-system. 
+To update a control plane components in a different namespace, use the 
+--namespace flag.
+
+Example:
+  $ osm update --namespace hello-world
+
+Multiple control plane installations can exist within a cluster. Each
+control plane is given a cluster-wide unqiue identifier called mesh name.
+A mesh name can be passed in via the --mesh-name flag. By default, the
```
michelleN, line 68: a user should not be able to change the mesh name when updating a control plane. That is the ID of the control plane essentially and the name of the Helm release being updated.

cmd/cli/update.go
```diff
@@ -0,0 +1,170 @@
...
+			return upd.run(config)
+		},
+	}
+
+	f := cmd.Flags()
+	f.StringVar(&upd.cfg.containerRegistry, "container-registry", "openservicemesh", "container registry that hosts control plane component images")
+	f.StringVar(&upd.cfg.osmImageTag, "osm-image-tag", "v0.2.0", "osm image tag")
+	f.StringVar(&upd.cfg.containerRegistrySecret, "container-registry-secret", "acr-creds", "name of Kubernetes secret for container registry credentials to be created if it doesn't already exist")
+	f.StringVar(&upd.cfg.chartPath, "osm-chart-path", "", "path to osm chart to override default chart")
+	f.StringVar(&upd.cfg.certManager, "certificate-manager", defaultCertManager, "certificate manager to use (tresor or vault)")
```
michelleN, line 119: We are not ready to support some of these flags in the update action. The only options that should be able to be configured are the ones defined in the osm configmap in charts/osm/templates/osm-configmap.yaml

nitishm:
> Thanks for working on this @nitishm ! Left a few comments in line. Would be great to separate this into two PRs. One for just adding the update action and restricting the flags/fields for update to those outlined in the osm configmap called osm-config (see charts/osm/templates/osm-configmap) and another PR for consolidating/DRY-ing up code. That would make this much easier and speedier to review. Lmk what you think.

@michelleN Moved the PR to #1625 

With the latest PR, DRY-ing the code does not make sense at this point but yes eventually we should abstract certain things away, into their own modules.

nojnhuh:
Thanks for working on this, @nitishm! It looks like you're continuing to work on this in #1625 and possibly follow-up PRs. Is it Ok to close this PR then?

PR #1485, opened 2020-08-10
envoy/rds: use route direction while creating routes

shashankram:
The current code relies on passing bools around for source/destination
route configs and if the route corresponds to a local cluster. This can
be confusing and hard to understand if one does not know how these
booleans are interpreted in the code.
This change simplies the logic by introducing the concept of a direction
for a route. Routes can be outbound (to reach a remote service), or
inbound (to accept traffic on the local cluster host). Since the route
configurations are already using Inbound/Outbound terminology, updates
the variable names to reflect this.

There is no functional change introduced in the code, and the routing
logic remains as before.

- New Functionality      [ ]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [X]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests / CI System      [ ]
- Other                  [ ]

`No`

### DIFF
--- a/pkg/envoy/rds/response.go
+++ b/pkg/envoy/rds/response.go
@@ -38,10 +38,10 @@ func NewResponse(ctx context.Context, catalog catalog.MeshCataloger, proxy *envo
 	}
 
 	routeConfiguration := []*xds_route.RouteConfiguration{}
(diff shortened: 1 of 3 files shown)


ksubrmnn:
Great change! Didn't realize I was getting this confused myself

PR #1490, opened 2020-08-10
Add Gloo Gateway to Ingress documentation

christian-posta:
Add Gloo integration docs for Ingress+OSM

### DIFF
--- a/docs/patterns/ingress.md
+++ b/docs/patterns/ingress.md
@@ -9,10 +9,11 @@ This document describes how to expose HTTP and HTTPS routes outside the cluster 
 - An ingress controller must be running in the cluster.
 
 ## Exposing an HTTP or HTTPS service using Ingress
-A service can expose HTTP or HTTPS routes outside the cluster using Kubernetes Ingress along with an ingress controller. Once an ingress resource is configured to expose HTTP routes outside the cluster to a service within the cluster, OSM will configure the sidecar proxy on pods to allow ingress traffic to the service based on the ingress routing rules defined by the Kubernetes Ingress resource.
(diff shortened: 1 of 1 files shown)


docs/patterns/ingress.md
```diff
@@ -167,8 +169,78 @@ spec:
...
+```
+
+For Gloo's ingress, we don't need any additional annotations, however, the `kubernetes.io/ingress.class: gloo` annotation is recommended. With Gloo, we configure the `Upstream` objects with the appropriate trust authority. In Gloo, the `Upstream` object represents a routable target (Kubernetes Service, Consul Service, Cloud Function, etc). 
+
+To prepare the root certifciate, we must do something similar to what we do for the Azure Application Gateway.
+
+```bash
+kubectl get secret/osm-ca-bundle -n osm-system -o jsonpath="{.data['ca\.crt']}" | base64 -d > osm-c-bundlea.pem
+
+glooctl create secret tls --name osm-ca-bundle --rootca osm-c-bundlea.pem
```
shashankram, line 189: OSM stores the root certificate in a secret `osm-cabundle` in the `osm-system` namespace by default. Is it possible to reference this secret directly with Gloo based ingress? This would be similar to how `nginx-ingress` is configured with OSM.

christian-posta: Yah Gloo doesn't seem to recognize that secret because it's not a 'tls' secret... let me check the code real quick to double check

shashankram: Sounds good. Were you able to test ingess with OSM and Gloo?

christian-posta: Yah, gloo expects secrets a certain way, so no updates on OSM side. We will work internally to make this a little easier though and send updated PR at a later date.

PR #1495, opened 2020-08-11
Updated Bookstore Demo Apps requesting custom port number

### LINKED ISSUE #1483: The ports 8080, 8081, 8082 and 8083 cannot be used with Azure Cloud Shell Web Preview
While following the [OSM Manual Demo Guide](https://github.com/openservicemesh/osm/blob/main/docs/example/README.md) using _Azure Cloud Shell_ with a Kubernetes cluster created with _Azure Kubernetes Service (AKS)_, I have not been able to view the following UIs using the _Web Preview_:
- http://localhost:8080 - **Bookbuyer**
- http://localhost:8081 - **bookstore-v1**
- http://localhost:8082 - **bookstore-v2** 
- http://localhost:8083 - **bookthief**

because the _Web Preview_ only supports the following port ranges:
```
Configure port to preview
Valid ports include [1025-8079] and [8091-49151]
```
Would it be OK to change the ports 8080, 8081, 8082 and 8083 into, for instance, 7080, 7081, 7082 and 7083 so it would easily work with the _Web Preview_ of _Azure Cloud Shell_?

tstilwell91:

fixes #1483
This allows users to specify port numbers for the Bookstore Demo Applications.  It will prompt with the default port numbers but these can be changed.  The port-forward-ui scripts can be run individually or from port-forward-all.sh

- New Functionality      [X]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [ ]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [X]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests / CI System      [ ]
- Other                  [ ]

No.

### DIFF
--- a/scripts/port-forward-all.sh
+++ b/scripts/port-forward-all.sh
@@ -1,10 +1,15 @@
 #!/bin/bash
 
+read -p "Enter BOOKBUYER pod local port:" -i "8080" -e BOOKBUYER_PORT
+read -p "Enter BOOKSTOREv1 pod local port:" -i "8081" -e BOOKSTOREv1_PORT
+read -p "Enter BOOKSTOREv2 pod local port:" -i "8082" -e BOOKSTOREv2_PORT
+read -p "Enter BOOKTHIEF pod local port:" -i "8083" -e BOOKTHIEF_PORT
+
 ./scripts/port-forward-zipkin.sh &
-./scripts/port-forward-bookbuyer-ui.sh &
-./scripts/port-forward-bookstore-ui-v2.sh &
-./scripts/port-forward-bookstore-ui-v1.sh &
-./scripts/port-forward-bookthief-ui.sh &
+./scripts/port-forward-bookbuyer-ui.sh $BOOKBUYER_PORT &
+./scripts/port-forward-bookstore-ui-v2.sh bookstore-v2 $BOOKSTOREv2_PORT &
+./scripts/port-forward-bookstore-ui-v1.sh bookstore-v1 $BOOKSTOREv1_PORT &
+./scripts/port-forward-bookthief-ui.sh $BOOKTHIEF_PORT &
 ./scripts/port-forward-osm-debug.sh &
 ./scripts/port-forward-grafana.sh &
 
--- a/scripts/port-forward-all.sh
+++ b/scripts/port-forward-all.sh
@@ -1,15 +1,10 @@
 #!/bin/bash
 
-read -p "Enter BOOKBUYER pod local port:" -i "8080" -e BOOKBUYER_PORT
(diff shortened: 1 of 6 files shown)


scripts/port-forward-all.sh
```diff
@@ -1,10 +1,15 @@
...
+
 ./scripts/port-forward-zipkin.sh &
-./scripts/port-forward-bookbuyer-ui.sh &
-./scripts/port-forward-bookstore-ui-v2.sh &
-./scripts/port-forward-bookstore-ui-v1.sh &
-./scripts/port-forward-bookthief-ui.sh &
+./scripts/port-forward-bookbuyer-ui.sh $BOOKBUYER_PORT &
+./scripts/port-forward-bookstore-ui-v2.sh bookstore-v2 $BOOKSTOREv2_PORT &
+./scripts/port-forward-bookstore-ui-v1.sh bookstore-v1 $BOOKSTOREv1_PORT &
+./scripts/port-forward-bookthief-ui.sh $BOOKTHIEF_PORT &
```
shashankram, line 12: Since these are being backgrounded and run somewhat in parallel, how would the prompts look like? Can we make this simpler:
- ./scripts port-forward-bookbuyer.sh will run the port forward on the default ports
- ./scripts-port-forward-bookbuyer.sh 7071 will override the port used for port forwarding

This will avoid the need for prompts and seems like a simpler solution. What do you think?

tstilwell91: Appreciate the quick feedback!  Sure thing.  I haven't experienced an issue with the prompts but it's probably better to get rid of them.  Do you have any suggestions for the port-forward-all.sh script?  Should I just leave it as is and if someone needs to customize the ports, then they can edit the script manually?

shashankram: Another option is to keep the code here as is, and make use of an environment variable in the service specific port forward scripts, something like
```
BOOKBUYER_LOCAL_PORT="${BOOKBUYER_LOCAL_PORT:-8080}"
POD="$(kubectl get pods --selector app=bookbuyer -n "$BOOKBUYER_NAMESPACE" --no-headers | grep 'Running' | awk '{print $1}')"

kubectl describe pod "$POD" -n "$BOOKBUYER_NAMESPACE"

kubectl port-forward "$POD" -n "$BOOKBUYER_NAMESPACE" $BOOKBUYER_LOCAL_PORT:80
```

Someone wanting to override this port could use `BOOKBUYER_LOCAL_PORT=7070 ./scripts/port-forward-all.sh`.
This would apply to the other `port-forward-*ui*.sh` scripts

tstilwell91:
Sorry I didn't realize changing the README.md would dismiss the review.  This document was the only place I could find that references these specific `port-forward` scripts.

PR #1497, opened 2020-08-11
fix(*): update chart values and metadata

michelleN:
* change ImagePullPolicy on all Deployments to IfNotPresent
* remove default imagePullSecrets, user can add imagePullSecrets
  manually if required

- New Functionality      [ ]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [ ]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests / CI System      [ ]
- Other                  [X]

Thanks @stefanprodan for pointing these bits out!

### DIFF
--- a/charts/osm/values.yaml
+++ b/charts/osm/values.yaml
@@ -5,10 +5,9 @@ OpenServiceMesh:
   replicaCount: 1
   image:
     registry: openservicemesh
-    pullPolicy: Always
+    pullPolicy: IfNotPresent
     tag: v0.3.0
-  imagePullSecrets:
-    - name: acr-creds
+  imagePullSecrets: {}
   sidecarImage: envoyproxy/envoy-alpine:v1.15.0
   prometheus:
     port: 7070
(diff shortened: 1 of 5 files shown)


charts/osm/values.yaml
```diff
@@ -5,8 +5,8 @@ OpenServiceMesh:
   replicaCount: 1
   image:
     registry: openservicemesh
-    pullPolicy: Always
-    tag: latest
+    pullPolicy: IfNotPresent
+    tag: v0.2.0
   imagePullSecrets:
     - name: acr-creds
```
stefanprodan, line 11: Can you please remove the ACR secret?

michelleN: yea let me make that section more generic

stefanprodan:
Can we change the pull policy on all container Grafana, Prometheus, zipkin) to IfNotPresent?

PR #1499, opened 2020-08-11
fix(cli): osm namespace add should fail when adding osm-system NS

Sheshagiri:
fixes: https://github.com/openservicemesh/osm/issues/1196
This has changes to return an error when a user tries to add a namespace which has osm control plane running in it.

- New Functionality      [ ]
- Documentation          [ ]
- Install                [X]
- Control Plane          [ ]
- CLI Tool               [x]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests / CI System      [ ]
- Other                  [ ]

todo:
- [x] update tests

### DIFF
--- a/cmd/cli/install.go
+++ b/cmd/cli/install.go
@@ -202,7 +202,7 @@ func (i *installCmd) run(config *helm.Configuration) error {
 	}
 
 	deploymentsClient = i.clientSet.AppsV1().Deployments(settings.Namespace()) // Get deployments for specified namespace
-	labelSelector = metav1.LabelSelector{MatchLabels: map[string]string{"app": OSMControllerName}}
+	labelSelector = metav1.LabelSelector{MatchLabels: map[string]string{"app": constants.OSMControllerName}}
 
 	listOptions = metav1.ListOptions{
 		LabelSelector: labels.Set(labelSelector.MatchLabels).String(),
--- a/cmd/cli/install.go
+++ b/cmd/cli/install.go
@@ -56,7 +56,6 @@ const (
 	defaultCertManager         = "tresor"
 	defaultVaultProtocol       = "http"
 	defaultMeshName            = "osm"
-	osmControllerLabel         = "osm-controller"
 	defaultCertValidityMinutes = int(1440) // 24 hours
 )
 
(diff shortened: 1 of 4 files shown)


cmd/cli/namespace_add.go
```diff
@@ -64,13 +65,26 @@ func (a *namespaceAddCmd) run() error {
 		ctx, cancel := context.WithCancel(context.Background())
 		defer cancel()
 
-		patch := `{"metadata":{"labels":{"` + constants.OSMKubeResourceMonitorAnnotation + `":"` + a.meshName + `"}}}`
-		_, err := a.clientSet.CoreV1().Namespaces().Patch(ctx, ns, types.StrategicMergePatchType, []byte(patch), metav1.PatchOptions{}, "")
-		if err != nil {
-			return errors.Errorf("Could not label namespace [%s]: %v", ns, err)
+		deploymentsClient := a.clientSet.AppsV1().Deployments(ns)
+		labelSelector := metav1.LabelSelector{MatchLabels: map[string]string{"app": "osm-controller"}}
```
ksubrmnn, line 69: we should probably make this a constant

shashankram:
@Sheshagiri, thanks for the change, would you mind adding tests prior to merging?

Sheshagiri:
> @Sheshagiri, thanks for the change, would you mind adding tests prior to merging?

Sure, I have a todo. I'll add those and mark this as ready for review.

cmd/cli/install.go
```diff
@@ -53,10 +53,10 @@ well as for adding a Kubernetes Namespace to the list of Namespaces a control
...
 `
 const (
-	defaultCertManager   = "tresor"
-	defaultVaultProtocol = "http"
-	defaultMeshName      = "osm"
-
+	defaultCertManager         = "tresor"
+	defaultVaultProtocol       = "http"
+	defaultMeshName            = "osm"
+	osmControllerLabel         = "osm-controller"
```
ksubrmnn, line 59: can you move this tp pkg/constants/constants.go?

Otherwise LGTM!

Sheshagiri: done :)

cmd/cli/namespace_test.go
```diff
@@ -93,6 +93,34 @@ var _ = Describe("Running the namespace add command", func() {
...
+		Context("given one namespace with osm-controller installed in it as an arg", func() {
+			BeforeEach(func() {
+				out = new(bytes.Buffer)
+				fakeClientSet = fake.NewSimpleClientset()
+				// mimic osm controller deployment in testNamespace
+				deploymentSpec := createDeploymentSpec(testNamespace, defaultMeshName)
+				fakeClientSet.AppsV1().Deployments(testNamespace).Create(context.TODO(), deploymentSpec, metav1.CreateOptions{})
+
+				nsSpec := createNamespaceSpec(testNamespace, "")
+				fakeClientSet.CoreV1().Namespaces().Create(context.TODO(), nsSpec, metav1.CreateOptions{})
```
michelleN, line 105: I was surprised to see that we're creating the namespace after creating the deployment in the namespace. Does this work because we're using `fakeClientSet` so it is not validating that the value of `testNamespace` exists before creating the Deployment with that Namespace?

PR #1518, opened 2020-08-13
zipkin : Updating configmap for zipkin

snehachhabria:
This PR adds updates the configmap to show zipkin related values only when zipkin tracking is enabled in OSM

- New Functionality      [ ]
- Documentation          [ ]
- Install                [X]
- Control Plane          [ ]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests / CI System      [ ]
- Other                  [ ]

### DIFF
--- a/charts/osm/templates/osm-configmap.yaml
+++ b/charts/osm/templates/osm-configmap.yaml
@@ -9,10 +9,12 @@ data:
   envoy_log_level: {{ .Values.OpenServiceMesh.envoyLogLevel | quote }}
   prometheus_scraping: "true"
 
+{{- if .Values.OpenServiceMesh.zipkin.enable }}
   zipkin_tracing: {{ .Values.OpenServiceMesh.zipkin.enable | quote }}
   zipkin_address: {{ .Values.OpenServiceMesh.zipkin.address | quote }}
   zipkin_port: {{ .Values.OpenServiceMesh.zipkin.port | quote }}
   zipkin_endpoint: {{ .Values.OpenServiceMesh.zipkin.endpoint | quote }}
+{{- end }}
 
 {{- if .Values.OpenServiceMesh.enableEgress }}
   mesh_cidr_ranges: {{ .Values.OpenServiceMesh.meshCIDRRanges | quote }}


pkg/configurator/client.go
```diff
@@ -138,9 +141,9 @@ func (c *Client) getConfigMap() *osmConfig {
 		UseHTTPSIngress:             getBoolValueForKey(configMap, useHTTPSIngressKey),
 
 		ZipkinTracing:  getBoolValueForKey(configMap, zipkinTracingKey),
-		ZipkinAddress:  getStringValueForKey(configMap, zipkinAddressKey),
-		ZipkinPort:     getIntValueForKey(configMap, zipkinPortKey),
-		ZipkinEndpoint: getStringValueForKey(configMap, zipkinEndpointKey),
```
shashankram, line 143: Can these helpers be modified to log a message and return default values for `string`, `int` types? 
Would that avoid the need for separate helpers considering the defaults are nothing but the zero values for those types?

shashankram:
> hey @snehachhabria - why are we updating this post-deploy. The service/port info is already available in the chart values file. Why is that not being passed in to osm-controller at deploy time? Could you explain a bit more about why the changes to `pkg/configuration/client.go` are needed?

Based on my chat with Sneha, the changes to `pkg/configuration/client.go` are not needed.

pkg/configurator/client.go
snehachhabria: I have removed the helpers and replaced them with the previous methods. PTAL

snehachhabria:
> hey @snehachhabria - why are we updating this post-deploy. The service/port info is already available in the chart values file. Why is that not being passed in to osm-controller at deploy time? Could you explain a bit more about why the changes to `pkg/configuration/client.go` are needed?

@michelleN I have reverted the changes in `pkg/configuration/client.go`, PTAL and let me know if you have any other questions. The changes I am making are related to zipkin and has nothing to do with osm-controller

PR #1532, opened 2020-08-14
docs(*): add design doc for uninstalling osm

michelleN:
This PR adds documentation for the design of `osm mesh delete`. While this command is already implemented, the document gives context on the feature, motivation, and future direction.

- New Functionality      [ ]
- Documentation          [X]
- Install                [ ]
- Control Plane          [ ]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests / CI System      [ ]
- Other                  [ ]

- Does this change contain code from or inspired by another project? If so, did you notify the maintainers and provide attribution? No. I took a look at how Kubernetes structures KEPs (Kubernetes Enhancement Proposals) but didn't end up copying the same structure.

### DIFF
--- a/docs/design/uninstall_osm.md
+++ b/docs/design/uninstall_osm.md
@@ -1,0 +1,66 @@
+# Uninstall OSM Control Plane from Cluster
+
+- **Revision:** 1
+- **Status:** Implemented
+
+## Table of Contents
+<!-- toc -->
+- [Summary](#summary)
+  - [Use Case](#use-case)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [Experience](#experience)
+  - [Solution](#solution)
+    - [Implementation Details](#implementation-details)
+  - [Testing](#testing)
+- [Historical Context](#historical-context)
+- [Looking Forward](#looking-forward)
+<!-- /toc -->
+
+## Summary
+A user should be able to delete all Kubernetes resources related to an OSM control plane installation using a CLI command.
+
+### Use Case
+A user can use `osm install` to install an OSM control plane in a Kubernetes cluster to test out OSM. Once tested, they may want to uninstall the control plane components.
+
+## Motivation
+Make it easier to cleanup OSM control planes in test clusters
+
+### Goals
+Delete all Kubernetes resources associated with OSM control plane.
+
+### Non goals
+- This version of the proposal will not handle removing any CRDs since these are cluster-wide resources that may be used by other OSM control planes, service mesh implementations or tools.
+- This version of the proposal will not handle removing Envoy proxies from application pods. The user must re-start application Pods once the `osm` control plane has been deleted. Traffic will be interrupted.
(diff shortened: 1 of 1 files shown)


docs/design/uninstall_osm.md
```diff
@@ -0,0 +1,66 @@
...
+
+## Motivation
+Make it easier to cleanup OSM control planes in test clusters
+
+### Goals
+Delete all Kubernetes resources associated with OSM control plane.
+
+### Non goals
+- This version of the proposal will not handle removing any CRDs since these are cluster-wide resources that may be used by other OSM control planes, service mesh implementations or tools.
+- This version of the proposal will not handle removing Envoy proxies from application pods. The user must re-start application Pods once the `osm` control plane has been deleted. Traffic will be interrupted.
```
nojnhuh, line 36: > Traffic will be interrupted.

Is it true that using a [PodDisruptionBudget](https://kubernetes.io/docs/concepts/workloads/pods/disruptions/) for highly-available workloads would mitigate any interruptions? If so, is that worth mentioning here?

docs/design/uninstall_osm.md
```diff
@@ -0,0 +1,66 @@
...
+CLI command:
+```bash
+osm mesh delete MESH_NAME
+```
+
+### Solution
+The CLI command will delete the `osm-controller` Deployment, Service and any other resources associated with the OSM control plane with the given MESH_NAME.
+
+#### Implementation Details
+The `osm` CLI uses Helm libraries and a Helm chart under the hood to install an osm control plane in a Kubernetes cluster. It also creates a namespace (`osm-system` by default) to house the Kubernetes resources defined in and installed using the Helm chart. The `osm mesh delete` command works by deleting the Helm release associated with the control plane. The Helm release lives in the same namespace as the OSM control plane installation. The namespace will not be deleted because we don't know what else for now the user has installed in that namespace. Because the SMI CRDs live in the chart's `crds/` directory, they are protected from any delete action in Helm.
```
nojnhuh, line 51: Nit: capitalize "OSM" for consistency.
```suggestion
The `osm` CLI uses Helm libraries and a Helm chart under the hood to install an OSM control plane in a Kubernetes cluster. It also creates a namespace (`osm-system` by default) to house the Kubernetes resources defined in and installed using the Helm chart. The `osm mesh delete` command works by deleting the Helm release associated with the control plane. The Helm release lives in the same namespace as the OSM control plane installation. The namespace will not be deleted because we don't know what else for now the user has installed in that namespace. Because the SMI CRDs live in the chart's `crds/` directory, they are protected from any delete action in Helm.
```

docs/design/uninstall_osm.md
michelleN: A PodDisruptionBudget would be helpful if we wanted to ensure a certain number of Pods are available by blocking voluntary disruptions like using `kubectl drain` / draining a node for repair or upgrade. In this case, a PodDisruptionBudget would not be able help deal with traffic interruption unfortunately. Traffic is interrupted because the Envoy Proxy remains in the Pod but doesn't have a management server to talk to.

PR #1542, opened 2020-08-17
Fix how service selectors are matched to pod labels

### LINKED ISSUE #1524: OSM assumes that the app label on a Service resource will have same name as the Service
**Bug description**:
When OSM matches Deployment resources against a service, it parses the app label and and interprets the label to refer to a k8s Service having the same name as the value for the app label.
https://github.com/openservicemesh/osm/blob/main/pkg/endpoint/providers/kube/client.go#L131-L141

This is incorrect, because such a service might not exist because Kubernetes allows Services to have app labels with a different name. Example:
```
apiVersion: v1
kind: Service
metadata:
  name: my-service
spec:
  selector:
    app: MyApp
  ports:
    - protocol: TCP
      port: 80
      targetPort: 9376
```

When the K8s Service name is different than its app label, OSM fails to match a Deployement to a Service, and things don't work as expected.

- SMI Traffic Access Policy  [X]

**Expected behavior**:
OSM should honor Kubernetes Service definition and not make assumptions on the app label being the same as the service name.

Deploy a Service where the app label is different than the Service name.

**How was OSM installed?**:
`osm install` command.

**Anything else we need to know?**: No

**Environment**:
- OSM version (use `osm version`): Version: dev; Commit: 6ded080; Date: 2020-08-13-14:57

Client Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.2", GitCommit:"59603c6e503c87169aea6106f57b9f242f64df89", GitTreeState:"clean", BuildDate:"2020-01-18T23:30:10Z", GoVersion:"go1.13.5", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"15", GitVersion:"v1.15.10", GitCommit:"ba1ce4122d67b0f7399a41afbbd8a7258d22ecb8", GitTreeState:"clean", BuildDate:"2020-04-21T00:13:25Z", GoVersion:"go1.12.12", Compiler:"gc", Platform:"linux/amd64"}

- Size of cluster (number of worker nodes in the cluster): 1 node
- Others: N/A

shashankram:
OSM does not correctly match services to pod labels based
on the service selector. This change resolves a few existing
bugs in relation to service selector matching.

- It resolves how service selectors are matched to pod labels
  by leveraging the k8s client-go API.

- Resolves incorrect selector matching which relies on label
  intersection between service and pod labels. K8s selectors
  are not computed using label intersection, its more complex
  and supports additional matching criteria. The unit tests
  prove that label intersection for selectors is not correct.

- Resolves incorrect service lookup by assuming a service with
  the same name as the pod label exits.

- Removes assumptions regarding resources to have the `app` label
  to perform selector matching.

This change does not alter other existing assumptions in the code,
but rather only addresses the existing bugs with service selector
matching.

The change also updates the CI's iteration threshold to be closer to the test
timeout interval.

Resolves #1524

- New Functionality      [ ]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [ ]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests / CI System      [ ]
- Other                  [X]

`No`


pkg/catalog/xds_certificates.go
```diff
@@ -151,11 +143,10 @@ func listServicesForPod(pod *v1.Pod, kubeClient kubernetes.Interface) ([]v1.Serv
...
 	}
 
-	podLabels := mapStringStringToSet(pod.Labels)
-
 	for _, svc := range svcList.Items {
-		serviceLabelSet := mapStringStringToSet(svc.Spec.Selector)
-		if serviceLabelSet.Intersect(podLabels).Cardinality() > 0 {
+		svcRawSelector := svc.Spec.Selector
+		selector := labels.Set(svcRawSelector).AsSelector()
+		if selector.Matches(labels.Set(pod.Labels)) {
```
eduser25, line 149: nice

PR #1577, opened 2020-08-18
ref(*): Integrate golangci-lint

nojnhuh:

This PR standardizes on `golangci-lint` as the primary static analysis tool for source code across the Makefile, other scripts, and CI. The initial settings only point out new issues (based on git history) and enable all linters (per @draychev). I expect we'll tweak settings as we go, but if we find it's too restrictive we can override the return code to 0 to prevent CI from failing while still showing issues it found.

Some standalone `go vet`, `go lint`, and `go fmt` checks have been removed since `golangci-lint` uses them under the hood.

One notable difference is that by default, `golangci-lint` ignores the ever-popular `exported method should have a comment or be unexported` and similar messages. Those messages are part of a set that are ignored by default (see `--exclude-use-default` in `golangci-lint run --help`), but we can disable those if we want to fail on those messages anyway.

- New Functionality      [ ]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [ ]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests / CI System      [X]
- Other                  [ ]

  No

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -31,31 +31,17 @@ jobs:
       - name: ShellCheck
         run: shellcheck -x $(find . -name '*.sh')
 
-  gofmt:
-    name: Go fmt
+  lint:
+    name: Lint
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
         uses: actions/checkout@v1
-      - name: Setup Go 1.14
-        uses: actions/setup-go@v1
+      - name: golangci-lint
+        uses: golangci/golangci-lint-action@v2
         with:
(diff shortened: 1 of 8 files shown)


nojnhuh:
> Does this flag existing issues when a new change modifies the same file but different lines?

I think what you're getting at is "If I change a line in a file and add no issues but there was an existing issue on a different line in that file, will that existing issue be flagged?" And to the best of my knowledge, the answer to that question is no.

PR #1583, opened 2020-08-18
chore(*): specify linters to run

michelleN:
+ we had go-vet, gofmt and go-lint enabled before
using golangci-lint, so I've re-enabled that setting
using golangci-lint. Let's iteratively add linters.

- New Functionality      [ ]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [ ]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests / CI System      [ ]
- Other                  [ ]

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -6,8 +6,15 @@ issues:
 
 linters:
   enable:
+    - goconst
+    - gocyclo
     - gofmt
+    - goimports
     - golint
+    - govet
+    - misspell
+    - unused
+    - whitespace
 
 linters-settings:
   goimports:
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -5,7 +5,9 @@ issues:
   new: true
 
 linters:
-  enable-all: true
+  enable:
+    - gofmt
+    - golint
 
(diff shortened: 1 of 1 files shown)


.golangci.yml
```diff
@@ -5,7 +5,9 @@ issues:
   new: true
 
 linters:
-  enable-all: true
+  enable:
+    - gofmt
+    - golint
```
shashankram, line 10: Why not run all?

shashankram: Upon further reading, I see the `enable-all` option being deprecated soon:
https://golangci-lint.run/usage/configuration/

Can we set the following to start with?
```
enable:
    - goconst
    - gocyclo
    - gofmt
    - goimports
    - golint
    - govet
    - misspell
    - unused
    - whitespace
```

michelleN: Yes happy to do that. Some of these are enabled by default but it doesn't hurt to explicitly have our list laid out here.

PR #1593, opened 2020-08-20
Replacing Zipkin -> Jagger for tracing

eduser25:
Note: Zipkin could refer both to the app and the protocol format
for tracing.

For things referring to the app specifically, this commit replaces
zipking with jaeger in all related forms (CRDs, deploy options, etc)

For OSM, and to keep the code as agnostic as possible,
the ZipkinCluster has become the TracingCluster (as anything
could be listening to tracing in Zipkin protocol and not necessarily
Jaeger or Zipkin), and the zipkin endpoint configuration is kept as such,
as the remote listener is still expected to listen in Zipkin format.

- Install                  [X]
- Metrics                [X]
- Other                   [X] (Tracing)

- Does this change contain code from or inspired by another project?
No


pkg/envoy/cds/response_test.go
```diff
@@ -75,7 +75,7 @@ var _ = Describe("CDS Response", func() {
 			// 1. Destination cluster
 			// 2. Source cluster
 			// 3. Prometheus cluster
-			// 4. Zipkin cluster
+			// 4.  cluster
```
shashankram, line 78:
```suggestion
			// 4.  Tracing cluster
```

pkg/configurator/client.go
```diff
@@ -78,16 +78,16 @@ type osmConfig struct {
...
+	// ZipkinAddress is the address of the zipkin-based listener cluster
 	ZipkinAddress string `yaml:"zipkin_address"`
 
-	// ZipkinPort is the port of the Zipkin cluster
+	// ZipkinPort remote port for the zipkin-based listener
 	ZipkinPort int `yaml:"zipkin_port"`
 
-	// ZipkinEndpoint is the endpoint of the Zipkin cluster
+	// ZipkinEndpoint is the protocol endpoint for the zipkin-based listener
 	ZipkinEndpoint string `yaml:"zipkin_endpoint"`
```
shashankram, line 91: To avoid ambiguity between Zipkin deployment and protocol, it would make sense to change to `s/Zipkin/Tracing` here similar to the other changes you've made.

charts/osm/values.yaml
```diff
@@ -38,24 +38,23 @@ OpenServiceMesh:
...
-  # namespace where OSM resides. Set this to false if Zipkin
-  # has already been installed or is not needed.
-  deployZipkin: false
+  # Set deployJaeger to true to deploy a Jaeger cluster in the
+  # namespace where OSM resides.
+  deployJaeger: true
 
-  # The following section configures Zipkin tracing for the Envoys
-  # connected to the OSM control plane.
-  zipkin:
```
snehachhabria, line 48: I think it was better to have this key and all the other related keys nested under it. Though I find the naming as zipkin little confusing as it will all be related to the jaeger deployment

charts/osm/values.yaml
```diff
@@ -38,24 +38,23 @@ OpenServiceMesh:
...
+    enable: true
 
-    # Informs Envoys where the Zipkin spans should be sent
-    address: "zipkin.osm-system.svc.cluster.local"
+    # Tracing destination cluster
+    address: "jaeger.osm-system.svc.cluster.local"
 
-    # Port of the Zipkin cluster
+    # Destination port for the Zipkin protocol listener
     port: 9411
```
snehachhabria, line 57: Will this be 16686 or 9411 still ?

pkg/configurator/client.go
eduser25: that I can do. Left it with Zipkin to clarify it was zipkin based, but Tracing looks good too.

charts/osm/values.yaml
eduser25: @shashankram just suggested to have the section renamed to Tracing (intead of zipkin).

The reason for the separation I believe is to separate the "deployment" option with the "endpoint location" config.
would you still believe `deployYagger` should be under `tracingEndpoint` ?

pkg/envoy/cds/response_test.go
eduser25: done

pkg/constants/constants.go
```diff
@@ -45,17 +45,17 @@ const (
...
 	EnvoyMetricsCluster = "envoy-metrics-cluster"
 
-	// EnvoyZipkinCluster is the name of the Zipkin cluster.
-	EnvoyZipkinCluster = "envoy-zipkin-cluster"
+	// EnvoyTracingCluster is the default name to refer to the tracing cluster.
+	EnvoyTracingCluster = "envoy-tracing-cluster"
 
-	// DefaultZipkinEndpoint is the Zipkin endpoint.
-	DefaultZipkinEndpoint = "/api/v2/spans"
+	// DefaultTracingEndpoint is the default endpoint route.
```
shashankram, line 51: What does `default endpoint route` imply here?

eduser25: I had to actually check, it's the default http path where where the metrics are pushed via POST
https://zipkin.io/zipkin-api/#/default/post_spans

Could be hardcoded maybe? it's part of the zipkin v2 api and this should not change I think

shashankram: My question was more to do with the term `endpoint route`, which seems a bit off in the context of tracing.

eduser25: `http path`? `http api path`?

PR #1594, opened 2020-08-20
test(kube endpoints): fix flaky test

nojnhuh:
This changes modifies a flaky test not to rely on the order of services
returned from `GetServicesForServiceAccount()`.

- New Functionality      [ ]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [ ]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests / CI System      [X]
- Other                  [ ]

No

### DIFF
--- a/pkg/endpoint/providers/kube/client_test.go
+++ b/pkg/endpoint/providers/kube/client_test.go
@@ -441,7 +441,8 @@ var _ = Describe("When getting a Service associated with a ServiceAccount", func
 			{Name: "test-1", Namespace: testNamespace},
 			{Name: "test-2", Namespace: testNamespace},
 		}
-		Expect(meshServices).To(Equal(expectedServices))
+		Expect(meshServices[0]).To(BeElementOf(expectedServices))
+		Expect(meshServices[1]).To(BeElementOf(expectedServices))
 
 		err = fakeClientSet.AppsV1().Deployments(testNamespace).Delete(context.Background(), deployment.Name, metav1.DeleteOptions{})
 		Expect(err).ToNot(HaveOccurred())


pkg/endpoint/providers/kube/client.go
```diff
@@ -113,7 +112,7 @@ func (c Client) ListEndpointsForService(svc service.MeshService) []endpoint.Endp
 // GetServicesForServiceAccount retrieves a list of services for the given service account.
 func (c Client) GetServicesForServiceAccount(svcAccount service.K8sServiceAccount) ([]service.MeshService, error) {
 	log.Info().Msgf("[%s] Getting Services for service account %s on Kubernetes", c.providerIdent, svcAccount)
-	services := mapset.NewSet()
+	var services []service.MeshService
```
shashankram, line 115: The mapset was used to avoid duplicates, this needs to be handled.

nojnhuh: Ok, I'll change it back and just add an extra step to sort the slice by namespace/name then.

shashankram: Why does the test rely on ordering though? Could you point me to the test code which is resulting in this?

nojnhuh: https://github.com/nojnhuh/osm/blob/35e671640f777b1ee8ceefcee48264473b46b330/pkg/endpoint/providers/kube/client_test.go#L444

shashankram: @nojnhuh we could change the test to something like
```
		expectedServices := []{..., ...}
		Expect(services[0]).To(BeElementOf(expectedServices))
		Expect(services[1]).To(BeElementOf(expectedServices))
```

nojnhuh: Has the change to handle multiple services been propagated throughout though? I know the plan was to just return the first element of the list while we were working on that, so the first service should always be the same one, right?

shashankram: The change to handle multiple services has not been handled throughout, but is in progress. However, this change doesn't relate to it. We should not be introducing new assumptions that are in the process of being removed.

Either of the following work:
1. Return sorted list
2. Update the test to not rely on ordering

nojnhuh: Ok, I'll change the tests.

PR #1624, opened 2020-08-26
trafficpolicy: consolidate routes between a source and destination

shashankram:
Currently if multiple routes exist between a source and destination
service, a TrafficTarget is created per route. This complicates
the logic in the control plane because routes in different
TrafficTarget objects need to be aggregated for a given source and
destination pair. For HTTP routes this is handled in RDS, although
a bit differently by aggregating the routes based on the hostname
of the destination service.

It is easier to implement the routing policies between a source
and destination when all the routes are consolidated in a single
TrafficTarget object, because at the time of building the route
configuration one is aware of all possible routes.

This change consolidates the routes between a source and destination
in a single TrafficTarget object. To consolidate the routes, the
source and destination services in a traffic target policy are
hashed to a key, to which all corresponding routes are associated.
This ensures that all routes between a source and destination get
associcated to the corresponding TrafficTarget policy. RDS is updated
to loop through all the routes when building the traffic policies.

- New Functionality      [ ]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [X]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests / CI System      [ ]
- Other                  [ ]

`No`


pkg/catalog/routes_test.go
```diff
@@ -30,27 +30,39 @@ var _ = Describe("Catalog tests", func() {
 		})
 	})
 
-	Context("Test getTrafficPolicyPerRoute", func() {
-		It("lists traffic policies", func() {
-			allTrafficPolicies, err := getTrafficPolicyPerRoute(mc, tests.RoutePolicyMap, tests.BookbuyerService)
+	Context("Test getTrafficPoliciesForService", func() {
+		It("should return all the traffic policies associated with a service", func() {
```
snehachhabria, line 34: Can we have another case in this test whereby we see another route being added to the same src:dest traffic policy

shashankram: Done.

PR #1625, opened 2020-08-26
Add update command with subset of flags

nitishm:

Added an `update` command to the OSM CLI as per the requirements of #1474 

- New Functionality      [ ]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [ ]
- CLI Tool               [X]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests / CI System      [ ]
- Other                  [ ]

No


nitishm:
@michelleN Here's the PR solely for the `update` command

nitishm:
I might need help in fleshing out the unit tests though.

nitishm:
@michelleN FYI

cmd/cli/update.go
```diff
@@ -0,0 +1,178 @@
...
+This command updates an osm control plane on the Kubernetes cluster.
+
+The default Kubernetes namespace that gets updated is called osm-system. 
+To update a control plane components in a different namespace, use the 
+--namespace flag.
+
+Example:
+  $ osm update --namespace hello-world
+
+Multiple control plane installations can exist within a cluster. Each
```
michelleN, line 67: The mesh name is not a property that can or should be updated. Can you update the description?

cmd/cli/update.go
```diff
@@ -0,0 +1,178 @@
...
+
+	// TODO (nitishm): Should we check if the controlplane deployment exists ?
+
+	upgradeClient := helm.NewUpgrade(config)
+	upgradeClient.Namespace = settings.Namespace()
+	if _, err = upgradeClient.Run(chartRequested.Name(), chartRequested, values); err != nil {
+		return err
+	}
+
+	fmt.Fprintf(u.out, "OSM successfully updated in namespace [%s]\n", settings.Namespace())
```
nitishm, line 160: @michelleN If we need to also print the Mesh Name we will need to do a lookup.  WDYT?

cmd/cli/update.go
```diff
@@ -0,0 +1,178 @@
...
+control plane is given a cluster-wide unqiue identifier called mesh name.
+A mesh name can be passed in via the --mesh-name flag. By default, the
+mesh-name name will be set to "osm." The mesh name must conform to same
+guidelines as a valid Kubernetes label value. Must be 63 characters or
+less and must be empty or begin and end with an alphanumeric character
+([a-z0-9A-Z]) with dashes (-), underscores (_), dots (.), and
+alphanumerics between.
+
+Example:
+  $ osm update --mesh-name "hello-osm"
```
michelleN, line 77: A better example might be to use `$ osm update MESH_NAME --enable-egress false

cmd/cli/update.go
```diff
@@ -0,0 +1,178 @@
+/*
+Copyright The Helm Authors.
+
```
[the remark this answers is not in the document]
nitishm: Not sure how it got there 🤔 . Will remove it.

michelleN:
@nitishm Please take a look at [mesh_delete_test.go](https://github.com/openservicemesh/osm/blob/main/cmd/cli/mesh_delete_test.go) or [install_test.go](https://github.com/openservicemesh/osm/blob/main/cmd/cli/install_test.go) to see how you could set up some unit tests for this change.

cmd/cli/update.go
```diff
@@ -0,0 +1,178 @@
...
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
+
+This file is inspired by the way Helm handles environment variables
```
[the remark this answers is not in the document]
nitishm: Done

cmd/cli/update.go
nitishm: Done

cmd/cli/update.go
nitishm: Done

cmd/cli/update.go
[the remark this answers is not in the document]
nitishm: Done

cmd/cli/update.go
nitishm: > Thanks. Left some comments inline.
> On another note: The issue defined was for `osm update MESH_NAME` but now that I am reading through it, I'm thinking we should nest the update command under the `osm mesh` command so: `osm mesh update MESH_NAME` what do you think?

IMO I agree. All mesh-related commands would best be under the `mesh`. This would include `install`, `delete` and `update`. Is that what you are thinking as well?

michelleN: I think install can remain top level as it is a command that will be used frequently. All other mesh related commands should be nested under `osm mesh`

nitishm: Sounds good. Do you want me to do it in this PR or a follow up PR?

cmd/cli/update.go
```diff
@@ -0,0 +1,124 @@
...
+To update a control plane components in a different namespace, use the 
+--namespace flag.
+
+Example:
+  $ osm update --namespace hello-world --enable-egress false
+`
+
+type updateCmd struct {
+	out       io.Writer
+	clientSet kubernetes.Interface
```
michelleN, line 32: I'm not seeing the clientSet being used anywhere in this code

nitishm: Will omit

cmd/cli/update.go
michelleN: In this PR please.

shalier (commented):
@nitishm Could you rebase on to the latest changes in main?

shalier (commented):
@nitishm Seems that when enabling multiple flags it writes onto the original configMap instead of the most recently updated configMap
`osm update --enable-permissive-traffic-policy true`
`osm update --enable-egress true`
will return
`data:`
  `egress: "true"`
...
  `permissive_traffic_policy_mode: "false"`
rather than 
`data:`
  `egress: "true"`
...
  `permissive_traffic_policy_mode: "true"`

shalier (commented):
@nitishm Seems that after enabling a flag, you're unable to disable it 
`osm update --enable-egress true`
will return in configMap
`data:`
  `egress: "true"`
but `osm update --enable-egress false` will still show in configMap as
`data:`
  `egress: "true"`

nitishm:
Hmm. Maybe it's not picking up the latest version of the chart 🤔. Will take a look at it and update the test cases as well.

nitishm:
So this is slightly more complex than what I thought initially. 
```go
func (u *updateCmd) resolveValues(currentValues map[string]interface{}) (map[string]interface{}, error) {
	valuesConfig := []string{
		fmt.Sprintf("OpenServiceMesh.enablePermissiveTrafficPolicy=%t", u.enablePermissiveTrafficPolicy),
		fmt.Sprintf("OpenServiceMesh.enableEgress=%t", u.enableEgress),
		fmt.Sprintf("OpenServiceMesh.meshCIDRRanges=%s", strings.Join(u.meshCIDRRanges, " ")),
	}

	for _, val := range valuesConfig {
		if err := strvals.ParseInto(val, currentValues); err != nil {
			return currentValues, err
		}
	}

	return currentValues, nil
}
```

`resolveValues()` overrides values from 
```go
	f.BoolVar(&upd.enablePermissiveTrafficPolicy, "enable-permissive-traffic-policy", false, "Enable permissive traffic policy mode")
	f.BoolVar(&upd.enableEgress, "enable-egress", false, "Enable egress in the mesh")
	f.StringSliceVar(&upd.meshCIDRRanges, "mesh-cidr", []string{}, "mesh CIDR range, accepts multiple CIDRs, required if enable-egress option is true")
```
, which means if a flag is not specified it takes the zero value for that flag. 

Either the user has to specify every flag in the command (which is unwieldy and affects the UX) or we need a different way of differentiating which fields are being updated vs the ones that should carry the current value forward.

Any ideas on how we can do this @michelleN @shalier? One thought is to use the `--set` form of setting config values similar to Helm.

shalier (commented):
@nitishm I agree that using something like --set in Helm for updating the config values sounds like a good way to go about this

cmd/cli/update_test.go
```diff
@@ -0,0 +1,103 @@
+package main
+
+import (
+	"bytes"
+	"io/ioutil"
+
+	. "github.com/onsi/ginkgo"
```
khenidak, line 7: please stop using this approach. it is an anti pattern

cmd/cli/update.go
```diff
@@ -0,0 +1,111 @@
...
+	f.StringVar(&upd.chartPath, "osm-chart-path", "", "path to osm chart to override default chart")
+	f.BoolVar(&upd.enablePermissiveTrafficPolicy, "enable-permissive-traffic-policy", false, "Enable permissive traffic policy mode")
+	f.BoolVar(&upd.enableEgress, "enable-egress", false, "Enable egress in the mesh")
+	f.StringSliceVar(&upd.meshCIDRRanges, "mesh-cidr", []string{}, "mesh CIDR range, accepts multiple CIDRs, required if enable-egress option is true")
+
+	return cmd
+}
+
+func (u *updateCmd) run(config *helm.Configuration) error {
+	var chartRequested *chart.Chart
```
khenidak, line 62: This is a general question. Not a comment. Generally speaking you want to have run sequenced as:
1. default input if needed
2. validate input as needed
3. run..

This even works better if and when you choose to use versioned options/flags.

Do you know why aren't we doing that?

khenidak (changes requested):
Thank you for this. A comment and a question has been added

cmd/cli/update.go
ksubrmnn: Do you mean something like this, Kal? https://github.com/openservicemesh/osm/blob/b7337ab144d0126fa44312bc3001f563183ffb76/cmd/cli/install.go#L199

We use the model for install where there are many arguments that are passed in. We could do the same to validate if we can load the chart

nitishm:
I can rebase - but this is not functioning as intended - see @shalier 's comment https://github.com/openservicemesh/osm/pull/1625#pullrequestreview-532774230

nojnhuh:
> @nitishm Seems that when enabling multiple flags it writes onto the original configMap instead of the most recently updated configMap
> `osm update --enable-permissive-traffic-policy true`
> `osm update --enable-egress true`
> will return
> `data:`
> `egress: "true"`
> ...
> `permissive_traffic_policy_mode: "false"`
> rather than
> `data:`
> `egress: "true"`
> ...
> `permissive_traffic_policy_mode: "true"`

I wouldn't necessarily call this an issue given how `helm upgrade` handles values. Multiple flags can be added to a single `osm update` command and they should all take effect, so in this case I would do a single `osm update --enable-permissive-traffic-policy --enable-egress`.

> @nitishm Seems that after enabling a flag, you're unable to disable it
> `osm update --enable-egress true`
> will return in configMap
> `data:`
> `egress: "true"`
> but `osm update --enable-egress false` will still show in configMap as
> `data:`
> `egress: "true"`

This looks like a `cobra` quirk: https://github.com/spf13/cobra/issues/613. I would expect `osm update --enable-egress=false` to work (or implicitly by omitting the flag entirely).

> Either the user has to specify every flag in the command (which is unwieldy and affects the UX) or we need a different way of differentiating which fields are being updated vs the ones that should carry the current value forward.

Is the main question here whether omitting a flag should mean "take the CLI's default value" vs. "take the value of the existing release?" I don't have a strong opinion, so I would favor whatever is simplest to implement and document. `helm upgrade --reuse-values` looks helpful here, but I worry about that causing issues while the chart and its values are still changing quite frequently since it effectively ignores the new values.yaml in favor of the old one.

PR #1631, opened 2020-08-27
ref(pkg/envoy/route): clean up constants

michelleN:
* removed unused constants
* update case for constants to match comment

- New Functionality      [ ]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [X]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests / CI System      [ ]
- Other                  [ ]

### DIFF
--- a/pkg/envoy/route/config.go
+++ b/pkg/envoy/route/config.go
@@ -35,20 +35,14 @@ const (
 	//OutboundRouteConfigName is the name of the route config that the envoy will identify
 	OutboundRouteConfigName = "RDS_Outbound"
 
-	// maxRegexProgramSize is the max supported regex complexity
-	maxRegexProgramSize = 1024
-
-	//InboundVirtualHost is the name of the virtual host on the inbound route configuration
+	// inboundVirtualHost is the name of the virtual host on the inbound route configuration
 	inboundVirtualHost = "inbound_virtualHost"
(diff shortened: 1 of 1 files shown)


pkg/envoy/route/config.go
```diff
@@ -29,27 +29,21 @@ const (
...
+	// InboundVirtualHost is the name of the virtual host on the inbound route configuration
+	InboundVirtualHost = "inbound_virtualHost"
 
-	//InboundVirtualHost is the name of the virtual host on the inbound route configuration
-	inboundVirtualHost = "inbound_virtualHost"
-
-	//OutboundVirtualHost is the name of the virtual host on the outbound route configuration
-	outboundVirtualHost = "outbound_virtualHost"
+	// OutboundVirtualHost is the name of the virtual host on the outbound route configuration
+	OutboundVirtualHost = "outbound_virtualHost"
```
shashankram, line 42: Why are these 2 constants being exported with this change? The first letter should be lower case to avoid exporting these.

michelleN: They shouldn't be. I changed them back and updated the comments instead. thanks.

PR #1645, opened 2020-08-28
catalog: add an API to retrieve endpoints for service FQDN

eduser25:
One of our ongoing proposals in order to implement per-service filtering
at listener level will require being able to identify the traffic destination
of an ongoing request.

For that to be able to happen, we need to be able to do L3/L4 mapping to
a service.

In cases where the service is behind a cluster IP (in K8s), the FQDN resolvable IP is that
of the Cluster IP, henceforth we will require to get the Cluster IP from a specific
service when programming the outbound listener.

This commit introduces a new API called GetResolvableServiceEndpoints,
which post-conditions to return the expected resolvable endpoint where a service
is made available from a client application perspective.

To properly test, needs some larger changes that I will follow up
with on a commit on top.

- Control Plane          [X]

No

### DIFF
--- a/pkg/catalog/endpoint.go
+++ b/pkg/catalog/endpoint.go
@@ -21,7 +21,7 @@ func (mc *MeshCatalog) ListEndpointsForService(svc service.MeshService) ([]endpo
 	return endpoints, nil
 }
 
-// GetResolvableServiceEndpoints returns the resolvable set of endpoint destinations where the service is made available at.
+// GetResolvableServiceEndpoints returns the resolvable set of endpoint over which a service is accessible using its FQDN
 func (mc *MeshCatalog) GetResolvableServiceEndpoints(svc service.MeshService) ([]endpoint.Endpoint, error) {
 	// TODO: Move the implmentation of this function to be provider-specific. Currently, the providers might
 	// not have access to some common structures in order to perform these operations in an optimal way
@@ -38,19 +38,20 @@ func (mc *MeshCatalog) GetResolvableServiceEndpoints(svc service.MeshService) ([
 	if len(service.Spec.ClusterIP) == 0 {
 		// If no cluster IP, use final endpoint as resolvable destinations
 		endpoints, err = mc.ListEndpointsForService(svc)
-	} else {
-		ip := net.ParseIP(service.Spec.ClusterIP)
-		if ip == nil {
-			log.Error().Msgf("Could not parse Cluster IP %s", service.Spec.ClusterIP)
-			return nil, errParseClusterIP
-		}
+	}
 
-		for _, svcPort := range service.Spec.Ports {
-			endpoints = append(endpoints, endpoint.Endpoint{
-				IP:   ip,
-				Port: endpoint.Port(svcPort.Port),
-			})
-		}
+	// Cluster IP is present
+	ip := net.ParseIP(service.Spec.ClusterIP)
+	if ip == nil {
+		log.Error().Msgf("Could not parse Cluster IP %s", service.Spec.ClusterIP)
+		return nil, errParseClusterIP
+	}
+
+	for _, svcPort := range service.Spec.Ports {
+		endpoints = append(endpoints, endpoint.Endpoint{
+			IP:   ip,
+			Port: endpoint.Port(svcPort.Port),
+		})
 	}
 
 	return endpoints, err
--- a/pkg/catalog/endpoint.go
+++ b/pkg/catalog/endpoint.go
@@ -21,25 +21,28 @@ func (mc *MeshCatalog) ListEndpointsForService(svc service.MeshService) ([]endpo
 	return endpoints, nil
 }
 
-// GetServiceEndpoints returns the highest abstract set of endpoint destinations where the service is made available at.
-// If no LB/virtual IPs are assigned to the service, GetServiceEndpoints will return ListEndpointsForService
-func (mc *MeshCatalog) GetServiceEndpoints(svc service.MeshService) ([]endpoint.Endpoint, error) {
+// GetResolvableServiceEndpoints returns the resolvable set of endpoint destinations where the service is made available at.
+func (mc *MeshCatalog) GetResolvableServiceEndpoints(svc service.MeshService) ([]endpoint.Endpoint, error) {
+	// TODO: Move the implmentation of this function to be provider-specific. Currently, the providers might
+	// not have access to some common structures in order to perform these operations in an optimal way
 	var endpoints []endpoint.Endpoint
 	var err error
 
 	// Check if the service has been given Cluster IP
-	// TODO: push this in providers. Providers currently do not have services cache.
 	service := mc.GetSMISpec().GetService(svc)
 	if service == nil {
 		log.Error().Msgf("Could not find service %s", svc.String())
-		return []endpoint.Endpoint{}, errServiceNotFound
+		return nil, errServiceNotFound
 	}
 
-	if len(service.Spec.ClusterIP) > 0 {
+	if len(service.Spec.ClusterIP) == 0 {
+		// If no cluster IP, use final endpoint as resolvable destinations
+		endpoints, err = mc.ListEndpointsForService(svc)
+	} else {
 		ip := net.ParseIP(service.Spec.ClusterIP)
 		if ip == nil {
-			log.Error().Msgf("Could not parse IP %s", service.Spec.ClusterIP)
-			return []endpoint.Endpoint{}, errParseClusterIP
+			log.Error().Msgf("Could not parse Cluster IP %s", service.Spec.ClusterIP)
+			return nil, errParseClusterIP
 		}
 
 		for _, svcPort := range service.Spec.Ports {
@@ -48,8 +51,6 @@ func (mc *MeshCatalog) GetServiceEndpoints(svc service.MeshService) ([]endpoint.
 				Port: endpoint.Port(svcPort.Port),
 			})
 		}
-	} else {
-		endpoints, err = mc.ListEndpointsForService(svc)
 	}
 
 	return endpoints, err
--- a/pkg/catalog/endpoint.go
+++ b/pkg/catalog/endpoint.go
@@ -37,7 +37,7 @@ func (mc *MeshCatalog) GetResolvableServiceEndpoints(svc service.MeshService) ([
 
 	if len(service.Spec.ClusterIP) == 0 {
 		// If no cluster IP, use final endpoint as resolvable destinations
-		endpoints, err = mc.ListEndpointsForService(svc)
+		return mc.ListEndpointsForService(svc)
 	}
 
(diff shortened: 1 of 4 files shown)


pkg/catalog/endpoint.go
```diff
@@ -18,3 +20,37 @@ func (mc *MeshCatalog) ListEndpointsForService(svc service.MeshService) ([]endpo
...
+func (mc *MeshCatalog) GetServiceEndpoints(svc service.MeshService) ([]endpoint.Endpoint, error) {
+	var endpoints []endpoint.Endpoint
+	var err error
+
+	// Check if the service has been given Cluster IP
+	// TODO: push this in providers. Providers currently do not have services cache.
+	service := mc.GetSMISpec().GetService(svc)
+	if service == nil {
+		log.Error().Msgf("Could not find service %s", svc.String())
+		return []endpoint.Endpoint{}, errServiceNotFound
```
shashankram, line 35:
```suggestion
		return nil, errServiceNotFound
```

pkg/catalog/endpoint.go
```diff
@@ -18,3 +20,37 @@ func (mc *MeshCatalog) ListEndpointsForService(svc service.MeshService) ([]endpo
...
+	}
+
+	if len(service.Spec.ClusterIP) > 0 {
+		ip := net.ParseIP(service.Spec.ClusterIP)
+		if ip == nil {
+			log.Error().Msgf("Could not parse IP %s", service.Spec.ClusterIP)
+			return []endpoint.Endpoint{}, errParseClusterIP
```
shashankram, line 42:
```suggestion
			return nil, errParseClusterIP
```

pkg/catalog/endpoint.go
```diff
@@ -18,3 +20,37 @@ func (mc *MeshCatalog) ListEndpointsForService(svc service.MeshService) ([]endpo
...
+
+	// Check if the service has been given Cluster IP
+	// TODO: push this in providers. Providers currently do not have services cache.
+	service := mc.GetSMISpec().GetService(svc)
+	if service == nil {
+		log.Error().Msgf("Could not find service %s", svc.String())
+		return []endpoint.Endpoint{}, errServiceNotFound
+	}
+
+	if len(service.Spec.ClusterIP) > 0 {
```
shashankram, line 38: Can we invert this condition to:
```
    if no cluster ip:
        return ListEndpointsForSercice()

   // parse cluster ip and return
```

pkg/catalog/types.go
```diff
@@ -70,9 +70,13 @@ type MeshCataloger interface {
 	// ListSMIPolicies lists SMI policies.
 	ListSMIPolicies() ([]*split.TrafficSplit, []service.WeightedService, []service.K8sServiceAccount, []*spec.HTTPRouteGroup, []*target.TrafficTarget, []*corev1.Service)
 
-	// ListEndpointsForService returns the list of provider endpoints corresponding to a service
+	// ListEndpointsForService returns the list of individual instance endpoint destinations that comprise the backed service
 	ListEndpointsForService(service.MeshService) ([]endpoint.Endpoint, error)
 
+	// GetServiceEndpoint returns the highest abstract set of endpoint destinations where the service is made available at.
+	// If no LB/virtual IPs are assigned to the service, GetServiceEndpoints will return ListEndpointsForService
+	GetServiceEndpoints(service.MeshService) ([]endpoint.Endpoint, error)
```
shashankram, line 78: This would clear the ambiguity with `ListEndpointsForService`, where we attempt to convey that `GetResolvableEndpointsForService` returns only endpoints that are resolvable using service FQDN.

```suggestion
	GetResolvableEndpointsForService(service.MeshService) ([]endpoint.Endpoint, error)
```

pkg/catalog/endpoint.go
```diff
@@ -18,3 +20,37 @@ func (mc *MeshCatalog) ListEndpointsForService(svc service.MeshService) ([]endpo
 	}
 	return endpoints, nil
 }
+
+// GetServiceEndpoints returns the highest abstract set of endpoint destinations where the service is made available at.
+// If no LB/virtual IPs are assigned to the service, GetServiceEndpoints will return ListEndpointsForService
+func (mc *MeshCatalog) GetServiceEndpoints(svc service.MeshService) ([]endpoint.Endpoint, error) {
```
nojnhuh, line 26: Is this something that should work the same for any `MeshCataloger` implementation? If so, does it make sense to make this a function that takes a `MeshCataloger` as a parameter vs. implementing this as a method on the struct since it only calls methods that are already part of the interface? 

i.e.:
```go
func GetServiceEndpoints(mc MeshCataloger, svc service.MeshService) ([]endpoint.Endpoint, error) {
```

That way we don't have to add another method to the interface and other implementations could take advantage of this function.

eduser25:
I missed explaining purpose for this API.
Addressing all, plus adding more context and explanation to the change.

pkg/catalog/types.go
eduser25: +1, much better

pkg/catalog/endpoint.go
eduser25: tl'dr: providers don't have easy access to service's data, this is an easy-to-fix implementation that can go away as soon as providers can get services, which we'd propose as a refactor that can happen parallely while this commit unblocks a lot of other work that depends on it.

---

what you said is 100% correct. Your suggestions are our original plan, we did reason all that, we however saw some problems with the implementation that are trying to circumvent. 

Originally the implementation of this api should be offloaded to endpoint provider, as having cluster/virtual/proxy/lb/etc.... IP depends on the provider of the service, and the current implementation is leaking k8s stuff into the catalog.

However, in the case of K8s provider, the provider would either (1) require access to the services cache (one under SMIspec right now)  or (2) have it's own service cache or (3) issue API call to get service details (to get the Cluster IP).

More or less, we deemeed that since:
- Endpoint providers do not have easy access to the available service cache under SMIspec, unless we refactor and introduce new shared constructs not currently present (that's some major refactor probs)
- Creating a new service cache for endpoint providers is far from optimal
- issues a new API call for each service seems a no-go (even though to some degree, this might already happen)
and
- the only actual provider currently working and being offered is K8s

we propose this implementations which leaks some k8s in the catalog, but is deffinitely easy to clean up and pushed down to providers for them to implement they can query a services cache.

pkg/catalog/endpoint.go
```diff
@@ -18,3 +20,38 @@ func (mc *MeshCatalog) ListEndpointsForService(svc service.MeshService) ([]endpo
 	}
 	return endpoints, nil
 }
+
+// GetResolvableServiceEndpoints returns the resolvable set of endpoint destinations where the service is made available at.
```
shashankram, line 24:
```suggestion
// GetResolvableServiceEndpoints returns the resolvable set of endpoint over which a service is accessible using its FQDN
```

pkg/catalog/endpoint.go
```diff
@@ -18,3 +20,38 @@ func (mc *MeshCatalog) ListEndpointsForService(svc service.MeshService) ([]endpo
...
+	service := mc.GetSMISpec().GetService(svc)
+	if service == nil {
+		log.Error().Msgf("Could not find service %s", svc.String())
+		return nil, errServiceNotFound
+	}
+
+	if len(service.Spec.ClusterIP) == 0 {
+		// If no cluster IP, use final endpoint as resolvable destinations
+		endpoints, err = mc.ListEndpointsForService(svc)
+	} else {
```
shashankram, line 41: Can we change this to make it less nested.
```
if no clusterIP:
    return ListEndpointsForService(svc)

// other stuff
...
return endpoints, nil
```

pkg/catalog/types.go
```diff
@@ -70,9 +70,15 @@ type MeshCataloger interface {
 	// ListSMIPolicies lists SMI policies.
 	ListSMIPolicies() ([]*split.TrafficSplit, []service.WeightedService, []service.K8sServiceAccount, []*spec.HTTPRouteGroup, []*target.TrafficTarget, []*corev1.Service)
 
-	// ListEndpointsForService returns the list of provider endpoints corresponding to a service
+	// ListEndpointsForService returns the list of individual instance endpoint destinations that comprise the backed service
```
shashankram, line 73:
```suggestion
	// ListEndpointsForService returns the list of individual endpoints backing a service
```

pkg/catalog/types.go
```diff
@@ -70,9 +70,15 @@ type MeshCataloger interface {
...
 	ListEndpointsForService(service.MeshService) ([]endpoint.Endpoint, error)
 
+	// GetResolvableServiceEndpoints returns the resolvable set of endpoint destinations where the service is made available at.
```
shashankram, line 76:
```suggestion
	// GetResolvableServiceEndpoints returns the resolvable set of endpoint over which a service is accessible using its FQDN
```

pkg/catalog/endpoint.go
```diff
@@ -38,19 +38,20 @@ func (mc *MeshCatalog) GetResolvableServiceEndpoints(svc service.MeshService) ([
 	if len(service.Spec.ClusterIP) == 0 {
 		// If no cluster IP, use final endpoint as resolvable destinations
 		endpoints, err = mc.ListEndpointsForService(svc)
```
shashankram, line 40: return here.

pkg/catalog/endpoint.go
```diff
@@ -18,3 +20,39 @@ func (mc *MeshCatalog) ListEndpointsForService(svc service.MeshService) ([]endpo
 	}
 	return endpoints, nil
 }
+
+// GetResolvableServiceEndpoints returns the resolvable set of endpoint over which a service is accessible using its FQDN
+func (mc *MeshCatalog) GetResolvableServiceEndpoints(svc service.MeshService) ([]endpoint.Endpoint, error) {
+	// TODO: Move the implmentation of this function to be provider-specific. Currently, the providers might
```
shashankram, line 26: I would raise an issue for this and reference it in the comment.

pkg/catalog/endpoint_test.go
```diff
@@ -21,4 +21,17 @@ var _ = Describe("Test catalog functions", func() {
 			Expect(actual).To(Equal(expected))
 		})
 	})
+
+	Context("Testing GetResolvableServiceEndpoints()", func() {
+		It("returns the endpoint for the service", func() {
+			actual, err := mc.GetResolvableServiceEndpoints(tests.BookstoreService)
```
shashankram, line 27: Could you add test for the case where the service does not have cluster IP set but has endpoints backing it?

eduser25: I have few on stash. Working on it.

pkg/catalog/endpoint.go
eduser25: https://github.com/openservicemesh/osm/issues/1655

PR #1654, opened 2020-09-01
add --set to the osm cli

### LINKED ISSUE #1644: Add flag to `osm install` to override arbitrary chart values
Currently, whenever a new chart parameter is created, it must be surfaced as a flag to `osm install` to be controllable from the CLI. Adding a generic `--set` or similar flag like the one on [`helm install`](https://helm.sh/docs/helm/helm_install/) that takes a key/value pair would allow values to still be controlled from the CLI without requiring any additional code changes.

**Possible use cases**

- Rarely used or temporary values
- Testing alpha-level features controlled by a feature flag without having to add an explicitly hidden CLI flag.

rajdhandus:
Adds ability to pass in arbitrary arguments to osm cli via the --set arg
and resolves #1644

- New Functionality      [ x]
- Documentation          [ ]
- Install                [x ]
- Control Plane          [ ]
- CLI Tool               [x ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests / CI System      [ ]
- Other                  [ ]

no & no


cmd/cli/install.go
```diff
@@ -148,17 +150,24 @@ func newInstallCmd(config *helm.Configuration, out io.Writer) *cobra.Command {
...
 
 	return cmd
 }
 
 func (i *installCmd) run(config *helm.Configuration) error {
+
+	parsedArgs, err := i.parseSetOptions()
+	if err != nil {
+		return err
+	}
```
nojnhuh, line 163: Can we move this into `resolveValues()` so that wraps all of the value parsing/merging? That would also eliminate the need to pass in an argument which might clean up the tests a bit too.

cmd/cli/install.go
```diff
@@ -189,7 +198,7 @@ func (i *installCmd) loadOSMChart() error {
 	return nil
 }
 
-func (i *installCmd) resolveValues() (map[string]interface{}, error) {
+func (i *installCmd) resolveValues(parsedArgs []string) (map[string]interface{}, error) {
 	finalValues := map[string]interface{}{}
 	valuesConfig := []string{
 		fmt.Sprintf("OpenServiceMesh.image.registry=%s", i.containerRegistry),
```
nojnhuh, line 204: I think if `valuesConfig` was `i.setOptions` + all of the existing `fmt.Sprintf` calls for the explicit flags (in that order), then you wouldn't have to manually override the value from `--set` with the explicit flag since Helm would do that based on the order (not 100% sure about that). That would eliminate the need for `overrideLegacyArgs()` where the variable/chart value association is duplicated, and maybe even `parseSetOptions()`.

cmd/cli/install.go
```diff
@@ -148,17 +150,24 @@ func newInstallCmd(config *helm.Configuration, out io.Writer) *cobra.Command {
 	f.BoolVar(&inst.enableMetricsStack, "enable-metrics-stack", true, "Enable metrics (Prometheus and Grafana) deployment")
 	f.StringVar(&inst.meshName, "mesh-name", defaultMeshName, "name for the new control plane instance")
 	f.BoolVar(&inst.deployJaeger, "deploy-jaeger", true, "Deploy Jaeger in the namespace of the OSM controller")
+	f.StringArrayVar(&inst.setOptions, "set", []string{}, "set values on the command line (can specify multiple or separate values with commas: key1=val1, key2=val2)")
```
nojnhuh, line 153: > `key1=val1, key2=val2`

I'm not certain a space between these will work. Not a big deal if it doesn't, but I would take that out of the help message just to make it a bit clearer. I see the help message for Helm's `--set` doesn't include the space.

cmd/cli/install_test.go
```diff
@@ -979,9 +1091,10 @@ var _ = Describe("Test osm image pull policy cli option", func() {
 		installCmd := &installCmd{
 			osmImagePullPolicy: "Always",
 		}
-
-		vals, err := installCmd.resolveValues()
+		parsedSetOptions, parseErr := installCmd.parseSetOptions()
+		vals, err := installCmd.resolveValues(parsedSetOptions)
 		Expect(err).NotTo(HaveOccurred())
+		Expect(parseErr).NotTo(HaveOccurred())
```
nojnhuh, line 1098: I haven't taken a super close look at each of these tests, but can you verify

1. There's at least one test where no `--set` options are passed
1. There's at least one test for the comma-separated case like in the help message for the flag

nojnhuh (dismissed):
Thanks @rajdhandus!

I left a few comments inline. Mostly things to double-check or try out to see if it simplifies things a bit.

cmd/cli/install.go
rajdhandus: I had originally coded it this way primarily to let `validateOptions()` act on the merged arguments... I have incorporated your comments in a different way to remove the argument and clean up the tests.. please check and let me know..

cmd/cli/install.go
rajdhandus: Same here.. `validateOptions` operates on the instance variables.. whereas `valuesConfig` array has formatted strings with both key and value combines... validating such a format makes it tricky... let me know

cmd/cli/install.go
rajdhandus: thanks.. taken care of..

cmd/cli/install_test.go
rajdhandus: #1 was already take care of..  for #2 I have added new test case

cmd/cli/install.go
nojnhuh: Makes sense. I'm not too concerned with validating options from `--set` at the moment since any value that requires validation has its own flag that users should use instead of `--set`. We can prioritize that if we find people are running into issues because they're using `--set` with invalid values.

@michelleN Would you still like to see `resolveValues()` be a separate step from `validateOptions()`?

cmd/cli/install.go
```diff
@@ -225,9 +234,114 @@ func (i *installCmd) resolveValues() (map[string]interface{}, error) {
 			return nil, err
 		}
 	}
+
+	for _, val := range parsedArgs {
+		// parses Helm strvals line and merges into a map for the final overrides for values.yaml
+		if err := strvals.ParseInto(val, finalValues); err != nil {
+			return nil, err
+		}
+	}
```
nojnhuh, line 243: @rajdhandus Just checking in to see if you're still working on this.

I think it would simplify things significantly if you iterated over `i.setOptions` here directly vs. parsing them manually first. I know that would mean those values don't get validated but I think that's a worthwhile tradeoff. In the event that a value is defined both with an explicit flag and `--set`, I think the explicit flag should take precedence as that will be validated. We should make that order of precedence clear in the `--help` and that `--set` should only be used when an explicit flag is not available.

nojnhuh:
Hey @rajdhandus, are you still planning to work on this? If not, I'll plan to take what you've done here so far and finish it up.

nojnhuh:
I made a couple small changes to this:

* Refactor how the values are merged before passing them to Helm so that `--set` only works for chart values that can't be specified with another flag, e.g. `--set OpenServiceMesh.meshName=name` will behave as if no `--mesh-name` was passed. This will prevent values from circumventing any custom validation tied to the value of the flag. I tried to make this behavior clear in the help text for the `--set`, flag, but any suggestions to improve it are very much welcome.

* Refactored `TestResolveValues()` to be table-driven, added tests equivalent to the Ginkgo-based ones initially added here.

cmd/cli/install.go
```diff
@@ -184,6 +185,7 @@ func newInstallCmd(config *helm.Configuration, out io.Writer) *cobra.Command {
 	f.StringVar(&inst.envoyLogLevel, "envoy-log-level", defaultEnvoyLogLevel, "Envoy log level is used to specify the level of logs collected from envoy and needs to be one of these (trace, debug, info, warning, warn, error, critical, off)")
 	f.BoolVar(&inst.enforceSingleMesh, "enforce-single-mesh", defaultEnforceSingleMesh, "Enforce only deploying one mesh in the cluster")
 	f.DurationVar(&inst.timeout, "timeout", 5*time.Minute, "Time to wait for installation and resources in a ready state, zero means no timeout")
+	f.StringArrayVar(&inst.setOptions, "set", nil, "set arbitrary chart values values not settable by another flag (can specify multiple or separate values with commas: key1=val1,key2=val2)")
```
ksubrmnn, line 188: Typo: values values

Also please capitalize this: set -> Set

cmd/cli/install.go
```diff
@@ -229,6 +231,14 @@ func (i *installCmd) loadOSMChart() error {
 
 func (i *installCmd) resolveValues() (map[string]interface{}, error) {
 	finalValues := map[string]interface{}{}
+
```
ksubrmnn, line 234: Can we validate that the input is in the correct format? Or does strvals do that for us (and return a good error message)?

ksubrmnn (changes requested):

cmd/cli/install.go
nojnhuh: Done

cmd/cli/install.go
nojnhuh: `strvals` will do that. It gives reasonable errors, but I wrapped the ones here with a hint that they came from `--set`. I also added a unit test that shows an example error.

ksubrmnn (approved):

eduser25 (approved):

michelleN (approved):
Thank you

Issue #1655, opened 2020-09-01
Move GetResolvableServiceEndpoints to be endpoint-provider specifc


eduser25:
https://github.com/openservicemesh/osm/pull/1645 introduced `GetResolvableServiceEndpoints` on the catalog, which attempts to provide a provider-agnostic way to get a resolvable destination for a Service on the mesh. What a "Resolvable Endpoint" is, and "how to get it" is after all dependent on the provider logically speaking.

The implementation however is not provided by an `endpoint provider` and instead is leaking k8s specific into the catalog. The reason it was done like this was primarily because providers do not have easy access to a Services cache, and the currently available service cache is settled under `SMIspec` under `catalog`.

Some possible or alternative solutions (with unweighted pros/cons)
- Make the Services cache/API under SMIspec accessible/shareable
- Allow providers to have their own Services cache
- Have providers get Service information via API calls

More context: https://github.com/openservicemesh/osm/pull/1645#discussion_r479540675

**Scope (please mark with X where applicable)**
- Envoy Control Plane        [X]

**Possible use cases**
Clean up on unclean interfaces, so providers can better fit in the overall architecture

eduser25:
fixed long ago https://github.com/openservicemesh/osm/pull/2191

PR #1663, opened 2020-09-03
Move Namespace controller implementation under k8sclient

eduser25:
We've assessed multiple problems related to performance, where
several xDS vertical paths might incur in meaningful delays to perform
certain operations due to API/REST calling under the hood.

These calls, though legitimate, are a product of some of the design
boundaries that the modules have set between the themselves.

In an effort to coalesce how multiple packages access wellknown K8s
resources - and very specifically emphasizing going through async/cache models so we
don't incur in additional API calls on operations - we propose to move some of
the major K8s accessors to a common client/class/interface that can
be passed or accessed by the modules that will need it.

Incoming commit order:
(1) Move Namespaces cache and monitor into k8sclient (90% of the code used
is shared with any other monitor)
(2) Create Services cache and monitor into k8sclient, push the interface
so SMISpec and Endpoint providers can use it

Other pending issues that are happening because of the above:
(3) perf degradation. The XDS verticals are taking a very long time because of
the API calls.
(4) https://github.com/openservicemesh/osm/issues/1655
....

- Control Plane          [X]
- Other                  [X]

No


cmd/osm-controller/osm-controller.go
```diff
@@ -140,8 +140,8 @@ func main() {
 	}
 	log.Info().Msgf("Initial ConfigMap %s: %s", osmConfigMapName, string(configMap))
 
-	namespaceController := namespace.NewNamespaceController(kubeClient, meshName, stop)
-	meshSpec, err := smi.NewMeshSpecClient(*smiKubeConfig, kubeClient, osmNamespace, namespaceController, stop)
+	k8sclient := k8s.NewK8sClient(kubeClient, meshName, stop)
```
shashankram, line 143: I'd prefer renaming `k8s` to `kube` throughout this change.

pkg/kubernetes/client.go
```diff
@@ -11,78 +9,89 @@ import (
...
-	k8s "github.com/openservicemesh/osm/pkg/kubernetes"
 )
 
-var (
-	resyncPeriod = 30 * time.Second
-)
+// NewK8sClient returns a new Client which means to provide access to locally-cached k8s resources
+func NewK8sClient(kubeClient kubernetes.Interface, meshName string, stop chan struct{}) Client {
+	// Initialize client object
+	cl := Client{
```
shashankram, line 17: nit: s/cl/client/g

pkg/kubernetes/client.go
```diff
@@ -11,78 +9,89 @@ import (
...
+
+	return cl
 }
 
-// NewNamespaceController implements namespace.Controller and creates the Kubernetes client to manage namespaces.
-func NewNamespaceController(kubeClient kubernetes.Interface, meshName string, stop chan struct{}) Controller {
-	// Only monitor namespaces that are labeled with this OSM's mesh name
-	monitorNamespaceLabel := map[string]string{constants.OSMKubeResourceMonitorAnnotation: meshName}
+// Initializes Namespace monitoring
+func initNamespaceMonitor(c *Client) {
```
shashankram, line 36:
```suggestion
func (c *Client) initNamespaceMonitor() {
```

pkg/kubernetes/client.go
```diff
@@ -11,78 +9,89 @@ import (
...
 
-	if c.informer == nil {
+	if c.informers == nil {
 		return errInitInformers
 	}
 
-	go c.informer.Run(stop)
-	log.Info().Msgf("Waiting namespace.Monitor informer cache sync")
-	if !cache.WaitForCacheSync(stop, c.informer.HasSynced) {
+	for k, v := range c.informers {
```
shashankram, line 62:
```suggestion
	for name, informer := range c.informers {
```

pkg/kubernetes/types.go
```diff
@@ -27,10 +29,28 @@ const (
...
 )
 
 // Event is the combined type and actual object we received from Kubernetes
 type Event struct {
 	Type  EventType
 	Value interface{}
 }
+
+// InformerCollection stores the different Informers we keep for K8s resources
+type InformerCollection map[string]cache.SharedIndexInformer
```
shashankram, line 47: To prevent incorrect lookups from the wrong cache, I'd change it to how this is in `smi/types.go`
```
// InformerCollection is a struct of the Kubernetes informers used in OSM
type InformerCollection struct {
	Services       cache.SharedIndexInformer
	TrafficSplit   cache.SharedIndexInformer
	HTTPRouteGroup cache.SharedIndexInformer
	TCPRoute       cache.SharedIndexInformer
	TrafficTarget  cache.SharedIndexInformer
	Backpressure   cache.SharedIndexInformer
}

// CacheCollection is a struct of the Kubernetes caches used in OSM
type CacheCollection struct {
	Services       cache.Store
	TrafficSplit   cache.Store
	HTTPRouteGroup cache.Store
	TCPRoute       cache.Store
	TrafficTarget  cache.Store
	Backpressure   cache.Store
}
```

shashankram (commented):
Looks good overall, few comments inline.

cmd/osm-controller/osm-controller.go
eduser25: `kubeclient` already exists... hence the `k8s`. I do accept other suggestions though.

shashankram: This is okay then. Should avoid package aliasing to k8s unless there is a conflict.

pkg/kubernetes/errors.go
```diff
@@ -0,0 +1,9 @@
+package kubernetes
+
+import "github.com/pkg/errors"
+
+var (
+	errSyncingCaches     = errors.New("Failed initial cache sync for Namespace informers")
+	errInitInformers     = errors.New("Namespace informers are not initialized")
```
shashankram, line 7: In light of this informer not being specific to `Namespaces`, and because additional context can be a part of the log message:
```suggestion
	errInitInformers     = errors.New("Informer not initialized")
```

pkg/namespace/errors.go
```diff
@@ -1,9 +1,3 @@
 package namespace
 
-import "github.com/pkg/errors"
-
-var (
-	errSyncingCaches     = errors.New("Failed initial cache sync for Namespace informers")
-	errInitInformers     = errors.New("Namespace informers are not initialized")
-	errListingNamespaces = errors.New("Failed to list monitored namespaces")
-)
+var ()
```
shashankram, line 3: This file should be removed.

shashankram (commented):
Is there a reason `pkg/namespace` needs to exist? We could even move the interface definition under `pkg/kubernetes`.

eduser25:
i actually want to do the move, but wanted to make it on separate commits, to keep the changes not too large.

shashankram (approved):

draychev (approved):

PR #1694, opened 2020-09-11
Debugger tests fixes/improvements

eduser25:
- Removed fakes, added mocks for pkg/debugger interface types
- Moved available tests to use mocks
- Added tests for debugger/server and debugger/certificate
- Created/rebased tests to use `testing` instead of `Ginko`

`7.7%` -> `36.0%`

- Tests / CI System      [X]

No


eduser25:
Will rebase w/o using Ginko

ksubrmnn (approved):

snehachhabria (dismissed):

pkg/debugger/policy_test.go
```diff
@@ -1,71 +1,38 @@
 package debugger
 
 import (
-	"fmt"
 	"net/http/httptest"
 	"testing"
-	"time"
 
+	gomock "github.com/golang/mock/gomock"
```
shashankram, line 7: we don't need an import alias here.

shashankram (approved):

snehachhabria (approved):

pkg/debugger/policy_test.go
eduser25: fixing

shashankram: Don't worry about it, can be handled later. We have other occurrences of this.

ksubrmnn (approved):

PR #1696, opened 2020-09-15
cmd/certificate: handle error scenarios when fetching secret

shashankram:
This change refactors the `getCertFromKubernetes` function to
handle error scenarios. The function is only expected to error
if it finds a secret with an invalid root certificate. It also
helps remove the call to `log.Fatal` while parsing the root
certificate.

Also updates the existing test context (Ginkgo based) to test error scenarios.

- New Functionality      [ ]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [ ]
- CLI Tool               [ ]
- Certificate Management [X]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests / CI System      [ ]
- Other                  [ ]

`No`

### DIFF
--- a/cmd/osm-controller/certificates.go
+++ b/cmd/osm-controller/certificates.go
@@ -48,7 +48,11 @@ func getTresorOSMCertificateManager(kubeClient kubernetes.Interface, enableDebug
 	// load the CA from the given k8s secret within the namespace where OSM is install.d
 	// An empty string or nil value would not load or save/load CA.
 	if caBundleSecretName != "" {
-		rootCert = getCertFromKubernetes(kubeClient, osmNamespace, caBundleSecretName)
+		rootCert, err = getCertFromKubernetes(kubeClient, osmNamespace, caBundleSecretName)
+		if err != nil {
+			log.Error().Err(err).Msgf("Error retrieving root certificate from secret %s/%s", osmNamespace, caBundleSecretName)
+			return nil, nil, err
+		}
 	}
 
 	if rootCert == nil {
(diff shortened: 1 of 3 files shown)


cmd/osm-controller/certificates.go
```diff
@@ -75,57 +79,47 @@ func getTresorOSMCertificateManager(kubeClient kubernetes.Interface, enableDebug
...
+// getCertFromKubernetes returns a Certificater type corresponding to the root certificate.
+// The function returns an error only if a secret is found with invalid data.
+func getCertFromKubernetes(kubeClient kubernetes.Interface, namespace, secretName string) (certificate.Certificater, error) {
 	rootCertSecret, err := kubeClient.CoreV1().Secrets(namespace).Get(context.Background(), secretName, metav1.GetOptions{})
 	if err != nil {
-		log.Warn().Msgf("Error retrieving root certificate rootCertSecret %q from namespace %q; Will create a new one", secretName, osmNamespace)
-		return nil
+		// It is okay for this secret to be missing, in which case a new CA will be created along with a k8s secret
+		log.Debug().Msgf("Could not retrieve root certificate secret %q from namespace %q", secretName, osmNamespace)
+		return nil, nil
```
snehachhabria, line 89:
```suggestion
		return nil, err
```

snehachhabria: Why don't you want to error in this scenario as well?

shashankram: This isn't an error, check the function description and existing behavior.

shashankram: Also comment on L87.

snehachhabria (approved):

cmd/osm-controller/certificates.go
shashankram: The reason this isn't an error because the Get() api is used to only check if the secret exists, if it doesn't its ok (no need to log an error because its not an error from the controller's perspective). This is more of a helper function to parse the root certificate from the secret, the caller decides how to use the returned root cert (and creates a root cert secret if it doesn't exist).

eduser25 (approved):

PR #1705, opened 2020-09-16
kubernetes/Controller: Fix return type during initialization

shashankram:
Fixes the return value and signature when a new Kubernetes
Controller is being initialized by returning the interface
and an error.

- New Functionality      [ ]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [ ]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests / CI System      [ ]
- Other                  [X]

`No`

### DIFF
--- a/pkg/kubernetes/client.go
+++ b/pkg/kubernetes/client.go
@@ -14,8 +14,8 @@ import (
 	"github.com/openservicemesh/osm/pkg/service"
 )
 
-// NewKubernetesClient returns a new Client which means to provide access to locally-cached k8s resources
-func NewKubernetesClient(kubeClient kubernetes.Interface, meshName string, stop chan struct{}) Client {
+// NewKubernetesClient returns a new kubernetes.Controller which means to provide access to locally-cached k8s resources
+func NewKubernetesClient(kubeClient kubernetes.Interface, meshName string, stop chan struct{}) (Controller, error) {
 	// Initialize client object
 	client := Client{
 		kubeClient:    kubeClient,
@@ -30,10 +30,11 @@ func NewKubernetesClient(kubeClient kubernetes.Interface, meshName string, stop 
 	client.initServicesMonitor()
 
 	if err := client.run(stop); err != nil {
-		log.Fatal().Err(err).Msg("Could not start Kubernetes Namespaces client")
+		log.Error().Err(err).Msg("Could not start Kubernetes Namespaces client")
+		return nil, err
 	}
 
-	return client
+	return client, nil
 }
 
 // Initializes Namespace monitoring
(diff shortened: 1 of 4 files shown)


pkg/smi/client_test.go
```diff
@@ -45,7 +45,10 @@ func bootstrapClient() (MeshSpec, *fakeKubeClientSet, error) {
 	smiTrafficSpecClientSet := testTrafficSpecClient.NewSimpleClientset()
 	smiTrafficTargetClientSet := testTrafficTargetClient.NewSimpleClientset()
 	osmPolicyClientSet := osmPolicyClient.NewSimpleClientset()
-	kubernetesClient := k8s.NewKubernetesClient(kubeClient, meshName, stop)
+	kubernetesClient, err := k8s.NewKubernetesClient(kubeClient, meshName, stop)
+	if err != nil {
+		log.Fatal().Err(err).Msg("Error initializing kubernetes controller")
```
nojnhuh, line 50: Instead of `log.Fatal` which calls `os.Exit(1)`, could we use `GinkgoT().Fatal()` or something similar so the failure is registered by the testing framework?

pkg/kubernetes/client.go
```diff
@@ -14,8 +14,8 @@ import (
 	"github.com/openservicemesh/osm/pkg/service"
 )
 
-// NewKubernetesClient returns a new Client which means to provide access to locally-cached k8s resources
-func NewKubernetesClient(kubeClient kubernetes.Interface, meshName string, stop chan struct{}) Client {
+// NewKubernetesClient returns a new kubernetes.Controller which means to provide access to locally-cached k8s resources
+func NewKubernetesClient(kubeClient kubernetes.Interface, meshName string, stop chan struct{}) (Controller, error) {
```
nojnhuh, line 18: Does changing the name of this function to `NewKubernetesController` make it any more clear?

eduser25 (approved):

snehachhabria (approved):

pkg/smi/client_test.go
shashankram: Not familiar with `Ginko's fatal`, but seems like we should leverage this across the test code. Would you mind creating an issue to handle this repo wide?

pkg/kubernetes/client.go
shashankram: I thought about this, but kept it as is since a few other interfaces are instantiated using `NewXXXClient`. How about I handle these renames generically in a follow-up?

pkg/smi/client_test.go
nojnhuh: What I understand is `GinkgoT()` is just a way to get a handle to the current `*testing.T`, so it's basically the equivalent of `t.Fatal()`. Yes, will open an issue.

PR #1712, opened 2020-09-16
Speedup cleanup

eduser25:
Run kubectl in parallel and wait for all of them to return.

`--wait` [waits for resources to be gone](https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands)

- Tests / CI System      [X]

No

### DIFF
--- a/demo/clean-kubernetes.sh
+++ b/demo/clean-kubernetes.sh
@@ -6,9 +6,11 @@ set -aueo pipefail
 source .env
 
 for ns in "$BOOKWAREHOUSE_NAMESPACE" "$BOOKBUYER_NAMESPACE" "$BOOKSTORE_NAMESPACE" "$BOOKTHIEF_NAMESPACE" "$K8S_NAMESPACE"; do
-    kubectl delete namespace "$ns" --ignore-not-found
+    kubectl delete namespace "$ns" --ignore-not-found &
 done
 
 # Clean up Hashicorp Vault deployment
-kubectl delete deployment vault -n "$K8S_NAMESPACE" --ignore-not-found
-kubectl delete service vault -n "$K8S_NAMESPACE" --ignore-not-found
+kubectl delete deployment vault -n "$K8S_NAMESPACE" --ignore-not-found &
+kubectl delete service vault -n "$K8S_NAMESPACE" --ignore-not-found &
+
+wait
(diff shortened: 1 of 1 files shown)


snehachhabria (approved):

eduser25:
I'm assuming kubectl handles multiple concurrent calls gracefully (and `K8S_NAMESPACE` and vault deployment deletion should not present issues)

demo/clean-kubernetes.sh
```diff
@@ -6,9 +6,11 @@ set -aueo pipefail
 source .env
 
 for ns in "$BOOKWAREHOUSE_NAMESPACE" "$BOOKBUYER_NAMESPACE" "$BOOKSTORE_NAMESPACE" "$BOOKTHIEF_NAMESPACE" "$K8S_NAMESPACE"; do
-    kubectl delete namespace "$ns" --ignore-not-found
+    kubectl delete namespace "$ns" --ignore-not-found &
```
shashankram, line 9: Since the demo relies on cleanup to happen first before setting up the run, not waiting for this to complete will result in race conditions where a namsepace is being created while it is pending deletion.

shashankram (changes requested):

demo/clean-kubernetes.sh
shashankram: This is regarding a race where the namespace is marked for deletion but doesn't get deleted, seen this previously. This issue is present even without your change, wondering if it can be addressed.

shashankram, line 9: We should just delete the namespaces in parallel:
`kubectl delete ns1 ns1 ...`

eduser25: surprising k8s doesn't serialize the events.
should I infer that the `wait` for the bash commands will just then guarantee they are "marked for deletion" but not deleted?

shashankram: That makes sense.

The only change I would make is to delete multiple namespaces via a single command.

eduser25: this works serially on the back apparently: https://stackoverflow.com/questions/43675231/kubernetes-delete-all-jobs-in-bulk
I've found a wait to wait for resource to be truly deleted, kubectl `--wait` will do the job
https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands

shashankram: Where does it say its serial? Let's skip the wait for now, probably going to slow it down further.

eduser25: Delete using visitor pattern
https://github.com/kubernetes/kubectl/blob/master/pkg/cmd/delete/delete.go#L293
Reading down it still looks serial to me, the api doesn't say though. Seeing the visitor callback uses append to a slice in the outer scope, i'd say can't be parallel

nojnhuh (approved):

draychev (approved):
This looks good to me!

Issue #1575, opened 2020-08-18
Program envoy to use list of services returned by GetServicesFromEnvoyCertificate


ksubrmnn:
**Please describe the Improvement and/or Feature Request**

Envoy should program itself for each of the services returned by GetServicesFromEnvoyCertificate
**Scope (please mark with X where applicable)**
- New Functionality          [ ]
- Install                    [ ]
- SMI Traffic Access Policy  [ ]
- SMI Traffic Specs Policy   [ ]
- SMI Traffic Split Policy   [ ]
- Permissive Traffic Policy  [ ]
- Ingress                    [ ]
- Egress                     [ ]
- Envoy Control Plane        [X]
- CLI Tool                   [ ]
- Metrics                    [ ]
- Certificate Management     [ ]
- Sidecar Injection          [ ]
- Logging                    [ ]
- Debugging                  [ ]
- Tests                      [ ]
- CI System                  [ ]
- Project Release            [ ]

**Possible use cases**

This will allow Envoy to be programmed for the multiple services that can be associated with a pod.

ksubrmnn:
Search for 1575 in the repo to see instances of where this should change

snehachhabria:
Might be handled by #2064

shashankram:
This will be handled by #2064.

PR #1725, opened 2020-09-18
Introducing per-service filtering

eduser25:
Introduces per-destination filter-chain matching on outbound

This change will allow setting specific network filtering on per service 
destination, which enables finer configuration per-service level,
as well as the use of different network filters to allow not only L7 traffic
through. 

Additional benefits:
- Since we are filtering all permitted traffic, we can generalize and filter
 any remaining traffic as Egress (when enabled), which simplifies `Egress`
to not need a CIDR anymore. (TODO: cleanup CIDR flags/code)
- Additional work that might benefit from it: per-service route table
on RDS, TCP routing, ....

Additionally:
- Fixing the listener tests required adding the long-awaited catalog mock.
Will add more tests in subsequent commits.

- New Functionality      [X]
- Control Plane          [X]
- Networking             [X]
- SMI Policy             [X]

No


pkg/envoy/cds/response.go
```diff
@@ -66,8 +66,8 @@ func NewResponse(catalog catalog.MeshCataloger, proxy *envoy.Proxy, _ *xds_disco
 	}
 	clusterFactories[localCluster.Name] = localCluster
 
-	if cfg.IsEgressEnabled() {
-		// Add a pass-through cluster for egress
+	if cfg.IsEgressEnabled() || cfg.IsPermissiveTrafficPolicyMode() {
```
eduser25, line 69: @shashankram , I'm not sure this is correct. I understood permissive as allowing traffic both inmesh and outmesh. If this is not correct let me know.

pkg/envoy/lds/listener.go
```diff
@@ -200,3 +150,148 @@ func getCIDRRange(cidr string) (*xds_core.CidrRange, error) {
 
 	return cidrRange, nil
 }
+
+// getFilterForService builds a network filter action for traffic destined to a specific service
+func getFilterForService(meshSvc service.MeshService, cfg configurator.Configurator) (string, *any.Any, error) {
```
eduser25, line 155: could use dstService here maybe

pkg/envoy/lds/listener.go
```diff
@@ -200,3 +150,148 @@ func getCIDRRange(cidr string) (*xds_core.CidrRange, error) {
...
+			log.Error().Err(err).Msgf("Error marshalling HTTPConnManager object")
+			return "", nil, err
+		}
+	}
+
+	return wellknownFilterName, marshalledFilter, nil
+}
+
+// getFilterChainMatchForService builds a filter chain to match the destination traffic.
+// Filter Chain matches currently match on destination IP for possible service endpoint leafs
```
eduser25, line 193:
```suggestion
// Filter Chain currently match on destination IP for possible service endpoints
```

pkg/envoy/cds/response.go
shashankram: The passthrough cluster is a special cluster used specifically for egress, so the conditional should not include permissive mode.

Please revert this to as it was before.

pkg/envoy/cds/response.go
```diff
@@ -66,8 +66,8 @@ func NewResponse(catalog catalog.MeshCataloger, proxy *envoy.Proxy, _ *xds_disco
...
+		// Add a pass-through cluster for egress or permissive
```
shashankram, line 70: Same as above.

pkg/envoy/lds/listener.go
```diff
@@ -200,3 +150,148 @@ func getCIDRRange(cidr string) (*xds_core.CidrRange, error) {
...
+
+		// Filterchain for current dest
+		filterChain := xds_listener.FilterChain{
+			Name:             keyService.String(),
+			Filters:          []*xds_listener.Filter{},
+			FilterChainMatch: &xds_listener.FilterChainMatch{},
+		}
+
+		// Get filter for service
+		filterName, filterProto, err := getFilterForService(keyService, cfg)
```
eduser25, line 263: I can make this return the filter object filled already, instead of the inner elements

pkg/envoy/cds/response.go
eduser25: will do

eduser25:
Will also address setting Permissive to TCP proxy, we do have some drawbacks to take care of.

eduser25:
Fixes for CI/demo expected values will need to be addressed

.github/workflows/main.yml
```diff
@@ -217,7 +217,7 @@ jobs:
         env:
           CERT_MANAGER: "cert-manager" # enables jetstack/cert-manager integration
           BOOKSTORE_SVC: "bookstore-v1"
-          BOOKTHIEF_EXPECTED_RESPONSE_CODE: "404"
+          BOOKTHIEF_EXPECTED_RESPONSE_CODE: "0"
```
snehachhabria, line 220: why does this become '0' ? Shouldn't it still be 404

pkg/envoy/lds/listener.go
```diff
@@ -60,50 +46,14 @@ func newOutboundListener(cfg configurator.Configurator) (*xds_listener.Listener,
...
-
-	// With egress, a filter chain to match TLS traffic is added to the outbound listener.
-	// In-mesh traffic will always be HTTP so this filter chain will not match for in-mesh.
-	// HTTPS egress traffic will match this filter chain and will be proxied to its original
-	// destination.
-	egressFilterChain, err := buildEgressFilterChain()
+	serviceFilterChains, err := getFilterChains(catalog, cfg, localSvc)
 	if err != nil {
-		return err
+		log.Error().Err(err).Msgf("Error getting filter chain for Egress")
```
shashankram, line 51: Log message needs to be updated to reflect call to `getFilterChains`, which doesn't seem specific to egress.

pkg/envoy/lds/listener.go
```diff
@@ -200,3 +150,124 @@ func getCIDRRange(cidr string) (*xds_core.CidrRange, error) {
 
 	return cidrRange, nil
 }
+
+// getFilterForService builds a network filter action for traffic destined to a specific service
+func getFilterForService(dstSvc service.MeshService, cfg configurator.Configurator) (*xds_listener.Filter, error) {
```
shashankram, line 155: To make it clear these are outbound filters, I'd rename this and other similar functions related to outbound filter processing.

```suggestion
func getOutboundFilterForService(dstSvc service.MeshService, cfg configurator.Configurator) (*xds_listener.Filter, error) {
```

pkg/envoy/lds/listener.go
```diff
@@ -200,3 +150,124 @@ func getCIDRRange(cidr string) (*xds_core.CidrRange, error) {
...
+
+	return &xds_listener.Filter{
+		Name:       wellknown.HTTPConnectionManager,
+		ConfigType: &xds_listener.Filter_TypedConfig{TypedConfig: marshalledFilter},
+	}, nil
+}
+
+// getFilterChainMatchForService builds a filter chain to match the destination traffic.
+// Filter Chain currently match on destination IP for possible service endpoints
+func getFilterChainMatchForService(dstSvc service.MeshService, catalog catalog.MeshCataloger, cfg configurator.Configurator) (*xds_listener.FilterChainMatch, error) {
```
shashankram, line 174:
```suggestion
func getOutboundFilterChainMatchForService(dstSvc service.MeshService, catalog catalog.MeshCataloger, cfg configurator.Configurator) (*xds_listener.FilterChainMatch, error) {
```

pkg/envoy/lds/listener.go
```diff
@@ -200,3 +150,124 @@ func getCIDRRange(cidr string) (*xds_core.CidrRange, error) {
...
+	ret := &xds_listener.FilterChainMatch{}
```
shashankram, line 175: s/ret/filterMatch or s/ret/fcMatch or s/ret/match

pkg/envoy/lds/listener.go
```diff
@@ -200,3 +150,124 @@ func getCIDRRange(cidr string) (*xds_core.CidrRange, error) {
...
+			PrefixLen: &wrapperspb.UInt32Value{
+				Value: singleIpv4Mask,
+			},
+		})
+	}
+
+	return ret, nil
+}
+
+func getFilterChains(catalog catalog.MeshCataloger, cfg configurator.Configurator, localSvc []service.MeshService) ([]*xds_listener.FilterChain, error) {
```
shashankram, line 200:
```suggestion
func getOutboundFilterChains(catalog catalog.MeshCataloger, cfg configurator.Configurator, localSvc []service.MeshService) ([]*xds_listener.FilterChain, error) {
```

pkg/envoy/lds/listener.go
```diff
@@ -200,3 +150,124 @@ func getCIDRRange(cidr string) (*xds_core.CidrRange, error) {
...
+	var ret []*xds_listener.FilterChain
+	var dstServicesSet map[service.MeshService]struct{} = make(map[service.MeshService]struct{}) // Set, avoid dups
+
+	outboundSvc, err := catalog.ListAllowedOutboundServices(localSvc[0])
```
shashankram, line 204: Since we are making an assumption around a proxy being associated with a single service (as is the case now), I'd reference issue https://github.com/openservicemesh/osm/issues/1575

pkg/envoy/lds/listener.go
```diff
@@ -200,3 +150,124 @@ func getCIDRRange(cidr string) (*xds_core.CidrRange, error) {
...
+	if err != nil {
+		log.Error().Err(err).Msgf("Error getting allowed outbound services for %s", localSvc[0].String())
+		return nil, err
+	}
+
+	// Transform into set, when listing apex services we might face repetitions
```
shashankram, line 210: Maybe use a mapset instead.

pkg/envoy/lds/listener.go
```diff
@@ -200,3 +150,124 @@ func getCIDRRange(cidr string) (*xds_core.CidrRange, error) {
...
+	for _, meshSvc := range outboundSvc {
+		dstServicesSet[meshSvc] = struct{}{}
+	}
+
+	// Getting apex services referring to the outbound services
+	splitServices := catalog.GetSMISpec().ListTrafficSplitServices()
+	for _, svc := range splitServices {
+		for _, outSvc := range outboundSvc {
+			if svc.Service == outSvc {
```
shashankram, line 219: Let's add a comment on why this check is necessary. Updating L215 should be ok.

shashankram (commented):
Looks good overall. I'll pull this change and test it.

.github/workflows/main.yml
eduser25: nope... 404 is returned by RDS right now, as there's no route for traffic that's actually listened to
with this change, there's no filter chain that matches bookthief's traffic destination (because it's not permitted), the stream doesn't even reach RDS in this case.

pkg/envoy/lds/listener.go
eduser25: arf, I had the comment to begin with, I removed it as I thought I had too much chatter on the func. Readding.

.github/workflows/main.yml
snehachhabria: but isn't that incorrect? we should be able to show that if a traffic target isn't applied a service can't access and that should be depicted by a 503 or 404. A 0 doesn't seem right to me

pkg/envoy/lds/listener.go
eduser25: mapset is a [map underneath](https://github.com/gizak/golang-set/blob/master/set.go#L38), I imagine you want it for readability at this point; I personally dislike adding deps if not strictly needed.

shashankram: for readability and helpers it provides. We already use it across the code.

eduser25: I know, and I never liked it. Want me to change it here though? I'm ok with either in any case

pkg/envoy/lds/listener.go
```diff
@@ -200,3 +150,127 @@ func getCIDRRange(cidr string) (*xds_core.CidrRange, error) {
...
+	}
+
+	return filterMatch, nil
+}
+
+func getOutboundFilterChains(catalog catalog.MeshCataloger, cfg configurator.Configurator, localSvc []service.MeshService) ([]*xds_listener.FilterChain, error) {
+	var filterChains []*xds_listener.FilterChain
+	var dstServicesSet map[service.MeshService]struct{} = make(map[service.MeshService]struct{}) // Set, avoid dups
+
+	// Assuming single service in pod til #1682, #1575 get addressed
```
snehachhabria, line 204:
```suggestion
	// Assuming single service in pod till #1682, #1575 get addressed
```

.github/workflows/main.yml
eduser25: Several discussions stirred from this:
- SMI is vague, in which it says traffic is to be "denied", but doesn't specifically say what kind of deny (network deny, L7 deny through http code)
- We might proceed with this at the expense of the behavior change, however we will validate this is in line with expected behavior with customers as it does indeed reduce the visibility as of why a connection is rejected, for example.

snehachhabria (approved):

pkg/envoy/lds/listener.go
shashankram: not necessarily for this PR but in a follow-up would be good.

shashankram (dismissed):

shashankram (approved):

snehachhabria (approved):

Issue #1016, opened 2020-07-07
Finer granularity for Prometheus monitored namespaces


snehachhabria:
**Please describe the Improvement and/or Feature Request**
As of today, all pods that are a part of the mesh enable the collection of metrics by default. We want to make this configurable, whereby only the specified namespaces (via the config map) are annotated with Prometheus annotations for metrics collection 

Part of #966 

**Possible use cases**
This will give users the ability to collect metrics  only for specified namespaces

shashankram:
This should be driven through `osm cli` if annotations are required on prometheus pods post install.

shashankram:
> This should be driven through `osm cli` if annotations are required on prometheus pods post install.

Needs a bit of discussion over implementation details.

shashankram:
Prioritizing #1619 till we iron out the proposal.

snehachhabria:
@shashankram is the proposal for this ironed out ? We might need to prioritize the work on this. cc @delqn

shashankram:
I'll look into this and come up with a proposal.

PR #1731, opened 2020-09-18
metrics: support configuring select namespaces

shashankram:
Adds support to configure namespaces for metrics
collection.

- osm cli command to enable/disable metrics on
  select namespaces. Existing mesh participating
  pods will be enabled for metrics.

- injector changes to auto enable metrics on
  pods belonging to namespaces for which metrics
  are enabled.

`osm metrics` command is limited to namespace scope
configurations at this time. It should be possible
in the future to enable metrics for select pods.

Part of #1016

- New Functionality      [X]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [ ]
- CLI Tool               [X]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [X]
- SMI Policy             [ ]
- Security               [ ]
- Tests / CI System      [ ]
- Other                  [ ]

`No`

### DIFF
--- a/cmd/cli/metrics.go
+++ b/cmd/cli/metrics.go
@@ -1,0 +1,25 @@
+package main
+
+import (
+	"io"
+
+	"github.com/spf13/cobra"
+)
+
+const metricsDescription = `
+This command consists of multiple subcommands related to managing metrics
+associated with osm.
+`
+
+func newMetricsCmd(out io.Writer) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "metrics",
+		Short: "manage metrics",
+		Long:  metricsDescription,
+		Args:  cobra.NoArgs,
+	}
+	cmd.AddCommand(newMetricsEnable(out))
+	cmd.AddCommand(newMetricsDisable(out))
+
+	return cmd
+}
--- a/cmd/cli/metrics_disable.go
+++ b/cmd/cli/metrics_disable.go
@@ -51,7 +51,6 @@ func newMetricsDisable(out io.Writer) *cobra.Command {
 		},
 	}
 
-	//add mesh name flag
 	f := cmd.Flags()
 	f.StringSliceVar(&disableCmd.namespaces, "namespace", []string{}, "One or more namespaces to disable metrics on")
 
--- a/cmd/cli/metrics_disable.go
+++ b/cmd/cli/metrics_disable.go
@@ -1,0 +1,133 @@
+package main
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"strings"
+
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/kubernetes"
+
+	"github.com/openservicemesh/osm/pkg/constants"
+)
+
+const metricsDisableDescription = `
+This command will disable metrics on all pods belonging to the given
+namespace or set of namespaces.
+`
+
+type metricsDisableCmd struct {
+	out        io.Writer
+	namespaces []string
+	clientSet  kubernetes.Interface
+}
+
+func newMetricsDisable(out io.Writer) *cobra.Command {
+	disableCmd := &metricsDisableCmd{
+		out: out,
+	}
+
+	cmd := &cobra.Command{
+		Use:   "disable ...",
+		Short: "disable metrics",
+		Long:  metricsDisableDescription,
+		Args:  cobra.NoArgs,
+		RunE: func(_ *cobra.Command, args []string) error {
+			config, err := settings.RESTClientGetter().ToRESTConfig()
+			if err != nil {
+				return errors.Errorf("Error fetching kubeconfig")
+			}
+
+			clientset, err := kubernetes.NewForConfig(config)
+			if err != nil {
+				return errors.Errorf("Could not access Kubernetes cluster. Check kubeconfig")
+			}
+			disableCmd.clientSet = clientset
+			return disableCmd.run()
+		},
+	}
+
+	//add mesh name flag
+	f := cmd.Flags()
+	f.StringSliceVar(&disableCmd.namespaces, "namespace", []string{}, "One or more namespaces to disable metrics on")
+
+	return cmd
+}
+
+func (cmd *metricsDisableCmd) run() error {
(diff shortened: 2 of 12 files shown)


cmd/cli/metrics_disable.go
```diff
@@ -0,0 +1,133 @@
...
+			clientset, err := kubernetes.NewForConfig(config)
+			if err != nil {
+				return errors.Errorf("Could not access Kubernetes cluster. Check kubeconfig")
+			}
+			disableCmd.clientSet = clientset
+			return disableCmd.run()
+		},
+	}
+
+	//add mesh name flag
```
ksubrmnn, line 54: Do we still need this comment?

cmd/cli/metrics_enable.go
```diff
@@ -0,0 +1,137 @@
...
+	}
+
+	//add mesh name flag
+	f := cmd.Flags()
+	f.StringSliceVar(&enableCmd.namespaces, "namespace", []string{}, "One or more namespaces to enable metrics on")
+
+	return cmd
+}
+
+func (cmd *metricsEnableCmd) run() error {
```
ksubrmnn, line 62: I think this would also have to check if [enablePrometheus](https://github.com/openservicemesh/osm/blob/main/charts/osm/values.yaml#L35) is enabled

ksubrmnn (changes requested):

cmd/cli/metrics_enable.go
ksubrmnn: Or if we also support byo prometheus, we might have to check for that.

cmd/cli/metrics_disable.go
shashankram: Removed.

cmd/cli/metrics_enable.go
shashankram: Don't think its necessary, because these are just annotations being added. If the user doesn't deploy prometheus or disables prometheus scraping via osm-config configMap its expected for the metrics to not be scraped. Running the CLI just informs the controller the scope of scraping (namespace in this case).

ksubrmnn: Sure, I think this is okay for now, but we should eventually consider returning some sort of message or at least document clearly so that the user understands that this could be a no-op if Prometheus is not deployed. In other words, we need to document clearly that Prometheus deployment is a pre-req because "enable metrics" is vague unless you know the implementation.

shashankram: This will go into the documentation that I am going to write as a part of #1016, which is why I am not going to mark the issue as resolved.

shashankram: Btw the command states that metrics will be enabled on pods, it doesn't say it will deploy prometheus. I could have changed the command to `osm metrics-collection` but it seemed longer than necessary. WDYT?

cmd/cli/metrics_enable.go
```diff
@@ -0,0 +1,137 @@
...
+	"github.com/spf13/cobra"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/kubernetes"
+
+	"github.com/openservicemesh/osm/pkg/constants"
+)
+
+const metricsEnableDescription = `
+This command will enable metrics on all pods belonging to the given namespace
```
ksubrmnn, line 19: Maybe: `This command will enable metrics for an existing Prometheus deployment for all pods...`

shashankram: It doesn't have to be existing, the user could deploy Prometheus whenever they want metrics collection to happen (post running this command)?

ksubrmnn: Oh that's true. Maybe just add "this command does not deploy Prometheus"

cmd/cli/metrics_enable.go
snehachhabria: Agree with Shashank. Adding the annotations will actually not enable any metrics collection until the prometheus scraping via osm-config is enabled (for either OSM Prometheus or BYO)

cmd/cli/metrics.go
```diff
@@ -0,0 +1,25 @@
...
+
+import (
+	"io"
+
+	"github.com/spf13/cobra"
+)
+
+const metricsDescription = `
+This command consists of multiple subcommands related to managing metrics
+associated with osm.
```
snehachhabria, line 11: Should we indicate that its prometheus format metrics only as of today

cmd/cli/metrics_disable.go
```diff
@@ -0,0 +1,132 @@
...
+	"github.com/spf13/cobra"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/kubernetes"
+
+	"github.com/openservicemesh/osm/pkg/constants"
+)
+
+const metricsDisableDescription = `
+This command will disable metrics on all pods belonging to the given
```
snehachhabria, line 19: same as above, format of metrics is prometheus

cmd/cli/metrics_enable.go
snehachhabria, line 19:
```suggestion
This command will enable prometheus format metrics scraping from all pods belonging to the given namespace
```

shashankram: Added a note to the command documentation.

ksubrmnn (dismissed):

cmd/cli/metrics_enable.go
```diff
@@ -0,0 +1,137 @@
...
+or set of namespaces. Newly created pods belonging to namespaces that
+are enabled for metrics will be automatically enabled with metrics.
```
snehachhabria, line 21:
```suggestion
are enabled for metrics scraping will be automatically enabled with metrics scraping.
```

cmd/cli/metrics_enable.go
ksubrmnn: Addressed in latest commit

cmd/cli/metrics.go
shashankram: The CLI is configuring prometheus scraping, not necessarily the format.
It would make more sense to note that it enables prometheus scraping rather than talking about formats.

snehachhabria (approved):

ksubrmnn (approved):

PR #1737, opened 2020-09-21
cli: restrict metrics command to monitored namespaces

shashankram:
Ensures that metrics can only be enabled or disabled on
namespaces that are monitored.

Updates existing code in mesh_list* to reuse code to
list all the mesh names so that a namespace can be
validated for monitoring.

Part of #1016

- New Functionality      [ ]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [ ]
- CLI Tool               [X]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests / CI System      [ ]
- Other                  [ ]

`No`

### DIFF
--- a/cmd/cli/mesh_list.go
+++ b/cmd/cli/mesh_list.go
@@ -51,7 +51,7 @@ func newMeshList(out io.Writer) *cobra.Command {
 }
 
 func (l *meshListCmd) run() error {
-	list, err := l.selectMeshes()
+	list, err := getControllerDeployments(l.clientSet)
 	if err != nil {
 		return errors.Errorf("Could not list deployments %v", err)
 	}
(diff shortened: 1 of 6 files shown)


cmd/cli/metrics_disable.go
```diff
@@ -64,10 +64,21 @@ func (cmd *metricsDisableCmd) run() error {
...
 		defer cancel()
 
-		if _, err := cmd.clientSet.CoreV1().Namespaces().Get(ctx, ns, metav1.GetOptions{}); err != nil {
+		namespace, err := cmd.clientSet.CoreV1().Namespaces().Get(ctx, ns, metav1.GetOptions{})
+		if err != nil {
 			return errors.Errorf("Failed to retrieve namespace [%s]: %v", ns, err)
 		}
 
+		// Check if the namespace belongs to a mesh, if not return an error
+		monitored, err := isMonitoredNamespace(*namespace, getMeshNames(cmd.clientSet))
```
eduser25, line 73: I imagine using `metav1.GetOptions{}` to match meshnames (if even entirely possible) will just complicate your life

cmd/cli/mesh_list.go
```diff
@@ -72,11 +72,24 @@ func (l *meshListCmd) run() error {
...
 	labelSelector := metav1.LabelSelector{MatchLabels: map[string]string{"app": constants.OSMControllerName}}
 	listOptions := metav1.ListOptions{
 		LabelSelector: labels.Set(labelSelector.MatchLabels).String(),
 	}
 	return deploymentsClient.List(context.TODO(), listOptions)
 }
+
+// getMeshNames returns a list of mesh names corresponding to meshes within the cluster
+func getMeshNames(clientSet kubernetes.Interface) []string {
+	var meshList []string
```
eduser25, line 87: If the meshLists are actual strong keys, I'd say to go for a set and have `meshExists` just do a lookup instead.
This is CLI and doesn't matter that much, but I'll probably try to be more insistent with this on the control plane; We have way too many loops over lists that can be indexed.

eduser25 (approved):
lgtm

cmd/cli/metrics_disable.go
shashankram: You mean to list namespaces whose labels match against a list of mesh names?

snehachhabria (dismissed):

cmd/cli/mesh_list.go
shashankram: Using a set makes sense here, will update.

eduser25 (approved):

snehachhabria (approved):

PR #1755, opened 2020-09-25
test(pkg/utils): Add tests and refactor existing tests with Testify

### LINKED ISSUE #1673: add unit tests for pkg/utils
This is a package in OSM that contains some utility or helper functions that can be used across different packages.
https://github.com/openservicemesh/osm/tree/main/pkg/utils

shalier:
Add unit tests to utils as well as refactor current tests to use Testify 
and resolves #1673

- New Functionality      [ ]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [ ]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests / CI System      [ X]
- Other                  [ ]

No


pkg/envoy/ads/server.go
```diff
@@ -53,7 +53,7 @@ func (s *Server) Start(ctx context.Context, cancel context.CancelFunc, port int,
 	}
 
 	xds_discovery.RegisterAggregatedDiscoveryServiceServer(grpcServer, s)
-	go utils.GrpcServe(ctx, grpcServer, lis, cancel, ServerType)
+	go utils.GrpcServe(ctx, grpcServer, lis, cancel, ServerType, nil)
```
shashankram, line 56: You don't want to be passing a nil channel, writing to it will block.

ksubrmnn: Agreed. Let's use this channel and set `s.ready = true` if it's successful

pkg/utils/grpc_test.go
```diff
@@ -0,0 +1,76 @@
...
+	cache := make(map[certificate.CommonName]certificate.Certificater)
+	certManager := tresor.NewFakeCertManager(&cache, 1*time.Hour)
+	adsCert, err := certManager.GetRootCertificate()
+	assert.Nil(err)
+
+	serverType := "ADS"
+	flags := pflag.NewFlagSet(`osm-controller`, pflag.ExitOnError)
+	port := flags.Int("port", constants.OSMControllerPort, "Aggregated Discovery Service port number.")
+
+	grpcServer, lis, err := NewGrpc(serverType, *port, adsCert.GetCertificateChain(), adsCert.GetPrivateKey(), adsCert.GetIssuingCA())
```
ksubrmnn, line 66: Just check that this err is nil

ksubrmnn (changes requested):
Great job, @shalier!!

pkg/utils/grpc.go
```diff
@@ -45,10 +45,13 @@ func NewGrpc(serverType string, port int, certPem, keyPem, rootCertPem []byte) (
 }
 
 // GrpcServe starts the gRPC server passed.
-func GrpcServe(ctx context.Context, grpcServer *grpc.Server, lis net.Listener, cancel context.CancelFunc, serverType string) {
+func GrpcServe(ctx context.Context, grpcServer *grpc.Server, lis net.Listener, cancel context.CancelFunc, serverType string, errorCh chan interface{}) {
 	go func() {
 		if err := grpcServer.Serve(lis); err != nil {
 			log.Error().Err(err).Msgf("[grpc][%s] Error serving gRPC request", serverType)
+			if errorCh != nil {
```
ksubrmnn, line 52: Looks good!

pkg/utils/uuid_test.go
```diff
@@ -0,0 +1,38 @@
...
+	for _, uuid := range listUUID {
+		result := IsValidUUID(uuid)
+
+		if result == false {
+			assert.False(result)
+		} else {
+			assert.True(result)
+		}
+	}
+
```
ksubrmnn, line 37: nit: unnecessary newline

ksubrmnn (changes requested):

pkg/utils/uuid_test.go
```diff
@@ -0,0 +1,37 @@
...
+)
+
+func TestNewUUIDStr(t *testing.T) {
+	assert := assert.New(t)
+
+	output := NewUUIDStr()
+	outputType := reflect.TypeOf(output)
+	stringType := reflect.TypeOf("")
+
+	assert.IsType(stringType, outputType)
```
ksubrmnn, line 18: I think instead of this you can check if the string is non-empty and that should be enough. I don't think checking the return type here is necessary.

pkg/utils/userip_test.go
```diff
@@ -0,0 +1,31 @@
...
+	ctxs := []context.Context{
+		context.Background(),
+		peer.NewContext(context.TODO(), &peer.Peer{Addr: tests.NewMockAddress("9.8.7.6")}),
+	}
+
+	for _, ctx := range ctxs {
+		result := GetIPFromContext(ctx)
+
+		if result == nil {
+			assert.Nil(result)
```
ksubrmnn, line 25: I don't think you need this condition and check. I think instead you'd probably just want to assert that result is not nil

pkg/utils/json_test.go
```diff
@@ -1,19 +1,40 @@
...
+		{[]byte("should error"),
+			"",
+			nil},
+	}
+
+	for _, pjt := range prettyJSONtests {
+		prettyJSON, err := PrettyJSON(pjt.input, pjt.prefix)
 
-})
+		if err != nil {
```
ksubrmnn, line 32: You don't need to assert.NotNil err here since we already know err is nil.

You could add a field to prettyJSONTest that is a bool for error and check if the error should have occurred. We need to ensure that the errors are happening for the correct cases

pkg/utils/uuid_test.go
```diff
@@ -0,0 +1,37 @@
...
+	assert := assert.New(t)
+
+	listUUID := []string{
+		uuid.New().String(),
+		uuid.New().String() + "!xyz"}
+
+	for _, uuid := range listUUID {
+		result := IsValidUUID(uuid)
+
+		if result == false {
```
ksubrmnn, line 31: This check isn't needed since we already know result is false in this block. I would recommend using the table driven style where you create an object that maps the input (the string) and the expected output as a boolean

pkg/utils/grpc_test.go
```diff
@@ -0,0 +1,75 @@
...
+
+	newGrpcTests := []newGrpcTest{
+		{"abc", 123, emptyByteArray},
+		{"ADS", 8081, emptyByteArray},
+		{"ADS", 8080, certPem},
+	}
+
+	for _, gt := range newGrpcTests {
+		resServer, resListener, err := NewGrpc(gt.serverType, gt.port, gt.certPem, keyPem, rootPem)
+		if err != nil {
```
ksubrmnn, line 42: Please add the expected err value to newGrpcTest so that we can check that the correct tests are erroring

pkg/utils/mtls_test.go
```diff
@@ -0,0 +1,96 @@
...
+
+	setupMutualTLStests := []setupMutualTLStest{
+		{emptyByteArray, goodKeyPem, goodCA},
+		{goodCertPem, goodKeyPem, emptyByteArray},
+		{goodCertPem, goodKeyPem, goodCA},
+	}
+
+	for _, smt := range setupMutualTLStests {
+		result, err := setupMutualTLS(true, serverType, smt.certPem, smt.keyPem, smt.ca)
+		if err != nil {
```
ksubrmnn, line 50: Please update the error checking logic

pkg/utils/mtls_test.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+
+	for _, vct := range validateClientTests {
+		result, err := ValidateClient(vct.ctx, vct.commonNames)
+		if err != nil {
+			assert.Equal(result, certificate.CommonName(""))
+			assert.True(errors.Is(err, vct.expectedError))
+		} else {
+			assert.NotNil(result)
+			assert.Nil(err)
```
ksubrmnn, line 93: you don't need to check the err again

pkg/utils/mtls_test.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+
+	for _, smt := range setupMutualTLStests {
+		result, err := setupMutualTLS(true, serverType, smt.certPem, smt.keyPem, smt.ca)
+		if err != nil {
+			assert.Nil(result)
+			assert.NotNil(err)
+		} else {
+			assert.NotNil(result)
+			assert.Nil(err)
```
ksubrmnn, line 55: No need to check this err again

pkg/utils/grpc_test.go
```diff
@@ -0,0 +1,75 @@
...
+			assert.Nil(resServer)
+			assert.Nil(resListener)
+			assert.NotNil(err)
+		} else {
+			assert.NotNil(resServer)
+			assert.NotNil(resListener)
+			assert.Nil(err)
```
ksubrmnn, line 49: No need to check this error again

ksubrmnn (changes requested):

pkg/utils/json_test.go
```diff
@@ -1,19 +1,41 @@
...
+	}
 
-})
+	for _, pjt := range prettyJSONtests {
+		prettyJSON, err := PrettyJSON(pjt.input, pjt.prefix)
+		if err != nil {
+			assert.Contains(err.Error(), pjt.expectedError)
+			assert.Nil(prettyJSON)
+		} else {
+			assert.Equal(pjt.expectedPrettyJSON, prettyJSON)
```
ksubrmnn, line 38: Can you check that expectedError is empty in this else block?

pkg/utils/mtls_test.go
```diff
@@ -0,0 +1,95 @@
...
+		{goodCertPem, goodKeyPem, emptyByteArray, "[grpc][mTLS][ADS] Failed to append client certs"},
+		{goodCertPem, goodKeyPem, goodCA, ""},
+	}
+
+	for _, smt := range setupMutualTLStests {
+		result, err := setupMutualTLS(true, serverType, smt.certPem, smt.keyPem, smt.ca)
+		if err != nil {
+			assert.Nil(result)
+			assert.Contains(err.Error(), smt.expectedError)
+		} else {
```
ksubrmnn, line 54: check expectedError

pkg/utils/mtls_test.go
```diff
@@ -0,0 +1,95 @@
...
+		{peer.NewContext(context.TODO(), &peer.Peer{AuthInfo: tests.NewMockAuthInfo(cert)}), badCommonNameMapping, status.Error(codes.Unauthenticated, "disallowed subject common name")},
+		{peer.NewContext(context.TODO(), &peer.Peer{AuthInfo: tests.NewMockAuthInfo(cert)}), goodCommonNameMapping, nil},
+	}
+
+	for _, vct := range validateClientTests {
+		result, err := ValidateClient(vct.ctx, vct.commonNames)
+		if err != nil {
+			assert.Equal(result, certificate.CommonName(""))
+			assert.True(errors.Is(err, vct.expectedError))
+		} else {
```
ksubrmnn, line 91: check expectedError

ksubrmnn (dismissed):

pkg/utils/grpc_test.go
```diff
@@ -0,0 +1,76 @@
...
+	assert := assert.New(t)
+
+	cache := make(map[certificate.CommonName]certificate.Certificater)
+	certManager := tresor.NewFakeCertManager(&cache, 1*time.Hour)
+	adsCert, err := certManager.GetRootCertificate()
+	assert.Nil(err)
+
+	serverType := "ADS"
+	flags := pflag.NewFlagSet(`osm-controller`, pflag.ExitOnError)
+	port := flags.Int("port", constants.OSMControllerPort, "Aggregated Discovery Service port number.")
```
shashankram, line 65: No reason to use `pflag` to initialize the port. 
for testing you can use any port. 

Ex. `port := 9999`

pkg/utils/grpc_test.go
```diff
@@ -0,0 +1,76 @@
...
+	grpcServer, lis, err := NewGrpc(serverType, *port, adsCert.GetCertificateChain(), adsCert.GetPrivateKey(), adsCert.GetIssuingCA())
+	assert.Nil(err)
+
+	ctx, cancel := context.WithCancel(context.Background())
+	errorCh := make(chan interface{}, 1)
+	go GrpcServe(ctx, grpcServer, lis, cancel, serverType, errorCh)
+	defer cancel()
+	time.Sleep(50 * time.Millisecond)
```
shashankram, line 73: Why is this required, seems like a possibility for races.

shashankram (approved):

ksubrmnn (approved):

Issue #1683, opened 2020-09-10
Revisit mocking on dynamic operations


eduser25:
Packages like `catalog/fake.go`provide a common initialization for testing some package component while mocking most dependencies of the component itself. This testing instance is oftentimes kept during different test instances.

The problem arises when different tests using the same context instance expect dynamic results results from the mocks that might operate underneath. In this cases, the initialization the initialization has to code some dynamic behavior to act accordingly (and to some extent even mock real functionality), mocking dynamically with the semantics of the expected behavior of the class we want to test on top.

Even though this works, it introduces dynamic behavior on the mocks themselves, when ideally and for clarity and ease of testing, mocks could just expect specific static structs to be returned at each individual test.

One suggestion would be to remove the context and share no context altogether between tests while exposing access to the component mocks, this way every test would be required to properly mock the expected responses for each mock expected to run during the test.

An alternative would be to use real packages (if even possible) for the dependencies which are very dynamic - downside is though, this would introduce additional coupling, more dependencies and distant failures for unit tests. 

- Tests                      [X]

eduser25:
This is more architectural and related to tests. We'll reopen based on discussion.

PR #1769, opened 2020-09-29
Move remaining XDS vertical API calls to use kubecontroller caches

eduser25:
GetPodFromCertificate and ListPods were two remaining APIs used by
pretty much all verticals which were still going through kubeclient API
REST calls.

In E2E tests, with a higher count of pods we've seen slowness issues,
which ranged k8s API throttling, go-routine contention (as envoy updates pile up)
and webhooks timeout.

These additions remove any network/API dependency from the XDS paths, making
verticals' execution range between 50-100ms in demo conditions and debug logging.
(Down from 7-10seconds per vertical)

- Control Plane          [X]
- Tests / CI System      [X]

No

### DIFF
--- a/pkg/catalog/fake.go
+++ b/pkg/catalog/fake.go
@@ -57,9 +57,7 @@ func NewFakeMeshCatalog(kubeClient kubernetes.Interface) *MeshCatalog {
 		// play pretend this call queries a controller cache
 		var services []*corev1.Service
 
-		// Things worked before because internally, the APIs were calling kubeClient APIs with "namespace"
-		// parametre. Some tests have been using catalog APIs bypassing the namespace monitoring altogether.
-		// For now, this assumes that catalog tests use monitored namespaces at all times
+		// This assumes that catalog tests use monitored namespaces at all times
 		svcList, _ := kubeClient.CoreV1().Services("").List(context.Background(), metav1.ListOptions{})
 		for idx := range svcList.Items {
 			services = append(services, &svcList.Items[idx])
@@ -76,7 +74,6 @@ func NewFakeMeshCatalog(kubeClient kubernetes.Interface) *MeshCatalog {
 
 		return vv
 	}).AnyTimes()
-
 	mockKubeController.EXPECT().ListPods().DoAndReturn(func() []*v1.Pod {
 		vv, err := kubeClient.CoreV1().Pods("").List(context.Background(), metav1.ListOptions{})
 		if err != nil {
--- a/pkg/catalog/fake.go
+++ b/pkg/catalog/fake.go
@@ -52,23 +52,17 @@ func NewFakeMeshCatalog(kubeClient kubernetes.Interface) *MeshCatalog {
 	mockIngressMonitor.EXPECT().GetIngressResources(gomock.Any()).Return(nil, nil).AnyTimes()
 	mockIngressMonitor.EXPECT().GetAnnouncementsChannel().Return(testChan).AnyTimes()
 
-	// Monitored namespaces is made a set to make sure we don't repeat namespaces on mock
-	listExpectedNs := tests.GetUnique([]string{
-		tests.BookstoreService.Namespace,
-		tests.BookbuyerService.Namespace,
-		tests.BookstoreApexService.Namespace,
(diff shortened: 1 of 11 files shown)


eduser25:
Additionally, bookbuyer in demo should start seeing success after 6-10 iterations, down from 30

pkg/catalog/fake.go
```diff
@@ -52,23 +52,17 @@ func NewFakeMeshCatalog(kubeClient kubernetes.Interface) *MeshCatalog {
...
 		// play pretend this call queries a controller cache
 		var services []*corev1.Service
 
-		for _, ns := range listExpectedNs {
-			svcList, _ := kubeClient.CoreV1().Services(ns).List(context.Background(), metav1.ListOptions{})
-			for _, svcItem := range svcList.Items {
-				services = append(services, &svcItem)
-			}
+		// Things worked before because internally, the APIs were calling kubeClient APIs with "namespace"
+		// parametre. Some tests have been using catalog APIs bypassing the namespace monitoring altogether.
```
shashankram, line 61: nit
```suggestion
		// parameter. Some tests have been using catalog APIs bypassing the namespace monitoring altogether.
```

pkg/catalog/fake.go
```diff
@@ -52,23 +52,17 @@ func NewFakeMeshCatalog(kubeClient kubernetes.Interface) *MeshCatalog {
...
 	mockKubeController.EXPECT().ListServices().DoAndReturn(func() []*corev1.Service {
 		// play pretend this call queries a controller cache
 		var services []*corev1.Service
 
-		for _, ns := range listExpectedNs {
-			svcList, _ := kubeClient.CoreV1().Services(ns).List(context.Background(), metav1.ListOptions{})
-			for _, svcItem := range svcList.Items {
-				services = append(services, &svcItem)
-			}
+		// Things worked before because internally, the APIs were calling kubeClient APIs with "namespace"
```
shashankram, line 60: This comment is not very helpful. Once this is merged `Things worked before because...` lacks context.

pkg/catalog/fake.go
```diff
@@ -82,6 +76,20 @@ func NewFakeMeshCatalog(kubeClient kubernetes.Interface) *MeshCatalog {
...
+		if err != nil {
+			return nil
+		}
+
+		var podRet []*v1.Pod = []*v1.Pod{}
+		for idx := range vv.Items {
+			podRet = append(podRet, &vv.Items[idx])
+		}
+		return podRet
+	}).AnyTimes()
```
shashankram, line 91: Why is this needed?

shashankram (dismissed):

shashankram:
> Additionally, bookbuyer in demo should start seeing success after 6-10 iterations, down from 30

That's a nice improvement.

pkg/catalog/fake.go
eduser25: I left it for review purposes mostly. I imagine that if I stayed with the "assumption" part, reviews might have asked "why change it".

pkg/catalog/fake.go
eduser25: This is needed for the tests run catalog functions, as I can't interface with the kubecontroller inside it. I changed some tests to just run a separate kubecontroller as they just test xds_cert functions, and they don't need catalog.
But for the ones running catalog, they need this.

pkg/catalog/fake.go
eduser25: will change it to just leave the assumption part, should be explanatory enough

pkg/catalog/fake.go
shashankram: I'd drop a comment here and file an issue to track such custom fake code needing enhancements.

eduser25: +1, could we maybe reuse https://github.com/openservicemesh/osm/issues/1683 and generalize the issue?

pkg/catalog/fake.go
eduser25: done

shashankram (approved):

nojnhuh (approved):

PR #1777, opened 2020-09-30
Allow explicitly ignoring a namespace for injection

shashankram:
This change provides a mechanism to explicitly ignore
namespaces to prevent sidecar injection from happening.
A new cli command `osm namespace ignore` is added for this
purpose.

Additionally, the mutatingwebhookconfiguration is updated
to only accept admission requests belonging to namespaces
that are not explicitly ignored and not osm control plane
namespace.

Enhancements such as checking if a namespace is ignored before 
enabling monitoring will be considered in a subsequent change.

Part of #1759

- New Functionality      [X]
- Documentation          [X]
- Install                [ ]
- Control Plane          [X]
- CLI Tool               [X]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests / CI System      [ ]
- Other                  [ ]

`No`

### DIFF
--- a/charts/osm/templates/mutatingwebhook.yaml
+++ b/charts/osm/templates/mutatingwebhook.yaml
@@ -17,3 +17,14 @@ webhooks:
   namespaceSelector:
     matchLabels:
       openservicemesh.io/monitored-by: {{.Values.OpenServiceMesh.meshName}}
+    matchExpressions:
+      # This label is explicitly set to ignore a namespace
+      - key: "openservicemesh.io/ignore"
+        operator: DoesNotExist
+
+      # This label is set by Helm when it creates a namespace (https://github.com/helm/helm/blob/release-3.2/pkg/action/install.go#L292)
+      # It ensures that pods in the control plane namespace are never injected with a sidecar
+      - key: "name"
+        operator: NotIn
+        values:
+        - {{.Release.Namespace}}
--- a/cmd/cli/namespace_ignore.go
+++ b/cmd/cli/namespace_ignore.go
@@ -1,0 +1,91 @@
+package main
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"strings"
+
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/kubernetes"
+)
+
+const namespaceIgnoreDescription = `
+This command will configure a namespace or a set of namespaces from
+participating in the mesh. Automatic sidecar injection on pods
+belonging to the given namespace or set of namespaces will be prevented.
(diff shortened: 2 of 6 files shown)


draychev (approved):
LGTM

eduser25 (dismissed):

cmd/cli/namespace_ignore.go
```diff
@@ -0,0 +1,91 @@
...
+
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/kubernetes"
+)
+
+const namespaceIgnoreDescription = `
+This command will configure a namespace or a set of namespaces from
```
nojnhuh, line 17:
```suggestion
This command will prevent a namespace or a set of namespaces from
```

nojnhuh (commented):
@shashankram Looks good, just a couple questions:

1. Can you clarify how this is different from implicitly ignoring the namespace by not adding the `sidecar-injection: enabled` label? 
1. And what happens when a namespace has both the `sidecar-injection: enabled` and `ignore: true` labels?

nojnhuh (approved):
@shashankram Makes sense, I forgot about the per-pod injection labels. Thanks!

eduser25 (approved):

charts/osm/templates/mutatingwebhook.yaml
```diff
@@ -17,3 +17,14 @@ webhooks:
...
     matchLabels:
       openservicemesh.io/monitored-by: {{.Values.OpenServiceMesh.meshName}}
+    matchExpressions:
+      # This label is explicitly set to ignore a namespace
+      - key: "openservicemesh.io/ignore"
+        operator: DoesNotExist
+
+      # This label is set by Helm when it creates a namespace (https://github.com/helm/helm/blob/release-3.2/pkg/action/install.go#L292)
+      # It ensures that pods in the control plane namespace are never injected with a sidecar
+      - key: "name"
```
ritazh, line 27: @shashankbarsin From the link referenced, this `name` label is set when `--create-namespace` is used to create the namespace via Helm. This is not true if we are installing into an existing namespace. We could consider deploying a namespace resource e.g. `osm-system` with the `openservicemesh.io/ignore` label as part of the chart. WDYT?

shashankram: I thought about this, we could very well create the namespace using a chart instead of using the Helm go client to auto-create the namespace as is the case now.

@ritazh would you mind raising an issue to handle the namespace creation in the cli as you have suggested and we can discuss over it?

ritazh: Added https://github.com/openservicemesh/osm/issues/1817 to track

PR #1794, opened 2020-10-06
configurator: add config option for service cert validity duration

shashankram:
**Description**:
Adds a new config option to retrieve the service certificate validity
duration. This is a prerequisite to make the service cert validity
duration dynamically configurable via the osm-config ConfigMap.

Part of #1757

**Affected area**:

- New Functionality      [ ]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [ ]
- CLI Tool               [ ]
- Certificate Management [X]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests                  [ ]
- CI System              [ ]
- Performance            [ ]
- Other                  [ ]

`No`

### DIFF
--- a/pkg/configurator/methods.go
+++ b/pkg/configurator/methods.go
@@ -3,8 +3,14 @@ package configurator
 import (
 	"encoding/json"
 	"fmt"
+	"time"
 
 	"github.com/openservicemesh/osm/pkg/constants"
+)
+
+const (
+	// defaultServiceCertValidityDuration is the default validity duration for service certificates
+	defaultServiceCertValidityDuration = 24 * time.Hour
 )
 
 // The functions in this file implement the configurator.Configurator interface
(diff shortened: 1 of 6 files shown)


snehachhabria (approved):

pkg/configurator/methods.go
```diff
@@ -3,10 +3,16 @@ package configurator
...
 	"encoding/json"
 	"fmt"
+	"time"
 
 	"github.com/openservicemesh/osm/pkg/constants"
 )
 
+const (
+	// defaultServiceCertValidityDuration is the default validity duration for service certificates
+	defaultServiceCertValidityDuration = 24 * time.Hour
```
eduser25, line 13: semi-related: Is `constants.go` expected to be used only if multiple modules need to read values? I imagine the declaration here is closed inside configurator and we do not expect someone else to read

eduser25 (approved):

pkg/configurator/methods.go
shashankram: That is correct.

PR #1822, opened 2020-10-13
tests/e2e: client-server deployment test

eduser25:
- Test runs HTTP from multiple source client deployments to one
destination server deployment.
- Number of clients and replicaset can be tunned in the test (it
is not parametrized through test flags though)

By default:
- 5 clients and 5 replicas
- 1 server, with 5 replicas
- It then adds TrafficTargets from all clients to the server,
- Finally tests HTTP traffic from all client pods to the server (FQDN)

(It currently does not verify that all servers reply, TrafficSplit test
has some logic to do that that could be reused here if needed)

**Affected area**:
- Tests                  [X]

No

### DIFF
--- a/tests/e2e/e2e_deployment_client_server_test.go
+++ b/tests/e2e/e2e_deployment_client_server_test.go
@@ -1,0 +1,171 @@
+package e2e
+
+import (
+	"context"
+	"fmt"
+	"sync"
+	"time"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var _ = Describe("Test HTTP traffic from N deployment client -> 1 deployment server", func() {
+	Context("DeploymentsClientServer", func() {
+		destApp := "server"
+		sourceAppBaseName := "client"
+		var sourceNamespaces []string = []string{}
+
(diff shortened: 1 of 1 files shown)


tests/e2e/e2e_deployment_client_server_test.go
```diff
@@ -0,0 +1,171 @@
...
+				// Verify success conditions
+				for _, ns := range results {
+					for _, podResult := range ns {
+						if podResult.Err != nil || podResult.StatusCode != 200 {
+							return false
+						}
+					}
+				}
+				return true
+			}, 5, 150*time.Second)
```
shashankram, line 166: Could you declare the values here as a constant or at least a variable or with inline comments, so that it can understandable.

shashankram:
> (It currently does not verify that all servers reply, TrafficSplit test
> has some logic to do that that could be reused here if needed)

Please open an issue to track this.

tests/e2e/e2e_deployment_client_server_test.go
eduser25: > Please open an issue to track this.
Have a list if things on my whiteboard. Going to take care of it today.

shashankram (approved):

snehachhabria (approved):

eduser25:
> Please open an issue to track this.

#1827

PR #1823, opened 2020-10-13
tests/e2e: client-server deployment traffic split test

eduser25:
- Test runs HTTP from multiple source client deployments to one
destination traffic split, backed by multiple server services
- Number of clients and servers backing the traffic split is tunable

By default:
- 2 Clients with 5 replica
- 5 Servers with 2 replica
- Adds traffic targets
- Creates traffic split service and adds all servers as traffic split backends.

It then tests traffic from all clients to the traffic split FQDN.
- Server instances are instrumented to reply with their pod name. The test will
also verify all server backing the traffic split are actually and eventually
reached.

**Affected area:**
- Tests [X]

No

### DIFF
--- a/tests/e2e/e2e_trafficsplit_test.go
+++ b/tests/e2e/e2e_trafficsplit_test.go
@@ -1,0 +1,253 @@
+package e2e
+
+import (
+	"context"
+	"fmt"
+	"sync"
+	"time"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var _ = Describe("Test HTTP from N Clients deployments to 1 Server deployment backed with Traffic split test", func() {
+	Context("ClientServerTrafficSplit", func() {
+		const (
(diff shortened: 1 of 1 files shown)


tests/e2e/e2e_trafficsplit_test.go
```diff
@@ -0,0 +1,253 @@
...
+					}
+				}
+				td.T.Logf("Unique servers replied %d/%d",
+					len(serversSeen), numberOfServerServices*serverReplicaSet)
+
+				// Success conditions:
+				// - All clients have been answered consecutively 5 successful HTTP requests
+				// - We have seen all servers from the traffic split reply at least once
+				return curlSuccess && (len(serversSeen) == numberOfServerServices*serverReplicaSet)
+			}, 5, 150*time.Second)
```
shashankram, line 248: please use variables/constants/comments for these values.

eduser25: I want to file an issue to parametrize those. It's on my radar.

eduser25: https://github.com/openservicemesh/osm/issues/1825

snehachhabria (approved):

PR #1831, opened 2020-10-14
demo: move mesh delete command into the cleanup script

shashankram:
**Description**:
It is useful to run the cleanup script by itself. The mesh
delete command must also be run to cleanup mesh resources
created by install. This change moves the mesh delete command
into the cleanup script so that this single cleanup script
can be used to clean up all resources deployed by the demo.

**Affected area**:

- New Functionality      [ ]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [ ]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests                  [ ]
- CI System              [ ]
- Performance            [ ]
- Other                  [X]

`No`

### DIFF
--- a/demo/clean-kubernetes.sh
+++ b/demo/clean-kubernetes.sh
@@ -5,6 +5,8 @@ set -aueo pipefail
 # shellcheck disable=SC1091
 source .env
 
+bin/osm mesh uninstall -f --mesh-name "$MESH_NAME" --namespace "$K8S_NAMESPACE"
+
 for ns in "$BOOKWAREHOUSE_NAMESPACE" "$BOOKBUYER_NAMESPACE" "$BOOKSTORE_NAMESPACE" "$BOOKTHIEF_NAMESPACE" "$K8S_NAMESPACE"; do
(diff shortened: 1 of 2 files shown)


demo/clean-kubernetes.sh
```diff
@@ -5,6 +5,8 @@ set -aueo pipefail
 # shellcheck disable=SC1091
 source .env
 
+bin/osm mesh uninstall -f --mesh-name "$MESH_NAME" --namespace "$K8S_NAMESPACE"
```
michelleN, line 8: Should we add a `make build-osm` before running the uninstall command?

shashankram: We could but unnecessary in the context of current usage (always invoked via demo script). Is it ok to assume the cleanup script for the demo is run post install, at which time the cli binary is built already? I don't see value in even running this without the demo being run prior.

michelleN (approved):

nojnhuh (approved):

PR #1834, opened 2020-10-14
Update refs to v0.4.2

eduser25:
Updating references to 0.4.2 on charts, install and values


nojnhuh:
Could you also update the chart README with https://github.com/norwoodj/helm-docs? @michelleN has used that before. I know it's not documented in the release guide but we should add it there.

nojnhuh (approved):

snehachhabria (approved):

Issue #1798, opened 2020-10-06
Installing a traffic split makes source pods lose access to individual backend services


eduser25:
Filing as discussed with @ksubrmnn for doc/tracking purposes (the issue is known).

When Installing a traffic split, the source services loose access to the individual backed resources targeted by the traffic target itself.

- Traffic split in action:
```bash-4.3# curl traffic-split.server:80/tags
{
  "X_POD_NAME": "server0-54cc66b797-zvvhq"
}
bash-4.3# curl traffic-split.server:80/tags
{
  "X_POD_NAME": "server0-54cc66b797-flm7r"
}
bash-4.3# curl traffic-split.server:80/tags
{
  "X_POD_NAME": "server2-6bfd5c68cc-zt56j"
}
...
```
-  Test traffic from sources to individual services is still allowed:
```
bash-4.3# curl server0.server:80/tags -v
< HTTP/1.1 404 Not Found
< server: envoy
bash-4.3# curl server1.server:80/tags -v
< HTTP/1.1 404 Not Found
< server: envoy
```

The issue seems to linger somewhere around RDS, where only the Traffic Split gets on the route table  (not showing virtual-hosts which contain only traffic split dns resolvable names - hence the 404 for individual services later):
![image](https://user-images.githubusercontent.com/5503924/95259841-ad805380-07dc-11eb-9865-050895bdd003.png)

- SMI Traffic Access Policy  [X]
- SMI Traffic Split Policy   [X]
- Tests                      [X]

**Expected behavior**:
Services ought to not loose access to individual backend services allowed by traffic target (unless SMI were to specifically conceptually forbid by adding a split).

**How was OSM installed?**:
E2E testing, using OSM CLI installer

**Environment**:
- OSM version (use `osm version`): ToT
- Kubernetes version (use `kubectl version`): 1.18
- Size of cluster (number of worker nodes in the cluster): 10~15 pods

PR #1846, opened 2020-10-14
Build TrafficRoutes

ksubrmnn:
**Description**:
This PR creates new APIs/structs to build traffic routes in an extensible way as a first step to v1alpha3 traffic split. Only the API work has been done here, the routes are not yet consumed in any place in the code. 

This is a draft PR. There are still a few things to be resolved

**Affected area**:

- New Functionality      [ ]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [X]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests                  [ ]
- CI System              [ ]
- Performance            [ ]
- Other                  [ ]

### DIFF
--- a/pkg/catalog/routes.go
+++ b/pkg/catalog/routes.go
@@ -39,12 +39,18 @@ func (mc *MeshCatalog) ListTrafficPolicies(service service.MeshService) ([]traff
 		return nil, err
 	}
 
-	allTrafficPolicies, err := getTrafficPoliciesForService(mc, allRoutes, service)
+	trafficTargetsMap, err := getTrafficPoliciesForService(mc, allRoutes, service)
 	if err != nil {
 		log.Error().Err(err).Msgf("Could not get all traffic policies")
 		return nil, err
 	}
-	return allTrafficPolicies, nil
+
+	var matchedTrafficTargets []trafficpolicy.TrafficTarget
+	for _, trafficTarget := range trafficTargetsMap {
+		matchedTrafficTargets = append(matchedTrafficTargets, trafficTarget)
+	}
+
+	return matchedTrafficTargets, nil
 }
 
 // This function returns the list of connected services.
@@ -245,15 +251,11 @@ func getTrafficTargetFromSrcDstHash(hash string, name string, httpRoutes []traff
 
 // getTrafficPoliciesForService returns a list of TrafficTarget policies associated with a given MeshService.
 // The function consolidates all the routes between a source and destination in a single TrafficTarget object.
-func getTrafficPoliciesForService(mc *MeshCatalog, routePolicies map[trafficpolicy.TrafficSpecName]map[trafficpolicy.TrafficSpecMatchName]trafficpolicy.HTTPRoute, meshService service.MeshService) ([]trafficpolicy.TrafficTarget, error) {
+func getTrafficPoliciesForService(mc *MeshCatalog, routePolicies map[trafficpolicy.TrafficSpecName]map[trafficpolicy.TrafficSpecMatchName]trafficpolicy.HTTPRoute, meshService service.MeshService) (map[string]trafficpolicy.TrafficTarget, error) {
 	// 'srcDstTrafficTargetMap' is used to consolidate all routes from a source to a destination service.
 	// For the same source to destination if multiple routes are specified, all the routes are
 	// a part of a single TrafficTarget associated with that source and destination.
 	srcDstTrafficTargetMap := make(map[string]trafficpolicy.TrafficTarget)
-
-	// 'matchedTrafficTargets' is the list of all computed TrafficTarget policies that the given 'meshService`
-	// is a part of.
-	var matchedTrafficTargets []trafficpolicy.TrafficTarget
 
 	for _, trafficTargets := range mc.meshSpec.ListTrafficTargets() {
 		log.Debug().Msgf("Discovered TrafficTarget resource: %s/%s", trafficTargets.Namespace, trafficTargets.Name)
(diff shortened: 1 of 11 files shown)


pkg/catalog/routes.go
```diff
@@ -333,12 +335,31 @@ func getTrafficPoliciesForService(mc *MeshCatalog, routePolicies map[trafficpoli
...
 	}
 
 	log.Debug().Msgf("Traffic policies for service %s: %+v", meshService, matchedTrafficTargets)
-	return matchedTrafficTargets, nil
+	return matchedTrafficTargets, nil*/
+
+	return srcDstTrafficTargetMap, nil
+}
+
+func (mc *MeshCatalog) getTrafficRoutesForService(svc service.MeshService, trafficTargets map[string]trafficpolicy.TrafficTarget) []trafficpolicy.TrafficRoutes {
```
ksubrmnn, line 348: The arg for this function is what is returned by getTrafficPoliciesForService(). I will refactor a bit here after an initial review

pkg/catalog/routes.go
```diff
@@ -397,3 +418,75 @@ func listTrafficTargetPermutations(name string, srcServiceList []service.MeshSer
...
+		routeWeightedSvcs = append(routeWeightedSvcs, rWS)
+	}
+
+	return routeWeightedSvcs
+}
+
+func getRouteKey(route trafficpolicy.HTTPRoute, routeMap map[*trafficpolicy.HTTPRoute][]service.WeightedService) *trafficpolicy.HTTPRoute {
+	var matchingKey *trafficpolicy.HTTPRoute
+
+	for key := range routeMap {
```
eduser25, line 484: The map serves no purpose here I believe. 

You just do a linear search on the map to return the key, to later use the key as a map lookup index.
I have not looked into it deeply, but I'd probably suggest to find a way to index `HTTPRoute` by some usable key.

pkg/catalog/routes.go
```diff
@@ -397,3 +418,75 @@ func listTrafficTargetPermutations(name string, srcServiceList []service.MeshSer
...
+
+	// Retrieve the weighted clusters from traffic split
+	// Assumes no nested splits for now
+	// Assumes service would only be root once (not sure if this is validated)
+	servicesList := mc.meshSpec.ListTrafficSplitServices()
+	for _, activeService := range servicesList {
+		if activeService.RootService == svc.Name {
+			weightedSvcs = append(weightedSvcs, activeService)
+		}
+	}
```
eduser25, line 433: Correct me if wrong, this patch does not address https://github.com/openservicemesh/osm/issues/1798 right? (just making sure here)

pkg/catalog/routes.go
```diff
@@ -397,3 +418,75 @@ func listTrafficTargetPermutations(name string, srcServiceList []service.MeshSer
...
+func (mc *MeshCatalog) getRouteWeightedServices(src service.MeshService, dest service.MeshService, trafficTargets map[string]trafficpolicy.TrafficTarget) []trafficpolicy.RouteWeightedServices {
+	routeWeightedSvcs := []trafficpolicy.RouteWeightedServices{}
+
+	weightedBackends := mc.getWeightedBackendsForService(dest)
+
+	routeServicesMap := make(map[*trafficpolicy.HTTPRoute][]service.WeightedService)
+
+	for _, backend := range weightedBackends {
+		hash := hashSrcDstService(src, backend.Service)
+		routes := trafficTargets[hash].HTTPRoutes
```
eduser25, line 454: shouldn't `trafficTargets[hash]` be safeguarded and check it exists before accessing an element into it?
even if it were to be guaranteed by some other precondition, best practices to err check

pkg/tests/fixtures.go
```diff
@@ -57,8 +64,11 @@ const (
 	// WildcardWithHeadersMatchName is the name of the match object.
 	WildcardWithHeadersMatchName = "allow-everything-on-header"
 
-	// Weight is the percentage of the traffic to be sent this way in a traffic split scenario.
-	Weight = 100
+	// Weight1 is the weight of the traffic to be sent this way in a traffic split scenario.
+	Weight1 = 100
+
+	// Weight2is the weight of the traffic to be sent this way in a traffic split scenario.
```
snehachhabria, line 70:
```suggestion
	// Weight2 is the weight of the traffic to be sent this way in a traffic split scenario.
```

pkg/tests/fixtures.go
```diff
@@ -261,6 +264,35 @@ var (
 		},
 	}
 
+	// BookstoreV2TrafficTarget is a traffic target SMI object.
+	BookstoreV2TrafficTarget = target.TrafficTarget{
```
snehachhabria, line 268: Should we add another fixture and test where bookstorev1 and bookstorev2 have the same service account ?

pkg/catalog/routes.go
```diff
@@ -333,12 +335,31 @@ func getTrafficPoliciesForService(mc *MeshCatalog, routePolicies map[trafficpoli
...
+		trafficRoute := trafficpolicy.TrafficRoutes{
+			//Name:                  target.Name,
+			Source:                target.Source,
+			Destination:           target.Destination,
+			RouteWeightedServices: mc.getRouteWeightedServices(target.Source, target.Destination, trafficTargets),
+		}
+		trafficRoutes = append(trafficRoutes, trafficRoute)
+	}
+
+	//TODO: Add roots of traffic splits (will need some slight testing changes)
```
snehachhabria, line 361: Discussed offline to get all unique apex services and then build the traffic routes for these apex services

pkg/catalog/routes.go
```diff
@@ -333,12 +335,31 @@ func getTrafficPoliciesForService(mc *MeshCatalog, routePolicies map[trafficpoli
 		}
 	}
 
-	for _, trafficTarget := range srcDstTrafficTargetMap {
+	/*for _, trafficTarget := range srcDstTrafficTargetMap {
```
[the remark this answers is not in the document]
ksubrmnn: The function takes MeshService so the hash is already namespaced

See the [test](https://github.com/openservicemesh/osm/blob/main/pkg/catalog/routes_test.go#L315) for an example

Issue #1845, opened 2020-10-14
Integrate security scanning 


shashankram:
**Please describe the Improvement and/or Feature Request**
Scan source code for security vulnerabilities and integrate scanning with CI using gosec.

**Scope (please mark with X where applicable)**
- New Functionality          [X]
- Install                    [ ]
- SMI Traffic Access Policy  [ ]
- SMI Traffic Specs Policy   [ ]
- SMI Traffic Split Policy   [ ]
- Permissive Traffic Policy  [ ]
- Ingress                    [ ]
- Egress                     [ ]
- Envoy Control Plane        [ ]
- CLI Tool                   [ ]
- Metrics                    [ ]
- Certificate Management     [ ]
- Sidecar Injection          [ ]
- Logging                    [ ]
- Debugging                  [ ]
- Tests                      [ ]
- CI System                  [ ]
- Project Release            [ ]

PR #1850, opened 2020-10-15
Add security code checker using gosec linter

shashankram:
**Description**:
Enables security code checking using gosec linter. 
Some of the existing false positives have been annotated
so that gosec does not flag these when run locally.

Individual commits in the PR have more details 
regarding changes in that commit.

Part of #1845 

**Affected area**:

- New Functionality      [ ]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [ ]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [X]
- Tests                  [ ]
- CI System              [ ]
- Performance            [ ]
- Other                  [X]

`No`

### DIFF
--- a/pkg/health/health.go
+++ b/pkg/health/health.go
@@ -43,6 +43,9 @@ func (httpProbe HTTPProbe) Probe() (int, error) {
 	client := &http.Client{}
 
 	if httpProbe.Protocol == ProtocolHTTPS {
+		// Certificate validation is to be skipped for HTTPS probes
+		// similar to how k8s api server handles HTTPS probes.
+		// #nosec G402
 		transport := &http.Transport{
 			TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
 		}
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -8,6 +8,7 @@ linters:
     - gofmt
     - goimports
     - golint
+    - gosec
     - govet
     - misspell
     - unused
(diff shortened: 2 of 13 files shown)


eduser25 (approved):

pkg/health/health.go
```diff
@@ -43,6 +43,9 @@ func (httpProbe HTTPProbe) Probe() (int, error) {
 	client := &http.Client{}
 
 	if httpProbe.Protocol == ProtocolHTTPS {
+		// Certificate validation is to be skipped for HTTPS probes
+		// similar to how k8s api server handles HTTPS probes.
+		// #nosec G402
```
nojnhuh, line 48: I appreciate the explanations for each override in the commit messages. I think a comment like this right next to the override is also good just to shortcut having to `git blame` to see why the override is there. golangci-lint also has lots of ways to configure overrides in its config file if that's easier than tagging individual rule violations: https://golangci-lint.run/usage/configuration/. Each of these seem isolated enough that that might not be useful for these, but worth knowing about in any case.

shashankram: These are the very few individual false positives that are better off explicitly annotating. There is one more that I need to clean up G104 in a subsequent change, but its isn't a problem as of now due to the linter running only on new changes.

I also found it to be rather too verbose to explain each override inline in the code, and it can be rather distracting from a linting perspective.

nojnhuh (approved):

snehachhabria (approved):

PR #1852, opened 2020-10-15
pkg/injector: Add webhook time tracking facility

eduser25:
Adding a small piece to help track time spent by individual routines
handling webhooks.

It parses the timeout value from the webhook call, and based on some
static thresholds will log differently depending on how much time is
taking vs. the timeout value given by the url.

**Affected area**:

- Control Plane          [X]
- Performance            [X]

No

### DIFF
--- a/pkg/injector/errors.go
+++ b/pkg/injector/errors.go
@@ -4,5 +4,5 @@ import "github.com/pkg/errors"
 
 var (
 	errNamespaceNotFound   = errors.New("namespace not found")
-	errParseWebhookTimeout = errors.New("Could not read webhook timeout")
+	errParseWebhookTimeout = errors.New("could not read webhook timeout")
 )
--- a/pkg/injector/errors.go
+++ b/pkg/injector/errors.go
@@ -3,5 +3,6 @@ package injector
 import "github.com/pkg/errors"
 
 var (
-	errNamespaceNotFound = errors.New("namespace not found")
+	errNamespaceNotFound   = errors.New("namespace not found")
+	errParseWebhookTimeout = errors.New("Could not read webhook timeout")
 )
--- a/pkg/injector/profiling.go
+++ b/pkg/injector/profiling.go
@@ -3,8 +3,6 @@ package injector
 import (
 	"net/http"
 	"time"
-
-	"github.com/rs/zerolog"
 )
 
 // Helper to parse timeout variable from webhook URL
@@ -24,19 +22,11 @@ func readTimeout(req *http.Request) (*time.Duration, error) {
 }
 
 // Time tracking function for webhook processing.
+// Gompares duration and prints timeout values
 func webhookTimeTrack(start time.Time, timeout time.Duration) {
 	elapsed := time.Since(start)
-	var logEv *zerolog.Event
 	percentOfTimeout := float64(elapsed.Microseconds()) / float64(timeout.Microseconds())
 
-	if percentOfTimeout < 0.75 {
-		logEv = log.Debug()
-	} else if percentOfTimeout < 1 {
-		logEv = log.Warn()
-	} else {
-		// Error logging when going beyond timeout value to process a webhook
-		logEv = log.Error()
-	}
-	logEv.Msgf("Mutate Webhook took %v to execute (%.2f of it's timeout, %v)",
+	log.Debug().Msgf("Mutate Webhook took %v to execute (%.2f of it's timeout, %v)",
 		elapsed, percentOfTimeout, timeout)
 }
--- a/pkg/injector/profiling.go
+++ b/pkg/injector/profiling.go
@@ -1,0 +1,42 @@
+package injector
+
+import (
+	"net/http"
+	"time"
+
+	"github.com/rs/zerolog"
+)
+
+// Helper to parse timeout variable from webhook URL
+func readTimeout(req *http.Request) (*time.Duration, error) {
+	durationValue, found := req.URL.Query()[webhookTimeoutStr]
+	if !found || len(durationValue) != 1 {
+		log.Error().Msgf("Webhook timeout value not found in request")
+		return nil, errParseWebhookTimeout
+	}
+
+	val, err := time.ParseDuration(durationValue[0])
+	if err != nil {
+		log.Error().Msgf("Error parsing timeout value as duration: %v", err)
+		return nil, errParseWebhookTimeout
+	}
+	return &val, nil
+}
+
+// Time tracking function for webhook processing.
+func webhookTimeTrack(start time.Time, timeout time.Duration) {
+	elapsed := time.Since(start)
+	var logEv *zerolog.Event
+	percentOfTimeout := float64(elapsed.Microseconds()) / float64(timeout.Microseconds())
+
+	if percentOfTimeout < 0.75 {
+		logEv = log.Debug()
+	} else if percentOfTimeout < 1 {
+		logEv = log.Warn()
+	} else {
+		// Error logging when going beyond timeout value to process a webhook
+		logEv = log.Error()
+	}
+	logEv.Msgf("Mutate Webhook took %v to execute (%.2f of it's timeout, %v)",
+		elapsed, percentOfTimeout, timeout)
+}
--- a/pkg/injector/profiling.go
+++ b/pkg/injector/profiling.go
@@ -22,7 +22,7 @@ func readTimeout(req *http.Request) (*time.Duration, error) {
 }
 
 // Time tracking function for webhook processing.
-// Gompares duration and prints timeout values
+// Will calculate elapsed time since start and log debug how much time spent executing
 func webhookTimeTrack(start time.Time, timeout time.Duration) {
 	elapsed := time.Since(start)
 	percentOfTimeout := float64(elapsed.Microseconds()) / float64(timeout.Microseconds())
--- a/pkg/injector/profiling.go
+++ b/pkg/injector/profiling.go
@@ -7,7 +7,7 @@ import (
 
 // Helper to parse timeout variable from webhook URL
 func readTimeout(req *http.Request) (*time.Duration, error) {
-	durationValue, found := req.URL.Query()[webhookTimeoutStr]
+	durationValue, found := req.URL.Query()[webhookMutateTimeoutKey]
 	if !found || len(durationValue) != 1 {
 		log.Error().Msgf("Webhook timeout value not found in request")
 		return nil, errParseWebhookTimeout
--- a/pkg/injector/webhook.go
+++ b/pkg/injector/webhook.go
@@ -51,7 +51,7 @@ const (
 	WebhookHealthPath = "/healthz"
 
 	// webhookTimeoutStr is the url variable name for timeout
-	webhookTimeoutStr = "timeout"
+	webhookMutateTimeoutKey = "timeout"
 )
 
 // NewWebhook starts a new web server handling requests from the injector MutatingWebhookConfiguration
--- a/pkg/injector/webhook.go
+++ b/pkg/injector/webhook.go
@@ -8,10 +8,12 @@ import (
 	"io/ioutil"
 	"net/http"
 	"strings"
+	"time"
 
 	"github.com/google/uuid"
 
 	"github.com/pkg/errors"
+	"github.com/rs/zerolog"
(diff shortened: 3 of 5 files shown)


pkg/injector/errors.go
```diff
@@ -3,5 +3,6 @@ package injector
 import "github.com/pkg/errors"
 
 var (
-	errNamespaceNotFound = errors.New("namespace not found")
+	errNamespaceNotFound   = errors.New("namespace not found")
+	errParseWebhookTimeout = errors.New("Could not read webhook timeout")
```
eduser25, line 7:
```suggestion
	errParseWebhookTimeout = errors.New("could not read webhook timeout")
```

eduser25:
log example:
```
{"level":"debug","component":"sidecar-injector","time":"2020-10-15T02:55:38Z","file":"webhook.go:166","message":"Mutate Webhook for pod server/server took 1.841792726s time to execute (0.06 of it's timeout, 30s)"}
```

pkg/injector/webhook.go
```diff
@@ -47,6 +49,9 @@ const (
 
 	// WebhookHealthPath is the HTTP path at which the health of the webhook can be queried
 	WebhookHealthPath = "/healthz"
+
+	// webhookTimeoutStr is the url variable name for timeout
```
shashankram, line 53: What does "url variable name for timeout" mean here?

pkg/injector/webhook.go
```diff
@@ -128,9 +133,54 @@ func (wh *webhook) healthHandler(w http.ResponseWriter, req *http.Request) {
...
+		logEv = log.Debug()
+	} else if percentOfTimeout < 1 {
+		logEv = log.Warn()
+	} else {
+		// Error logging when going beyond timeout value to process a webhook
+		logEv = log.Error()
+	}
+	logEv.Msgf("Mutate Webhook for %s took %v to execute (%.2f of it's timeout, %v)",
+		(*resourceName), elapsed, percentOfTimeout, timeout)
+}
```
shashankram, line 168: Please move these to a new file called `profile.go`

pkg/injector/webhook.go
```diff
@@ -180,20 +230,30 @@ func (wh *webhook) mutateHandler(w http.ResponseWriter, req *http.Request) {
 	log.Debug().Msg("Done responding to admission request")
 }
 
-func (wh *webhook) mutate(req *v1beta1.AdmissionRequest) *v1beta1.AdmissionResponse {
+func (wh *webhook) mutate(req *v1beta1.AdmissionRequest) (*v1beta1.AdmissionResponse, string) {
```
shashankram, line 233: what string is being returned here is unclear.

pkg/injector/webhook_test.go
```diff
@@ -534,7 +534,7 @@ var _ = Describe("Testing Injector Functions", func() {
 		}
 
 		// Action !!
-		actual := wh.mutate(nil)
+		actual, _ := wh.mutate(nil)
```
shashankram, line 537: please test the additional values being returned in the same or new test.

pkg/injector/webhook.go
```diff
@@ -162,7 +212,7 @@ func (wh *webhook) mutateHandler(w http.ResponseWriter, req *http.Request) {
 		log.Error().Err(err).Msg("Error decoding admission request")
 		admissionResp.Response = toAdmissionError(err)
 	} else {
-		admissionResp.Response = wh.mutate(admissionReq.Request)
+		admissionResp.Response, webhookResourceName = wh.mutate(admissionReq.Request)
```
shashankram, line 215: where is `webhookResourceName` being used?

shashankram (changes requested):
Thanks for the change.

I'd like to see a more concrete description of the time tracking facility being introduced since its hard to reason from the code exactly what's going on.

pkg/injector/webhook.go
eduser25: webhook urls come with a timeout variable in it, ie: `URL=/mutate?timeout=30s`

pkg/injector/webhook.go
eduser25: that's probably what I like the least. I wanted something to identify the webhook to see what actual resource took more/less to execute. I'm basically returning the podname in this case.

If you some better idea or other thoughts let me know.

pkg/injector/webhook.go
eduser25: it's read and used by the defer function. Used to log what resource took how much time.

pkg/injector/webhook.go
eduser25: fair enough

pkg/injector/webhook_test.go
eduser25: This test just enters the function and immediately exits at the very first condition (nil request). Didn't bother as it is not relevant at all in this instance.

eduser25: > its hard to reason from the code exactly what's going on.

Let's see if I can help.
- on mutate handler, we read timeout from the url.
- if reading the url succeeded, we defer a time-tracking function that will run when `mutateHandler` is done and will print how much time we spent in the handler itself.

Most of the ... ugly code is getting the podname. We could obviate that code and it'd look way cleaner, but the log message would be less relevant without knowing which the resource.

eduser25:
> its hard to reason from the code exactly what's going on.

Let's see if I can help.
- on mutate handler, we read timeout from the url.
- if reading the url succeeded, we defer a time-tracking function that will run when `mutateHandler` is done and will print how much time we spent in the handler itself.

Most of the ... ugly code is getting the podname. We could obviate that code and it'd look way cleaner, but the log message would be less relevant without knowing which the resource.

pkg/injector/webhook.go
shashankram: Lets rename this to `webhookMutateTimeoutKey` or something along that line.

pkg/injector/webhook.go
shashankram: We will not know the pod name at this time because the pod isn't even  created, so some arbitrary autogenerated name might not be useful.

pkg/injector/webhook.go
```diff
@@ -128,9 +133,54 @@ func (wh *webhook) healthHandler(w http.ResponseWriter, req *http.Request) {
...
+	}
+	logEv.Msgf("Mutate Webhook for %s took %v to execute (%.2f of it's timeout, %v)",
+		(*resourceName), elapsed, percentOfTimeout, timeout)
+}
+
 func (wh *webhook) mutateHandler(w http.ResponseWriter, req *http.Request) {
 	log.Info().Msgf("Request received: Method=%v, URL=%v", req.Method, req.URL)
 
+	// Til we parse resource name, we will set it to something generic in case
+	// we are done with the webhook before we even get to parse it
```
shashankram, line 174: This comment isn't clear to me.

pkg/injector/webhook.go
```diff
@@ -128,9 +133,54 @@ func (wh *webhook) healthHandler(w http.ResponseWriter, req *http.Request) {
...
+	webhookResourceName := "<unparsed>"
+	// Read timeout from request
+	timeoutValue, err := readTimeout(req)
+	if err != nil {
+		log.Error().Msgf("Could not read timeout from request url: %v", err)
```
shashankram, line 179: Under what circumstances will this error occur? Is there a scenario where the request will not have a timeout?

pkg/injector/webhook.go
```diff
@@ -128,9 +133,54 @@ func (wh *webhook) healthHandler(w http.ResponseWriter, req *http.Request) {
...
+}
+
 func (wh *webhook) mutateHandler(w http.ResponseWriter, req *http.Request) {
 	log.Info().Msgf("Request received: Method=%v, URL=%v", req.Method, req.URL)
 
+	// Til we parse resource name, we will set it to something generic in case
+	// we are done with the webhook before we even get to parse it
+	webhookResourceName := "<unparsed>"
+	// Read timeout from request
+	timeoutValue, err := readTimeout(req)
```
shashankram, line 177:
```suggestion
	reqTimeout, err := readTimeout(req)
```

pkg/injector/webhook.go
```diff
@@ -128,9 +133,54 @@ func (wh *webhook) healthHandler(w http.ResponseWriter, req *http.Request) {
...
+	if err != nil {
+		log.Error().Msgf("Could not read timeout from request url: %v", err)
+	} else {
+		defer webhookTimeTrack(&webhookResourceName, time.Now(), *timeoutValue)
```
shashankram, line 181:
```suggestion
		defer webhookTimeTrack(&webhookResourceName, time.Now(), *reqTimeout)
```

pkg/injector/webhook.go
```diff
@@ -128,9 +133,54 @@ func (wh *webhook) healthHandler(w http.ResponseWriter, req *http.Request) {
...
+
+	val, err := time.ParseDuration(durationValue[0])
+	if err != nil {
+		log.Error().Msgf("Error parsing timeout value as duration: %v", err)
+		return nil, errParseWebhookTimeout
+	}
+	return &val, nil
+}
+
+// Time tracking function for webhook processing.
```
shashankram, line 152: Please describe in comments how the time thresholds are being used.

pkg/injector/webhook.go
```diff
@@ -128,9 +133,54 @@ func (wh *webhook) healthHandler(w http.ResponseWriter, req *http.Request) {
...
+
+	if percentOfTimeout < 0.75 {
+		logEv = log.Debug()
+	} else if percentOfTimeout < 1 {
+		logEv = log.Warn()
+	} else {
+		// Error logging when going beyond timeout value to process a webhook
+		logEv = log.Error()
+	}
+	logEv.Msgf("Mutate Webhook for %s took %v to execute (%.2f of it's timeout, %v)",
```
shashankram, line 166: Why are these custom logs necessary? K8s events should be logging timeouts anyway.

pkg/injector/webhook.go
```diff
@@ -180,20 +230,30 @@ func (wh *webhook) mutateHandler(w http.ResponseWriter, req *http.Request) {
...
 
+	// Gather a meaningful resource name for this request
+	var pName string
+	if len(pod.Name) != 0 {
+		pName = pod.Name
+	} else {
+		pName = pod.GenerateName
+	}
+	outResourceName = fmt.Sprintf("%s/%s", req.Namespace, pName)
+
```
shashankram, line 256: Generating a pod name is not very meaningful. How would one compare this with actual k8s pods?

shashankram:
> > its hard to reason from the code exactly what's going on.
> 
> Let's see if I can help.
> 
> * on mutate handler, we read timeout from the url.
> * if reading the url succeeded, we defer a time-tracking function that will run when `mutateHandler` is done and will print how much time we spent in the handler itself.
> 
> Most of the ... ugly code is getting the podname. We could obviate this part of the code and it'd look way cleaner, but the log message would be less relevant without knowing which resource came triggered the webhook.

I am wondering why this tracking facility is even required. The PR does not describe why this functionality is required.
When the webhook experiences a timeout, k8s API server will log an event regarding the timeout, so why is this even needed.

pkg/injector/webhook.go
eduser25: This one I do not know. I'm being cautions and just avoiding the call If it can't read it.

pkg/injector/webhook.go
eduser25: synced offlined. We agreed timeout messaging/events are going to be seen in K8s events to begin with, so arguably we do not need to strictly log error on those from this side of the fence.
We however might want to keep at least debug log for this messages to debug/profile timings without requiring to plug in pprof in the picture.

draychev (dismissed):
This seems very useful!

shashankram:
@eduser25 are you planning to iterate on the changes we discussed and agreed upon or do you prefer to merge this as is?

shashankram (approved):

pkg/injector/profiling.go
```diff
@@ -0,0 +1,32 @@
+package injector
```
shashankram, line 1: missing tests.

shashankram (approved):

snehachhabria (approved):

PR #1865, opened 2020-10-16
ref(routes): ListTrafficTargetPermutation using TrafficTarget fields …

ksubrmnn:
…directly

**Description**:
Refactors ListTrafficTargetPermutation to include all of the logic for getting services from service accounts listed in TrafficTarget. This makes the code a bit more modular and easier to read.

**Affected area**:

- New Functionality      [ ]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [X]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests                  [ ]
- CI System              [ ]
- Performance            [ ]
- Other                  [ ]

No

### DIFF
--- a/pkg/catalog/routes.go
+++ b/pkg/catalog/routes.go
@@ -7,6 +7,7 @@ import (
 
 	mapset "github.com/deckarep/golang-set"
 	"github.com/pkg/errors"
+	target "github.com/servicemeshinterface/smi-sdk-go/pkg/apis/access/v1alpha2"
 	corev1 "k8s.io/api/core/v1"
 
 	"github.com/openservicemesh/osm/pkg/constants"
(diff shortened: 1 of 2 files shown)


snehachhabria (approved):

pkg/catalog/routes.go
```diff
@@ -381,19 +364,40 @@ func getDefaultWeightedClusterForService(meshService service.MeshService) servic
...
+	if destErr != nil {
+		log.Error().Msgf("TrafficTarget %s/%s could not get destination services for service account %s", trafficTarget.Namespace, trafficTarget.Name, dstNamespacedServiceAcc.String())
+		return nil, destErr
+	}
+
 	trafficPolicies := make([]trafficpolicy.TrafficTarget, 0, len(srcServiceList)*len(destServiceList))
 
 	for _, destService := range destServiceList {
 		for _, srcService := range srcServiceList {
 			trafficTarget := trafficpolicy.TrafficTarget{
```
eduser25, line 393: local scope keeps it safe, but just to avoid confusion between the param and the local var maybe:
```suggestion
			tempTrafficTarget := trafficpolicy.TrafficTarget{
				Name:        utils.GetTrafficTargetName(trafficTarget.Name, srcService, destService),
				Destination: destService,
				Source:      srcService,
			}
			trafficPolicies = append(trafficPolicies, tempTrafficTarget)
```

pkg/catalog/routes.go
```diff
@@ -261,30 +262,12 @@ func getTrafficPoliciesForService(mc *MeshCatalog, routePolicies map[trafficpoli
...
-			namespacedServiceAccount := service.K8sServiceAccount{
-				Namespace: trafficSources.Namespace,
-				Name:      trafficSources.Name,
-			}
-
-			srcServiceList, srcErr := mc.GetServicesForServiceAccount(namespacedServiceAccount)
-			if srcErr != nil {
-				log.Error().Msgf("TrafficTarget %s/%s could not get source services for service account %s", trafficTargets.Namespace, trafficTargets.Name, fmt.Sprintf("%s/%s", trafficSources.Namespace, trafficSources.Name))
-				return nil, srcErr
+			trafficTargetPermutations, err := mc.listTrafficTargetPermutations(*trafficTargets, trafficSources, trafficTargets.Spec.Destination)
```
eduser25, line 266: nit: you could keep using the pointer probably, to avoid the compiler making a copy of the struct if not needed

eduser25 (approved):

pkg/catalog/routes.go
ksubrmnn: Agreed. I will update this in a different PR I'm working on.

pkg/catalog/routes.go
ksubrmnn: Yeah, I thought about this. It's a trade off for performance vs restricting the function to not make any changes to the traffic target object. I guess it might not matter if ListTrafficTargets() returns pointers anyway. I can update in a separate PR.

PR #1867, opened 2020-10-19
tests(e2e): add traffic split test using shared service account

nojnhuh:
**Description**: This test is based off the existing TrafficSplit test. The only difference is the server backends all use the same ServiceAccount for this test vs. each having their own ServiceAccount as in the existing test. This change adds a new test instead of modifying the existing test to preserve readability of each scenario.

**Affected area**:

- New Functionality      [ ]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [ ]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests                  [X]
- CI System              [ ]
- Performance            [ ]
- Other                  [ ]

No

### DIFF
--- a/tests/e2e/e2e_trafficsplit_same_sa_test.go
+++ b/tests/e2e/e2e_trafficsplit_same_sa_test.go
@@ -1,0 +1,257 @@
+package e2e
+
+import (
+	"context"
+	"fmt"
+	"sync"
+	"time"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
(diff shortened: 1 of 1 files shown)


snehachhabria (approved):
Thanks a lot for adding this e2e! I'm glad that both the scenarios of multiple service accounts and shared service accounts with traffic split it being tested

eduser25 (approved):

PR #1868, opened 2020-10-19
feat(metrics): Disable Prometheus by default

### LINKED ISSUE #1753: disable metrics stack by default in osm install and in osm chart
At the moment, we deploy Prometheus by default which enables us to populate the pre-built Grafana dashboards via the `osm dashboard` command. Most folks already have a Prometheus instance running in their cluster and may want to leverage their own and not spin up a new instance which requires more compute. This issue is to disable the metrics stack (Prometheus and Grafana) in the default instance. There are still flags you can use to toggle this option on. The demo documentation should still provide the same experience leveraging the metrics stack but should show the commands to turn these features on via flags (or helm chart values).

We do not currently have any metrics related osm cli commands so this change will not break any current behavior. In the future, we can discuss whether it makes more sense to spin up a Prometheus by default so that CLI commands can leverage them.

ksubrmnn:
**Description**:
This PR disables Prometheus by default in `osm install` but deploys prometheus by default for the demo only.

Fixes #1753 #1893 #1897 

**Affected area**:

- New Functionality      [ ]
- Documentation          [ ]
- Install                [X]
- Control Plane          [ ]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [X]
- SMI Policy             [ ]
- Security               [ ]
- Tests                  [ ]
- CI System              [ ]
- Performance            [ ]
- Other                  [ ]

No

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -90,17 +90,17 @@ export BOOKWAREHOUSE_NAMESPACE=bookwarehouse
 # Default: false
 # export ENABLE_EGRESS=true
 
-# optional: ENABLE_GRAFANA (true/false)
+# optional: DEPLOY_GRAFANA (true/false)
 # Default: true
-# export ENABLE_GRAFANA=true
+# export DEPLOY_GRAFANA=true
 
 # optional: ENABLE_FLUENTBIT (true/false)
 # Default: false
 # export ENABLE_FLUENTBIT=true
 
-# optional: ENABLE_PROMETHEUS_DEPLOYMENT (true/false)
+# optional: DEPLOY_PROMETHEUS (true/false)
 # Default: true
-# export ENABLE_PROMETHEUS_DEPLOYMENT=true
+# export DEPLOY_PROMETHEUS=true
 
 # optional: ENABLE_PROMETHEUS_SCRAPING (true/false)
 # Default: true
--- a/.env.example
+++ b/.env.example
@@ -98,13 +98,13 @@ export BOOKWAREHOUSE_NAMESPACE=bookwarehouse
 # Default: false
 # export ENABLE_FLUENTBIT=true
 
-# optional: ENABLE_PROMETHEUS (true/false)
+# optional: ENABLE_PROMETHEUS_DEPLOYMENT (true/false)
 # Default: true
-# export ENABLE_PROMETHEUS=true
+# export ENABLE_PROMETHEUS_DEPLOYMENT=true
 
 # optional: ENABLE_PROMETHEUS_SCRAPING (true/false)
-# Default: false
-# export ENABLE_PROMETHEUS_SCRAPING=false
+# Default: true
+# export ENABLE_PROMETHEUS_SCRAPING=true
 
 # optional: Maximum of iterations to test for expected return codes. 0 means unlimited.
 # export CI_MAX_ITERATIONS_THRESHOLD=0
--- a/.env.example
+++ b/.env.example
@@ -91,12 +91,20 @@ export BOOKWAREHOUSE_NAMESPACE=bookwarehouse
 # export ENABLE_EGRESS=true
 
 # optional: ENABLE_GRAFANA (true/false)
-# Default: false
+# Default: true
 # export ENABLE_GRAFANA=true
 
 # optional: ENABLE_FLUENTBIT (true/false)
 # Default: false
 # export ENABLE_FLUENTBIT=true
+
+# optional: ENABLE_PROMETHEUS (true/false)
+# Default: true
+# export ENABLE_PROMETHEUS=true
+
+# optional: ENABLE_PROMETHEUS_SCRAPING (true/false)
+# Default: false
+# export ENABLE_PROMETHEUS_SCRAPING=false
 
 # optional: Maximum of iterations to test for expected return codes. 0 means unlimited.
 # export CI_MAX_ITERATIONS_THRESHOLD=0
--- a/charts/osm/README.md
+++ b/charts/osm/README.md
@@ -21,7 +21,7 @@ A Helm chart to install the OSM control plane on Kubernetes
 | OpenServiceMesh.enableFluentbit | bool | `false` |  |
 | OpenServiceMesh.enablePermissiveTrafficPolicy | bool | `false` |  |
 | OpenServiceMesh.enablePrometheusDeployment | bool | `false` |  |
-| OpenServiceMesh.enablePrometheusScraping | bool | `true` | Automatically set to true if `enablePrometheusDeployment` is true |
+| OpenServiceMesh.enablePrometheusScraping | bool | `true` | Cannot be false if `enablePrometheusDeployment` is true |
 | OpenServiceMesh.envoyLogLevel | string | `"error"` |  |
 | OpenServiceMesh.grafana.port | int | `3000` |  |
 | OpenServiceMesh.image.pullPolicy | string | `"IfNotPresent"` |  |
--- a/charts/osm/README.md
+++ b/charts/osm/README.md
@@ -17,11 +17,11 @@ A Helm chart to install the OSM control plane on Kubernetes
 | OpenServiceMesh.enableBackpressureExperimental | bool | `false` |  |
 | OpenServiceMesh.enableDebugServer | bool | `false` |  |
 | OpenServiceMesh.enableEgress | bool | `false` |  |
-| OpenServiceMesh.enableGrafana | bool | `false` |  |
+| OpenServiceMesh.deployGrafana | bool | `false` |  |
 | OpenServiceMesh.enableFluentbit | bool | `false` |  |
 | OpenServiceMesh.enablePermissiveTrafficPolicy | bool | `false` |  |
-| OpenServiceMesh.enablePrometheusDeployment | bool | `false` |  |
-| OpenServiceMesh.enablePrometheusScraping | bool | `true` | Cannot be false if `enablePrometheusDeployment` is true |
+| OpenServiceMesh.deployPrometheus | bool | `false` |  |
+| OpenServiceMesh.enablePrometheusScraping | bool | `true` | Cannot be false if `deployPrometheus` is true |
 | OpenServiceMesh.envoyLogLevel | string | `"error"` |  |
 | OpenServiceMesh.grafana.port | int | `3000` |  |
 | OpenServiceMesh.image.pullPolicy | string | `"IfNotPresent"` |  |
--- a/charts/osm/README.md
+++ b/charts/osm/README.md
@@ -20,7 +20,8 @@ A Helm chart to install the OSM control plane on Kubernetes
 | OpenServiceMesh.enableGrafana | bool | `false` |  |
 | OpenServiceMesh.enableFluentbit | bool | `false` |  |
 | OpenServiceMesh.enablePermissiveTrafficPolicy | bool | `false` |  |
-| OpenServiceMesh.enablePrometheus | bool | `true` |  |
+| OpenServiceMesh.enablePrometheus | bool | `false` |  |
+| OpenServiceMesh.enablePrometheusScraping | bool | `false` | Automatically set to true if `enablePrometheus` is true |
 | OpenServiceMesh.envoyLogLevel | string | `"error"` |  |
 | OpenServiceMesh.grafana.port | int | `3000` |  |
 | OpenServiceMesh.image.pullPolicy | string | `"IfNotPresent"` |  |
--- a/charts/osm/README.md
+++ b/charts/osm/README.md
@@ -20,8 +20,8 @@ A Helm chart to install the OSM control plane on Kubernetes
 | OpenServiceMesh.enableGrafana | bool | `false` |  |
 | OpenServiceMesh.enableFluentbit | bool | `false` |  |
 | OpenServiceMesh.enablePermissiveTrafficPolicy | bool | `false` |  |
-| OpenServiceMesh.enablePrometheus | bool | `false` |  |
-| OpenServiceMesh.enablePrometheusScraping | bool | `false` | Automatically set to true if `enablePrometheus` is true |
+| OpenServiceMesh.enablePrometheusDeployment | bool | `false` |  |
+| OpenServiceMesh.enablePrometheusScraping | bool | `true` | Automatically set to true if `enablePrometheusDeployment` is true |
 | OpenServiceMesh.envoyLogLevel | string | `"error"` |  |
 | OpenServiceMesh.grafana.port | int | `3000` |  |
 | OpenServiceMesh.image.pullPolicy | string | `"IfNotPresent"` |  |
--- a/charts/osm/templates/osm-configmap.yaml
+++ b/charts/osm/templates/osm-configmap.yaml
@@ -8,7 +8,7 @@ data:
   egress: {{ .Values.OpenServiceMesh.enableEgress | quote }}
   envoy_log_level: {{ .Values.OpenServiceMesh.envoyLogLevel | quote }}
   enable_debug_server: {{ .Values.OpenServiceMesh.enableDebugServer | quote }}
-  prometheus_scraping: "true"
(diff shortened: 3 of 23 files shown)


shashankram (changes requested):
Could you reference https://github.com/openservicemesh/osm/issues/1753 in the PR if related.

Also we need to update the following docs to reflect whether Prometheus and Grafana are deployed/enabled by default or not. Folks trying the manual demo aren't sure what's the case going by the docs:
https://github.com/openservicemesh/osm/blob/main/docs/example/README.md
https://github.com/openservicemesh/osm/blob/main/docs/patterns/observability.md

snehachhabria:
I noticed that the `prometheus_scraping` is set to true : https://github.com/openservicemesh/osm/blob/d3ade1e9be0738645d1eede55cede06f67479c17/charts/osm/templates/osm-configmap.yaml#L11 , i think this should this be set to false by default and have it as true only for the demo

charts/osm/templates/osm-configmap.yaml
```diff
@@ -8,7 +8,7 @@ data:
   egress: {{ .Values.OpenServiceMesh.enableEgress | quote }}
   envoy_log_level: {{ .Values.OpenServiceMesh.envoyLogLevel | quote }}
   enable_debug_server: {{ .Values.OpenServiceMesh.enableDebugServer | quote }}
-  prometheus_scraping: "true"
+  prometheus_scraping: {{ .Values.OpenServiceMesh.enablePrometheus | quote }}
```
snehachhabria, line 11: The `.Values.OpenServiceMesh.enablePrometheus` is used to deploy Prometheus resources in osm but this field needs to be set true for BYO Prometheus, so I am not 100% sure if we should link the two.

docs/example/README.md
```diff
@@ -208,7 +208,7 @@ Wait for the changes to propagate and observe the counters increment for booksto
 - http://localhost:8082 - **bookstore-v2**
 
 ## Inspect Dashboards
-OSM ships with a set of pre-configured Grafana dashboards. **NOTE** If you still have the additional terminal still running the `./scripts/port-forward-all.sh` script, go ahead and `CTRL+C` to terminate the port forwarding. The `osm dashboard` port redirection will not work simultaneously with the port forwarding script still running. The `osm dashboard` can be viewed with the following command:
+OSM can be configured to deploy Grafana dashboards using the `--enable-grafana` flag in `osm install`. **NOTE** If you still have the additional terminal still running the `./scripts/port-forward-all.sh` script, go ahead and `CTRL+C` to terminate the port forwarding. The `osm dashboard` port redirection will not work simultaneously with the port forwarding script still running. The `osm dashboard` can be viewed with the following command:
```
SanyaKochhar, line 211: I was wondering whether it may be useful to mention the flags in the [Install OSM Control Plane](https://github.com/ksubrmnn/osm/blob/disable_prometheus/docs/example/README.md#install-osm-control-plane) section so that users don't reach the end of the demo and only then realize they can't use metrics if they hadn't enabled them already. Would resolve #1893 and #1897

charts/osm/templates/osm-configmap.yaml
ksubrmnn: I see. I can make a separate flag?

snehachhabria: yes, a separate flag would help making it configurable for the demo vs a regular install

michelleN (dismissed):

docs/example/README.md
ksubrmnn: Addressed

charts/osm/values.yaml
```diff
@@ -32,7 +32,8 @@ OpenServiceMesh:
   enablePermissiveTrafficPolicy: false
   enableBackpressureExperimental: false
   enableEgress: false
-  enablePrometheus: true
+  enablePrometheus: false
```
shashankram, line 35: Can we rename this to `enablePrometheusDeployment`

charts/osm/values.yaml
```diff
@@ -32,7 +32,8 @@ OpenServiceMesh:
...
+  enablePrometheusScraping: false
```
shashankram, line 36: Let's keep the default to true to simplify user experience. Even though the scraping is set to on, it will only scrape config if metrics is explicitly enabled on a namespace.

.env.example
```diff
@@ -91,13 +91,21 @@ export BOOKWAREHOUSE_NAMESPACE=bookwarehouse
...
 # optional: ENABLE_FLUENTBIT (true/false)
 # Default: false
 # export ENABLE_FLUENTBIT=true
 
+# optional: ENABLE_PROMETHEUS (true/false)
+# Default: true
+# export ENABLE_PROMETHEUS=true
+
+# optional: ENABLE_PROMETHEUS_SCRAPING (true/false)
+# Default: false
```
snehachhabria, line 106:
```suggestion
# Default: true
```

.env.example
```diff
@@ -91,13 +91,21 @@ export BOOKWAREHOUSE_NAMESPACE=bookwarehouse
 # export ENABLE_EGRESS=true
 
 # optional: ENABLE_GRAFANA (true/false)
-# Default: false
+# Default: true
```
snehachhabria, line 94: Isn't the default false ? OSM by default won't deploy Prometheus or Grafana

.env.example
```diff
@@ -91,13 +91,21 @@ export BOOKWAREHOUSE_NAMESPACE=bookwarehouse
...
 # export ENABLE_GRAFANA=true
 
 # optional: ENABLE_FLUENTBIT (true/false)
 # Default: false
 # export ENABLE_FLUENTBIT=true
 
+# optional: ENABLE_PROMETHEUS (true/false)
+# Default: true
```
snehachhabria, line 102: Same as above, Also we should probably rename this to DEPLOY_PROMETHEUS, WDYT ?

.env.example
```diff
@@ -91,13 +91,21 @@ export BOOKWAREHOUSE_NAMESPACE=bookwarehouse
...
+# export ENABLE_PROMETHEUS=true
+
+# optional: ENABLE_PROMETHEUS_SCRAPING (true/false)
+# Default: false
+# export ENABLE_PROMETHEUS_SCRAPING=false
```
snehachhabria, line 107:
```suggestion
# export ENABLE_PROMETHEUS_SCRAPING=true
```

charts/osm/README.md
```diff
@@ -20,7 +20,8 @@ A Helm chart to install the OSM control plane on Kubernetes
 | OpenServiceMesh.enableGrafana | bool | `false` |  |
 | OpenServiceMesh.enableFluentbit | bool | `false` |  |
 | OpenServiceMesh.enablePermissiveTrafficPolicy | bool | `false` |  |
-| OpenServiceMesh.enablePrometheus | bool | `true` |  |
+| OpenServiceMesh.enablePrometheus | bool | `false` |  |
+| OpenServiceMesh.enablePrometheusScraping | bool | `false` | Automatically set to true if `enablePrometheus` is true |
```
snehachhabria, line 24:
```suggestion
| OpenServiceMesh.enablePrometheusScraping | bool | `true` |  |
```

demo/run-osm-demo.sh
```diff
@@ -29,8 +29,10 @@ CTR_TAG="${CTR_TAG:-$(git rev-parse HEAD)}"
 IMAGE_PULL_POLICY="${IMAGE_PULL_POLICY:-Always}"
 ENABLE_DEBUG_SERVER="${ENABLE_DEBUG_SERVER:-true}"
 ENABLE_EGRESS="${ENABLE_EGRESS:-false}"
-ENABLE_GRAFANA="${ENABLE_GRAFANA:-false}"
+ENABLE_GRAFANA="${ENABLE_GRAFANA:-true}"
 ENABLE_FLUENTBIT="${ENABLE_FLUENTBIT:-false}"
+ENABLE_PROMETHEUS="${ENABLE_PROMETHEUS:-true}"
+ENABLE_PROMETHEUS_SCRAPING="${ENABLE_PROMETHEUS_SCRAPING:-false}"
```
snehachhabria, line 35:
```suggestion
ENABLE_PROMETHEUS_SCRAPING="${ENABLE_PROMETHEUS_SCRAPING:-true}"
```

docs/osm_config_map.md
```diff
@@ -10,7 +10,7 @@ OSM deploys a configMap `osm-config` as a part of its control plane (in the same
 | egress | bool | true, false| `"false"` | Enables egress in the mesh. |
 | enable_debug_server | bool | true, false| `"true"` | Enables a debug endpoint on the osm-controller pod to list information regarding the mesh such as proxy connections, certificates, and SMI policies. |
 | envoy_log_level | string | trace, debug, info, warning, warn, error, critical, off | `"error"` | Sets the logging verbosity of Envoy proxy sidecar, only applicable to newly created pods joining the mesh. |
-| prometheus_scraping | bool | true, false | `"true"` | Enables Prometheus metrics scraping on sidecar proxies. |
+| prometheus_scraping | bool | true, false | `"false"` | Enables Prometheus metrics scraping on sidecar proxies. |
```
snehachhabria, line 13:
```suggestion
| prometheus_scraping | bool | true, false | `"true"` | Enables Prometheus metrics scraping on sidecar proxies. |
```

docs/patterns/observability/metrics.md
```diff
@@ -115,7 +116,7 @@ Metrics scraping can be configured using the `osm metrics` command. By default,
 For metrics to be scraped, the following prerequisites must be met:
 - The namespace must be a part of the mesh, ie. it must be labeled with the `openservicemesh.io/monitored-by` label with an appropriate mesh name. This can be done using the `osm namespace add` command.
 - A running service able to scrap Prometheus endpoints. OSM provides configuration for an [automatic bringup of Prometheus](#automatic-bring-up); alternatively users can [bring their own Prometheus](#byo-bring-your-own).
-- The `prometheus_scraping` config key in osm-controller's `osm-config` ConfigMap must be set to `"true"`, which is the default configuration.
+- The `prometheus_scraping` config key in osm-controller's `osm-config` ConfigMap must be set to `"true"`. It is `"false"` by default.
```
snehachhabria, line 119: This can be reverted, as we want the default as true

docs/patterns/observability/metrics.md
```diff
@@ -35,7 +36,7 @@ The following section will document the additional steps needed to allow an alre
 
 - Already running an accessible Prometheus instance *outside* of the mesh.
 - A running OSM control plane instance, deployed without metrics stack.
-   - OSM controls the Envoy's Prometheus listener aperture through `prometheus_scraping: "true"`, under OSM configmap. By default this is set to true, but do double check it has been enabled on the OSM configmap, or else Prometheus might not be able to reach the pods.
+   - OSM controls the Envoy's Prometheus listener aperture through `prometheus_scraping: "true"`, under OSM configmap. By default this is set to false (unless OSM was installed with Prometheus enabled.)
```
snehachhabria, line 39: Revert this change too please

.env.example
ksubrmnn: These defaults are for the demo script, not the osm cli itself

.env.example
ksubrmnn: These defaults are for the demo script, not the osm cli itself

.env.example
ksubrmnn: Renamed to ENABLE_PROMETHEUS_DEPLOYMENT to match enable grafana. I can change both flags to deploy_* in the future if that's what we want

snehachhabria: This should be good for now

.env.example
```diff
@@ -91,13 +91,21 @@ export BOOKWAREHOUSE_NAMESPACE=bookwarehouse
 # export ENABLE_EGRESS=true
 
 # optional: ENABLE_GRAFANA (true/false)
```
snehachhabria, line 93:
```suggestion
# optional: ENABLE_GRAFANA_DEPLOYMENT (true/false)
```

charts/osm/README.md
```diff
@@ -20,7 +20,8 @@ A Helm chart to install the OSM control plane on Kubernetes
 | OpenServiceMesh.enableGrafana | bool | `false` |  |
```
snehachhabria, line 20:
```suggestion
| OpenServiceMesh.enableGrafanaDeployment | bool | `false` |  |
```

charts/osm/values.schema.json
```diff
@@ -22,7 +22,8 @@
                 "enablePermissiveTrafficPolicy",
                 "enableBackpressureExperimental",
                 "enableEgress",
-                "enablePrometheus",
+                "enablePrometheusDeployment",
+                "enablePrometheusScraping",
                 "enableGrafana",
```
snehachhabria, line 27:
```suggestion
                "enableGrafanaDeployment",
```

demo/README.md
```diff
@@ -32,6 +32,10 @@ From the root of this repository execute:
 ./demo/run-osm-demo.sh
 ```
 
+### Observability
+By default, Prometheus is deployed by the demo script. To turn this off. Set the variable `ENABLE_PROMETHEUS` in your `.env` file to false.
```
snehachhabria, line 36:
```suggestion
By default, Prometheus is deployed by the demo script. To turn this off. Set the variable `ENABLE_PROMETHEUS_DEPLOYMENT` in your `.env` file to false. 
```

demo/README.md
```diff
@@ -32,6 +32,10 @@ From the root of this repository execute:
 ./demo/run-osm-demo.sh
 ```
 
+### Observability
+By default, Prometheus is deployed by the demo script. To turn this off. Set the variable `ENABLE_PROMETHEUS` in your `.env` file to false. 
+By default, Grafana is deployed by the demo script. To turn this off. Set the variable `ENABLE_PROMETHEUS` in your `.env` file to false.
```
snehachhabria, line 37:
```suggestion
By default, Grafana is deployed by the demo script. To turn this off. Set the variable `ENABLE_GRAFANA_DEPLOYMENT` in your `.env` file to false. 
```

docs/example/README.md
```diff
@@ -37,9 +37,16 @@ The OSM Manual Install Demo Guide is designed to quickly allow you to demo and e
...
-osm install
-```
+
+1.  By default, OSM does not enable Prometheus or Grafana.
+    ```bash
+    osm install
+    ```
+1. To enable Prometheus and Grafana, use their respective flags
+    ```bash
+    osm install --enable-prometheus-deployment true --enable-grafana true
```
snehachhabria, line 47:
```suggestion
    osm install --enable-prometheus-deployment true --enable-grafana-deployment true
```

.env.example
ksubrmnn: Changed this to deploy_grafana

cmd/cli/install.go
```diff
@@ -350,6 +356,12 @@ func (i *installCmd) validateOptions() error {
 		}
 	}
 
+	if i.deployPrometheus {
+		if !i.enablePrometheusScraping {
+			return errors.Errorf("Prometheus scraping cannot be disabled when deploying Prometheus.")
```
snehachhabria, line 361: Nice!

snehachhabria (approved):

cmd/cli/install.go
shashankram: What is the reason to add this check? I can imagine a use case to deploy prometheus, but only turn on scraping much later on via the ConfigMap.

ksubrmnn: That's fair. Mostly I wanted to protect the customer from doing something they may not have been aware of. This can be worked around by creating the deployment and then turning it off in the config map. Or we can just remove this check. Depends on what we think is best for the user.

shashankram: I'd suggest we remove the error, and print a message stating `prometheus_scraping` needs to be enabled for metrics to actually be scraped. Deployment of prometheus should be decoupled from scraping configuration.

snehachhabria (approved):

shashankram (approved):

PR #1869, opened 2020-10-19
tests/e2e: rework NS cleanup

eduser25:
Removed some of the cumbersomeness from the namespace cleaning
code.

All created testing NS, for a given test, will now be tagged with `osmTest`
as label, and the random seed used by Ginkgo as value.

On test cleanup, only tests matching that specific label (and label value)
will be cleaned up.

When a test error occurs and cleanup can't be done (because of some major
error condition that prevents cleanup from executing altogether),
stale test applications and config can be left behind.

From now on, this can be easily addressed and cleanup using something like:

`kubectl delete namespaces -l osmTest`

Additionally, with Kind, will avoid K8s test cleanup if the kind setup
is about to be trashed anyway (saving some precious time on big tests).

**Affected area**:

- Tests                  [X]
- CI System              [X]

No

### DIFF
--- a/tests/e2e/common.go
+++ b/tests/e2e/common.go
@@ -26,6 +26,7 @@ import (
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/intstr"
@@ -48,6 +49,8 @@ const (
 	defaultMeshName = "osm-system"
 	// default image tag
 	defaultImageTag = "latest"
+	// test tag prefix, for NS labeling
+	osmTest = "osmTest"
 )
 
 // OsmTestData stores common state, variables and flags for the test at hand
@@ -78,9 +81,6 @@ type OsmTestData struct {
 	client          *kubernetes.Clientset
 	smiClients      *smiClients
 	clusterProvider *cluster.Provider // provider, used when kindCluster is used
-
-	// Tracks namespaces to cleanup when test finishes (if cleanup is enabled)
-	cleanupNamespaces map[string]bool
 }
 
 // Function to run at init before Ginkgo has called parseFlags
@@ -102,6 +102,14 @@ func registerFlags(td *OsmTestData) {
 	flag.StringVar(&td.osmNamespace, "osmNamespace", utils.GetEnv("K8S_NAMESPACE", defaultMeshName), "OSM mesh name")
 }
 
+// GetTestSelectorMap returns a string-based selector used to refer/select all namespace
+// resources for this test
+func (td *OsmTestData) GetTestSelectorMap() map[string]string {
+	return map[string]string{
+		osmTest: fmt.Sprintf("%d", GinkgoRandomSeed()),
+	}
+}
+
 // AreRegistryCredsPresent checks if Registry Credentials are present
 // It's usually used to factor if a docker registry secret and ImagePullSecret
 // should be installed when creating namespaces and application templates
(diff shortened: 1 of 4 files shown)


tests/e2e/common.go
```diff
@@ -94,6 +94,14 @@ func registerFlags(td *OsmTestData) {
 	flag.StringVar(&td.osmNamespace, "osmNamespace", utils.GetEnv("K8S_NAMESPACE", defaultMeshName), "OSM mesh name")
 }
 
+// GetTestSelectorMap returns a string-based selector used to refer/select all namespace
+// resources for this test
+func (td *OsmTestData) GetTestSelectorMap() map[string]string {
```
nojnhuh, line 99:
```suggestion
func (td *OsmTestData) GetTestNamespaceSelectorMap() map[string]string {
```

tests/e2e/common.go
```diff
@@ -478,14 +482,13 @@ func (td *OsmTestData) CreateMultipleNs(nsName ...string) error {
 // CreateNs creates a Namespace. Will automatically add Docker registry creds if provided
 func (td *OsmTestData) CreateNs(nsName string, labels map[string]string) error {
 	if labels == nil {
-		labels = map[string]string{}
+		labels = td.GetTestSelectorMap()
+	} else {
+		for k, v := range td.GetTestSelectorMap() {
+			labels[k] = v
+		}
```
nojnhuh, line 489: Could this loop be run unconditionally so we could get rid of the if/else?

tests/e2e/common.go
```diff
@@ -675,25 +677,46 @@ const (
 
 // Cleanup is Used to cleanup resorces once the test is done
 func (td *OsmTestData) Cleanup(ct CleanupType) {
-	// In-cluster Test resources cleanup(namespace, crds, specs and whatnot) here
-	if td.cleanupTest {
-		var nsList []string
-		for ns := range td.cleanupNamespaces {
-			err := td.DeleteNs(ns)
+	// Small speedup: avoid graceful/wait delete of k8s resources on kind if the cluster is destined to be
+	// trashed anyway
```
nojnhuh, line 681: +1 I've been meaning to make this change.

nojnhuh (dismissed):
I left a couple tiny nits but LGTM.

tests/e2e/common.go
```diff
@@ -675,25 +677,46 @@ const (
...
+	if td.cleanupTest &&
+		(!td.kindCluster ||
+			(td.kindCluster &&
+				(ct == Test && !td.cleanupKindClusterBetweenTests) ||
+				(ct == Suite && !td.cleanupKindCluster))) {
```
shashankram, line 686: Too many nested conditionals makes this hard to read. Would you dropping a comment explaining under what scenarios this block gets hit. I see the comment in L680 but it doesn't reflect the meaning behind the conditionals.

tests/e2e/common.go
eduser25: hmmm... yeah can do. I'll do.

tests/e2e/common.go
eduser25: the comment is just one line above... is it not good enough?

shashankram (approved):

tests/e2e/common.go
eduser25: I'll try to clarify the comment a bit more with additional context on what the conditions do

snehachhabria (approved):

nojnhuh (dismissed):

eduser25:
Ops. Need a rebase since @nojnhuh merged a new test.

eduser25:
```
/bin/tar --posix --use-compress-program zstd -T0 -cf cache.tzst -P -C /home/runner/work/osm/osm --files-from manifest.txt
Warning: uploadChunk (start: 67108864, end: 99847018) failed: read ECONNRESET
/home/runner/work/_actions/actions/cache/v2/dist/save/index.js:3136
                        throw new Error(`Cache upload failed because file read failed with ${error.message}`);
```
hmm... will wait sometime

eduser25:
image
have to wait

eduser25:
After cert manager test:
```
Deleting namespace osm-system
STEP: [Cleanup] waiting for osmTest:1603242475 test NS cleanup
Poll err: timed out waiting for the condition
```
https://github.com/openservicemesh/osm/issues/1880

snehachhabria (approved):

nojnhuh (approved):

PR #1872, opened 2020-10-20
tests(e2e): add permissive traffic policy mode test

### LINKED ISSUE #1815: Add an e2e test for permissive traffic policy mode

We should add a scenario to the new e2e framework testing the permissive traffic policy mode that allows services to communicate by default vs. disallowing by default.

The general flow for the test should be:
1. Install OSM with permissive mode enabled
1. Install a simple client/server topology
1. Verify the client can reach the server
1. Disable permissive mode
1. Verify the client cannot reach the server.

The same check that traffic is denied when permissive mode is disabled at the end of the test in e2e_pod_client_server_test.go can be removed once the above scenario is implemented.

- Tests                      [X]

nojnhuh:
**Description**: This change adds an e2e test for the permissive traffic policy mode. It tests that two services can communicate without any SMI policies when permissive mode is enabled, disables permissive mode, then tests that the services cannot communicate.

Fixes #1815

**Affected area**:

- New Functionality      [ ]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [ ]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests                  [X]
- CI System              [ ]
- Performance            [ ]
- Other                  [ ]

No

### DIFF
--- a/tests/e2e/e2e_permissive_test.go
+++ b/tests/e2e/e2e_permissive_test.go
@@ -1,0 +1,107 @@
+package e2e
+
+import (
+	"fmt"
+	"strings"
+	"time"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("Permissive Traffic Policy Mode", func() {
+	Context("PermissiveMode", func() {
+		const sourceNs = "client"
+		const destNs = "server"
+		var ns []string = []string{sourceNs, destNs}
+
+		It("Tests HTTP traffic for client pod -> server pod with permissive mode", func() {
+			// Install OSM
+			installOpts := td.GetOSMInstallOpts()
+			installOpts.enablePermissiveMode = true
+			Expect(td.InstallOSM(installOpts)).To(Succeed())
+			Expect(td.WaitForPodsRunningReady(td.osmNamespace, 60*time.Second, 1)).To(Succeed())
+
+			// Create Test NS
+			for _, n := range ns {
+				Expect(td.CreateNs(n, nil)).To(Succeed())
+				Expect(td.AddNsToMesh(true, n)).To(Succeed())
+			}
+
+			// Get simple pod definitions for the HTTP server
+			svcAccDef, podDef, svcDef := td.SimplePodApp(
+				SimplePodAppDef{
+					name:      "server",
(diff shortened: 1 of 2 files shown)


snehachhabria (approved):

tests/e2e/e2e_permissive_test.go
```diff
@@ -0,0 +1,107 @@
...
+			_, err = td.CreateServiceAccount(sourceNs, &svcAccDef)
+			Expect(err).NotTo(HaveOccurred())
+			srcPod, err := td.CreatePod(sourceNs, podDef)
+			Expect(err).NotTo(HaveOccurred())
+			_, err = td.CreateService(sourceNs, svcDef)
+			Expect(err).NotTo(HaveOccurred())
+
+			Expect(td.WaitForPodsRunningReady(sourceNs, 90*time.Second, 1)).To(Succeed())
+
+			req := HTTPRequestDef{
```
eduser25, line 68: personal POV:

I'd add more complexity (maybe we can do it on another test? but it will be a superset of this; but I'd use multiple deployments and multiple pods per svc, etc). 
Permissive pushes everything everywhere, I'm just thinking of potential issues that would not show up until we put some complexity on the mesh itself.

That's not to say this test is not good, but if we have a larger test that covers more, I wonder if we'll really need a smaller one at that point (in terms of timing and number of tests, etc).

eduser25 (dismissed):

tests/e2e/e2e_permissive_test.go
nojnhuh: Yeah I agree. I like the idea of still keeping the smallest possible example and running a more complex test in addition. That could help us better pinpoint problems as they arise.

eduser25: +1, agreed having smaller test too.

eduser25 (approved):

eduser25:
After cert manager test:
```
Could not confirm all namespace deletion in time: timed out waiting for the condition
```
https://github.com/openservicemesh/osm/issues/1880

snehachhabria (approved):

PR #1878, opened 2020-10-20
cli: expose portforwarding capability and refactor dashboard cmd

shashankram:
**Description**:
This change exposes port forwarding capability generically
(previously implemented in the dashboard cmd) 
so that other commands that require to port forward to a pod
can leverage this. It updates the dashboard cmd to leverage
the generic port forwarder.

A subsequent change that requires port forwarding from the cli
will also leverage the newly added port forwarding APIs.

There also doesn't seem to be a trivial way to mock
port forwarding to a pod, so this will likely have to be
tested using e2e for automated tests.

**Affected area**:

- New Functionality      [ ]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [ ]
- CLI Tool               [X]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests                  [ ]
- CI System              [ ]
- Performance            [ ]
- Other                  [ ]

`No`

### DIFF
--- a/cmd/cli/dashboard.go
+++ b/cmd/cli/dashboard.go
@@ -106,7 +106,10 @@ func (d *dashboardCmd) run() error {
 	// Select pod/s given the service data available
 	set := labels.Set(svc.Spec.Selector)
 	listOptions := metav1.ListOptions{LabelSelector: set.AsSelector().String()}
-	pods, _ := v1ClientSet.Pods(settings.Namespace()).List(context.TODO(), listOptions)
+	pods, err := v1ClientSet.Pods(settings.Namespace()).List(context.TODO(), listOptions)
+	if err != nil {
+		return errors.Errorf("Error listing pods: %s", err)
+	}
 
 	// Will select first running Pod available
 	var grafanaPod *corev1.Pod
--- a/cmd/cli/dashboard.go
+++ b/cmd/cli/dashboard.go
@@ -1,26 +1,24 @@
 package main
 
 import (
-	"bytes"
 	"context"
 	"fmt"
 	"io"
 	"net/http"
 	"os"
 	"os/signal"
-	"syscall"
 
 	"github.com/pkg/browser"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 	"helm.sh/helm/v3/pkg/action"
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/util/httpstream"
 	"k8s.io/client-go/kubernetes"
 	v1 "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/client-go/rest"
-	"k8s.io/client-go/tools/portforward"
 	"k8s.io/client-go/transport/spdy"
 )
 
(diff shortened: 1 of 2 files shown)


cmd/cli/dashboard.go
```diff
@@ -108,67 +106,43 @@ func (d *dashboardCmd) run() error {
 	// Select pod/s given the service data available
 	set := labels.Set(svc.Spec.Selector)
 	listOptions := metav1.ListOptions{LabelSelector: set.AsSelector().String()}
-	pods, err := v1ClientSet.Pods(settings.Namespace()).
-		List(context.TODO(), listOptions)
+	pods, _ := v1ClientSet.Pods(settings.Namespace()).List(context.TODO(), listOptions)
```
nojnhuh, line 109: Shouldn't we be handling the error here? This looks like it would be fine in the case where the list comes back empty, but if some other error occurs where `pods` comes back `nil`, then dereferencing `pods.Items` will panic.

cmd/cli/portforward.go
```diff
@@ -0,0 +1,110 @@
...
+		}
+	}()
+
+	// Process signals and call readyFunc
+	select {
+	case <-pf.readyChan:
+		if err := readyFunc(pf); err != nil {
+			return err
+		}
+		return nil
```
nojnhuh, line 95: Nit:
```suggestion
		return readyFunc(pf)
```

cmd/cli/dashboard.go
```diff
@@ -108,67 +106,43 @@ func (d *dashboardCmd) run() error {
...
-		<-d.sigintChan // Blocking
-		fmt.Fprintf(d.out, "[+] SIGINT/TERM received, closing")
-		close(stopChan)
-	}()
-
-	// This routine blocks on readyChan till forwarding is setup and ready.
-	go func() {
-		<-readyChan
-
-		fmt.Fprintf(d.out, "[+] Port forwarding successful (localhost:%d)\n", d.localPort)
```
nojnhuh, line 161: Is there any output like this when the port forward is successful? I could see that being helpful just to communicate that things are working fine and nothing is hung up.

cmd/cli/dashboard.go
```diff
@@ -108,67 +106,43 @@ func (d *dashboardCmd) run() error {
...
+	})
+	if err != nil {
+		return errors.Errorf("Port forwarding failed: %s", err)
 	}
 
+	// The command should only exit when a signal is received from the OS.
+	// Exiting before will result in port forwarding to stop causing the browser
+	// if open to not render the dashboard.
+	sigChan := make(chan os.Signal, 1)
+	signal.Notify(sigChan, os.Interrupt)
```
nojnhuh, line 144: Is there a particular reason for not also waiting on `SIGTERM` like before? I don't disagree with the change if that was intentional, just curious.

nojnhuh (changes requested):

cmd/cli/dashboard.go
shashankram: I don't think it is necessary because in case of an error with port forwarding the output will be relayed back unlike before.

cmd/cli/dashboard.go
shashankram: Good catch.

cmd/cli/dashboard.go
shashankram: SIGTERM doesn't work on Windows.

cmd/cli/portforward.go
shashankram: Done.

cmd/cli/dashboard.go
nojnhuh: I agree that it's not necessary, but my first thought whenever I run a command that produces no output is "what did I do wrong to make this hang," so it's just a usability thing. At least in the case that the browser isn't being opened.

shashankram: You should still see:
`[+] Starting Dashboard forwarding` regardless.

nojnhuh (approved):

snehachhabria (approved):

eduser25 (approved):

Issue #1884, opened 2020-10-21
cli: investigate usage of global namespace flag associated with osm-controller's namespace


shashankram:
**Please describe the Improvement and/or Feature Request**
The cli has a global namespace flag set to be osm-controller's namespace, which is unused by several cli commands and causes confusion when individual cli commands also expose a `--namespace` flag with their commands. Moreover, the command help does not render the `namespace` flag for the command, but rather just the global namespace flag.

Investigate the usage of this flag and see if it needs to be renamed.
https://github.com/openservicemesh/osm/blob/main/cmd/cli/osm.go#L35

**Scope (please mark with X where applicable)**
- New Functionality          [ ]
- Install                    [ ]
- SMI Traffic Access Policy  [ ]
- SMI Traffic Specs Policy   [ ]
- SMI Traffic Split Policy   [ ]
- Permissive Traffic Policy  [ ]
- Ingress                    [ ]
- Egress                     [ ]
- Envoy Control Plane        [ ]
- CLI Tool                   [X]
- Metrics                    [ ]
- Certificate Management     [ ]
- Sidecar Injection          [ ]
- Logging                    [ ]
- Debugging                  [ ]
- Tests                      [ ]
- CI System                  [ ]
- Project Release            [ ]

shashankram:
Resolved by https://github.com/openservicemesh/osm/commit/a91ccad1c275e4faef28bc8e063fee4f5380e4fb

PR #1887, opened 2020-10-21
cli: rename global namespace flag

shashankram:
**Description**:
This change is an RFC to discuss the proposal
to rename the global namespace flag exposed in
osm cli, which reflects the control plane namespace.
The reason to do so is because local command flags
could also expose a namespace flag, which don't
show up in the help text for the command. Moreover,
the global namespace flag isn't used by several
commands (ex. osm metrics, osm proxy etc.).

The proposal is to rename the global flag to represent
OSM's namespace to `osm-namespace` so that local
commands wanting to expose `namespace` as a flag can
do so with command specific description.

Part of #1884

**Affected area**:

- New Functionality      [ ]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [ ]
- CLI Tool               [X]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests                  [ ]
- CI System              [ ]
- Performance            [ ]
- Other                  [ ]

`No`

### DIFF
--- a/cmd/cli/install.go
+++ b/cmd/cli/install.go
@@ -32,10 +32,10 @@ wide Kubernetes resources.
 
 The default Kubernetes namespace that gets created on install is called
 osm-system. To create an install control plane components in a different
-namespace, use the --namespace flag.
+namespace, use the global --osm-namespace flag.
 
 Example:
-  $ osm install --namespace hello-world
+  $ osm install --osm-namespace hello-world
 
 Multiple control plane installations can exist within a cluster. Each
 control plane is given a cluster-wide unqiue identifier called mesh name.
@@ -346,5 +346,5 @@ func errMeshAlreadyExists(name string) error {
 }
 
 func errNamespaceAlreadyHasController(namespace string) error {
-	return errors.Errorf("Namespace %s has an osm controller. Please specify a new namespace using --namespace", namespace)
+	return errors.Errorf("Namespace %s has an osm controller. Please specify a new namespace using --osm-namespace", namespace)
 }
--- a/demo/clean-kubernetes.sh
+++ b/demo/clean-kubernetes.sh
@@ -5,7 +5,7 @@ set -aueo pipefail
 # shellcheck disable=SC1091
 source .env
 
-bin/osm mesh uninstall -f --mesh-name "$MESH_NAME" --namespace "$K8S_NAMESPACE"
+bin/osm mesh uninstall -f --mesh-name "$MESH_NAME" --osm-namespace "$K8S_NAMESPACE"
 
 for ns in "$BOOKWAREHOUSE_NAMESPACE" "$BOOKBUYER_NAMESPACE" "$BOOKSTORE_NAMESPACE" "$BOOKTHIEF_NAMESPACE" "$K8S_NAMESPACE"; do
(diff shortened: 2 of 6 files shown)


michelleN (commented):
I like this change a lot. Makes it much more clear what namespace means. I have been wondering whether we really need this flag on every command. If that is overkill, we should add it per command rather than globally. It seems like we use to it to target a control plane but I'm wondering if that is even necessary since we now ensure mesh name is unique to the cluster.

shashankram:
> I like this change a lot. Makes it much more clear what namespace means. I have been wondering whether we really need this flag on every command. If that is overkill, we should add it per command rather than globally. It seems like we use to it to target a control plane but I'm wondering if that is even necessary since we now ensure mesh name is unique to the cluster.

Very valid point you have there. I considered removing this global flag altogether, then noticed several cli commands are leveraging it via the `settings.Namespace()` api. So moving forward we could have individual commands expose their own namespace flag, including the flag for the controller namespace if necessary. Right now it seemed less destructive to rename the global flag over removing it. What do you think @michelleN?

michelleN (commented):
Regardless of whether we'd get rid of the flag globally or keep it, I'd give this change a 👍 . Looks like we'd just need to clean up documentation related to this change.

shashankram:
> Agreed. Regardless of whether we'd get rid of the flag globally or keep it, I'd give this change a 👍 and would support moving forward with it. Looks like we'd just need to clean up documentation related to this change.

I might have missed cleaning up something, is there any documentation in particular that needs to reflect this?

snehachhabria (approved):

michelleN:
@shashankram - nvm just thought there might be some --namespace references in the manual demo guide but there isn't.

michelleN (approved):

michelleN (changes requested):
The CLI help text just needs to be updated. https://github.com/openservicemesh/osm/blob/main/cmd/cli/install.go#L35-L38

shashankram:
> The CLI help text just needs to be updated. https://github.com/openservicemesh/osm/blob/main/cmd/cli/install.go#L35-L38

@michelleN it already is in this PR: https://github.com/openservicemesh/osm/pull/1887/files#diff-4afa5892801237e567fde7607da3a3923912b6cffb91cb33e0df10a9e6d4bb95R35

michelleN (approved):

nojnhuh (changes requested):
Found a couple more spots using `--namespace` for the `osm metrics` commands:

- https://github.com/openservicemesh/osm/blob/main/demo/configure-app-namespaces.sh#L17

- 4 places: https://github.com/openservicemesh/osm/blob/main/docs/patterns/observability.md#configuring-prometheus-metrics-scraping

shashankram:
> Found a couple more spots using `--namespace` for the `osm metrics` commands:
> 
> * https://github.com/openservicemesh/osm/blob/main/demo/configure-app-namespaces.sh#L17
> * 4 places: https://github.com/openservicemesh/osm/blob/main/docs/patterns/observability.md#configuring-prometheus-metrics-scraping

Those use the local command flags.

nojnhuh (approved):
Yep, my mistake, never mind.

PR #1889, opened 2020-10-22
tests(e2e): add helm install e2e test

### LINKED ISSUE #1814: Add helm install to e2e test framework

In addition to installation via the osm cli, `helm install` needs to be added as another installation process to the e2e test framework regardless of specific scenarios we are testing.

- Tests                      [X]

**Possible use cases**

michelleN:
**Description**:
This PR adds a test that installs the local osm chart with the locally built osm images using the helm directly to ensure that helm install with the default configuration works. The test uses Helm's `--wait` flag to ensure that all Pods installed (but most importantly osm-controller) are in a ready state before exiting.

resolves #1814

-->
**Affected area**:

- New Functionality      [ ]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [ ]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests                  [ ]
- CI System              [ ]
- Performance            [ ]
- Other                  [ ]

No


tests/e2e/e2e_helm_install_test.go
```diff
@@ -0,0 +1,16 @@
...
+import (
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("Test osm control plane installation with Helm", func() {
+	Context("Using default values", func() {
+		It("installs osm control plane successfully", func() {
+			// Install OSM with Helm
+			Expect(td.HelmInstallOSM()).To(Succeed())
```
ritazh, line 12: Can we also explicitly test if all of the osm components are running as expected?

- ensure osm controller pod status is in running state, this would catch any issue that cause the readiness checker to fail, e.g. cert issues, rbac issues, etc
- osm service exists
- crds exist, this would catch any change in our crds
- configmap exists with the default values, this would catch any default behavior changes in the chart
- mutatingwebhookconfig exists

tests/e2e/common.go
```diff
@@ -190,38 +190,46 @@ func (td *OsmTestData) GetOSMInstallOpts() InstallOSMOpts {
...
+	release := "helm-install-osm"
+
+	values := fmt.Sprintf("OpenServiceMesh.image.registry=%s,OpenServiceMesh.image.tag=%s,meshName=%s", td.ctrRegistryServer, td.osmImageTag, release)
+	args := []string{"install", release, "../../charts/osm", "--set", values, "--namespace", namespace, "--create-namespace", "--wait"}
+	stdout, stderr, err := td.RunLocal("helm", args)
+	if err != nil {
+		td.T.Logf("stdout:\n%s", stdout)
+		return errors.Errorf("failed to run helm install with osm chart: %s", stderr)
+	}
+	if err = td.deleteHelmRelease(release, namespace); err != nil {
```
ritazh, line 210: We should have a test explicitly to test helm uninstall. Should I open a separate issue for this?

michelleN: yes please

tests/e2e/e2e_helm_install_test.go
michelleN: I added a schema file for values so that will help us catch config related issues around values like if a required value were missing, we'd get notified on the helm install test. Also does pattern matching on certificate-manager type, envoy log level and things like that.

michelleN: I can still add a test for the configmap default values though. Can't hurt.

eduser25:
Few items, food for thought only:

(1)
IMHO, I think installing with helm could be an option of `InstallOSM` which takes in `InstallOSMOpts` struct as param, and `helmInstall` could just be a flag in it. This would allow the entire suite to run with installation through charts with helm if necessary (_if and only if_ we would want this to happen). 

The `InstallOSM` code though is pretty much a heavily conditioned function, ideally we would have factory patterns to get different installations, etc... but we don't have any of that as of now, so I don't have a strong feeling about adding it all there at this time.

(2) https://github.com/openservicemesh/osm/pull/1894 introduces a NoInstall option, where the tests will assume an OSM instance running is running in some namespace, so the Install API will immediately return and assume OSM it's already running.
This was done in light of some environments... installing OSM thorugh some 3rd party addon, so we can run most of the test suite on it, while [skipping](https://github.com/openservicemesh/osm/pull/1894/files#diff-fdf4dda496428ea04d5b0c8befbc8cc3f7a25611fe491ed2f675776b6f46a761R268) tests that require modified installation behavior, (like cert-manager, hashi-vault, fluentbit test, or this one - with the difference this one will fail as it is in those conditions, as it doesn't call `InstallOSM`.

I think however we can go ahead with this commit for the time being, and I'll rebase and add additional flags to this test to skip it if NoInstall is signaled to the test. We can open an issue to rework the Install API to absorb installing through helm in the future.

tests/e2e/e2e_helm_install_test.go
eduser25, line 12: most of the tests will run a `WaitForPodsReady` function, and wait for the right number of pods expected to be ready in `osmNamespace`; this ensures to some degree that things deployed and are running.

eduser25: oh nvm, I see you expect that from `--wait` correct?

michelleN: I'm using the `--wait` flag on the helm install command which ensures that the Pods installed via the helm chart are in a ready state. I feel like that is a more targeted approach. What do you think? Are there benefits of using `WaitForPodsReady` over the wait flag?

michelleN:
@eduser25 Love (1). I think that's a great way to go. I can also see lots of benefits to (2).

tests/e2e/e2e_helm_install_test.go
eduser25: I would assume helm knows better which and how many pods does it have to see running? Normally we have to write down the number of pods _we_ expect to see running based on the installation we code, but I'd assume helm does and knows that better. 
No problem with it for me.

tests/e2e/e2e_helm_install_test.go
```diff
@@ -0,0 +1,36 @@
+package e2e
+
+import (
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("Test osm control plane installation with Helm", func() {
```
nojnhuh, line 8: @michelleN Could you update the test to add it to one of the tiers we defined in #1902 (specifically tier 2)?
```suggestion
var _ = DescribeTier2("Test osm control plane installation with Helm", func() {
```

eduser25 (approved):

nojnhuh (approved):

ritazh (approved):
lgtm

ksubrmnn (approved):

ritazh:
Regarding this test being a part of tier 2, is there a reason we feel this should not be part of PR gate? I would like to advocate for users who use `helm install` directly to make sure we don't merge PRs in main that could impact this experience.

nojnhuh:
> Regarding this test being a part of tier 2, is there a reason we feel this should not be part of PR gate? I would like to advocate for users who use `helm install` directly to make sure we don't merge PRs in main that could impact this experience.

My take on this is that a `helm install` test is unlikely to be broken if all the other PR gate tests pass given the CLI is a thin wrapper over `helm install`.

michelleN:
I think what @ritazh is trying to prevent is what if the values in the helm chart are changed or something about the chart is changed that is overwritten via the cli. I do feel that the one test that does a simple helm install is valuable is a good tier 1 test.

nojnhuh:
I think I understand what you mean, but can you give an example? If it's an issue of potential mismatch between the values in the chart and the values we pass via the CLI, then could that be checked with a unit test instead of an e2e test?

michelleN:
There was an issue for example where the key `certificateManager` in the values file was misspelled. The template using that key called the correctly spelled key for certificateManager. The `osm` CLI passed the correctly spelled key in as a value during install as well. But if you did a `helm install`, the osm-controller would never go into a ready state because the key was incorrectly spelled in the values file and thus never passed in as a paramter to osm controller. Had there been a simple helm install e2e test in place, we would have caught that.

There are a few ways we can handle that specific scenario though.
(1) we could make sure every reference to a value has a default specified in the template (I think that might be overkill or not a good thing to do in some situations)
(2) I recently added the schema file to make sure we have all the required keys defined and that schema is validate on helm install. That helps but may not cover all the corner cases which is why I think having a simple helm install e2e is nice. 

I do not think that running all the test scenarios (permissive mode, egress, split, etc) using `osm install` AND `helm install` as a PR gate makes much sense. It seems redundant imo but I do think we should run all the scenarios via `osm install` AND `helm install` as a release gate.

nojnhuh:
I definitely agree scenarios like that should be tested. I can think of how we can write unit tests for that, but I don't think I would consider that specific scenario after one pass at adding unit tests, so I can see where an e2e test for that would be beneficial.

PR #1890, opened 2020-10-22
feat(charts): validate chart values w/ json schema

michelleN:
**Description**:
This PR adds a values.schema.json file to indicate which values are required, type, description, examples, defaults. The schema file will help prevent us from accidentally changing the values file and breaking things as well as ensuring we are specifying the required values any time this chart is installed.

From the docs (https://helm.sh/docs/topics/charts/#schema-files): 
> This schema will be applied to the values to validate it. Validation occurs when any of the following commands are invoked:
> - helm install
> - helm upgrade
> - helm lint
> - helm template

**Affected area**:

- New Functionality      [ ]
- Documentation          [ ]
- Install                [X ]
- Control Plane          [ ]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests                  [ ]
- CI System              [ ]
- Performance            [ ]
- Other                  [ ]

No

### DIFF
--- a/charts/osm/values.schema.json
+++ b/charts/osm/values.schema.json
@@ -11,7 +11,6 @@
             "type": "object",
             "title": "The OpenServiceMesh schema",
             "description": "The top level required key in the values file.",
-            "default": {},
             "required": [
                 "replicaCount",
                 "image",
@@ -38,7 +37,6 @@
                     "type": "integer",
                     "title": "The replicaCount schema",
                     "description": "The number of replicas of the osm-controller Pod.",
-                    "default": 1,
                     "examples": [
                         1
                     ]
@@ -48,7 +46,6 @@
                     "type": "object",
                     "title": "The image schema",
                     "description": "The details of the images to run.",
-                    "default": {},
                     "examples": [
                         {
                             "registry": "openservicemesh",
@@ -67,7 +64,6 @@
                             "type": "string",
                             "title": "The registry schema",
                             "description": "The registry of the images to run.",
-                            "default": "openservicemesh",
                             "examples": [
                                 "openservicemesh"
                             ]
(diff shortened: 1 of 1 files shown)


charts/osm/values.schema.json
```diff
@@ -0,0 +1,283 @@
...
+    "required": [
+        "OpenServiceMesh"
+    ],
+    "properties": {
+        "OpenServiceMesh": {
+            "$id": "#/properties/OpenServiceMesh",
+            "type": "object",
+            "title": "The OpenServiceMesh schema",
+            "description": "The top level required key in the values file.",
+            "default": {},
```
nojnhuh, line 14: I noticed the example in Helm's docs doesn't include defaults. How much value is there in adding them here? And what could go wrong if these defaults get out of sync with values.yaml?

charts/osm/values.schema.json
```diff
@@ -0,0 +1,283 @@
...
+                    "examples": [
+                        false
+                    ]
+                },
+                "envoyLogLevel": {
+                    "$id": "#/properties/OpenServiceMesh/properties/envoyLogLevel",
+                    "type": "string",
+                    "title": "The envoyLogLevel schema",
+                    "description": "Envoy log level.",
+                    "default": "error",
```
nojnhuh, line 225: A `pattern` matching the acceptable values might be beneficial here.

charts/osm/values.schema.json
michelleN: That's a really good point. I had them in here because they're an option in json schema but Helm is not using them. Let me go ahead and take them out.

nojnhuh (approved):

shashankram (approved):

PR #1894, opened 2020-10-22
tests/e2e: add NoInstall option

eduser25:
Adds new e2e install option `NoInstall`.

NoInstall means to run the tests assuming there is already
a running, permanent instance of OSM in namespace `osmNamespace`, part
of the test suite flags.

OSM will therefore not be installed/cleaned up between tests,
all other resources used by the tests (clients, other namespaces,
SMI defs, etc) will get cleaned though.

Since a single install is assumed, it assumes inmutability of the
install, so tests that _require_ or test specific parts of the
installation will be skipped.

Dynamic configuration (configmap) is however supported, and tests
that would usually call Install will instead set/reset the dynamic
config map values that the install command would originally set.

**Affected area**:

- New Functionality      [X]
- Tests                  [X]
- CI System              [X]

No

### DIFF
--- a/tests/e2e/common.go
+++ b/tests/e2e/common.go
@@ -14,11 +14,13 @@ import (
 	"strings"
 	"time"
 
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
 	"github.com/docker/docker/client"
 	"github.com/jetstack/cert-manager/pkg/apis/certmanager/v1alpha2"
 	cmmeta "github.com/jetstack/cert-manager/pkg/apis/meta/v1"
 	certman "github.com/jetstack/cert-manager/pkg/client/clientset/versioned"
-	. "github.com/onsi/ginkgo"
 	"github.com/pkg/errors"
 	"helm.sh/helm/v3/pkg/action"
 	"helm.sh/helm/v3/pkg/chart/loader"
@@ -46,9 +48,11 @@ const (
 	// contant, default name for the Registry Secret
 	registrySecretName = "acr-creds"
 	// constant, default name for the mesh
-	defaultMeshName = "osm-system"
+	defaultOsmNamespace = "osm-system"
 	// default image tag
 	defaultImageTag = "latest"
+	// default cert manager
+	defaultCertManager = "tresor"
 	// test tag prefix, for NS labeling
 	osmTest = "osmTest"
 )
@@ -62,6 +66,32 @@ func DescribeTierN(tier uint) func(string, func()) bool {
 var DescribeTier1 = DescribeTierN(1)
 var DescribeTier2 = DescribeTierN(2)
 
+// InstallType defines several OSM test deployment scenarios
+type InstallType string
+
+const (
+	// SelfInstall uses current kube cluster, installs OSM using CLI
+	SelfInstall InstallType = "SelfInstall"
+	// KindCluster Creates Kind cluster on docker and uses it as cluster, OSM installs through CLI
+	KindCluster InstallType = "KindCluster"
+	// NoInstall uses current kube cluster, assumes an OSM is present in `osmNamespace`
+	NoInstall InstallType = "NoInstall"
+)
+
+// Verifies the instType string flag option is a valid enum type
+func verifyValidInstallType(t InstallType) error {
+	switch t {
+	case SelfInstall:
+		fallthrough
+	case KindCluster:
+		fallthrough
+	case NoInstall:
+		return nil
+	default:
+		return errors.Errorf("%s is not a valid OSM install type", string(t))
+	}
+}
+
 // OsmTestData stores common state, variables and flags for the test at hand
 type OsmTestData struct {
 	T GinkgoTInterface // for common test logging
@@ -70,6 +100,7 @@ type OsmTestData struct {
 	waitForCleanup bool // Forces test to wait for effective deletion of resources upon cleanup
 
 	// OSM install-time variables
+	instType     InstallType // Install type.
 	osmNamespace string
 	osmImageTag  string
 
@@ -79,7 +110,6 @@ type OsmTestData struct {
 	ctrRegistryServer   string // server name. Has to be network reachable
 
 	// Kind cluster related vars
-	kindCluster                    bool   // Create and use a kind cluster
 	clusterName                    string // Kind cluster name (used if kindCluster)
 	cleanupKindClusterBetweenTests bool   // Clean and re-create kind cluster between tests
 	cleanupKindCluster             bool   // Cleanup kind cluster upon test finish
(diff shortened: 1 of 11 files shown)


tests/e2e/common.go
```diff
@@ -215,10 +244,27 @@ func (td *OsmTestData) GetOSMInstallOpts() InstallOSMOpts {
 	}
 }
 
-// InstallOSM installs OSM. Right now relies on externally calling the binary and a subset of possible opts
-// TODO: refactor install to be able to call it directly here vs. exec-ing CLI.
+// InstallOSM installs OSM. The behavior of this function is dependant on
```
nojnhuh, line 249:
```suggestion
// InstallOSM installs OSM. The behavior of this function is dependent on
```

tests/e2e/common.go
```diff
@@ -122,11 +153,9 @@ func (td *OsmTestData) AreRegistryCredsPresent() bool {
 func (td *OsmTestData) InitTestData(t GinkgoTInterface) error {
 	td.T = t
 
-	if len(td.ctrRegistryServer) == 0 {
-		td.T.Errorf("Did not read any container registry (is CTR_REGISTRY set?)")
-	}
+	Expect(verifyValidInstallType(td.instType)).To(Succeed())
```
nojnhuh, line 156: Should we just return an error here instead if this fails so we can tell which test actually failed from the ginkgo back trace?

tests/e2e/e2e_trafficsplit_same_sa_test.go
```diff
@@ -53,7 +53,6 @@ var _ = Describe("Test TrafficSplit where each backend shares the same ServiceAc
 		It("Tests HTTP traffic from Clients to the traffic split Cluster IP", func() {
 			// Install OSM
 			Expect(td.InstallOSM(td.GetOSMInstallOpts())).To(Succeed())
-			Expect(td.WaitForPodsRunningReady(td.osmNamespace, 90*time.Second, 1)).To(Succeed())
```
snehachhabria, line 56: Instead of removing this can we have a check to see there's a pod and its in running state ? 

Applies to all places

tests/e2e/common.go
```diff
@@ -215,10 +244,27 @@ func (td *OsmTestData) GetOSMInstallOpts() InstallOSMOpts {
...
-// TODO: refactor install to be able to call it directly here vs. exec-ing CLI.
+// InstallOSM installs OSM. The behavior of this function is dependant on
+// installType and instOpts
 func (td *OsmTestData) InstallOSM(instOpts InstallOSMOpts) error {
-	if td.kindCluster {
+	if td.instType == NoInstall {
+		if instOpts.certManager != defaultCertManager ||
+			instOpts.deployPrometheus ||
+			instOpts.deployGrafana ||
+			instOpts.deployJaeger {
```
nojnhuh: Should we explicitly define defaults for these values so there isn't a hidden assumption here that the default is false?

tests/e2e/e2e_trafficsplit_same_sa_test.go
eduser25: check `common.go`, patch adds it at the end of `InstallOSM` as it was called basically everywhere

tests/e2e/common.go
```diff
@@ -89,7 +119,8 @@ func registerFlags(td *OsmTestData) {
 	flag.BoolVar(&td.cleanupTest, "cleanupTest", true, "Cleanup test resources when done")
 	flag.BoolVar(&td.waitForCleanup, "waitForCleanup", true, "Wait for effective deletion of resources")
 
-	flag.BoolVar(&td.kindCluster, "kindCluster", false, "Creates kind cluster")
+	flag.StringVar((*string)(&td.instType), "installType", string(SelfInstall), "Type of install/deployment for OSM")
```
nojnhuh, line 122: The Makefile still defines `-kindCluster`, so we should update that.

tests/e2e/common.go
```diff
@@ -46,13 +48,41 @@ const (
...
+	NoInstall InstallType = "NoInstall"
+)
+
+// Verifies the instType string flag option is a valid enum type
+func verifyValidInstallType(t InstallType) error {
+	switch t {
+	case SelfInstall:
+		fallthrough
+	case KindCluster:
+		fallthrough
```
nojnhuh, line 78: Not sure I've ever actually seen `fallthrough` used before in the wild. 😄

tests/e2e/e2e_egress_test.go
```diff
@@ -15,9 +15,7 @@ var _ = Describe("HTTP and HTTPS Egress", func() {
 		It("Allows egress traffic when enabled", func() {
 			// Install OSM
 			installOpts := td.GetOSMInstallOpts()
-			installOpts.egressEnabled = true
```
eduser25, line 18: oh wow... my mistake.

tests/e2e/common.go
```diff
@@ -313,7 +359,7 @@ func (td *OsmTestData) InstallOSM(instOpts InstallOSMOpts) error {
 		return errors.Wrap(err, "failed to run osm install")
 	}
 
-	return nil
+	return td.WaitForPodsRunningReady(td.osmNamespace, 90*time.Second, 1)
```
nojnhuh, line 362: We should keep the expected number of pods in a variable that gets updated to reflect the options since installing vault adds a pod, cert-manager adds 3, etc.

nojnhuh (changes requested):

tests/e2e/common.go
eduser25: ah yes. flew over my head.

tests/e2e/common.go
eduser25: I did not delete the cert-manager wait. Test speciic ones can coexist, while this just ensures at least 1 was seen (which is sort of the default scenario)

tests/e2e/common.go
eduser25: +1, will do

tests/e2e/common.go
eduser25: y... yes, not sure why did I do that

eduser25: also fixed the `UpdateOSMConfig`

tests/e2e/common.go
nojnhuh: Would keeping track here eliminate the need to have the test-specific ones (at least for the control plane namespace)? Looks like tests should continue to work the same way as before, so that's what's really important. Changing this could be a future PR if we want that.

eduser25: ok fair enough

eduser25: I'll open an issue to see if we want this addressed differently post-patch (have Install get the waits on variable or else).

nojnhuh (approved):

shashankram (approved):

nojnhuh (approved):

PR #1927, opened 2020-10-27
fix(build): ensure controller version is not a chart version

nojnhuh:
**Description**:
Currently, we make tags of the form `v.X.Y.Z` for binary releases and
`chart/X.Y.Z` for the Helm chart. The version is passed to the
osm-controller binary via a linker flag. If both a chart and regular
semver tag exist on the same commit, the chart tag may be passed instead
of the regular version tag. This change ensures only our regular semver
tags are used for the versions embedded in the binaries.

**Affected area**:

- New Functionality      [ ]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [ ]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests                  [ ]
- CI System              [ ]
- Performance            [ ]
- Other                  [X] (Build)

No

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,7 @@ GOBIN  = $(GOPATH)/bin
 GOX    = go run github.com/mitchellh/gox
 
 CLI_VERSION ?= dev
-CONTROLLER_VERSION = $$(git describe --abbrev=0 --tags)
+CONTROLLER_VERSION = $$(git describe --abbrev=0 --tags --match "v*")
 BUILD_DATE=$$(date +%Y-%m-%d-%H:%M)
 GIT_SHA=$$(git rev-parse HEAD)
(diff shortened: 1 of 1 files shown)


snehachhabria (approved):

Makefile
```diff
@@ -12,7 +12,7 @@ GOBIN  = $(GOPATH)/bin
 GOX    = go run github.com/mitchellh/gox
 
 CLI_VERSION ?= dev
-CONTROLLER_VERSION = $$(git describe --abbrev=0 --tags)
+CONTROLLER_VERSION = $$(git describe --abbrev=0 --tags --match "v*")
```
shashankram, line 15: Would be good to drop a comment regarding the match.

shashankram (approved):

PR #1929, opened 2020-10-27
bug(pkg/routes): Fixing access to backend services 

snehachhabria:
**Description**:

This PR fixes : #1798 

**Affected area**:

- New Functionality      [ ]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [X]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [X]
- Security               [ ]
- Tests                  [ ]
- CI System              [ ]
- Performance            [ ]
- Other                  [ ]

### DIFF
--- a/pkg/catalog/routes.go
+++ b/pkg/catalog/routes.go
@@ -133,19 +133,21 @@ func hostnamesTostr(hostnames []string) string {
 // GetHostnamesForService returns the hostnames for a service.
 // The hostname can be the FQDN for the service, and can include ports as well.
 // Ex. bookstore.default, bookstore.default:80, bookstore.default.svc, bookstore.default.svc:80 etc.
-func (mc *MeshCatalog) GetHostnamesForService(meshService service.MeshService) (string, error) {
+func (mc *MeshCatalog) GetHostnamesForService(meshService service.MeshService) ([]string, error) {
 	log.Trace().Msgf("Finding domain for service %s", meshService)
+	var svcHostnames []string
 
 	if mc.configurator.IsPermissiveTrafficPolicyMode() {
 		hostnames, err := mc.getServiceHostnames(meshService)
 		if err != nil {
 			log.Error().Err(err).Msgf("Error getting service hostnames for MeshService %s", meshService)
-			return "", err
+			return svcHostnames, err
 		}
-		return hostnamesTostr(hostnames), nil
+		return []string{hostnamesTostr(hostnames)}, nil
 	}
 
 	// Retrieve the domain name from traffic split
+	// Domains will be derived from both the root service and backend
 	servicesList := mc.meshSpec.ListTrafficSplitServices()
 	for _, activeService := range servicesList {
 		if activeService.Service == meshService {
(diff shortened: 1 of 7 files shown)


pkg/catalog/routes.go
```diff
@@ -133,19 +133,21 @@ func hostnamesTostr(hostnames []string) string {
...
 
 	if mc.configurator.IsPermissiveTrafficPolicyMode() {
 		hostnames, err := mc.getServiceHostnames(meshService)
 		if err != nil {
 			log.Error().Err(err).Msgf("Error getting service hostnames for MeshService %s", meshService)
-			return "", err
+			return svcHostnames, err
 		}
-		return hostnamesTostr(hostnames), nil
+		return []string{hostnamesTostr(hostnames)}, nil
```
shashankram, line 146: This seems hacky because it is going to return something like []string{"foo,bar,baz"}. Can we avoid this?

pkg/catalog/routes.go
```diff
@@ -155,12 +157,20 @@ func (mc *MeshCatalog) GetHostnamesForService(meshService service.MeshService) (
...
+				return svcHostnames, err
 			}
-			return hostnamesTostr(hostnames), nil
+			hostnames, err := mc.getServiceHostnames(activeService.Service)
+			if err != nil {
+				log.Error().Err(err).Msgf("Error getting service hostnames for Backend service %s", activeService.Service)
+				return svcHostnames, err
+			}
+
+			svcHostnames = append(svcHostnames, hostnamesTostr(rootHostnames))
```
shashankram, line 171: I'd move this to L165 right after `rootHostNames` is computed.

pkg/envoy/rds/response.go
```diff
@@ -56,14 +56,17 @@ func NewResponse(catalog catalog.MeshCataloger, proxy *envoy.Proxy, _ *xds_disco
 			return nil, err
 		}
 
-		// All routes from a given source to destination are part of 1 traffic policy between the source and destination.
-		for _, httpRoute := range trafficPolicy.HTTPRoutes {
-			if isSourceService {
-				aggregateRoutesByHost(outboundAggregatedRoutesByHostnames, httpRoute, weightedCluster, hostnames)
-			}
+		// if the service is referenced in a traffic split there will be more than one hostname (root service and backend)
+		for _, hostname := range hostnames {
```
shashankram, line 60: The API `GetHostnamesForService` was never meant to return something like: 
`[]string{"foo,bar", "baz,bar"}`

We should fix this API and come up with a cleaner fix. Refer to https://github.com/openservicemesh/osm/issues/1316 which tracks this enhancement. This change doesn't align with #1316 because the API is now returning multiple hostname encodings as a list of strings.

I also see that the same route is applied to all each item in the list returned by `GetHostnamesForService`, so why not just return a list where each element does not have a comma separated list of hostnames?

shashankram (changes requested):

snehachhabria:
Closing this PR because of the alternative PR #1944

PR #1944, opened 2020-10-28
bug(pkg/*) : Fixing access to backend services

snehachhabria:
**Description**:
 Fixes : #1798  and #1316 

**Affected area**:

- New Functionality      [ ]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [X]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [X]
- Security               [ ]
- Tests                  [ ]
- CI System              [ ]
- Performance            [ ]
- Other                  [ ]

### DIFF
--- a/pkg/catalog/types.go
+++ b/pkg/catalog/types.go
@@ -102,7 +102,6 @@ type MeshCataloger interface {
 	GetServicesForServiceAccount(service.K8sServiceAccount) ([]service.MeshService, error)
 
 	// GetResolvableHostnamesForUpstreamService returns the hostnames over which an upstream service is accessible from a downstream service
-	// TODO(ref: Issue #1316): return a list of strings
 	GetResolvableHostnamesForUpstreamService(downstream, upstream service.MeshService) ([]string, error)
 
 	//GetWeightedClusterForService returns the weighted cluster for a service
--- a/pkg/catalog/types.go
+++ b/pkg/catalog/types.go
@@ -103,7 +103,7 @@ type MeshCataloger interface {
 
 	// GetResolvableHostnamesForUpstreamService returns the hostnames over which an upstream service is accessible from a downstream service
 	// TODO(ref: Issue #1316): return a list of strings
-	GetResolvableHostnamesForUpstreamService(downstream, upstream service.MeshService) (string, error)
+	GetResolvableHostnamesForUpstreamService(downstream, upstream service.MeshService) ([]string, error)
 
 	//GetWeightedClusterForService returns the weighted cluster for a service
 	GetWeightedClusterForService(service service.MeshService) (service.WeightedCluster, error)
--- a/pkg/envoy/route/config.go
+++ b/pkg/envoy/route/config.go
@@ -12,7 +12,6 @@ import (
 
 	"github.com/openservicemesh/osm/pkg/constants"
 	"github.com/openservicemesh/osm/pkg/envoy"
-	"github.com/openservicemesh/osm/pkg/kubernetes"
 	"github.com/openservicemesh/osm/pkg/service"
 	"github.com/openservicemesh/osm/pkg/trafficpolicy"
 )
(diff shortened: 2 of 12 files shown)


pkg/envoy/route/config.go
```diff
@@ -65,25 +64,25 @@ func UpdateRouteConfiguration(domainRoutesMap map[string]map[string]trafficpolic
...
-	for i := range domains {
-		domains[i] = strings.TrimSpace(domains[i])
+func createVirtualHostStub(namePrefix string, host string, domains set.Set) *xds_route.VirtualHost {
+	var domainsSlice []string
+	for domainIntf := range domains.Iter() {
+		domainsSlice = append(domainsSlice, strings.TrimSpace(domainIntf.(string)))
 	}
 
-	name := fmt.Sprintf("%s|%s", namePrefix, kubernetes.GetServiceFromHostname(domains[0]))
+	log.Trace().Msgf("TEST domains %v", domains)
```
shashankram, line 81: can be removed.

pkg/kubernetes/util.go
```diff
@@ -51,5 +51,9 @@ func GetHostnamesForService(service *corev1.Service, sameNamespace bool) []strin
 func GetServiceFromHostname(host string) string {
 	// The service name is the first string in the host name for a service.
 	// Ex. service.namespace, service.namespace.cluster.local
-	return strings.Split(host, ".")[0]
+	service := strings.Split(host, ".")[0]
+
+	// For services that are not namespaced the service name contains the port as well
+	// Ex. service:port
+	return strings.Split(service, ":")[0]
```
shashankram, line 58: Good find.

pkg/kubernetes/util_test.go
```diff
@@ -56,7 +56,7 @@ var _ = Describe("Hostnames for a kubernetes service", func() {
 		})
 		It("Returns the service name from its hostname", func() {
 			hostname := fmt.Sprintf("%s:%d", service, tests.ServicePort)
-			Expect(GetServiceFromHostname(hostname)).To(Equal(fmt.Sprintf("%s:%d", service, tests.ServicePort)))
+			Expect(GetServiceFromHostname(hostname)).To(Equal(service))
```
shashankram, line 59: Should we add another test for the case where the hostname is of the form `service:port`

pkg/trafficpolicy/types.go
```diff
@@ -27,8 +27,9 @@ type TrafficTarget struct {
 	HTTPRoutes  []HTTPRoute         `json:"http_route:omitempty"`
 }
 
-// RouteWeightedClusters is a struct of an HTTPRoute and associated weighted clusters
+// RouteWeightedClusters is a struct of an HTTPRoute, associated weighted clusters and the domains
 type RouteWeightedClusters struct {
 	HTTPRoute        HTTPRoute `json:"http_route:omitempty"`
 	WeightedClusters set.Set   `json:"weighted_clusters:omitempty"`
+	Domains          set.Set   `json:"domains:omitempty"`
```
shashankram, line 34: We stopped using the term `Domains` when we are actually referring to hostnames.

```suggestion
	Hostnames          set.Set   `json:"domains:omitempty"`
```

pkg/catalog/types.go
```diff
@@ -103,7 +103,7 @@ type MeshCataloger interface {
 
 	// GetResolvableHostnamesForUpstreamService returns the hostnames over which an upstream service is accessible from a downstream service
 	// TODO(ref: Issue #1316): return a list of strings
```
shashankram, line 105: This comment related to #1316 is addressed with this PR.

pkg/kubernetes/util_test.go
snehachhabria: Doesn't this line already cover the case of  `service:port`  ? So adding another test might not be needed

shashankram (approved):
And e2e test for this would be nice to have. Would you mind filing an issue for this.

shashankram (approved):
Thanks for the fix. We should note somewhere (either code or commit message) that this is a short term fix and that issue <issue-id> will rework this.

snehachhabria:
> Thanks for the fix. We should note somewhere (either code or commit message) that this is a short term fix and that issue will rework this.

I will include it in the commit message

eduser25 (approved):
have the test for e2e for this changes. lgtm

PR #1949, opened 2020-10-28
fix(charts): Add volume mount for Fluent Bit

SanyaKochhar:
**Description**: Adds var/lib/docker/containers volume for Fluent Bit to accommodate symlink variations

**Affected area**:

- New Functionality      [ ]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [ ]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests                  [ ]
- CI System              [ ]
- Performance            [ ]
- Other                  [x]

No

### DIFF
--- a/charts/osm/templates/osm-deployment.yaml
+++ b/charts/osm/templates/osm-deployment.yaml
@@ -87,6 +87,9 @@ spec:
           - name: var-log-pods
             mountPath: /var/log/pods
             readOnly: true
+          - name: var-lib-containers
+            mountPath: /var/lib/docker/containers
+            readOnly: true
        {{- end }}
     {{- if .Values.OpenServiceMesh.enableFluentbit }}
       volumes:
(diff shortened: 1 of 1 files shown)


ksubrmnn (approved):

draychev (approved):
We have `var-log-containers`, `var-log-pods`, and now `var-lib-containers`.  We need all three, correct?

SanyaKochhar:
> We have `var-log-containers`, `var-log-pods`, and now `var-lib-containers`. We need all three, correct?

@draychev I added var/lib/containers because apparently symlinks may work differently or break depending on which cluster this is running on. I haven't had issues yet but @sozercan did while setting up Fluent Bit for another project and suggested this change

michelleN (approved):

Issue #1965, opened 2020-10-30
Encode ServiceAccount in service certificates


shashankram:
**Please describe the Improvement and/or Feature Request**
Current, service certificates have a CN of the form <service>.<namespace>.svc.cluster.local, add the ServiceAccount for the service so that service identities can be mapped to TrafficTarget's source and destination identities (which use Kubernetes ServiceAccounts).
The CN for a service certificate is assigned here: https://github.com/openservicemesh/osm/blob/main/pkg/catalog/certificates.go#L10

Required by https://github.com/openservicemesh/osm/issues/1964

**Scope (please mark with X where applicable)**
- New Functionality          [ ]
- Install                    [ ]
- SMI Traffic Access Policy  [ ]
- SMI Traffic Specs Policy   [ ]
- SMI Traffic Split Policy   [ ]
- Permissive Traffic Policy  [ ]
- Ingress                    [ ]
- Egress                     [ ]
- Envoy Control Plane        [ ]
- CLI Tool                   [ ]
- Metrics                    [ ]
- Certificate Management     [X]
- Sidecar Injection          [ ]
- Logging                    [ ]
- Debugging                  [ ]
- Tests                      [ ]
- CI System                  [ ]
- Project Release            [ ]

PR #1968, opened 2020-10-30
identity: add abstraction for service identity

shashankram:
**Description**:
Adds a new `ServiceIdentity` type that is to be used to provide
identity to services managed by OSM. Provides an API to use
Kubernetes ServiceAccount as service identities.

The kubernetes based service identity makes use of a trust domain
that determines the domain within which an identity can be trusted.
Within the same cluster, a Kubernetes service identity would be
"service-account.namespace.cluster.local". The trust domain provides
the ability to extend a `ServiceIdentity` beyond kubernetes domains.

Part of #1965

**Affected area**:

- New Functionality      [ ]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [ ]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests                  [ ]
- CI System              [ ]
- Performance            [ ]
- Other                  [X]

`No`

### DIFF
--- a/pkg/identity/kubernetes.go
+++ b/pkg/identity/kubernetes.go
@@ -1,0 +1,17 @@
+package identity
+
+import (
+	"strings"
+
+	"github.com/openservicemesh/osm/pkg/service"
+)
+
+const (
+	identityDelimiter = "."
+)
+
+// GetKubernetesServiceIdentity returns the ServiceIdentity based on Kubernetes ServiceAccount and a trust domain
+func GetKubernetesServiceIdentity(svcAccount service.K8sServiceAccount, trustDomain string) ServiceIdentity {
+	si := strings.Join([]string{svcAccount.Name, svcAccount.Namespace, trustDomain}, identityDelimiter)
+	return ServiceIdentity(si)
+}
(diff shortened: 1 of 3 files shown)


snehachhabria (approved):

draychev (approved):
This looks good to me.
I find `trustDomain` to be a bit ambiguous - but looking at the unit tests it becames clear what this does.

shashankram:
> This looks good to me.
> I find `trustDomain` to be a bit ambiguous - but looking at the unit tests it becames clear what this does.

I should have added the same details present in the commit message regarding the `trustDomain` to the code. Will follow this up shortly.

michelleN (commented):
👍 Looks great @shashankram. I'd prefer if we merged this with a functional change that implements ServiceIdentity though. Non blocking comment but just a preference.

shashankram:
> 👍 Looks great @shashankram. I'd prefer if we merged this with a functional change that implements ServiceIdentity though. Non blocking comment but just a preference.

The functional change using this will be a follow-up PR. Preferred to separate this out because the functional change leveraging this could be significantly large. Tagged this PR against #1965 so that we can link the functional changes to this PR for future reference.

Issue #1909, opened 2020-10-23
Add reconciler for osm resources


ritazh:
**Please describe the Improvement and/or Feature Request**

- webhook mutatingwebhookconfiguration and secret
   Watch for changes in osm-webhook mwhc and secret resources to reconcile/ensure the CA cert in the mwhc is using the most recent CA cert from the secret.
- configmap
- custom resources

## Definition of **done**: 
This issue should result in a documentation that describes what resources osm owns and wants to reconcile. Then we can create followup issues for any reconcilers.

**Scope (please mark with X where applicable)**
- New Functionality          [ ]
- Install                    [ ]
- SMI Traffic Access Policy  [ ]
- SMI Traffic Specs Policy   [ ]
- SMI Traffic Split Policy   [ ]
- Permissive Traffic Policy  [ ]
- Ingress                    [ ]
- Egress                     [ ]
- Envoy Control Plane        [ ]
- CLI Tool                   [ ]
- Metrics                    [ ]
- Certificate Management     [X]
- Sidecar Injection          [ ]
- Logging                    [ ]
- Debugging                  [ ]
- Tests                      [ ]
- CI System                  [ ]
- Project Release            [ ]

**Possible use cases**

shashankram:
> Watch for changes in osm-webhook mwhc and secret resources to reconcile/ensure the CA cert in the mwhc is using the most recent CA cert from the secret.

The CA bundle in the mutatingwebhookconfigration gets patched as a part of the webhook's initialization: https://github.com/openservicemesh/osm/blob/0be40ed8bdb7dd64866f3c6812d73eb0380f5d15/pkg/injector/webhook.go#L80

The CA bundle gets created at the time of osm-controller startup, at which time the webhook gets initialized. So I am unsure why a reconciler is neccessary for this. 

@ritazh could you clarify where this would be applicable?

ritazh:
@shashankram sure. for example, you can modify the mutatingwebhookconfigration resource by using `kubectl edit`

shashankram:
> @shashankram sure. for example, you can modify the mutatingwebhookconfigration resource by using `kubectl edit` directly to updat/remove the caBundle.

[...]

PR #1976, opened 2020-11-03
feat(pkg/reconciler) : Adding a reconciler in OSM for the CA bundle in the mutating webhook configuration

snehachhabria:
**Description**:

This PR added a controller that watched for mutating webhook resources in OSM and updates only OSM's mutating webhook with the latest CA bundle, whenever it is missing ( due to deletion) 

Fixes : #1909 

**Affected area**:

- New Functionality      [X]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [X]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests                  [ ]
- CI System              [ ]
- Performance            [ ]
- Other                  [ ]


cmd/osm-controller/osm-controller.go
```diff
@@ -98,6 +110,15 @@ func init() {
...
 	flags.BoolVar(&optionalFeatures.Backpressure, "enable-backpressure-experimental", false, "Enable experimental backpressure feature")
+
+	// k8s controller manager options
+	flag.StringVar(&metricsAddr, "metrics-addr", ":8080", "The address the metric endpoint binds to.")
+	flag.BoolVar(&enableLeaderElection, "enable-leader-election", false,
+		"Enable leader election for controller manager. Enabling this will ensure there is only one active controller manager.")
+
+	_ = clientgoscheme.AddToScheme(scheme)
+	_ = v1beta1.AddToScheme(scheme)
+	// +kubebuilder:scaffold:scheme
```
shashankram, line 121: Could you describe as code comments what controller manager is, what metrics address means here, and so on.

cmd/osm-controller/osm-controller.go
```diff
@@ -229,6 +250,43 @@ func main() {
...
+		os.Exit(1)
+	}
+	// +kubebuilder:scaffold:builder
+
+	log.Info().Msg("starting manager")
+	if err := mgr.Start(ctrl.SetupSignalHandler()); err != nil {
+		log.Error().Err(err).Msg("problem running manager")
+		os.Exit(1)
+	}
+	log.Info().Msg("Successfully running controller manager")
```
shashankram, line 288: Would be better to move this initialization to the a separate function.

cmd/osm-controller/osm-controller.go
```diff
@@ -229,6 +250,43 @@ func main() {
...
+		LeaderElection:     enableLeaderElection,
+		Port:               9443,
+		Namespace:          osmNamespace,
+	})
+	if err != nil {
+		log.Error().Err(err).Msg("Error starting up controller manager")
+		os.Exit(1)
+	}
+
+	log.Info().Msg("Successfully setup controller for reconcile")
```
shashankram, line 267:
```suggestion
	log.Info().Msg("Successfully setup controller for resource reconciliation")
```

pkg/reconciler/mutatingwebhook/mutatingwebhook_controller.go
```diff
@@ -0,0 +1,81 @@
+package reconciler
```
shashankram, line 1: Would it make sense to rename this file to `reconcile.go`? This could be a pattern we follow for the different resource reconciliation, ex: `reconciler/<resource>/reconcile.go`. If the same file name cannot be reused within a package, we could keep it as is.

pkg/reconciler/mutatingwebhook/mutatingwebhook_controller.go
```diff
@@ -0,0 +1,81 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/openservicemesh/osm/pkg/certificate"
+	"github.com/openservicemesh/osm/pkg/constants"
+	"github.com/openservicemesh/osm/pkg/logger"
+)
+
+var log = logger.New("mutatingwebhook-reconciler")
```
shashankram, line 18: I'd create a logger for the reconciler package to be consistent with existing logging facility, where we use package scoped loggers.

pkg/reconciler/mutatingwebhook/mutatingwebhook_controller.go
```diff
@@ -0,0 +1,81 @@
...
+type MutatingWebhookConfigrationReconciler struct {
+	client.Client
+	Scheme       *runtime.Scheme
+	OsmWebhook   string
+	OsmNamespace string
+	CertManager  certificate.Manager
+}
+
+// +kubebuilder:rbac:groups=admissionregistration.k8s.io,resources=mutatingwebhookconfigurations,verbs=get;list;watch;update;
+// +kubebuilder:rbac:groups=admissionregistration.k8s.io,resources=mutatingwebhookconfigurations/status,verbs=get;update;patch
```
shashankram, line 30: Why is this required?

pkg/reconciler/mutatingwebhook/mutatingwebhook_controller.go
```diff
@@ -0,0 +1,81 @@
...
+		return ctrl.Result{}, client.IgnoreNotFound(err)
+	}
+
+	// reconcile only for OSM mutatingWebhookConfiguration
+	var shouldUpdate bool
+	if r.OsmWebhook == instance.Name {
+		//check if CA bundle exists on webhook
+		for idx, webhook := range instance.Webhooks {
+			// CA bundle missing for webhook, update webhook to include the latest CA bundle
+			if webhook.Name == "osm-inject.k8s.io" && webhook.ClientConfig.CABundle == nil {
```
shashankram, line 48: Could you make `mutatingWebhookName` an exported constant in the `webhook.go` file and use it as `injector.MutatingWebhookName` here.

pkg/reconciler/mutatingwebhook/mutatingwebhook_controller.go
```diff
@@ -0,0 +1,81 @@
...
+				log.Trace().Msgf("CA bundle missing for webhook : %s ", req.Name)
+				shouldUpdate = true
+				cn := certificate.CommonName(fmt.Sprintf("%s.%s.svc", constants.OSMControllerName, r.OsmNamespace))
+				cert, err := r.CertManager.GetCertificate(cn)
+				if err != nil {
+					return ctrl.Result{}, errors.Errorf("Unable to update mutating webhook, Error getting certificate for the mutating webhook: %+v", err)
```
shashankram, line 54: Could you augment the log message to something as:
```suggestion
					return ctrl.Result{}, errors.Errorf("Error updating mutating webhook, unable to get certificate for the mutating webhook %s: %s" webhookName, err)
```

pkg/reconciler/mutatingwebhook/mutatingwebhook_controller.go
```diff
@@ -0,0 +1,81 @@
...
+				}
+				instance.Webhooks[idx].ClientConfig.CABundle = cert.GetCertificateChain()
+			}
+		}
+	}
+
+	if !shouldUpdate {
+		log.Trace().Msgf("Mutating webhook configuration already compliant %s", req.Name)
```
shashankram, line 62: Please log the name of the mutatingwebhookconfiguration.

pkg/reconciler/mutatingwebhook/mutatingwebhook_controller.go
```diff
@@ -0,0 +1,81 @@
...
+		return ctrl.Result{}, nil
+	}
+
+	if err := r.Update(ctx, instance); err != nil {
+		log.Error().Err(err).Msgf("failure reading object %s", req.NamespacedName.String())
```
shashankram, line 67:
```suggestion
		log.Error().Err(err).Msgf("Error updating mutatingwebhookconfiguration %s: %s", webhookConfigName, err)
```

pkg/reconciler/mutatingwebhook/mutatingwebhook_controller.go
```diff
@@ -0,0 +1,81 @@
...
+		return ctrl.Result{Requeue: false}, nil
+	}
+
+	log.Trace().Msgf("Successfully updated mutating webhook configuration CA bundle for : %s ", req.Name)
```
shashankram, line 71:
```suggestion
	log.Trace().Msgf("Successfully updated mutating webhook configuration CA bundle for mutatingwebhookconfiguration %s ", req.Name)
```

Is `req.Name` the name of the webhook config?

pkg/reconciler/mutatingwebhook/suite_test.go
```diff
@@ -0,0 +1,69 @@
...
+		[]Reporter{printer.NewlineReporter{}})
+}
+
+var _ = BeforeSuite(func(done Done) {
+	logf.SetLogger(zap.New())
+
+	By("bootstrapping test environment")
+	testEnv = &envtest.Environment{
+		CRDDirectoryPaths: []string{filepath.Join("..", "config", "crd", "bases")},
+	}
```
shashankram, line 40: What's happening here? Is this specific to your environment?

snehachhabria: This is specific to the kubernetes test environment to start/stop a kubernetes control plane for testing the controllers

pkg/reconciler/mutatingwebhook/mutatingwebhook_controller.go
snehachhabria: yes `req.Name` is the name for the requesting mutatingwebhook i.e. the webhook which triggered the reconciler

pkg/reconciler/mutatingwebhook/mutatingwebhook_controller.go
snehachhabria: it is already being logged with `req.Name`

pkg/reconciler/mutatingwebhook/mutatingwebhook_controller.go
snehachhabria: I've updated the logger to be package scoped

pkg/reconciler/mutatingwebhook/mutatingwebhook_controller.go
snehachhabria: we are using kubebuilder to test the controllers, and these lines add the necessary permissions in kubebuilders to listen on events related to muutatingwebhooks

pkg/reconciler/mutatingwebhook/mutatingwebhook_controller.go
snehachhabria: seems like it worked, so I have updated the naming

cmd/osm-controller/osm-controller.go
snehachhabria: done

cmd/osm-controller/osm-controller.go
snehachhabria: added some comments

pkg/reconciler/mutatingwebhook/suite_test.go
shashankram: What does kubernetes control plane refer to here?

pkg/reconciler/mutatingwebhook/mutatingwebhook_controller.go
shashankram: I don't understand this since I am not familiar with kubebuilder, but if you think a few comments on this could help readability they would be welcome.

pkg/reconciler/mutatingwebhook/mutatingwebhook_controller.go
shashankram: ok, I'd mention the resource name as well as suggested in the comment.

pkg/reconciler/mutatingwebhook/suite_test.go
snehachhabria: Check these two docs : https://book-v1.book.kubebuilder.io/beyond_basics/running_tests.html , https://book.kubebuilder.io/reference/envtest.html. Do you think it would be good to reference these in the code as comments so that everyone can use it as a reference ?

pkg/reconciler/mutatingwebhook/mutatingwebhook_controller.go
snehachhabria: I'm not sure what comments I can add, but do you think a reference to a doc would help ?

shashankram (approved):

pkg/reconciler/mutatingwebhook/reconcile.go
```diff
@@ -0,0 +1,82 @@
...
+	instance := &v1beta1.MutatingWebhookConfiguration{}
+
+	if err := r.Get(ctx, req.NamespacedName, instance); err != nil {
+		log.Error().Err(err).Msgf("failure reading object %s ", req.NamespacedName.String())
+		return ctrl.Result{}, client.IgnoreNotFound(err)
+	}
+
+	// reconcile only for OSM mutatingWebhookConfiguration
+	var shouldUpdate bool
+	if r.OsmWebhook == instance.Name {
```
ritazh, line 45: Should we check for if `req.NamespacedName` is OSM mutatingWebhookConfig at the beginning of this function?

pkg/reconciler/mutatingwebhook/reconcile.go
```diff
@@ -0,0 +1,82 @@
...
+	}
+
+	if !shouldUpdate {
+		log.Trace().Msgf("Mutatingwebhookconfiguration %s already compliant", req.Name)
+		return ctrl.Result{}, nil
+	}
+
+	if err := r.Update(ctx, instance); err != nil {
+		log.Error().Err(err).Msgf("Error updating mutatingwebhookconfiguration %s: %s", req.Name, err)
+		return ctrl.Result{Requeue: false}, nil
```
ritazh, line 69: If the update fails, do we not want to retry?

cmd/osm-controller/osm-controller.go
```diff
@@ -353,3 +383,44 @@ func getOSMControllerPod(kubeClient kubernetes.Interface) (*corev1.Pod, error) {
...
 }
+
+//Setting up k8s controller manager to reconcile OSM resources
+func createControllerManagerForOSMResources(certManager certificate.Manager) error {
+	log.Info().Msg("Setting up controller manager to reconcile OSM resources")
+	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
+		Scheme:             scheme,
+		MetricsBindAddress: metricsAddr,
+		LeaderElection:     enableLeaderElection,
+		Port:               9443,
```
ritazh, line 394: Do you need `Port`?

snehachhabria: That's right, I have removed it

pkg/reconciler/mutatingwebhook/reconcile.go
snehachhabria: I thought a mutatingwebhookconfiguration is namespace independent and hence checking against the OSM namespace would be redundant for this particular scenario

pkg/reconciler/mutatingwebhook/reconcile.go
snehachhabria: We could update to retry, and that would help for transient issues but it there's an actual issue then we would be stuck in a reconcile loop. WDYT ?

pkg/reconciler/mutatingwebhook/reconcile.go
ritazh: I was suggesting that we check the request mutatingwebhookconfig's name `req.NamespacedName` at the beginning of the function so that we can skip for other mutatingwebhookconfigs.

pkg/reconciler/mutatingwebhook/reconcile.go
ritazh: If there is an error, then we should return the error so that it can be retried. 
```go
return ctrl.Result{}, client.IgnoreNotFound(err)
```

pkg/reconciler/mutatingwebhook/reconcile.go
snehachhabria: okay makes sense, let me update this. I will use `req.Name` though, since there's no namespace for mutatingwebhookconfigurations

ritazh: SG! Both should be fine.

pkg/reconciler/mutatingwebhook/reconcile.go
ritazh: Do we also need to check if the secret is deleted?

snehachhabria: I've updated the retry logic when the update fails. 

As of the secret being deleted, the GetCertificate handles that already so I think we should be okay

shashankram (approved):

ritazh (approved):
lgtm

PR #1984, opened 2020-11-04
fix(install): clean up resources when install fails

### LINKED ISSUE #1530: Clean up resources if Helm installation fails
**Bug description**:
If helm installation fails, resources are not cleaned up such as clusterrolebindings etc. We are left with a half-deployed state.

**Expected behavior**:
If install fails, the cluster should be as it was pre-installation
TBD
**How was OSM installed?**:
TBD
TBD
**Environment**:
- Others:

nojnhuh:
**Description**:
This change adds the behavior of the `--atomic` flag to our invocation
of `helm install` in the `osm install` command. With `--atomic`, Helm
will uninstall the release if it fails to install. `--atomic` also
implies `--wait` which will wait for pods, pod workloads, services,
PVCs, and CRDs to become ready before exiting.

A `--timeout` flag is also added to control the maximum amount of time
to wait for the resources to install and become ready. Its value is
passed through to Helm and behaves exactly the same as the `--timeout`
flag for `helm install`.

Fixes #1530

**Affected area**:

- New Functionality      [ ]
- Documentation          [ ]
- Install                [X]
- Control Plane          [ ]
- CLI Tool               [X]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests                  [ ]
- CI System              [ ]
- Performance            [ ]
- Other                  [ ]

No


cmd/cli/install.go
```diff
@@ -160,6 +161,7 @@ func newInstallCmd(config *helm.Configuration, out io.Writer) *cobra.Command {
 	f.BoolVar(&inst.deployJaeger, "deploy-jaeger", true, "Deploy Jaeger in the namespace of the OSM controller")
 	f.StringVar(&inst.envoyLogLevel, "envoy-log-level", "error", "Envoy log level is used to specify the level of logs collected from envoy and needs to be one of these (trace, debug, info, warning, warn, error, critical, off)")
 	f.BoolVar(&inst.enforceSingleMesh, "enforce-single-mesh", false, "Enforce only deploying one mesh in the cluster")
+	f.DurationVar(&inst.timeout, "timeout", 5*time.Minute, "time to wait for installation and resources in a ready state, zero means no timeout")
```
ksubrmnn, line 164: Can you capitalize the description to match the others? Or lowercase the others, whichever is standard

ksubrmnn (changes requested):
One nit, otherwise great change!

cmd/cli/install.go
shashankram: I think we need to follow a convention in the cli help text for flags. I noticed this inconsistency as well across cli commands.

tests/e2e/common.go
```diff
@@ -425,7 +426,7 @@ func (td *OsmTestData) InstallOSM(instOpts InstallOSMOpts) error {
 		return errors.Wrap(err, "failed to run osm install")
 	}
 
-	return td.WaitForPodsRunningReady(td.osmNamespace, 90*time.Second, 1)
+	return nil
```
eduser25, line 429: With this we are not checking if OSM actually came up. I'd be good to have the test fail here if OSM does not come up (ie. ErrPullImage) instead of later when we expect the applications or traffic to happen, which will be misleading when debugging.

Also not saying current implementation is best (as, if there are other containers in the NS, this check doesn't really work), but it did work in most circumstances.

nojnhuh: We are checking that the osm-controller pod is running, but it's now checked by Helm during `osm install` before that command exits, making this wait in the framework redundant. Any extra wait calls for pods in the osm-system namespace in the individual tests are also made redundant by this change.

ksubrmnn (approved):

shashankram (approved):

PR #1987, opened 2020-11-04
test(cli): Consolidate resolveValues tests

ksubrmnn:
**Description**:
Previously there were tests that checked if certain values resolved properly. Some tests checked for both true and false, but this is not necessary to check since we just need to see if the values are set. Also this led to a high number of similar tests that made the file difficult to read through.

I created one test to check if all values are resolved properly. 

**Affected area**:

- New Functionality      [ ]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [ ]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests                  [X]
- CI System              [ ]
- Performance            [ ]
- Other                  [ ]

No


cmd/cli/install_test.go
```diff
@@ -1407,57 +784,37 @@ var _ = Describe("Test envoy log level types", func() {
...
+		// parses Helm strvals line and merges into a map
+		err := strvals.ParseInto(val, expectedValues)
+		assert.Nil(err)
+	}
 
-		vals, err := installCmd.resolveValues()
-		Expect(err).NotTo(HaveOccurred())
+	vals, err := installCmd.resolveValues()
+	assert.Nil(err)
+	assert.NotNil(vals)
```
nojnhuh, line 814: Is this redundant given the equality assertion below this one?

shashankram (commented):
I think it makes sense to clean up a lot of duplicate tests that test `resolveValues()`. Is this PR only removing tests that are resolving chart values or also tests that rely on validation testing for cli parameters that `resolveValues()` doesn't handle?

Also I would be interested in knowing what the test coverage for the file is before and after this change (hoping it isn't going down).

cmd/cli/install_test.go
```diff
@@ -1407,57 +784,37 @@ var _ = Describe("Test envoy log level types", func() {
...
+	valuesConfig := []string{
+		fmt.Sprintf("OpenServiceMesh.imagePullSecrets[0].name=%s", testRegistrySecret),
+		fmt.Sprintf("OpenServiceMesh.vault.host=%s", testVaultHost),
+		fmt.Sprintf("OpenServiceMesh.vault.token=%s", testVaultToken),
+	}
+	for _, val := range valuesConfig {
+		// parses Helm strvals line and merges into a map
+		err := strvals.ParseInto(val, expectedValues)
+		assert.Nil(err)
+	}
```
nojnhuh, line 810: What's the purpose of this step vs. adding them to `expectedValues` by default?

cmd/cli/install_test.go
```diff
@@ -1620,3 +977,77 @@ func createDeploymentSpec(namespace, meshName string) *v1.Deployment {
 		},
 	}
 }
+
+func getDefaultInstallCmd(writer *bytes.Buffer) installCmd {
```
nojnhuh, line 981: Do these values need to stay in sync with the defaults in install.go? If so, can we define and make use of the defaults there and reference them here in the tests so we don't have to manually keep them in sync?

ksubrmnn:
> I think it makes sense to clean up a lot of duplicate tests that test `resolveValues()`. Is this PR only removing tests that are resolving chart values or also tests that rely on validation testing for cli parameters that `resolveValues()` doesn't handle?
> 
> Also I would be interested in knowing what the test coverage for the file is before and after this change (hoping it isn't going down).

Coverage is 40.3% on main and with this PR (so no change). This test doesn't remove any of the interesting checks (like meshName existing, envoy log level invalid), etc. Those are handled in run/validateOptions and those tests have not been removed.

cmd/cli/install_test.go
```diff
@@ -1407,57 +784,37 @@ var _ = Describe("Test envoy log level types", func() {
...
-			serviceCertValidityDuration: "24h",
-		}
+	// Fill out fields which are empty by default
+	installCmd.containerRegistrySecret = testRegistrySecret
+	installCmd.vaultHost = testVaultHost
+	installCmd.vaultToken = testVaultToken
 
-		vals, err := installCmd.resolveValues()
-		Expect(err).NotTo(HaveOccurred())
+	expectedValues := getDefaultValues()
```
nojnhuh, line 798: I see both `getDefaultValues()` and `getDefaultInstallCmd()` are only used once in this test. Are they likely to be used in other tests in the future? If not, I would personally find it easier to read if the values were hardcoded right in the test vs. in separate functions.

shashankram:
> > I think it makes sense to clean up a lot of duplicate tests that test `resolveValues()`. Is this PR only removing tests that are resolving chart values or also tests that rely on validation testing for cli parameters that `resolveValues()` doesn't handle?
> > Also I would be interested in knowing what the test coverage for the file is before and after this change (hoping it isn't going down).
> 
> Coverage is 40.3% on main and with this PR (so no change). This test doesn't remove any of the interesting checks (like meshName existing, envoy log level invalid), etc. Those are handled in run/validateOptions and those tests have not been removed.

Thanks for clarifying.

cmd/cli/install_test.go
ksubrmnn: I want to keep getDefaultValues() as just what you would get from the default values for the install cmd. That way those functions can be reused in a follow up PR to replace all the lines where we copy and paste the default config.

cmd/cli/install_test.go
ksubrmnn: Yep, plan to use these in the future! For example, the first couple tests just test the default params. Then subsequent tests often just change on param. it'll help us easily read in a test exactly which fields are being tested since we seem to arbitrarily specify random fields in any given test.

cmd/cli/install_test.go
ksubrmnn: Done

nojnhuh (approved):

shashankram (approved):

draychev (approved):
Fantastic! Thank you!

PR #1992, opened 2020-11-05
clean-up : fixing test naming

snehachhabria:
**Description**:
Most of the tests in `suite_test.go`  were named as `TestEndpoints`, this PR cleans up and fixes the naming of the tests in the various packages appropriately. No functional changes, only renaming

**Affected area**:

- New Functionality      [ ]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [ ]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests                  [X]
- CI System              [ ]
- Performance            [ ]
- Other                  [ ]

### DIFF
--- a/pkg/envoy/suite_test.go
+++ b/pkg/envoy/suite_test.go
@@ -7,7 +7,7 @@ import (
 	. "github.com/onsi/gomega"
 )
 
-func TestEnvoySds(t *testing.T) {
+func TestEnvoy(t *testing.T) {
 	RegisterFailHandler(Fail)
-	RunSpecs(t, "Envoy SDS Test Suite")
+	RunSpecs(t, "Envoy Test Suite")
 }
--- a/pkg/envoy/suite_test.go
+++ b/pkg/envoy/suite_test.go
@@ -7,7 +7,7 @@ import (
 	. "github.com/onsi/gomega"
 )
 
-func TestEndpoints(t *testing.T) {
+func TestEnvoySds(t *testing.T) {
 	RegisterFailHandler(Fail)
-	RunSpecs(t, "Test Suite")
+	RunSpecs(t, "Envoy SDS Test Suite")
 }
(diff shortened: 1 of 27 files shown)


pkg/envoy/suite_test.go
```diff
@@ -7,7 +7,7 @@ import (
 	. "github.com/onsi/gomega"
 )
 
-func TestEndpoints(t *testing.T) {
+func TestEnvoySds(t *testing.T) {
 	RegisterFailHandler(Fail)
-	RunSpecs(t, "Test Suite")
+	RunSpecs(t, "Envoy SDS Test Suite")
```
nojnhuh, line 12: Seems like we should omit SDS here.

pkg/featureflags/suite_test.go
```diff
@@ -7,7 +7,7 @@ import (
 	. "github.com/onsi/gomega"
 )
 
-func TestEndpoints(t *testing.T) {
+func TestFEatureFlags(t *testing.T) {
```
nojnhuh, line 10:
```suggestion
func TestFeatureFlags(t *testing.T) {
```

nojnhuh (changes requested):
Thanks for these updates! Left a couple nits inline.

pkg/envoy/suite_test.go
snehachhabria: There are a bunch of discovery services for Envoy and their tests are named as EnvoyRds, EnvoyEds, EnvoySds. So I'd prefer if we have this as is

nojnhuh: I see the pkg/envoy/sds package is named like this which makes sense, but it's not clear to me why pkg/envoy should also be named this way.

snehachhabria: You are right, I missed that this was for pkg/envoy. I have updated the name

nojnhuh (approved):

shashankram (approved):

PR #2000, opened 2020-11-06
configurator: optimize ConfigMap informer filtering

shashankram:
**Description**:
This change makes use of the filtering capability in informers
to filter API responses on the server side, by using a FieldSelector
to filter on the name of the ConfigMap as opposed to filtering all
objects on the client (osm-controller) side. The informer is already
namespaced, so the namespace and name should be enough to accurately
only have OSM's ConfigMap in the informer cache and events.

This change will also reduce the number of ConfigMap related events
the controller will receive when there are additional ConfigMaps in
the controller's namespace (ex. Grafana's ConfigMap when Grafana is
deployed as a part of control plane), and thus improving performance
in such scenarios. The informer resyncs are on a per resource basis,
so unnecessary events can take up processing resources.

**Affected area**:

- New Functionality      [ ]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [X]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests                  [ ]
- CI System              [ ]
- Performance            [X]
- Other                  [ ]

`No`

### DIFF
--- a/pkg/configurator/client.go
+++ b/pkg/configurator/client.go
@@ -2,10 +2,11 @@ package configurator
 
 import (
 	"fmt"
-	"reflect"
 	"strconv"
 
 	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes"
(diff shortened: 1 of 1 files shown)


nojnhuh (approved):

draychev (approved):
Good stuff!
Thank you @shashankram !

PR #2005, opened 2020-11-06
Chart changes with additional configurations

nshankar13:
**Description**:
Parameterize additional chart values

**Affected area**:

- New Functionality      [ ]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [ ]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests                  [ ]
- CI System              [ ]
- Performance            [ ]
- Other                  [X ]

### DIFF
--- a/charts/osm/templates/fluentbit-configmap.yaml
+++ b/charts/osm/templates/fluentbit-configmap.yaml
@@ -21,7 +21,7 @@ data:
       match      *
       regex       message /"level":"error"/
     [OUTPUT]
-      Name    {{ Values.OpenServiceMesh.fluentbitOutput }}
+      Name    {{ .Values.OpenServiceMesh.fluentbitOutput }}
       Match   *
       {{- if !.Values.OpenServiceMesh.customerLogAnalyticsWorkspace }}
       Customer_ID {{ .Values.OpenServiceMesh.customerID }}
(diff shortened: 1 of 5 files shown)


charts/osm/values.yaml
```diff
@@ -19,6 +19,7 @@ OpenServiceMesh:
     protocol: http
     token:
     role: openservicemesh
+  connectVault: true
```
shashankram, line 22: A comment on what this does will be helpful.

shashankram (approved):

charts/osm/values.yaml
nojnhuh: Even better than a comment, an entry in values.schema.json with a description.

charts/osm/templates/fluentbit-configmap.yaml
```diff
@@ -21,8 +21,13 @@ data:
       match      *
       regex       message /"level":"error"/
     [OUTPUT]
-      Name    stdout
+      Name    {{ .Values.OpenServiceMesh.fluentbitOutput }}
```
ritazh, line 24: Maybe this whole section under `[OUTPUT]` should be parameterized as each output plugin will have its own configurations. 
WDYT @SanyaKochhar

ritazh: On second thought, it's good to have the ability to set these fields individually.

ksubrmnn (approved):

snehachhabria (approved):

Issue #1578, opened 2020-08-18
Address golangci-lint issues


nojnhuh:
Currently, we only run `golangci-lint` against new changes. We should strive to remove that [option](https://github.com/openservicemesh/osm/blob/55af4af59d871b5be83b9012e9bda5d279a453a3/.golangci.yml#L5) from our config and enforce rules on the entire codebase once outstanding issues are fixed.

snehachhabria:
This was addressed with #2023

PR #2023, opened 2020-11-11
fix(*) : Fixing existing lint issues in the repo 

snehachhabria:
**Description**:

This PR fixes all the existing linter issues. 

According to the issue #1324 there were a bunch of lint issues on the repo, not all the errors mentioned in the issue appeared.  I have fixed all the errors that are appearing currently. Further the PR enables the linters check for the entire repo going forward.

Fixes : #1324 , #1578 

**Affected area**:

- New Functionality      [ ]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [ ]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests                  [ ]
- CI System              [X]
- Performance            [ ]
- Other                  [ ]


.golangci.yml
```diff
@@ -13,17 +13,30 @@ linters:
     - misspell
     - unused
     - whitespace
+  disable:
```
snehachhabria, line 16: errcheck is a program for checking for unchecked errors in go programs. Disabling this adds flexibility in the code base by preventing every function that returns an error from being assigned to a variable or being checked

.golangci.yml
```diff
@@ -13,17 +13,30 @@ linters:
...
+    - errcheck
 
 linters-settings:
   goimports:
     local-prefixes: github.com/openservicemesh/osm
 
 issues:
-  new: true
```
snehachhabria, line 22: going forward linters will check the entire repository

.golangci.yml
```diff
@@ -13,17 +13,30 @@ linters:
...
+    - path: pkg/certificate/providers/keyvault/
+      linters: 
+        - unused
+        - deadcode
+    # Exclude staticcheck messages for deprecated function, variable or constant
+    # This causes issues with package github.com/golang/protobuf/proto
+    - linters:
+        - staticcheck
+      text: "SA1019:"
+  exclude-use-default: false
```
snehachhabria, line 41: This flag will ensure that the `golint` check locally and on the CI return the same result

nojnhuh (dismissed):

.golangci.yml
shashankram: This is bad and good based on usage. Disabling this will result in code not handling returned errors not being flagged. On the other hand, its common not to check for errors when using an api like `Fprintf`.

pkg/strings/which.go
```diff
@@ -1,7 +1,9 @@
 package strings
 
+// Which is a struct of strings
```
shashankram, line 3:
```suggestion
// Which is a slice of strings
```

.golangci.yml
snehachhabria: That's true, I have been debating this and I'm not sure what we would prefer

snehachhabria: Based on an offline sync with @draychev we will attempt to enable errcheck in a separate PR

draychev (approved):
This looks great!
Feel free to merge and continue working on it.
I personally would prefer to have `errcheck` enabled (I practice this lint rule myself)

draychev (approved):

nojnhuh (approved):

PR #2035, opened 2020-11-11
Routes refactor

michelleN:
related to #2034

Issue #2034, opened 2020-11-11
refactor how inbound and outbound VirtualHosts are programmed on Envoy proxies


michelleN:
The way we currently couple SMI Traffic Split and SMI Traffic Target policies has proven difficult when trying to program the correct Envoy RouteConfiguration for inbound and outbound VirtualHosts to implement SMI Traffic Split v1alpha3. Until now, this was not an issue because HTTP route matching rules were only associated with Traffic Target resources. However, there are now a series of issues related to programming Virtual Hosts correctly on proxies since SMI Traffic Splits (as of v1alpha3) also have associated HTTP route matching rules.

In order to better predict behavior and also comply with the latest SMI Traffic Target and SMI Traffic Split APIs, we'll need to update the way we program inbound and outbound RouteConfiguration with regard to Traffic Targets and Traffic Splits.

When an SMI Traffic Target is applied, the proxies associated with the destination should be programmed with inbound VirtualHost with route matches corresponding to the route matches in the SMI Traffic Target resource. The proxies associated with any of the sources in the same SMI Traffic Target should be programmed with an outbound VirtualHost with an allow all route match (allow all paths/PathRegex and allow all Methods). What this means to the user is that when a Traffic Target is applied, the relevant clients (sources) will be able to discover and make requests to the destination and the server (destination) will be able to accept requests on the route matches associated with the Traffic Target for all hostnames associated with the Kubernetes Services related to the destination Service Account. (Note: RBAC for inbound connections still needs to be implemented - being tracked [here](https://github.com/openservicemesh/osm/issues/1964))

When an SMI TrafficSplit is applied, all clients will be programmed a Virtual Host corresponding to the route matching rules and weighted cluster split described in the TrafficSplit resource. If there is a RouteConfiguration that already exists for the apex service described in the Traffic Split, the routing rules and weighted cluster configuration will be merged in the routes list of the existing Virtual Host. For the user what this means is that any time a Traffic Split is applied, all services will be able to discover and make requests to the apex service. If the request matches the route matching rules described in the TrafficSplit resource, the requests will be load balanced according to the weighted clusters configuration. Access will be denied on the server side if there is no inbound VirtualHost programmed (because a valid TrafficTarget policy does not exist) and more granularly once RBAC is implemented(#1964).

See some more details [here](https://github.com/openservicemesh/osm/issues/705#issuecomment-716723318).

[...]

PR #2036, opened 2020-11-11
feat(pkg/catalog): Add routesFromRules func

michelleN:
This func allows one to get a list of trafficpolicy.HTTPRoute objects from a set of rules from an SMI Traffic Target resource. The larger change can be seen in #2035 where routesFromRules is named HTTPRoutesFromRules (renamed to allow for the function to later also return TCPRoutes from rules in the future).

There are no functional changes to the system in this PR. It is part of #2034 and a chunk of #2035.

### DIFF
--- a/pkg/catalog/routes.go
+++ b/pkg/catalog/routes.go
@@ -405,3 +405,33 @@ func (mc *MeshCatalog) listTrafficTargetPermutations(trafficTarget target.Traffi
 
 	return trafficPolicies, nil
 }
+
+// routesFromRules takes a set of traffic target rules and a namespace and returns a list of http routes (trafficpolicy.HTTPRoute)
+func (mc *MeshCatalog) routesFromRules(rules []target.TrafficTargetRule, namespace string) ([]trafficpolicy.HTTPRoute, error) {
+	routes := []trafficpolicy.HTTPRoute{}
+
+	specMatchRoute, err := mc.getHTTPPathsPerRoute() // returns map[traffic_spec_name]map[match_name]trafficpolicy.HTTPRoute
+	if err != nil {
+		return nil, err
+	}
+
+	if len(specMatchRoute) == 0 {
+		log.Debug().Msg("No elements in map[traffic_spec_name]map[match name]trafficpolicyHTTPRoute")
+		return routes, nil
+	}
+
+	for _, rule := range rules {
+		trafficSpecName := mc.getTrafficSpecName("HTTPRouteGroup", namespace, rule.Name)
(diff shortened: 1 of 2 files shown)


pkg/catalog/routes.go
```diff
@@ -405,3 +405,33 @@ func (mc *MeshCatalog) listTrafficTargetPermutations(trafficTarget target.Traffi
 
 	return trafficPolicies, nil
 }
+
+// routesFromRules takes a set of traffic target rules and a namespace and returns a list of http routes (trafficpolicy.HTTPRoute)
+func (mc *MeshCatalog) routesFromRules(rules []target.TrafficTargetRule, namespace string) ([]trafficpolicy.HTTPRoute, error) {
```
shashankram, line 410: Can we describe what namespace refers to here, and also name the `namespace` var accordingly, ex. `trafficTargetNamespace`.

pkg/catalog/routes_test.go
```diff
@@ -15,6 +16,56 @@ import (
...
+					Matches: []string{"hello"},
+				},
+			},
+			namespace:      tests.Namespace,
+			expectedRoutes: []trafficpolicy.HTTPRoute{},
+		},
+	}
+
+	for _, tc := range testCases {
+		routes, err := mc.routesFromRules(tc.rules, tc.namespace)
```
shashankram, line 62: It might be worth running each test case separately using `t.Run()` as done here https://github.com/openservicemesh/osm/blob/main/cmd/cli/metrics_test.go#L261

shashankram (approved):

pkg/catalog/routes_test.go
michelleN: Thanks!

shashankram (approved):

draychev (approved):

pkg/catalog/routes_test.go
```diff
@@ -15,6 +16,59 @@ import (
...
+			name: "http route group and match name exist",
+			rules: []target.TrafficTargetRule{
+				{
+					Kind:    "HTTPRouteGroup",
+					Name:    tests.RouteGroupName,
+					Matches: []string{tests.BuyBooksMatchName},
+				},
+			},
+			namespace: tests.Namespace,
+			expectedRoutes: []trafficpolicy.HTTPRoute{
```
ksubrmnn, line 39: Not worth dismissing the existing approvals, but in the next PR you can replace this with tests.BookstoreBuyHTTPRoute

ksubrmnn (approved):
One small comment you can address in the next PR

PR #2037, opened 2020-11-12
feat(pkg/catalog): Add GetServicesForServiceAccounts func 

michelleN:
- This PR contains a function called GetServicesForServiceAccounts that returns a list of services corresponding
to a list service accounts. Part of #2034. See this change as part of the bigger change in #2035. There
are no functional changes to the system in this PR.
- This PR also updates and existing test to use an existing var

### DIFF
--- a/pkg/catalog/routes.go
+++ b/pkg/catalog/routes.go
@@ -439,7 +439,7 @@ func (mc *MeshCatalog) routesFromRules(rules []target.TrafficTargetRule, traffic
 // GetServicesForServiceAccounts returns a list of services corresponding to a list service accounts
 //	TODO: Consider merging this function and mc.GetServicesForServiceAccount in future (#2038)
 func (mc *MeshCatalog) GetServicesForServiceAccounts(saList []service.K8sServiceAccount) []service.MeshService {
-	serviceList := []service.MeshService{}
+	serviceMap := map[service.MeshService]int{}
 
 	for _, sa := range saList {
 		services, err := mc.GetServicesForServiceAccount(sa)
@@ -447,8 +447,15 @@ func (mc *MeshCatalog) GetServicesForServiceAccounts(saList []service.K8sService
 			log.Error().Msgf("Error getting services linked to Service Account %s: %v", sa, err)
 			continue
 		} else {
-			serviceList = append(serviceList, services...)
+			for _, s := range services {
+				serviceMap[s] = 1
+			}
 		}
+	}
+
+	serviceList := []service.MeshService{}
+	for k := range serviceMap {
+		serviceList = append(serviceList, k)
 	}
 
 	return serviceList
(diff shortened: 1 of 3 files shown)


pkg/catalog/routes.go
```diff
@@ -435,3 +435,19 @@ func (mc *MeshCatalog) routesFromRules(rules []target.TrafficTargetRule, traffic
 
 	return routes, nil
 }
+
+// GetServicesForServiceAccounts returns a list of services corresponding to a list service accounts
+func (mc *MeshCatalog) GetServicesForServiceAccounts(saList []service.K8sServiceAccount) []service.MeshService {
```
ksubrmnn, line 440: Why this function instead of using this one? https://github.com/openservicemesh/osm/blob/59a2722b478433c790b0cc58d79c54f65280c937/pkg/catalog/service.go#L10

michelleN: Created this mainly so I could loop through the list and log the error instead of returning an error. The err being returned is currently being used by other functions in the codebase and I didn't want to change current behavior. Once the routes refactor is complete, we could merge the two functions and use one although I kinda like the way it reads now.

shashankram: It will help to track future refactoring like this via a comment referencing an issue so that it doesn't get left behind.

ksubrmnn: Okay that makes sense. I agree, if you add a comment, then this seems good to go

michelleN: all set.

shashankram (approved):

pkg/catalog/routes.go
```diff
@@ -435,3 +435,20 @@ func (mc *MeshCatalog) routesFromRules(rules []target.TrafficTargetRule, traffic
 
 	return routes, nil
 }
+
+// GetServicesForServiceAccounts returns a list of services corresponding to a list service accounts
+//	TODO: Consider merging this function and mc.GetServicesForServiceAccount in future (#2038)
+func (mc *MeshCatalog) GetServicesForServiceAccounts(saList []service.K8sServiceAccount) []service.MeshService {
+	serviceList := []service.MeshService{}
```
snehachhabria, line 442: Do you think we should have this as a map and then convert it to a slice to avoid duplicates ?

michelleN: yes good catch. will add logic to avoid duplicates.

pkg/catalog/routes.go
```diff
@@ -435,3 +435,28 @@ func (mc *MeshCatalog) routesFromRules(rules []target.TrafficTargetRule, traffic
 
 	return routes, nil
 }
+
+// GetServicesForServiceAccounts returns a list of services corresponding to a list service accounts
+//	TODO: Consider merging this function and mc.GetServicesForServiceAccount in future (#2038)
+func (mc *MeshCatalog) GetServicesForServiceAccounts(saList []service.K8sServiceAccount) []service.MeshService {
+	serviceMap := map[service.MeshService]int{}
```
snehachhabria, line 442: Should we use a set instead of a map ?

snehachhabria: An alternative could be 👍 
```suggestion
	serviceMap := map[service.MeshService]bool{}
```

This will need lesser memory as compared to int. My preference would still be a set

snehachhabria (approved):

ksubrmnn (approved):

PR #2046, opened 2020-11-12
osm-deployment: parametrize osm log level in values.yaml

eduser25:
Currently it's hardcoded in deployment.

This gives at least some short term flexibility till we add
configMap update, which has some ongoing discussion as
how to tackle.

- Install                [X]
- Control Plane          [X]

No

### DIFF
--- a/charts/osm/values.yaml
+++ b/charts/osm/values.yaml
@@ -43,7 +43,7 @@ OpenServiceMesh:
   meshName: osm
   useHTTPSIngress: false
   envoyLogLevel: error
-  osmLogLevel: trace
+  controllerLogLevel: trace
   enforceSingleMesh: false
 
   # Set deployJaeger to true to deploy a Jaeger cluster in the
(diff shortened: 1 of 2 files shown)


charts/osm/values.yaml
```diff
@@ -43,6 +43,7 @@ OpenServiceMesh:
   meshName: osm
   useHTTPSIngress: false
   envoyLogLevel: error
+  osmLogLevel: trace
```
shashankram, line 46: should we rename this to `controllerLogLevel`?

shashankram (approved):

charts/osm/values.yaml
ksubrmnn: Agreed, otherwise it may be confused with envoy log level

ksubrmnn (approved):

charts/osm/values.yaml
eduser25: ... I guess? we didn't call `envoyLogLevel` `sidecarProxyLogLevel`, but if you feel this is clearer, I'll change it gladly

eduser25: since we use it as `OpenServiceMesh.controllerLogLevel`, seems more than ok. changing it.

snehachhabria (approved):

shashankram (approved):

PR #2047, opened 2020-11-12
feat(*): add TrafficPolicy struct and helpers

michelleN:
- adds TrafficPolicy object to represent route configuration
between a source and a destination
- adds a buildTrafficPolicies function that takes a list of source services,
destination services, and routes and returns a list of traffic policies for each
source + destination combination
- adds a GetHostnamesForUpstreamService func to get a list of resolvable hostnames
for traffic policy

This PR includes no functional changes to the system. It is part of #2034. See #2035 for how it will integrate into the overall routes refactor work. As a follow up, `GetHostnamesForUpstreamService` should replace `GetResolvableHostnamesForUpstreamService` but the latter is left in the codebase for now to ensure nothing breaks during non functional changes and a clean transition to the new route programming logic (tracked via #2048) 

### DIFF
--- a/pkg/catalog/routes.go
+++ b/pkg/catalog/routes.go
@@ -486,6 +486,9 @@ func (mc *MeshCatalog) buildTrafficPolicies(sourceServices, destServices []servi
 				continue
 			}
 			routesClusters := []trafficpolicy.RouteWeightedClusters{}
+			// When TrafficSplit v1alpha3 is implemented (#705), weighted clusters information should be passed into this function as a parameter
+			//	but since we are not implementing TrafficSplit v1alpha3 and only dealing with TrafficTargets for the initial routes refactor(#2034),
+			//	we use the default weighted clusters configuration for the given destination service (destService)
 			weightedClusters := mapset.NewSet(getDefaultWeightedClusterForService(destService))
 
 			for _, route := range routes {
--- a/pkg/catalog/routes.go
+++ b/pkg/catalog/routes.go
@@ -460,3 +460,48 @@ func (mc *MeshCatalog) GetServicesForServiceAccounts(saList []service.K8sService
 
 	return serviceList
 }
+
+// GetHostnamesForUpstreamService returns the hostnames over which an upstream service is accessible from a downstream service
+// The hostname is the FQDN for the service, and can include ports as well.
+// Ex. bookstore.default, bookstore.default:80, bookstore.default.svc, bookstore.default.svc:80 etc.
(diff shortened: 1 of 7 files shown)


shashankram (approved):

pkg/catalog/routes.go
```diff
@@ -460,3 +460,48 @@ func (mc *MeshCatalog) GetServicesForServiceAccounts(saList []service.K8sService
 
 	return serviceList
 }
+
+// GetHostnamesForUpstreamService returns the hostnames over which an upstream service is accessible from a downstream service
+// The hostname is the FQDN for the service, and can include ports as well.
+// Ex. bookstore.default, bookstore.default:80, bookstore.default.svc, bookstore.default.svc:80 etc.
+// TODO: replace GetResolvableHostnamesForUpstreamService with this func once routes refactor is complete (#issue)
```
ksubrmnn, line 467: is there any issue number?

pkg/catalog/routes.go
```diff
@@ -460,3 +460,48 @@ func (mc *MeshCatalog) GetServicesForServiceAccounts(saList []service.K8sService
...
+// buildTrafficPolicies takes a list of source services, destination services and routes and returns a list of traffic policies for all
+//	combinations of the given source and destination services
+func (mc *MeshCatalog) buildTrafficPolicies(sourceServices, destServices []service.MeshService, routes []trafficpolicy.HTTPRoute) (policies []*trafficpolicy.TrafficPolicy) {
+	for _, sourceService := range sourceServices {
+		for _, destService := range destServices {
+			if sourceService == destService {
+				continue
+			}
+			routesClusters := []trafficpolicy.RouteWeightedClusters{}
+			weightedClusters := mapset.NewSet(getDefaultWeightedClusterForService(destService))
```
ksubrmnn, line 489: Maybe we need a comment here for why we are not getting the traffic split backends here

ksubrmnn (changes requested):

pkg/catalog/routes.go
michelleN: #2048

pkg/catalog/routes.go
michelleN: added

michelleN: added

shashankram (approved):

ksubrmnn (approved):

pkg/catalog/routes.go
ksubrmnn: Okay, I think that is missing from line 467 but it can be added later

PR #2055, opened 2020-11-13
fix(ci): fix go mod tidy check

nojnhuh:
**Description**:

CI was only checking the exit code of `go mod tidy`, but that is still 0 if it makes changes to the files. This change instead invokes the makefile target which will explicitly check if the files were modified by `go mod tidy` and fail if they were.

**Affected area**:

- New Functionality      [ ]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [ ]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests                  [ ]
- CI System              [X]
- Performance            [ ]
- Other                  [ ]

No

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -49,11 +49,7 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v1
       - name: go mod tidy
-        run: |
-          if ! go mod tidy ; then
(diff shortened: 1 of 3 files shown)


shashankram (approved):

draychev (approved):
Thank you @nojnhuh

Issue #2088, opened 2020-11-18
RBAC package and model to craft policies


shashankram:
**Please describe the Improvement and/or Feature Request**
https://github.com/openservicemesh/osm/issues/1964 implemented server side RBAC policies as a part of LDS. This needs to be extended out as a separate package to be able to build in XDS RBAC policy, preferably using a data model. This will also be used by RDS to build RBAC policies per route.

**Scope (please mark with X where applicable)**
- New Functionality          [X]
- Install                    [ ]
- SMI Traffic Access Policy  [ ]
- SMI Traffic Specs Policy   [ ]
- SMI Traffic Split Policy   [ ]
- Permissive Traffic Policy  [ ]
- Ingress                    [ ]
- Egress                     [ ]
- Envoy Control Plane        [X]
- CLI Tool                   [ ]
- Metrics                    [ ]
- Certificate Management     [ ]
- Sidecar Injection          [ ]
- Logging                    [ ]
- Debugging                  [ ]
- Tests                      [ ]
- CI System                  [ ]
- Project Release            [ ]

PR #2089, opened 2020-11-19
envoy/rbac: add generic RBAC policy builder

shashankram:
**Description**:
This change adds a new rbac pkg to build RBAC policies.
Currently this change doesn't affect existing functionality
and the goal is to have a generic rbac policy builder
that can be used when an RBAC policy needs to be built.

LDS builds RBAC policies to enforce RBAC on inbound
connections from downstreams. RDS will require to enable
RBAC on a per route basis in the future. This pkg will
serve as a modeler for RBAC policies.

Part of #2088

**Affected area**:

- New Functionality      [X]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [ ]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests                  [ ]
- CI System              [ ]
- Performance            [ ]
- Other                  [ ]

`No`

### DIFF
--- a/pkg/envoy/rbac/policy.go
+++ b/pkg/envoy/rbac/policy.go
@@ -1,0 +1,117 @@
+package rbac
+
+import (
+	xds_rbac "github.com/envoyproxy/go-control-plane/envoy/config/rbac/v3"
+	xds_matcher "github.com/envoyproxy/go-control-plane/envoy/type/matcher/v3"
+	"github.com/pkg/errors"
+)
+
+// Generate constructs an RBAC policy for the policy object on which this method is called
+func (p *Policy) Generate() (*xds_rbac.Policy, error) {
+	policy := &xds_rbac.Policy{}
+
+	// Construct the Principals ------------------------
+	var finalPrincipals []*xds_rbac.Principal
+
+	// Each RuleList follows OR semantics with other RuleList in the list of RuleList
+	for _, principalRuleList := range p.Principals {
+		// 'principalRuleList' corresponds to a single Principal in an RBAC policy.
+		// This Principal can be defined in terms of one of AND or OR rules.
+		// When AND/OR semantics are not required to define multiple rules corresponding
+		// to this principal, a single Rule in either the AndRules or OrRules will suffice.
+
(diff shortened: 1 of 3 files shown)


draychev (approved):
lgtm

pkg/envoy/rbac/policy.go
```diff
@@ -0,0 +1,117 @@
...
+	var finalPrincipals []*xds_rbac.Principal
+
+	// Each RuleList follows OR semantics with other RuleList in the list of RuleList
+	for _, principalRuleList := range p.Principals {
+		// 'principalRuleList' corresponds to a single Principal in an RBAC policy.
+		// This Principal can be defined in terms of one of AND or OR rules.
+		// When AND/OR semantics are not required to define multiple rules corresponding
+		// to this principal, a single Rule in either the AndRules or OrRules will suffice.
+
+		if len(principalRuleList.AndRules) != 0 && len(principalRuleList.OrRules) != 0 {
```
eduser25, line 23: Assuming we do NOT want to have And and Or in the same RuleList object in upcoming iterations, would it not be better to disallow it by design? (by typifying RuleList or polymorph it and front it with an interface?)

eduser25 (approved):
clean

snehachhabria (approved):

michelleN (approved):

pkg/envoy/rbac/policy.go
shashankram: Fronting it with an interface could be confusing for clients. Could you give an example of what you have in mind?

shashankram: One alternative I can think of is 
```
type RulesList struct {
    Rules []Rules
    semantic RuleSemantic // AND or OR
}
```
We can do this in the future when we are certain a combination of AND/OR rules is not going to be possible. I am not certain of that at this point given there are not many consumers of RBAC.

shashankram: Discussed with @eduser25 offline.

Issue #1934, opened 2020-10-27
Provided prometheus instance doesn't scale with default configs provided


eduser25:
Current prometheus, when enabling metrics for all namespaces in the mesh, with the given default configuration it comes with, can't scale more than 20~30 pods due to memory exhaustion (OOMKilled by cgroups/k8s):
```
        resources:
          limits:
            cpu: 500m
            memory: 512M
```
There's several aspects to tackle to make sure it better fits the given provisioning, so users do not face issues when ramping up with even a small number of pods.

- In the current default deployment, with default provisioned config, and with all the envoy metrics exposed (and 10s scrap interval), prometheus uses *avgd. 8.5~10M of resident memory per envoy*.

![image](https://user-images.githubusercontent.com/5503924/97360688-be603a00-185b-11eb-9f84-d3578829f597.png)
![image](https://user-images.githubusercontent.com/5503924/97360708-c5874800-185b-11eb-9dbd-fa4cfc36e3b3.png)

Actionable items:
- Revisit scrap interval, see how it affects (from self-testing, increasing it to 20s did not change much)
- Explore limiting the number of metrics scrapped by prometheus. We do not need all of, not for all envoys.
- Explore flushing to disk/storage backend, to avoid increased memory consumption.

**Affected area (please mark with X where applicable)**:
- Metrics                    [X]

**Expected behavior**:
Prometheus should not be OOM killed at 20~30 pod size.

**Steps to reproduce the bug (as precisely as possible)**:
Run and scale a setup

**How was OSM installed?**:
CLI

**Environment**:
- OSM version (use `osm version`): ToT
- Kubernetes version (use `kubectl version`): 1.18
- Size of cluster (number of worker nodes in the cluster): 20~30

eduser25:
topk 20 most expensive metrics (in terms of how many time series each one has) at 200~ pod
![image](https://user-images.githubusercontent.com/5503924/100031446-87257000-2daa-11eb-83e0-f7e1b477e6ef.png)

eduser25:
With current config, prometheus is killed due to memory exhaustion (in an enabled 4Gb instance) at around ~200 pods. 
- Assuming prometheus starts with roughly 120Mb of ram in use, that yields around *9.5Mb/envoy*, however
- There's a couple or three time-bucket series from API servers that scale pretty poorly too (apiserver_request_duration_seconds_bucket).

[...]

PR #2117, opened 2020-11-24
prometheus: drop time-bucket series, increase ram limit to 2G

eduser25:
(tentative)

We are using large amounts of labels to tag our metrics with addtional
dimensions of interest. This is good for visibility, however this explodes
in RAM consumption on Prometheus once scraped.

Most of the heavily unscaling metrics are the time-bucket series, which
already have large number of dimensions due to the buckets themselves.

- By filtering out time-bucket series (which might contain latency and
request total time buckets), we can cut the memory consumption around
66~70%, allowing larger scales to work under 2Gb, compared with same
scale crashing at 4Gb).
- Additionally, rose the RAM limit for prometheus to 2Gb.

Purely tentative, proper discussion on relabeling has to happen, this
is however required to even support slightly larger number of pods for
the time being.

More info:
#1934 

- Metrics                [X]
- Performance            [X]

No

### DIFF
--- a/charts/osm/templates/prometheus-configmap.yaml
+++ b/charts/osm/templates/prometheus-configmap.yaml
@@ -22,7 +22,11 @@ data:
           # TODO need to remove this when the CA and SAN match 
           insecure_skip_verify: true
         bearer_token_file: /var/run/secrets/kubernetes.io/serviceaccount/token
-        relabel_configs:
+        metric_relabel_configs:
+        - source_labels: [__name__]
+          regex: .*bucket.*
+          action: drop   
+        relabel_configs: 
         - source_labels: [__meta_kubernetes_namespace, __meta_kubernetes_service_name, __meta_kubernetes_endpoint_port_name]
           action: keep
           regex: default;kubernetes;https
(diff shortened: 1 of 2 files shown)


charts/osm/templates/prometheus-configmap.yaml
```diff
@@ -47,7 +51,11 @@ data:
       - job_name: 'kubernetes-pods'
         kubernetes_sd_configs:
         - role: pod
-        relabel_configs:
+        metric_relabel_configs:
+        - source_labels: [__name__]
+          regex: .*bucket.*
```
nojnhuh, line 56: We will need `osm_request_duration_ms_bucket` for SMI metrics eventually. Is there an easy way to ignore all but that one?

eduser25: I think the proper solution might be a ~whitelist~ allowed list for what we explicitly want. Didn't go the extra mile as I'm onto something else, but if we feel we don't want this at all, I will not push for this either.

nojnhuh: SMI metrics is far enough out that it shouldn't bock this change. Once we get to that point, it should be simple enough to get it to work.

nojnhuh (approved):

michelleN (approved):
Does this change have any implications on the grafana dashboards? would the screenshots in docs change at all?

eduser25:
> Does this change have any implications on the grafana dashboards? would the screenshots in docs change at all?

Yes it does. It kills the request and response latency stats as those are buckets. I wonder if we want this commit as is right now.
The problem is, the current labeling scales so incredibly poorly that I can't even use prometheus as it just runs out of memory even at 4Gb mem.

I can just cherrypick for my own testing, but I wonder what some other user might think if they happen to experience the same. Should we care for the time being? or should we open and mark this as a must-fix bug sometime soon? @michelleN

eduser25:
https://github.com/openservicemesh/osm/issues/2122

PR #2118, opened 2020-11-24
envoy/lds: refactor inbound filter chain building

shashankram:
**Description**:
Refactors code used to build the inbound HTTP filter chain
in preparation to support TCP filter chain.

Part of #1521

**Affected area**:

- New Functionality      [ ]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [X]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests                  [ ]
- CI System              [ ]
- Performance            [ ]
- Other                  [ ]

`No`

### DIFF
--- a/pkg/envoy/lds/inmesh.go
+++ b/pkg/envoy/lds/inmesh.go
@@ -9,7 +9,6 @@ import (
 	"github.com/pkg/errors"
 	"google.golang.org/protobuf/types/known/wrapperspb"
 
-	"github.com/openservicemesh/osm/pkg/configurator"
 	"github.com/openservicemesh/osm/pkg/envoy"
 	"github.com/openservicemesh/osm/pkg/envoy/route"
 	"github.com/openservicemesh/osm/pkg/service"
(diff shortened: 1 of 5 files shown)


eduser25 (approved):

pkg/envoy/lds/inmesh.go
```diff
@@ -27,36 +26,78 @@ var (
...
+	} else {
+		filterChains = append(filterChains, httpFilterChain)
+	}
+
+	// TODO: add TCP filter chain here
+
+	return filterChains
+}
+
+func (lb *listenerBuilder) getInboundHTTPFilters(proxyService service.MeshService) ([]*xds_listener.Filter, error) {
```
michelleN, line 44: Do you see this function ever taking a proxy identity instead of a service?

pkg/envoy/lds/response.go
```diff
@@ -63,21 +62,8 @@ func NewResponse(meshCatalog catalog.MeshCataloger, proxy *envoy.Proxy, _ *xds_d
...
-				log.Error().Err(err).Msgf("Error applying RBAC filter for service %s", proxyServiceName)
-				return nil, err
-			}
-			// RBAC filter should be the very first filter in the filter chain
-			meshFilterChain.Filters = append([]*xds_listener.Filter{rbacFilter}, meshFilterChain.Filters...)
-		}
-		inboundListener.FilterChains = append(inboundListener.FilterChains, meshFilterChain)
-	}
+	inboundMeshFilterChains := lb.getInboundMeshFilterChains(proxyServiceName)
+	inboundListener.FilterChains = append(inboundListener.FilterChains, inboundMeshFilterChains...)
```
michelleN, line 66: Is there a reason for doing an append here instead of setting inboundListener.FilterChains to inboundMeshFilterChains? I just worry about duplicates.

pkg/envoy/lds/inmesh.go
shashankram: Not at this moment, because the inbound filters make use of attributes such as SNI that are configured per service.

pkg/envoy/lds/response.go
shashankram: this just avoids relying relying on the order of building different filer chains (prometheus, ingress, in-mesh etc.).
I don't think there will be any duplicates here, and we have to use append regardless elsewhere since we continue to add filter chains to the same listener.

michelleN (approved):

PR #2120, opened 2020-11-24
fix(chart): allow empty OpenServiceMesh.image.pullPolicy

nojnhuh:
**Description**:
This change updates values.schema.json to allow defining an empty
OpenServiceMesh.image.pullPolicy, which behaves as described here:
https://kubernetes.io/docs/concepts/containers/images/#updating-images

**Affected area**:

- New Functionality      [ ]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [ ]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests                  [ ]
- CI System              [ ]
- Performance            [ ]
- Other                  [X]

No

### DIFF
--- a/charts/osm/values.schema.json
+++ b/charts/osm/values.schema.json
@@ -78,7 +78,7 @@
                             "type": "string",
                             "title": "The pullPolicy schema",
                             "description": "The image pull policy.",
-                            "pattern": "^(Always|Never|IfNotPresent)$",
+                            "pattern": "^(Always|Never|IfNotPresent)?$",
                             "examples": [
                                 "IfNotPresent"
                             ]


shashankram (approved):
What's the reason to make this change?

nojnhuh:
@shashankram I want to be able to set it to an empty value so when I'm playing with `helm upgrade`, I can set it to empty so I can upgrade from a tagged image to a `latest` image without modifying the pull policy and get the behavior I want. It's also valid in general so we should allow it.

shashankram:
> @shashankram I want to be able to set it to an empty value so when I'm playing with `helm upgrade`, I can set it to empty so I can upgrade from a tagged image to a `latest` image without modifying the pull policy and get the behavior I want. It's also valid in general so we should allow it.

Thanks, it would be worth mentioning in the commit message that this should be allowed in general.

eduser25 (approved):

PR #2123, opened 2020-11-25
pkg/kubernetes: reduce warn log to trace

eduser25:
As an increasing number of packages' channels are becoming unused
in favor of pubsub channels, these logs are quickly becoming a hog as warn.

We will however keep them as trace if we ever need them.

```
kubectl logs osm-controller-7545bfb54c-8x7vl | grep "Channel for provider Kubernetes is full" | wc -l
5069
```
- Other                  [X]

No


nojnhuh (approved):

shashankram:
> ```
> kubectl logs osm-controller-7545bfb54c-8x7vl | grep "Channel for provider Kubernetes is full" | wc -l
> 5069
> ```

Let's be direct in the commit message instead of hinting the motivation :-). Someone not familiar with `wc` isn't going to know what this implies.

shashankram (approved):

eduser25:
> Let's be direct in the commit message instead of hinting the motivation :-). Someone not familiar with `wc` isn't going to know what this implies.

adding context

PR #2129, opened 2020-12-01
cli: move portForwarder to a more generic location

eduser25:
This commit just moves PortForwarder to a more generic/utils
location in the repo.

Testing will require immediate use of port-forwarding; moving
it to a less bloated/more generic package hopes to reduced number
of imports and coupling between users of this feature.

**Affected area**:

- Other                  [X]

No

### DIFF
--- a/cmd/cli/dashboard.go
+++ b/cmd/cli/dashboard.go
@@ -7,7 +7,6 @@ import (
 	"os"
 	"os/signal"
 
-	"github.com/openservicemesh/osm/pkg/utils"
 	"github.com/pkg/browser"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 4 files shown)


snehachhabria (dismissed):

nojnhuh (approved):

draychev (dismissed):
This looks great. Thanks for refactoring this.
I wonder if at some future point it would make sense to move it to Kubernetes-specific utils package, since this is not Service Mesh specific.

shashankram:
@eduser25 thanks for the change. It seems to me it would be more appropriate to move this under `pkg/kubernetes`, since the portforwarding implementation is specific to kubernetes.

eduser25:
> @eduser25 thanks for the change. It seems to me it would be more appropriate to move this under `pkg/kubernetes`, since the portforwarding implementation is specific to kubernetes.

agreed, I'll do. I was not 100% convinced when moving that was the right location.
Should it be in it's own package/folder? seems to me small yet container enough that it wouldn't hurt maybe

eduser25:
tests didn't seem to trigger even with a reasonable change... ?

draychev (approved):

shashankram (approved):

PR #2133, opened 2020-12-01
kubernetes: add api to retrieve pods backing a service

shashankram:
**Description**:
Adds an api to the kubernetes.Controller interface to retreive
pods backing a service. This will be used by a subsequent change
to fetch the kubernetes ports for a service that used named
target ports.

**Affected area**:

- New Functionality      [ ]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [ ]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests                  [ ]
- CI System              [ ]
- Performance            [ ]
- Other                  [X]

`No`

### DIFF
--- a/pkg/kubernetes/client_test.go
+++ b/pkg/kubernetes/client_test.go
@@ -368,6 +368,7 @@ var _ = Describe("Test Namespace KubeController Methods", func() {
 		})
 
 		It("should correctly return the pods associated with a service", func() {
+			// Test creates 2 pods whose labels match the service's selectors
 			serviceChannel := events.GetPubSubInstance().Subscribe(announcements.ServiceAdded,
 				announcements.ServiceDeleted,
 				announcements.ServiceUpdated)
@@ -448,6 +449,7 @@ var _ = Describe("Test Namespace KubeController Methods", func() {
 		})
 
 		It("should correctly return the pods associated with a service", func() {
+			// Test creates 2 pods, only one whose labels match the service's selectors
 			serviceChannel := events.GetPubSubInstance().Subscribe(announcements.ServiceAdded,
 				announcements.ServiceDeleted,
 				announcements.ServiceUpdated)
--- a/pkg/kubernetes/client_test.go
+++ b/pkg/kubernetes/client_test.go
@@ -21,7 +21,8 @@ import (
 )
 
 var (
-	testMeshName = "mesh"
+	testMeshName      = "mesh"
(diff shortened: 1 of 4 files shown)


eduser25 (approved):

pkg/kubernetes/client_test.go
```diff
@@ -352,7 +352,179 @@ var _ = Describe("Test Namespace KubeController Methods", func() {
...
+		testMeshName := "foo"
+
+		BeforeEach(func() {
+			kubeClient = testclient.NewSimpleClientset()
+			kubeController, err = NewKubernetesController(kubeClient, testMeshName, make(chan struct{}))
+			Expect(err).ToNot(HaveOccurred())
+			Expect(kubeController).ToNot(BeNil())
+		})
+
+		It("should correctly return the pods associated with a service", func() {
```
ksubrmnn, line 370: What's the difference between this test and the test below? Can you include that in the text here? Or in a comment

ksubrmnn (changes requested):

pkg/kubernetes/client_test.go
shashankram: I left an additional comment.

eduser25 (approved):

ksubrmnn (approved):

draychev (changes requested):
Would you be able to add sample usage to this PR?

I suspect that you would not need the Pods.  You most likely need Endpoints -- would simplify this change quite a bit.

draychev (approved):
@shashankram if you have stacked PRs - go full steam ahead.
If Endpoints end up being a better fit for this - we can always come back and use these instead.

snehachhabria (approved):

michelleN (commented):
Does it make sense for your subsequent change to fetch endpoints rather than Pods backing a Service if you only need IP and port information?

shashankram:
> Does it make sense for your subsequent change to fetch endpoints rather than Pods backing a Service if you only need IP and port information?

I actually wanted this more than just fetching the endpoints (do some additional business logic and validation on the pod spec). This isn't required right away, so I can hold this and resurrect it when I need to leverage it.

shashankram:
needs rebase, will respin later when needed.

PR #2135, opened 2020-12-02
endpoint/providers: add api to retrieve port:protocol mapping for ser…

shashankram:
**Description**:
Adds an api to retrieve the port to protocol mapping for a service.
If an `AppProtocol` is unspecfied, it will default to `http` in lieu
of current OSM protocol support.
This will be used later to correctly filter inbound
traffic arriving on a port, as described in commit ff54bd77.

This change also refactors the tests which were relying on objects
created by previous tests causing test pollution, and some additional
cleanup.

Part of #1521

**Affected area**:

- New Functionality      [ ]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [X]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests                  [ ]
- CI System              [ ]
- Performance            [ ]
- Other                  [ ]

`No`

### DIFF
--- a/pkg/endpoint/providers/kube/client_test.go
+++ b/pkg/endpoint/providers/kube/client_test.go
@@ -27,165 +27,218 @@ var _ = Describe("Test Kube Client Provider", func() {
 		mockCtrl           *gomock.Controller
 		mockKubeController *k8s.MockController
 		mockConfigurator   *configurator.MockConfigurator
+
+		fakeClientSet *fake.Clientset
+		provider      endpoint.Provider
+		err           error
 	)
+
 	mockCtrl = gomock.NewController(GinkgoT())
 	mockKubeController = k8s.NewMockController(mockCtrl)
 	mockConfigurator = configurator.NewMockConfigurator(mockCtrl)
 
-	fakeClientSet := fake.NewSimpleClientset()
-	stopChann := make(chan struct{})
+	stopChan := make(chan struct{})
 	providerID := "provider"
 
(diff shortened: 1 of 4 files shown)


pkg/endpoint/providers/kube/client_test.go
```diff
@@ -27,166 +27,219 @@ var _ = Describe("Test Kube Client Provider", func() {
...
-			}
+			},
+		}
 
-			_, err = fakeClientSet.CoreV1().Endpoints(tests.BookbuyerService.Namespace).
-				Create(context.TODO(), endp, metav1.CreateOptions{})
-			Expect(err).To(BeNil())
+		_, err = fakeClientSet.CoreV1().Endpoints(tests.BookbuyerService.Namespace).Create(context.TODO(), endp, metav1.CreateOptions{})
+		Expect(err).To(BeNil())
+		<-provider.GetAnnouncementsChannel()
```
eduser25, line 89: droppable announcement. Might want to move it to pubsub, but let's deal with it separately.

eduser25 (commented):
lgtm, just waiting for that comment

eduser25 (approved):

draychev (approved):
👍

pkg/endpoint/providers/kube/client_test.go
shashankram: @eduser25 would you mind deprecating these across the code? I haven't been keeping up much with the channels that need to be deprecated.

PR #2136, opened 2020-12-02
tests/framework: Add GetTestFile

eduser25:
We might be interested in near future to save test results, data or logs on disk.
This commit introduces a helper for tests to prefix filenames and create test folder
upon use.

- Added a random uin64 identifier for a test, which will be used
to identify a certain test run. Note these are random, but not unique.
- GetTestFile is fed with a filename and returns a path to a file
inside test folder. If a folder for current test does not exist, it is
created in the process.

- Tests                  [X]

No


snehachhabria (approved):

tests/framework/common.go
```diff
@@ -215,8 +242,14 @@ func (td *OsmTestData) AreRegistryCredsPresent() bool {
 // Called by Gingkgo BeforeEach
 func (td *OsmTestData) InitTestData(t GinkgoTInterface) error {
 	td.T = t
+	r, err := rand.Int(rand.Reader, big.NewInt(math.MaxUint32))
```
michelleN, line 245: How come we're not using `time` to seed the random number generator here? https://stackoverflow.com/a/12321192

draychev (commented):
What's the use case for this?

shashankram:
> What's the use case for this?

Although the PR description has a one liner `Precursor to be able to save test data on disk.`, it would be nice to add more a bit more detail on what exactly this means.

tests/framework/common.go
eduser25: `rand.crypto` is supposed to be the safest, the reader from crypto is fed entropy from various sources I think.
In fact I was _forced_ to change it to `rand.crypto` as using `math.rand` actually triggers the linter....

eduser25:
> What's the use case for this?

> Although the PR description has a one liner Precursor to be able to save test data on disk., it would be nice to add more a bit more detail on what exactly this means.

Basically tests saving some results, logs or test data on disk. Honestly thought it'd be more obvious.

tests/framework/common.go
shashankram: For non crypto rand generations, using `math.Rand` is perfectly fine, you just need to silence the linter false alarms with `// #nosec` annotation

michelleN: @eduser25 - what is the linter error that you're getting?

shashankram (approved):

tests/framework/common.go
eduser25: @shashankram, I see. the linter doesn't actually tell you it is fine for use for non-crypto. 
@michelleN :
In some test file...
```
import (
	"fmt"
	"math/rand"

...

a := rand.Float32()
fmt.Printf("%i", a)
...
```

```
eserra@edu-linux ~/src/osm/tests (metrics) $ golangci-lint run --out-format=github-actions --timeout 5m
::error file=framework/common_profile.go,line=120,col=7::G404: Use of weak random number generator (math/rand instead of crypto/rand) (gosec)
```

shashankram: > @shashankram, I see. the linter doesn't actually tell you it is fine for use for non-crypto.
That's because the linter can't know whether you are using the api for crypto purpose. Doesn't mean you need to use crypto api for non crypto operations, you just need to suppress the false alarm, no other alternative with `gosec` linter enabled.

PR #2138, opened 2020-12-03
tests/framework: additions to framework to query from Prometheus/Grafana

eduser25:
This commit adds few high level helpers to allow querying for various resources
on both Prometheus and Grafana. Intention is to use them in our test infrastructure
to gather a holistic look on data and metrics on processes, namely control plane ones
(but not limited to).

The current set of helpers added are sufficient query and track
CPU and Memory trends, as well as graphing the overall control plane namespace
CPU and memory consumption through Grafana remote rendering, all at any point
during a test.

For graphing, Grafana remote rendering and specific dashboards which already
target the data to be graphed are needed.

This is part of an effort to better instrument our tests and potentially unlock minimal scale
testing using the very same tools we've written our tests with. 

Relevant APIs:
```
common_metrics.go:
  struct Prometheus {}
  (*Prometheus) VectorQuery(string, time) (float, err)                               // raw query call, since it's none-range, expects only one sample in vector form
  (*Prometheus) GetNumEnvoysInMesh() (float, err)                                    // returns number of current envoy in mesh as seen by Prom, ie. '20'
  (*Prometheus) GetMemRSSforContainer(ns, pod, container) (float, err)               // returns RSS footprint of container in bytes, ie. '52562'
  (*Prometheus) GetCPULoadAvgforContainer(ns, pod, container, avgTime) (float, err)  // returns CPU load avg for container given a timerange, ie. '0.62'
  (*Prometheus) GetCPULoadAvgforContainer(ns, pod, container) (float, float, float, err) // returns 1m, 5m and 15m CPU Load averages for a container

  struct Grafana {}
  (*Grafana) PanelPNGSnapshot(dashboard, panelId, timeMinutes, saveFilepath)        // Triggers a dashboard/panel render and stores it in 'saveFilepath'

common_apps.go:
  GetPrometheusPodHandle(ns, pod, port) (*Prometheus, err)  // port-forwards and gets handle generically for any Prometheus pod
  GetGrafanaPodHandle(ns, pod, port) (*Grafana, err)        // port-forwards and gets handle generically for any Grafana pod
  GetOSMPrometheusaHandle() (*Prometheus, err)              // GetPrometheusPodHandle wrapper for OSM-Prometheus
  GetOSMGrafanaHandle() (*Grafana, err)                     // GetGrafanaPodHandle  wrapper for OSM-Grafana
```

- Metrics                [X]
- Tests                  [X]
- Performance            [X]

No


tests/framework/common_apps.go
```diff
@@ -277,3 +323,88 @@ func (td *OsmTestData) SimpleDeploymentApp(def SimpleDeploymentAppDef) (corev1.S
...
+			"app": OsmPrometheusAppLabel,
+		},
+	})
+	if err != nil || len(prometheusPod) == 0 {
+		return nil, nil, errors.Errorf("Error getting Prometheus pods: %v (prom pods len: %d)", err, len(prometheusPod))
+	}
+	pHandle, err := Td.GetPrometheusPodHandle(prometheusPod[0].Namespace, prometheusPod[0].Name, DefaultOsmPrometheusPort)
+	if err != nil {
+		return nil, nil, err
+	}
```
nojnhuh, line 395: Would it make sense to include this chunk in `GetPrometheusPodHandle` so that function doesn't need any arguments (and likewise for Grafana)? I could see using Prometheus in a test without Grafana so that would eliminate the need to duplicate this chunk in the test.

tests/framework/common_apps.go
```diff
@@ -72,6 +102,22 @@ func (td *OsmTestData) GetMutatingWebhook(mwhcName string) (*v1beta1.MutatingWeb
 	return mwhc, nil
 }
 
+// GetPodsForAppLabel returns the Pods matching a specific `appLabel`
+func (td *OsmTestData) GetPodsForAppLabel(ns string, labelSel metav1.LabelSelector) ([]corev1.Pod, error) {
```
nojnhuh, line 106: Since this is generic for any label selector, not just an `app` label:
```suggestion
func (td *OsmTestData) GetPodsForLabels(ns string, labelSel metav1.LabelSelector) ([]corev1.Pod, error) {
```

nojnhuh (dismissed):
Left a couple nits, but LGTM. Will definitely use this for custom metrics tests.

tests/framework/common_apps.go
eduser25: good point.
One of the things I intended was to keep the semantics of OSM-specific things (fetching the pod with our label) in a separate function from the one that generically allows you to forward any prometheus instance.

Would at least split it into two different functions so they are independant, would that suffice?

nojnhuh: Yeah that makes sense

tests/framework/common_apps.go
eduser25: will fix.

nojnhuh (approved):

snehachhabria (approved):

PR #2140, opened 2020-12-03
envoy/lds: create inbound filter chains per service port

shashankram:
**Description**:
This change makes the inbound filter chains for in-mesh
and ingress a per port construct, so that protocol
specific filters and rules can be enforced per port.
Since protocol specific filters (ex. TCP proxy vs HTTP
Connection Manager) cannot be filters on the same filter
chain, a port specific filter chain is created to be
able to handle per port:protocol filtering.

Additionally, this change adds new tests for the ingress
filter chain and updates the existing tests for mesh filter
chains. The file `lds/response_test.go` was ingress specific
and has been rewritten to `lds/ingress_test.go`.

This is required for TCP filtering and routing: #1521

**Affected area**:

- New Functionality      [ ]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [X]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [X]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests                  [ ]
- CI System              [ ]
- Performance            [ ]
- Other                  [ ]

`No`


draychev (approved):

pkg/envoy/lds/ingress.go
```diff
@@ -57,27 +65,43 @@ func newIngressFilterChain(cfg configurator.Configurator, svc service.MeshServic
...
+				ingressFilterChainWithSNI := newIngressHTTPFilterChain(lb.cfg, svc, port)
+				ingressFilterChainWithSNI.Name = fmt.Sprintf("%s:%d", inboundIngressHTTPSFilterChain, port)
+				ingressFilterChainWithSNI.FilterChainMatch.ServerNames = []string{svc.ServerName()}
+				ingressFilterChains = append(ingressFilterChains, ingressFilterChainWithSNI)
+			}
+
+			// Filter chain without SNI matching enabled for HTTP clients and HTTPS clients that don't set the SNI
+			ingressFilterChainWithoutSNI := newIngressHTTPFilterChain(lb.cfg, svc, port)
+			ingressFilterChainWithoutSNI.Name = fmt.Sprintf("%s:%d", inboundIngressNonSNIFilterChain, port)
+			ingressFilterChains = append(ingressFilterChains, ingressFilterChainWithoutSNI)
```
eduser25, line 93: Wondering if you could benefit of a deepcopy (assuming reflect pkg will clone and repoint the cloned structs) of the result instead of calling `newIngressHTTPFilterChain` two times since I don't see difference in the invocations, hence I'd expect the results to be the same, avoiding k8s cache accesses if there were any.

Not a must-change though.

pkg/envoy/lds/ingress.go
```diff
@@ -57,27 +65,43 @@ func newIngressFilterChain(cfg configurator.Configurator, svc service.MeshServic
 	}
 }
 
-func getIngressFilterChains(svc service.MeshService, cfg configurator.Configurator) []*xds_listener.FilterChain {
+func (lb *listenerBuilder) getIngressFilterChains(svc service.MeshService) []*xds_listener.FilterChain {
```
eduser25, line 68: nit, unrelated to the commit, should it be `ForService`?

pkg/envoy/lds/ingress.go
```diff
@@ -57,27 +65,43 @@ func newIngressFilterChain(cfg configurator.Configurator, svc service.MeshServic
...
+		}
+	}
 
 	return ingressFilterChains
 }
 
-func getIngressTransportSocket(cfg configurator.Configurator, marshalledDownstreamTLSContext *any.Any) *xds_core.TransportSocket {
-	if cfg.UseHTTPSIngress() {
+func getIngressTransportSocket(forHTTPS bool, marshalledDownstreamTLSContext *any.Any) *xds_core.TransportSocket {
+	if forHTTPS {
```
eduser25, line 104: IMO, I'd put the condition outside. having a parametre to call a function and have it do nothing, better not call it from the start.

it throws dirt on the creation of the struct though, up to you.

eduser25 (approved):
lgtm, just minors

pkg/envoy/lds/ingress.go
shashankram: Good point, I haven't changed the existing behavior in this change. Eventually I foresee a use case to just parameterize the difference in the 2 chains and let the function build it as expected.

pkg/envoy/lds/ingress.go
shashankram: Sure, fair enough.

pkg/envoy/lds/ingress.go
eduser25: makes sense, I sort of imagined we would want to keep it for such

pkg/envoy/lds/ingress.go
shashankram: I'll leave this as is for now instead of refactoring code outside the scope of this PR. The existing code has generic functions that branch out based on `cfg.UseHTTPSIngress()`, so to do this differently the existing code (not changed in this PR) needs refactoring.

Issue #2141, opened 2020-12-03
Make osm component namespaces separate from Helm release namespace


ksubrmnn:
**Please describe the Improvement and/or Feature Request**

**Scope (please mark with X where applicable)**
- New Functionality          [ ]
- Install                    [X]
- SMI Traffic Access Policy  [ ]
- SMI Traffic Specs Policy   [ ]
- SMI Traffic Split Policy   [ ]
- Permissive Traffic Policy  [ ]
- Ingress                    [ ]
- Egress                     [ ]
- Envoy Control Plane        [ ]
- CLI Tool                   [ ]
- Metrics                    [ ]
- Certificate Management     [ ]
- Sidecar Injection          [ ]
- Logging                    [ ]
- Debugging                  [ ]
- Tests                      [ ]
- CI System                  [ ]
- Project Release            [ ]

**Possible use cases**

By making the osm-controller namespace separate from the helm namespace, users can hardcode the namespace of osm components to make deployments more predictable.

nojnhuh:
I'm not sure I understand the use case here. If a user knows the namespace of the helm release, then they also know the namespace of all the namespaced resources. How does it make deployments more predictable? To me it looks like this would only make it possible to store the release in a different namespace than the other resources. Is there a use case for that?

ksubrmnn:
@nojnhuh users can use the osmNamespace field in the values file to optionally hardcode the namespace of the osm-controller while still using the helm cli to specify the helm release namespace.

PR #2142, opened 2020-12-04
feat(osm/install): Decouple releaseNamespace and osmNamespace

ksubrmnn:
**Description**:
This PR allows the release-namespace for the osm installation to be different than the osm namespace.

By creating an osmNamespace field in the helm Values file, users can hardcode the namespace where the osm controller will be deployed. This way any OSM operators can allow users to specify where their helm release will live while still installing osm in a predictable namespace. 

**Affected area**:

- New Functionality      [ ]
- Documentation          [ ]
- Install                [X]
- Control Plane          [ ]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests                  [ ]
- CI System              [ ]
- Performance            [ ]
- Other                  [ ]

No


charts/osm/templates/_helpers.tpl
```diff
@@ -1,63 +1,4 @@
-{{/* vim: set filetype=mustache: */}}
-{{/*
-Expand the name of the chart.
```
ksubrmnn, line 3: I didn't see these used anymore, please let me know if I'm mistaken!

charts/osm/values.yaml
```diff
@@ -55,6 +55,7 @@ OpenServiceMesh:
   controllerLogLevel: trace
   enforceSingleMesh: false
   webhookConfigNamePrefix: osm-webhook
+  osmNamespace: ""
```
ritazh, line 58: Can we add a comment here and in the chart readme that if `osmNamespace` is not provided, then the release namespace is used as the namespace for where the osm controller will be installed?

SanyaKochhar (approved):

nshankar13 (approved):

shashankram:
@ksubrmnn are we good with https://github.com/openservicemesh/osm/issues/2141#issuecomment-740742818?

charts/osm/values.schema.json
```diff
@@ -397,6 +397,15 @@
                     "examples": [
                         true
                     ]
+                },
+                "osmNamespace": {
+                    "$id": "#/properties/OpenServiceMesh/properties/osmNamespace",
+                    "type": "string",
+                    "title": "The namespace of the osm controller",
+                    "description": "Indicates the namespace in which the osm controller will be installed.",
```
michelleN, line 405: nit:
title: "The namespace of the osm control plane components"
description: "Indicates the namespace in which the osm control plane components will be installed

ksubrmnn: I'll update this nit separately so I don't dismiss stale reviews.

ritazh (approved):

PR #2154, opened 2020-12-07
envoy/lds: add support for inbound TCP proxying

shashankram:
**Description**:
This change adds the plumbing necessary to be able to
proxy indbound TCP traffic for a service. Similar to
the inbound HTTP filter chain, a TCP filter chain
is adder per TCP port exposed on the service. This
is the first change in the series to support TCP traffic.
Subsequent changes will enable enforcing TCP port based
rules as a part of the RBAC policies.

Note that this change will not allow TCP clients to access
TCP based services yet because outbound TCP proxying with
TLS still needs to be implemented. This change does not
affect HTTP traffic.

Part of #1521

**Affected area**:

- New Functionality      [X]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [X]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [X]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests                  [ ]
- CI System              [ ]
- Performance            [ ]
- Other                  [ ]

`No`

### DIFF
--- a/pkg/envoy/lds/inmesh.go
+++ b/pkg/envoy/lds/inmesh.go
@@ -5,6 +5,7 @@ import (
 
 	xds_core "github.com/envoyproxy/go-control-plane/envoy/config/core/v3"
 	xds_listener "github.com/envoyproxy/go-control-plane/envoy/config/listener/v3"
+	xds_tcp_proxy "github.com/envoyproxy/go-control-plane/envoy/extensions/filters/network/tcp_proxy/v3"
 	"github.com/envoyproxy/go-control-plane/pkg/wellknown"
 	"github.com/golang/protobuf/ptypes"
 	"github.com/golang/protobuf/ptypes/any"
@@ -17,8 +18,10 @@ import (
 )
 
 const (
-	inboundMeshFilterChainName = "inbound-mesh-filter-chain"
-	httpAppProtocol            = "http"
+	inboundMeshHTTPFilterChainName = "inbound-mesh-http-filter-chain"
+	inboundMeshTCPFilterChain      = "inbound-mesh-tcp-filter-chain"
+	httpAppProtocol                = "http"
+	tcpAppProtocol                 = "tcp"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


pkg/envoy/lds/inmesh.go
```diff
@@ -50,6 +53,14 @@ func (lb *listenerBuilder) getInboundMeshFilterChains(proxyService service.MeshS
 			}
 			filterChains = append(filterChains, filterChainForPort)
 
+		case tcpAppProtocol:
+			filterChainForPort, err := lb.getInboundMeshTCPFilterChain(proxyService, port)
+			if err != nil {
+				log.Error().Err(err).Msgf("Error building inbound TCP filter chain for proxy:port %s:%d", proxyService, port)
+				continue // continue building filter chains for other ports on the service
```
eduser25, line 60: just because I'd like to hear what you had in mind, I imagine this is to at least not disrupt traffic for other applications if for some reason one inbound chain failed to generate?

shashankram: Yes, precisely.

pkg/envoy/lds/inmesh.go
```diff
@@ -140,6 +151,84 @@ func (lb *listenerBuilder) getInboundMeshHTTPFilterChain(proxyService service.Me
...
+			log.Error().Err(err).Msgf("Error applying RBAC filter for proxy service %s", proxyService)
+			return nil, err
+		}
+		// RBAC filter should be the very first filter in the filter chain
+		filters = append(filters, rbacFilter)
+	}
+
+	// Apply the TCP Proxy Filter
+	tcpProxy := &xds_tcp_proxy.TcpProxy{
+		StatPrefix:       "inbound-mesh-tcp-proxy",
```
eduser25, line 215: I think you will want to use `proxyService` in some form of prefix, to help identify the statistics for this filter later on

eduser25 (commented):
really nice, I only have one comment over stat name, otherwise lgtm

pkg/envoy/lds/inmesh.go
eduser25: discussed offline, we'll explore the effect of this going in the future.

eduser25 (approved):

pkg/envoy/lds/inmesh.go
shashankram: Discussed with @eduser25 offline and we have decided to look into the usage of stats prefixes in general: https://github.com/openservicemesh/osm/issues/2156

snehachhabria (approved):

michelleN (approved):

PR #2173, opened 2020-12-09
ref(pkg/catalog): update ListAllowedOutboundServicesForIdentity

michelleN:
**Description**:

This PR updates `ListAllowedOutboundServicesForIdentity` to handle permissive mode. When allowed outbound services are requested for a workload, all services running in the mesh will be returned.

**Affected area**:

- New Functionality      [ ]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [ ]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests                  [ ]
- CI System              [ ]
- Performance            [ ]
- Other                  [ ]

### DIFF
--- a/pkg/catalog/helpers_test.go
+++ b/pkg/catalog/helpers_test.go
@@ -24,7 +24,11 @@ import (
 	"github.com/openservicemesh/osm/pkg/tests"
 )
 
-func newFakeMeshCatalogForRoutes(t *testing.T) *MeshCatalog {
+type testParams struct {
+	permissiveMode bool
+}
+
+func newFakeMeshCatalogForRoutes(t *testing.T, testParams testParams) *MeshCatalog {
 	mockCtrl := gomock.NewController(t)
 	kubeClient := testclient.NewSimpleClientset()
 
@@ -124,6 +128,8 @@ func newFakeMeshCatalogForRoutes(t *testing.T) *MeshCatalog {
 	mockKubeController.EXPECT().IsMonitoredNamespace(tests.BookstoreV2Service.Namespace).Return(true).AnyTimes()
 	mockKubeController.EXPECT().IsMonitoredNamespace(tests.BookbuyerService.Namespace).Return(true).AnyTimes()
 	mockKubeController.EXPECT().ListMonitoredNamespaces().Return(listExpectedNs, nil).AnyTimes()
+
+	mockConfigurator.EXPECT().IsPermissiveTrafficPolicyMode().Return(testParams.permissiveMode).AnyTimes()
 
 	mockMeshSpec.EXPECT().GetAnnouncementsChannel().Return(announcementsChan).AnyTimes()
 	mockMeshSpec.EXPECT().ListTrafficTargets().Return([]*target.TrafficTarget{&tests.TrafficTarget}).AnyTimes()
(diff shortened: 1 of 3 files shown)


snehachhabria (approved):

pkg/catalog/routes.go
```diff
@@ -104,6 +104,15 @@ func (mc *MeshCatalog) ListAllowedOutboundServices(sourceService service.MeshSer
 
 // ListAllowedOutboundServicesForIdentity list the services the given service account is allowed to initiate outbound connections to
 func (mc *MeshCatalog) ListAllowedOutboundServicesForIdentity(identity service.K8sServiceAccount) []service.MeshService {
+	allowedServices := []service.MeshService{}
+
+	if mc.configurator.IsPermissiveTrafficPolicyMode() {
+		for _, svc := range mc.kubeController.ListServices() {
+			allowedServices = append(allowedServices, utils.K8sSvcToMeshSvc(svc))
+		}
```
snehachhabria, line 112: nit: I think we can have a method `ListMeshServices()` that handles the logic in these lines and returns `[]service.MeshService`. I see more use case in the future where this method might be useful

snehachhabria (approved):

pkg/catalog/helpers_test.go
```diff
@@ -24,7 +24,11 @@ import (
 	"github.com/openservicemesh/osm/pkg/tests"
 )
 
-func newFakeMeshCatalogForRoutes(t *testing.T) *MeshCatalog {
+type testParams struct {
+	permissiveMode bool
+}
+
+func newFakeMeshCatalogForRoutes(t *testing.T, testParams testParams) *MeshCatalog {
```
shashankram, line 31: In the future, lets remove these additional fakes and rely on building catalog using mocks of `smi.MeshSpec` and `kuberenetes.Controller`, or this function is going to bloat up and new fakes will be required based on the test variability.

This also makes it very clear in the test what the test is expecting and why, instead of the magic happening inside these fake functions.

pkg/catalog/helpers_test.go
```diff
@@ -125,6 +129,8 @@ func newFakeMeshCatalogForRoutes(t *testing.T) *MeshCatalog {
 	mockKubeController.EXPECT().IsMonitoredNamespace(tests.BookbuyerService.Namespace).Return(true).AnyTimes()
 	mockKubeController.EXPECT().ListMonitoredNamespaces().Return(listExpectedNs, nil).AnyTimes()
 
+	mockConfigurator.EXPECT().IsPermissiveTrafficPolicyMode().Return(testParams.permissiveMode).AnyTimes()
```
shashankram, line 132: Note that `AnyTimes()` is greedy, so you can't change the expectations for this mock call once `AnyTimes()` has been set.

This is another reason I suggest not initializing mocks in these fake functions but rather directly in the test.

shashankram (approved):

pkg/catalog/helpers_test.go
michelleN: I can definitely agree this may be brittle in the long run. Let's discuss this further bc the fake method is indeed creating and setting mocks. Do you have an example of the kind of set up you think is ideal?

shashankram: The fake is indeed creating mocks, but it is also relying on static fake SMI and Kube clients which are not extendable for tests. The problem with using static fake clients (smi, kubernetes) in table driven tests is that the APIs will return the same objects (ex. services, traffic targets etc.) every time. Also the fakes of these interfaces has proven to be very problematic especially when coupled with mocks. For ex. refer to the tests that use a combination of fakes + mocks in the catalog pkg.

PR #2176, opened 2020-12-09
ref(pkg/catalog): add ListMeshServices method 

michelleN:
**Description**:
addresses https://github.com/openservicemesh/osm/pull/2173#discussion_r539468262

**Affected area**:

- New Functionality      [ ]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [ ]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests                  [ ]
- CI System              [ ]
- Performance            [ ]
- Other                  [ ]

### DIFF
--- a/pkg/catalog/routes_test.go
+++ b/pkg/catalog/routes_test.go
@@ -5,11 +5,14 @@ import (
 	reflect "reflect"
 	"testing"
 
+	"github.com/golang/mock/gomock"
 	target "github.com/servicemeshinterface/smi-sdk-go/pkg/apis/access/v1alpha2"
 	"github.com/stretchr/testify/assert"
+	corev1 "k8s.io/api/core/v1"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"github.com/openservicemesh/osm/pkg/constants"
+	k8s "github.com/openservicemesh/osm/pkg/kubernetes"
 	"github.com/openservicemesh/osm/pkg/service"
 	"github.com/openservicemesh/osm/pkg/smi"
 	"github.com/openservicemesh/osm/pkg/tests"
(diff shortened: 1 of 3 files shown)


shashankram (approved):

pkg/catalog/types.go
```diff
@@ -72,6 +72,9 @@ type MeshCataloger interface {
 	// ListAllowedOutboundServiceAccounts lists the upstream service accounts the given service account can connect to
 	ListAllowedOutboundServiceAccounts(service.K8sServiceAccount) ([]service.K8sServiceAccount, error)
 
+	// ListMeshServices returns all services in the mesh
+	ListMeshServices() []service.MeshService
```
shashankram, line 76: What's the reason to expose this method in the interface? If it is internal to the `MeshCatalog`, it doesn't need to be exposed so wondering the use case.

michelleN: That is a very good point and I don't have a reason for exposing this at this time, so I will keep this private unless @snehachhabria objects.

shashankram (approved):

pkg/catalog/routes_test.go
```diff
@@ -421,6 +424,47 @@ func TestListAllowedOutboundServicesForIdentity(t *testing.T) {
...
+		kubeController: mockKubeController,
+	}
+
+	testCases := []struct {
+		name     string
+		services map[string]string // name: namespace
+	}{
+		{
+			name:     "services exist in mesh",
+			services: map[string]string{"bookstore": "bookstore-ns", "bookbuyer": "bookbuyer-ns", "bookwarehouse": "bookwarehouse"},
```
ksubrmnn, line 443: Could you directly use a list of MeshServices here?

ksubrmnn (approved):

pkg/catalog/routes_test.go
ksubrmnn: Also, this is more work, but it would be nice if the Mesh Services in pkg/test used individual namespaces so that we could just reuse those objects here and elsewhere

PR #2178, opened 2020-12-10
doc(charts): doc why failurePolicy must be Fail

michelleN:
**Description**:

failurePolicy in the MutatingWebhookConfiguration should
always be set to Fail so that no new resources get created
and thus bypass SMI Traffic Target policies if the for some
reason the webhook server is down. This is intentional and
guarentees that new pods in the monitored namespaces and those
intended to be injected with the sidecar are indeed injected.
Setting this value to Ignore may expose Pods to traffic
they should not be exposed to.

**Affected area**:

- New Functionality      [ ]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [ ]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests                  [ ]
- CI System              [ ]
- Performance            [ ]
- Other                  [ ]

No

### DIFF
--- a/charts/osm/templates/mutatingwebhook.yaml
+++ b/charts/osm/templates/mutatingwebhook.yaml
@@ -12,6 +12,8 @@ webhooks:
       namespace: {{.Release.Namespace}}
       path: /mutate-pod-creation
       port: 443
+  # failurePolicy should always be set to Fail to ensure no new resources get created without a sidecar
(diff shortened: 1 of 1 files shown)


ritazh (approved):
lgtm

PR #2187, opened 2020-12-11
feat(pkg/catalog): build inbound/outbound policies from traffic targets

michelleN:
This PR adds methods that build all outbound and inbound traffic policies for a given service account. This PR contains logic to build inbound and outbound policies based off SMI Traffic Targets. Follow up PRs will contain logic to handle building traffic policies in permissive traffic mode, update outbound traffic policies based on SMI Traffic Splits and update inbound traffic policies based on Ingress resources.

For each instance where the given service account is a source in an SMI Traffic Target, an outbound policies will be created for all services running workloads with the destination service account with an allow all route match indicating that the source can make requests to the destination. For each instance where the given service account is a destination in an SMI Traffic Target, inbound policies will be created such that there is an inbound policy for each service running workloads with the destination service account with rules for each route in the SMI Traffic Target and a list of allowed service accounts for each of those routes.

part of #2034

Signed-off-by: Michelle Noorali <michellemolu@gmail.com>

**Description**:

**Affected area**:

- New Functionality      [ ]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [X ]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [X ]
- Security               [ ]
- Tests                  [ ]
- CI System              [ ]
- Performance            [ ]
- Other                  [ ]


pkg/catalog/routes.go
```diff
@@ -22,6 +22,25 @@ const (
 	HTTPTraffic = "HTTPRouteGroup"
 )
 
+var allowAllRouteMatch trafficpolicy.HTTPRouteMatch = trafficpolicy.HTTPRouteMatch{
```
snehachhabria, line 25: nit : should we name this as `WildCardRouteMatch` ?

pkg/catalog/routes.go
```diff
@@ -514,9 +529,150 @@ func (mc *MeshCatalog) listMeshServices() []service.MeshService {
 	return services
 }
 
+func (mc *MeshCatalog) destinationServicesFromTrafficTarget(t *target.TrafficTarget) ([]service.MeshService, error) {
```
snehachhabria, line 532:
```suggestion
func (mc *MeshCatalog) getDestinationServicesFromTrafficTarget(t *target.TrafficTarget) ([]service.MeshService, error) {
```

pkg/catalog/routes.go
```diff
@@ -514,9 +529,150 @@ func (mc *MeshCatalog) listMeshServices() []service.MeshService {
...
+	sa := service.K8sServiceAccount{
+		Name:      t.Spec.Destination.Name,
+		Namespace: t.Spec.Destination.Namespace,
+	}
+	destServices, err := mc.GetServicesForServiceAccount(sa)
+	if err != nil {
+		return nil, errors.Errorf("Err finding Services for Service Account %#v: %v", sa, err)
```
snehachhabria, line 539:
```suggestion
		return nil, errors.Errorf("Error finding Services for Service Account %#v: %v", sa, err)
```

pkg/catalog/routes.go
```diff
@@ -514,9 +529,150 @@ func (mc *MeshCatalog) listMeshServices() []service.MeshService {
...
+func (mc *MeshCatalog) getHostnamesBasedOnNamespace(meshService service.MeshService) ([]string, []string, error) {
+	svc := mc.kubeController.GetService(meshService)
+	if svc == nil {
+		return nil, nil, errors.Errorf("Error fetching service %q", meshService)
+	}
+
+	hostnamesSameNamespace := kubernetes.GetLocalNamespaceHostnamesForService(svc)
+	hostnamesDiffNamespace := kubernetes.GetNamespaceScopedHostnamesForService(svc)
+
+	return hostnamesSameNamespace, hostnamesDiffNamespace, nil
```
snehachhabria, line 553: just out of curiosity: why do we want to differentiate local and namespaced hostnames ?

snehachhabria (dismissed):

pkg/catalog/routes.go
shashankram: s/Err/Error so that it can be easily grepped in the logs and is consistent with other error logs.

pkg/catalog/routes.go
michelleN: Great question. We'll use the InboundTrafficPolicy and OutboundTrafficPolicy objects to program virtual hosts. Because a domain has to be unique across virtual hosts, we'll want to create different policy objects for local hostnames vs namespaced hostnames so that we don't have to do any additional computation in rds/response.go at the time we're actually programming the virtual hosts.

pkg/catalog/routes.go
```diff
@@ -514,9 +529,150 @@ func (mc *MeshCatalog) listMeshServices() []service.MeshService {
...
+		Namespace: t.Spec.Destination.Namespace,
+	}
+	destServices, err := mc.GetServicesForServiceAccount(sa)
+	if err != nil {
+		return nil, errors.Errorf("Err finding Services for Service Account %#v: %v", sa, err)
+	}
+	return destServices, nil
+}
+
+func (mc *MeshCatalog) getHostnamesBasedOnNamespace(meshService service.MeshService) ([]string, []string, error) {
```
shashankram, line 544: Could you add a comment to this function given the function name has `BasedOnNamespace` but there is no parameter for a namespace.

pkg/catalog/routes.go
shashankram: +1, why can't this function automatically add the additional hostnames `service, service:port` as is done by `GetHostnamesForUpstreamService`?

pkg/catalog/routes.go
```diff
@@ -514,9 +529,150 @@ func (mc *MeshCatalog) listMeshServices() []service.MeshService {
...
+
+		hostnamesInNS, hostnamesNotInNS, err := mc.getHostnamesBasedOnNamespace(destService)
+		if err != nil {
+			continue
+		}
+
+		// create traffic policy with domains without namespace suffix
+		policyInNS := trafficpolicy.NewInboundTrafficPolicy(policyName(destService, true), hostnamesInNS)
+		// create traffic policy for domains with namespace suffix
+		policyNotInNS := trafficpolicy.NewInboundTrafficPolicy(policyName(destService, false), hostnamesNotInNS)
```
shashankram, line 584: Why are 2 separate policies being created? I understand these are being mapped to virtual hosts but I am unsure why there needs to be 2 of these here.

pkg/catalog/routes.go
```diff
@@ -514,9 +529,150 @@ func (mc *MeshCatalog) listMeshServices() []service.MeshService {
...
+}
+
 func isValidTrafficTarget(t *target.TrafficTarget) bool {
 	if t.Spec.Rules == nil || len(t.Spec.Rules) == 0 {
 		return false
 	}
 	return true
 }
+
+func policyName(svc service.MeshService, sameNamespace bool) string {
```
shashankram, line 672: please add a verb to the function name: `get, set, etc.` unless they are methods on types (ex. `svc.String()`)

pkg/catalog/routes.go
michelleN: updated to wildCardRouteMatch since we're only using this in this pkg for now

pkg/catalog/routes.go
michelleN: updated

pkg/catalog/routes.go
snehachhabria: +1

snehachhabria (dismissed):

pkg/catalog/routes.go
michelleN: Take the sample scenario below. If you have two clients (one in the same ns and one outside of the ns) trying to access a destination on different routes (traffic targets configured), you'll need to differentiate the inbound policies for the hostnames local to the ns and the namespaced scoped hostnames like so:
```
Bookbuyer in bookstore namespace
Bookbuyer in bookbuyer namespace
Bookstore in bookstore namespace 


TrafficTarget:
bookbuyer.bookbuyer → Bookstore.bookstore on the /buy route
Bookbuyer.bookstore → bookstore.bookstore on the /sell route


VirtualHosts on bookstore:
{
Domains: bookstore, bookstore:8888,
Rules:
	bookbuyer.bookstore SA can access /buy route
}
{
Domains: bookstore.bookstore, bookstore.bookstore:8888, bookstore.bookstore.svc, etc.
Rules:
	Bookbuyer.bookbuyer SA can access /sell route
	Bookbuyer.bookstore SA can access /buy route
}

````

Comment (shashankram) [reply pkg/catalog/routes.go:584]: @michelleN thanks for sharing the example. RBAC for the routes will be associated with routes, not the virtual host. So I would imagine the above example to look as follows:
```
Bookbuyer in bookstore namespace
Bookbuyer in bookbuyer namespace
Bookstore in bookstore namespace 


TrafficTarget:
bookbuyer.bookbuyer → Bookstore.bookstore on the /buy route
bookbuyer.bookstore → bookstore.bookstore on the /sell route


Inbound policies for bookstore.bookstore:
{
domains: bookstore, bookstore:8888, bookstore.bookstore, bookstore.bookstore:8888, bookstore.bookstore.svc, etc.
routes:
  - /buy
     rbac: source=bookbuyer.bookbuyer

  - /sell
     rbac: source=bookbuyer.bookstore
}
```

Essentially, the domains in the virtual_host don't need to be unique per route.

Comment (michelleN) [reply pkg/catalog/routes.go]: In this example you would be allowing bookbuyer.bookbuyer SA to make requests to the buy path even if the host header on the request is "bookstore" which it cannot be because it is in a different namespace so it would never get routed here and that makes sense to me.

File: pkg/catalog/routes.go [hunk 8]
```diff
@@ -514,9 +529,153 @@ func (mc *MeshCatalog) listMeshServices() []service.MeshService {
...
+
 func isValidTrafficTarget(t *target.TrafficTarget) bool {
 	if t.Spec.Rules == nil || len(t.Spec.Rules) == 0 {
 		return false
 	}
 	return true
 }
+
+// buildPolicyName creates a name for a policy associated with the given service
+func buildPolicyName(svc service.MeshService, namespaced bool) string {
```
ksubrmnn, line 675: Can you change this to `sameNamespace bool`

pkg/catalog/routes.go
```diff
@@ -514,9 +529,153 @@ func (mc *MeshCatalog) listMeshServices() []service.MeshService {
...
+
+		// create traffic policy with domains without namespace suffix
+		policyInNS := trafficpolicy.NewInboundTrafficPolicy(buildPolicyName(destService, true), hostnamesInNS)
+		// create traffic policy for domains with namespace suffix
+		policyNotInNS := trafficpolicy.NewInboundTrafficPolicy(buildPolicyName(destService, false), hostnamesNotInNS)
+
+		for _, sourceServiceAccount := range trafficTargetIdentitiesToSvcAccounts(t.Spec.Sources) {
+			for _, routeMatch := range routeMatches {
+				policyNotInNS.AddRule(*trafficpolicy.NewRouteWeightedCluster(routeMatch, wc), sourceServiceAccount)
+				if sourceServiceAccount.Namespace == destService.Namespace {
```
ksubrmnn, line 591: I'm assuming this logic right here is why you have the hostnames in two separate lists. That way a local service that is not attached to an authorized Service Account won't accidentally be allowed on the inbound path.

pkg/catalog/routes.go
michelleN: @snehachhabria and I had a discussion around these lines as well earlier and she also recommended building one inbound policy per service. I'll go ahead and update the code as well. If we end up running into scenarios later (wondering if we might need this logic for ingress but not sure), we can add this logic back in if needed.

pkg/catalog/routes.go
michelleN: yes but check out this [thread](https://github.com/openservicemesh/osm/pull/2187#discussion_r541103978). I went back and forth about it with @snehachhabria and @shashankram and am going to take this out bc of the scenario listed in the example. Lmk what you think.

pkg/tests/fixtures.go
```diff
@@ -150,6 +171,44 @@ var (
 		"bookstore-v2.default.svc.cluster.local:8888",
 	}
 
+	// BookstoreV2DefaultHostnames are the hostnames for the bookstore-v2 service in the default namespace
```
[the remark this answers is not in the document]
michelleN: default is the hostname

pkg/catalog/routes.go
michelleN: yes

shashankram (approved):

ksubrmnn (approved):

PR #2191, opened 2020-12-11
endpoint/kube: Cleanup and move remaining caches to kubecontroller

eduser25:
This commit removes a repeated cache (pods) and moves the endpoint cache
to kubecontroller in an attempt to unify code and  interfaces that handles 
kubernetes objects.

This commit also removes the mocks on `provider/kube` tests that were trying
to dynamically simulate kubecontroller logic (which is highly error-prone),
in favour of using the real kubecontroller object instead for simplicity.
Specific information for each category has been added on their respective
ginkgo `Defines`.

- Pod cache from provider has been removed. Kubecontroller held one
all along already.
- Endpoint cache has been moved from provider/kube to Kubecontroller.
- API to get endpoints for a service was added (GetEndpoints).
- Removed the Announcement channel for provider's interface and implementation,
  subscription/announcement capabilities are still available through pubsub.

Tests:
- Moved the tests that were using dynamic mocks to use a real kubecontroller object
instead.
- Added some more synchronization as there were still some races lurking around.

- Control Plane          [X]
- Tests                  [X]

No


pkg/catalog/catalog_test.go
```diff
@@ -22,7 +22,7 @@ var _ = Describe("Test catalog functions", func() {
 			chans := mc.getAnnouncementChannels()
 
 			// Currently returns len(Channels), see getAnnouncementChannels implementation for details.
-			expectedNumberOfChannels := 6
+			expectedNumberOfChannels := 5
```
shashankram, line 25: Can we list these 5 in comments?

shashankram (approved):

pkg/catalog/catalog_test.go
eduser25: will do

draychev (approved):

PR #2194, opened 2020-12-14
tests/e2e: add e2e test for TCP traffic

shashankram:
**Description**:
This change adds an e2e test for TCP traffic. The test verifies
that the inbound and outbound TCP filtering and proxying works
as expected. It uses `netcat` to initiate TCP connections
to the `tcp-echo-server` that echoes back the response.

Due to the absence of TCPRoute spec support at this moment, the
test uses a TrafficTarget without specific TCP rules to verify
inbound and outbound proxying of TCP traffic. The test will need
to be updated to be able to perform RBAC on TCP routes when the
TCPRoute spec is available to use.

It also updates the client container image for the test to rely
on an image with additional tools (ex. netcat in this case).

Part of #1521

**Affected area**:

- New Functionality      [ ]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [ ]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests                  [X]
- CI System              [ ]
- Performance            [ ]
- Other                  [ ]

`No`

### DIFF
--- a/tests/e2e/e2e_pod_client_server_test.go
+++ b/tests/e2e/e2e_pod_client_server_test.go
@@ -167,7 +167,7 @@ func testTCPTraffic(withSourceKubernetesService bool) {
 				SimplePodAppDef{
 					Name:        destName,
 					Namespace:   destName,
-					Image:       fmt.Sprintf("%s/tcp-echo-server:%s", installOpts.ContainerRegistryLoc, installOpts.OsmImagetag), // FIXME
+					Image:       fmt.Sprintf("%s/tcp-echo-server:%s", installOpts.ContainerRegistryLoc, installOpts.OsmImagetag),
 					Command:     []string{"/tcp-echo-server"},
 					Args:        []string{"--port", fmt.Sprintf("%d", destinationPort)},
 					Ports:       []int{destinationPort},
--- a/tests/e2e/e2e_pod_client_server_test.go
+++ b/tests/e2e/e2e_pod_client_server_test.go
@@ -21,12 +21,16 @@ var _ = OSMDescribe("Test HTTP traffic from 1 pod client -> 1 pod server",
 		Bucket: 1,
 	},
 	func() {
-		Context("SimpleClientServer with a Kubernetes Service for the Source", func() {
+		Context("SimpleClientServer with a Kubernetes Service for the Source: HTTP", func() {
 			testTraffic(true)
 		})
 
-		Context("SimpleClientServer without a Kubernetes Service for the Source", func() {
+		Context("SimpleClientServer without a Kubernetes Service for the Source: HTTP", func() {
 			testTraffic(false)
+		})
+
+		Context("SimpleClientServer without a Kubernetes Service for the Source: TCP", func() {
+			testTCPTraffic(false)
 		})
 	})
 
@@ -57,9 +61,9 @@ func testTraffic(withSourceKubernetesService bool) {
 
 			_, err := Td.CreateServiceAccount(destName, &svcAccDef)
 			Expect(err).NotTo(HaveOccurred())
-			dstPod, err := Td.CreatePod(destName, podDef)
+			_, err = Td.CreatePod(destName, podDef)
 			Expect(err).NotTo(HaveOccurred())
-			_, err = Td.CreateService(destName, svcDef)
+			dstSvc, err := Td.CreateService(destName, svcDef)
 			Expect(err).NotTo(HaveOccurred())
 
 			// Expect it to be up and running in it's receiver namespace
@@ -93,7 +97,7 @@ func testTraffic(withSourceKubernetesService bool) {
 				SourcePod:       srcPod.Name,
 				SourceContainer: sourceName,
 
-				Destination: fmt.Sprintf("%s.%s", dstPod.Name, dstPod.Namespace),
+				Destination: fmt.Sprintf("%s.%s", dstSvc.Name, dstSvc.Namespace),
 			}
 
 			srcToDestStr := fmt.Sprintf("%s -> %s",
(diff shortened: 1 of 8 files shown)


tests/e2e/e2e_pod_client_server_test.go
```diff
@@ -21,13 +21,17 @@ var _ = OSMDescribe("Test HTTP traffic from 1 pod client -> 1 pod server",
...
+		Context("SimpleClientServer with a Kubernetes Service for the Source: HTTP", func() {
 			testTraffic(true)
 		})
 
-		Context("SimpleClientServer without a Kubernetes Service for the Source", func() {
+		Context("SimpleClientServer without a Kubernetes Service for the Source: HTTP", func() {
 			testTraffic(false)
 		})
+
+		Context("SimpleClientServer without a Kubernetes Service for the Source: TCP", func() {
```
eduser25, line 32: `testTCPTraffic()`, though similar setup, doesn't really share any code with `testTraffic`, since we are voucing to keep a test-per-file too, I'd suggest to move it on a different file altogether and have the `*_http_test.go` and the `*_tcp_test.go` versions of the test on separate files

tests/framework/common_traffic.go
```diff
@@ -76,6 +98,31 @@ func (td *OsmTestData) HTTPRequest(ht HTTPRequestDef) HTTPRequestResult {
...
 
+// TCPRequest runs a synchronous TCP request to run the TCPRequestDef and return a TCPRequestResult
+func (td *OsmTestData) TCPRequest(req TCPRequestDef) TCPRequestResult {
+	commandArgs := fmt.Sprintf("echo \"%s\" | nc %s %d", req.Message, req.DestinationHost, req.DestinationPort)
+	command := []string{"sh", "-c", commandArgs}
+
+	stdout, stderr, err := td.RunRemote(req.SourceNs, req.SourcePod, req.SourceContainer, command)
+	if err != nil {
+		// Error codes from the execution come through err
+		// Curl 'Connection refused' err code = 7
```
eduser25, line 109: you might want to remove this

tests/e2e/e2e_pod_client_server_test.go
```diff
@@ -136,14 +140,133 @@ func testTraffic(withSourceKubernetesService bool) {
...
+				result := Td.TCPRequest(clientToServer)
+
+				if result.Err != nil {
+					Td.T.Logf("> (%s) TCP Req failed, response: %s, err: %s",
+						srcToDestStr, result.Response, result.Err)
+					return false
+				}
+
+				// Ensure the echo response contains request message
+				if !strings.Contains(result.Response, requestMsg) {
```
eduser25, line 233: nit: should we consider exact equal since we expect echo? (is nc returning a clean echo or do you have some other stdout text hanging around?)

eduser25 (commented):
lgtm, will wait to +1 to see if you agree to split on files or not

tests/e2e/e2e_pod_client_server_test.go
shashankram: The reason I left it here is because `testTCPTraffic` falls under the bucket of `e2e_pod_client_server_test` where simple client server traffic scenario is tested. We can break this into `e2e_pod_client_server_test_<http|tcp>` if necessary, but I don't see a strong reason to do so at this moment. Should be a straightforward change though (mostly moving files around). I'd rather address this in a follow-up if it must be done.

tests/e2e/e2e_pod_client_server_test.go
shashankram: nope, server prefixes the response with a prefix. So if client sends `foo`, server responds `echo response: foo`

eduser25 (approved):

nojnhuh (approved):

tests/framework/common_apps.go
```diff
@@ -38,8 +38,15 @@ const (
...
 	MemRSSPanel int = 13
+
 	// CPUPanel is the ID of the CPU panel on OSM's MeshDetails dashboard
 	CPUPanel int = 14
+
+	// AppProtocolHTTP is the HTTP application protocol
+	AppProtocolHTTP = "http"
+
+	// AppProtocolTCP is the TCP application protocol
+	AppProtocolTCP = "tcp"
```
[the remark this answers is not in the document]
shashankram: Would just be weird exporting stuff in `pkg/envoy` to an e2e test. This could very well be hardcoded to `http` inline in the e2e test code as well, considering its just a property of the test and is user driven. User can put whatever they choose, although the controller only support http and tcp at the moment.

tests/e2e/e2e_pod_client_server_test.go
```diff
@@ -136,14 +140,133 @@ func testTraffic(withSourceKubernetesService bool) {
...
+			}
+
+			destinationPort := 80
+
+			// Get simple pod definitions for the TCP server
+			svcAccDef, podDef, svcDef := Td.SimplePodApp(
+				SimplePodAppDef{
+					Name:        destName,
+					Namespace:   destName,
+					Image:       fmt.Sprintf("%s/tcp-echo-server:%s", installOpts.ContainerRegistryLoc, installOpts.OsmImagetag), // FIXME
```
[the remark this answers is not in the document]
shashankram: Missed removing this from debugging.

tests/framework/common.go
```diff
@@ -384,6 +384,42 @@ func (td *OsmTestData) DeleteHelmRelease(name, namespace string) error {
...
+	}
+
+	td.T.Log("Getting image data")
+	docker, err := client.NewClientWithOpts(client.WithAPIVersionNegotiation())
+	if err != nil {
+		return errors.Wrap(err, "failed to create docker client")
+	}
+	var imageIDs []string
+	for _, name := range imageNames {
+		imageName := fmt.Sprintf("%s/%s:%s", td.CtrRegistryServer, name, td.OsmImageTag)
```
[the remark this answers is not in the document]
shashankram: Done, thanks

nojnhuh (approved):

draychev (approved):

PR #2203, opened 2020-12-15
fix(ci): workaround DCO requirement for gh-pages branch

nojnhuh:
**Description**:
This change adds a step to the chart release CI workflow to force git to
signoff commits to pass the DCO check.

**Affected area**:

- New Functionality      [ ]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [ ]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests                  [ ]
- CI System              [X]
- Performance            [ ]
- Other                  [ ]

No

### DIFF
--- a/.github/workflows/release-chart.yml
+++ b/.github/workflows/release-chart.yml
@@ -9,6 +9,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
+      # Workaround DCO requirement for gh-pages branch
+      - run: git config alias.commit "commit -s"
       - name: Publish Helm charts
         uses: stefanprodan/helm-gh-pages@v1.1.0
         with:


SanyaKochhar (approved):

.github/workflows/release-chart.yml
```diff
@@ -9,6 +9,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
+      # Workaround DCO requirement for gh-pages branch
+      - run: git config alias.commit "commit -s"
       - name: Publish Helm charts
         uses: stefanprodan/helm-gh-pages@v1.1.0
```
phillipgibson, line 15: Should we update to v1.3.0 as well?

phillipgibson (approved):

.github/workflows/release-chart.yml
nojnhuh: Looking at the changes, I don't see anything that we need or would use. I could make that a separate change if there's something I'm missing.
https://github.com/stefanprodan/helm-gh-pages/releases

PR #2227, opened 2020-12-21
tests/e2e: move tcp test to own file and test permissive mode

shashankram:
**Description**:
Moves the TCP traffic test to its own file and adds a scenario
to test traffic in permissive mode.

Part of #1521

**Affected area**:

- New Functionality      [ ]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [ ]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests                  [X]
- CI System              [ ]
- Performance            [ ]
- Other                  [ ]

`No`


eduser25 (approved):

ksubrmnn (approved):

ksubrmnn:
Is there a reason that the client pod is not behind a service?

shashankram:
> Is there a reason that the client pod is not behind a service?

Yes, because client pods in kubernetes don't have to be backed by services. We had the limitation to create services for clients which has been removed, so it makes sense to test it the way real users would wire up their apps, which is without needing to create a service for client pods.

There are a few other tests that create services for client, but it didn't make sense to replicate this for all tests since we aren't testing anything additional.

PR #2228, opened 2020-12-21
Announcement Channels cleanup

eduser25:
With the introduction of pubsub, any module can post messages
N->N sender-receiver, which deprecates the use of unidirectional
channels for module specific events.

This commit cleans up the remains of the previously used subscription
system. None of the code removed right now was actually backed functionality,
as all event processing has already been migrated to pubsub.

- Cleanup [X]

No


pkg/catalog/catalog.go
```diff
@@ -49,30 +39,3 @@ func NewMeshCatalog(kubeController k8s.Controller, kubeClient kubernetes.Interfa
...
-		{"MeshSpec", mc.meshSpec.GetAnnouncementsChannel()},
-		{"CertManager", mc.certManager.GetAnnouncementsChannel()},
-		{"IngressMonitor", mc.ingressMonitor.GetAnnouncementsChannel()},
-		{"Ticker", ticking},
-		{"Services", mc.kubeController.GetAnnouncementsChannel(k8s.Services)},
-	}
-
-	if updateAtLeastEvery > 0 {
-		go func() {
-			ticker := time.NewTicker(updateAtLeastEvery)
```
eduser25, line 65: @draychev  I have an open item on me to reintroduce it when appropriate.

draychev (approved):

pkg/catalog/catalog.go
shashankram: Do we want to track this in an issue?

shashankram (approved):

pkg/catalog/catalog.go
eduser25: https://github.com/openservicemesh/osm/issues/2103

Issue #2264, opened 2021-01-05
validating webhook error: no such host


shalier:
**Bug description**:
The validating webhook is not working, getting the following error:
`error: configmaps "osm-config" could not be patched: Internal error occurred: failed calling webhook "osm-config-webhook.k8s.io": Post "https://osm-controller.osm-system.svc:443/validate-webhook?timeout=30s": dial tcp: lookup osm-controller.osm-system.svc on 192.168.65.2:53: no such host`

**Affected area (please mark with X where applicable)**:
- Install                    [ ]
- SMI Traffic Access Policy  [ ]
- SMI Traffic Specs Policy   [ ]
- SMI Traffic Split Policy   [ ]
- Permissive Traffic Policy  [ ]
- Ingress                    [ ]
- Egress                     [ ]
- Envoy Control Plane        [ ]
- CLI Tool                   [ ]
- Metrics                    [ ]
- Certificate Management     [ ]
- Sidecar Injection          [ ]
- Logging                    [ ]
- Debugging                  [ ]
- Tests                      [ ]
- CI System                  [ ]

**Expected behavior**:
Should catch updates to `osm-config` and validate changes

**Steps to reproduce the bug (as precisely as possible)**:

**How was OSM installed?**:

**Anything else we need to know?**:

**Environment**:
- OSM version (use `osm version`):
- Kubernetes version (use `kubectl version`):
- Size of cluster (number of worker nodes in the cluster):
- Others:

shalier:
Problems w/ Validating Webhook

1. There was a port conflict with the validating and mutating webhook (port: 443 | targetPort: 9090)
- Originally the validating webhook server was unable to start because of the port conflict. As a means of fixing this golang put the path for the validating webhook onto the mutating webhook server, which gave the appearance that the validating webhook was working.

- The port conflict was fixed by having the validating webhook on port 9093 - this revealed the underlying error: 
`error: configmaps "osm-config" could not be patched: Internal error occurred: failed calling webhook "osm-config-webhook.k8s.io": Post "https://osm-controller.osm-system.svc:443/validate-webhook?timeout=30s": dial tcp: lookup osm-controller.osm-system.svc on 192.168.65.2:53: no such host`

2. Further debugging (ty @shashankram) revealed that deleting the osm-controller pod fixes the validating webhook error.
- Note: Helm deploys resources in a certain order which can add to the deadlocks and races.

PR #2268, opened 2021-01-06
charts: disable validating webhook configuration

shashankram:
**Description**:
Since the validating webhook has been disabled by c48acb9c due to
issue #2264, the validating webhook config needs to be disabled as
well. Otherwise config-map update will be broken.

**Affected area**:

- New Functionality      [ ]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [ ]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests                  [ ]
- CI System              [ ]
- Performance            [ ]
- Other                  [X]

`No`

### DIFF
--- a/charts/osm/templates/validatingwebhook.yaml
+++ b/charts/osm/templates/validatingwebhook.yaml
@@ -1,3 +1,4 @@
+{{- /* disable validating webhook due to Issue #2264
 apiVersion: admissionregistration.k8s.io/v1beta1
 kind: ValidatingWebhookConfiguration
 metadata:
@@ -27,3 +28,4 @@ webhooks:
         - UPDATE
       resources:
         - configmaps
+*/}}


charts/osm/values.yaml
```diff
@@ -83,3 +83,7 @@ OpenServiceMesh:
     # Destination's API or collector endpoint where the spans will
     # be sent to
     endpoint: "/api/v2/spans"
+
+  # Set to true to enable the osm-config validator webhook
+  # Currently disabled due to Issue #2264
+  enableConfigValidator: false
```
nojnhuh, line 89: Any reason to create a new value here vs. just commenting out the config temporarily to mirror c48acb9c61c8650a7ff0b4516de001c3d1728387? I don't see why anyone would ever set this to `true` while the issue gets fixed.

shashankram: Is there a Helm equivalent of `#if 0` ?

nojnhuh: You can use `{{/* a comment */}}` or I believe `{{ if false }}` would work too.

shashankram: Perfect, will try this.

nojnhuh (approved):

ksubrmnn (approved):

draychev (approved):

PR #2270, opened 2021-01-07
metricsstore: integrate metrics store framework

shashankram:
**Description**:
This change integrates the metrics store framework
to allow osm-controller to publish metrics exposed
on the http server's `/metrics` path.

It introduces a `DefaultMetricsStore` that can be
used throughout the controller components. The metrics
store now publishes metrics on k8s api events.
It also updates the metrics to follow best practices
on constant labels to avoid setting the same label on
all metrics as per https://github.com/prometheus/client_golang/blob/v1.9.0/prometheus/metric.go#L87.

Part of #902 

**Affected area**:

- New Functionality      [ ]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [ ]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests                  [ ]
- CI System              [ ]
- Performance            [ ]
- Other                  [ ]


draychev (approved):

eduser25 (approved):
I still have still doubts how to expose apis per counter. As discussed offline, we'll tackle that further in docs and/or with other folks, but otherwise lgtm

shashankram:
> lgtm, I still have still doubts how to expose apis per counter. As discussed offline, we'll tackle that further in docs and/or with other folks

Yes, nothing is set in stone, this one API is just here to validate end to end usage. We will likely tweak it significantly to avoid `n` APIs where `n = #number of metrics`

Issue #2122, opened 2020-11-25
Fix label scalability/allow list for prometheus metrics


eduser25:
Too many labels causes time bucket series to scale poorly on prometheus backend. (#1934)

The difference in Mb/pod in Prometheus:
With time-bucket series: ~20MB/pod
Without time-bucket series: ~6MB/pod

We have temporarily disabled scraping `.*bucket.*` series (#2117) to allow prometheus to at least operate at reasonable numbers.

This however needs a scalable fix, probably in the form of only allow listing metrics of interest, or limitting the number of labels we scrap otherwise.

- Metrics                    [X]
- Performance           [X]

PR #2276, opened 2021-01-08
prometheus: limit scrapping to currently used metrics only

eduser25:
Prometheus is still (by default) scrapping a large number of metrics we
are currently not using on our deployments.

This does not mean to say they are usless (which they are not), but by
default adding all metrics is causing serious memory bottlnecks when
increasing scale even at lower numbers.

A previous commit hacked the bucket timeseries out of the equation, as they
amounted for roughly 75% of memory usage on prometheus (from 20Mb per pod to 6Mb)
https://github.com/openservicemesh/osm/pull/2117

This patch pinpoints the exact metrics we use on our deployments, and limits
ingestion to only metrics we we are currently using.
This brings down 6Mb to 2Mb per pod (additional 66%), further allowing operation
of prometheus under larger setups.

Fixes: #2122

**Affected area**:
- Performance            [X]

No

### DIFF
--- a/charts/osm/templates/prometheus-configmap.yaml
+++ b/charts/osm/templates/prometheus-configmap.yaml
@@ -25,8 +25,8 @@ data:
         bearer_token_file: /var/run/secrets/kubernetes.io/serviceaccount/token
         metric_relabel_configs:
         - source_labels: [__name__]
-          regex: .*bucket.*
-          action: drop   
+          regex: '(apiserver_watch_events_total|apiserver_admission_webhook_rejection_count)'
+          action: keep   
         relabel_configs: 
         - source_labels: [__meta_kubernetes_namespace, __meta_kubernetes_service_name, __meta_kubernetes_endpoint_port_name]
           action: keep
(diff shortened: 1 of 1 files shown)


charts/osm/templates/prometheus-configmap.yaml
```diff
@@ -128,32 +132,32 @@ data:
...
-          replacement: $1:$2
-        - action: labelmap
-          regex: __meta_kubernetes_service_label_(.+)
-        - source_labels: [__meta_kubernetes_namespace]
-          action: replace
-          target_label: kubernetes_namespace
-        - source_labels: [__meta_kubernetes_service_name]
-          action: replace
-          target_label: kubernetes_name
+      # - job_name: 'kubernetes-service-endpoints'
```
shashankram, line 135: Could you drop a comment in the file why these are commented out and what are the next steps.

shashankram (approved):

charts/osm/templates/prometheus-configmap.yaml
eduser25: I can do that. Basically I've seen we use no metrics from these endpoints, it's simple as that.
I'd like for @snehachhabria to take a look too as I'm not aware of these being used anywhere, but I'd rather have it confirmed.

shashankram: If we don't need it, might as well remove it. If we leave it here commented, let's document it in code and follow it up with an issue.

nojnhuh (approved):

PR #2284, opened 2021-01-11
metrics: add injector webhook metrics

eduser25:
Adds two new metrics:
- InjectorSidecarCount: keeps track of number of sidecar injector webhooks
dealt over time.
- InjectorRqTime: time taken to process injector webhooks

**Affected area**:

- Metrics                [X]

No

### DIFF
--- a/pkg/injector/profiling.go
+++ b/pkg/injector/profiling.go
@@ -1,32 +1,43 @@
 package injector
 
 import (
+	"fmt"
 	"net/http"
 	"time"
+
+	"github.com/openservicemesh/osm/pkg/metricsstore"
+)
+
+var (
+	defaultK8sTimeout = time.Duration(30 * time.Second)
 )
 
 // Helper to parse timeout variable from webhook URL
-func readTimeout(req *http.Request) (*time.Duration, error) {
+func readTimeout(req *http.Request) (time.Duration, error) {
 	durationValue, found := req.URL.Query()[webhookMutateTimeoutKey]
 	if !found || len(durationValue) != 1 {
 		log.Error().Msgf("Webhook timeout value not found in request")
-		return nil, errParseWebhookTimeout
+		return defaultK8sTimeout, errParseWebhookTimeout
 	}
 
 	val, err := time.ParseDuration(durationValue[0])
 	if err != nil {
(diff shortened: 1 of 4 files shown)


pkg/metricsstore/metricsstore.go
```diff
@@ -57,19 +66,45 @@ func init() {
...
+		Help:      "Counts the number of injector webhooks dealt with over time",
+	})
+
+	defaultMetricsStore.InjectorRqTime = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Namespace: metricsRootNamespace,
+			Subsystem: "injector",
+			Name:      "injector_rq_time",
+			Buckets:   []float64{.1, .25, .5, 1, 2.5, 5, 10, 20, 40},
+			Help:      "Histogram for time taken to process an injector webhooks",
```
shashankram, line 85:
```suggestion
			Help:      "Histogram for time taken to perform sidecar injection",
```

shashankram (approved):

pkg/injector/profiling.go
```diff
@@ -1,32 +1,43 @@
...
+func webhookTimeTrack(start time.Time, timeout time.Duration, success *bool) {
 	elapsed := time.Since(start)
 	percentOfTimeout := float64(elapsed.Microseconds()) / float64(timeout.Microseconds())
 
 	log.Debug().Msgf("Mutate Webhook took %v to execute (%.2f of it's timeout, %v)",
 		elapsed, percentOfTimeout, timeout)
+
+	metricsstore.DefaultMetricsStore.InjectorSidecarCount.Inc()
+	metricsstore.DefaultMetricsStore.InjectorRqTime.
+		WithLabelValues(fmt.Sprintf("%t", *success)).Observe(elapsed.Seconds())
```
nojnhuh, line 42: Out of context, this pointer dereference seems dangerous, but I think I understand why `success` is a pointer and why it's not an issue.

nojnhuh (approved):

pkg/injector/profiling.go
eduser25: golang should ensure to keep it around even out of scope as soon as dereference it, till the defer goes out (since it's the only pointer left to it).
I did this consciously expecting for this behavior at least.

PR #2286, opened 2021-01-11
envoy: update to v1.17 to leverage default_filter_chain matching

### LINKED ISSUE #2096: Move Egress filter chain to use v3 `default_filter_chain`
V3 supports `default_filter_chain`, which is the filter chain that is going to execute when non is matched on the `FilterChainMatch` list.

For egress, we are currently populating a full space of possible destinations, so anything not hitting a possible cluster endpoint IP is considered egress, as nothing else is assessed other than IP. In a sense, the space of things we hit is complete, so there cannot be partial filterchain matches, so egress works.

However, if we introduce more criteria for filterchain matching (ie: port), there could be partial matches, and egress will no longer be hit (this is very related to how envoy implements criteria matching).

`default_filter_chain` in this case is exactly what we want for egress.

- Egress                     [X]

shashankram:
**Description**:
Currently the egress filter chain is a part of `listener.filter_chains`
config. While this works for now, there is the risk of egress traffic
matching in-mesh filter chains due to the way envoy performs
filter chain matching as we augment the service filter chains
with additional match criteria such as ports. The `listener.default_filter_chain` 
was specifically introduced to avoid such unintentional matching.

This is also required to be able to do port:protocol matching
on the oundbound filter chains with the introduction of TCP
support to distinguish between protocol specific traffic by
its port.

Resolves #2096

**Affected area**:

- New Functionality      [ ]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [X]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests                  [ ]
- CI System              [ ]
- Performance            [ ]
- Other                  [ ]

`No`

### DIFF
--- a/pkg/envoy/lds/listener.go
+++ b/pkg/envoy/lds/listener.go
@@ -23,18 +23,9 @@ const (
 )
 
 func (lb *listenerBuilder) newOutboundListener() (*xds_listener.Listener, error) {
-	serviceFilterChains, err := lb.getOutboundFilterChains()
-	if err != nil {
-		log.Error().Err(err).Msgf("Error getting filter chains for outbound listener")
-		return nil, err
-	}
+	serviceFilterChains := lb.getOutboundFilterChainPerUpstream()
 
-	if len(serviceFilterChains) == 0 {
-		log.Info().Msgf("No filterchains for outbound services. Not programming Outbound listener.")
-		return nil, nil
-	}
-
-	return &xds_listener.Listener{
+	listener := &xds_listener.Listener{
 		Name:             outboundListenerName,
 		Address:          envoy.GetAddress(constants.WildcardIPAddr, constants.EnvoyOutboundListenerPort),
 		TrafficDirection: xds_core.TrafficDirection_OUTBOUND,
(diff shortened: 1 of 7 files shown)


pkg/envoy/lds/listener.go
```diff
@@ -51,7 +42,29 @@ func (lb *listenerBuilder) newOutboundListener() (*xds_listener.Listener, error)
...
-	}, nil
+	}
+
+	// Create filter chain for egress if egress is enabled
+	// This filterchain matches any traffic not filtered by allow rules, it will be treated as egress
+	// traffic when enabled
+	if lb.cfg.IsEgressEnabled() {
+		egressFilterChgain, err := buildEgressFilterChain()
+		if err != nil {
+			log.Error().Err(err).Msgf("Error getting filter chain for Egress")
```
snehachhabria, line 53:
```suggestion
			log.Error().Err(err).Msgf("Error building filter chain for Egress")
```

snehachhabria (approved):

eduser25 (approved):

pkg/envoy/lds/listener.go
shashankram: I'll consider updating this in a subsequent change or if I need to rebase.

PR #2295, opened 2021-01-13
catalog: add api to retrieve port:protocol mapping

shashankram:
**Description**:
This change adds an api to retrieve a service's port
to application protocol mapping, where the port is
the one used by clients to send requests to. This
port can be different than the actual ports exposed
by the service's application binary, ie. the api
returns the a service's `port` as opposed to its
`targetPort`.

This api will be used to program outbound filter chains
for upstream services which need to match on the ports
the clients send traffic to, and the protocol mapping
to be used for determining the type of traffic being
sent on that port.

For backwards compatibility, the default appProtocol
for a port is assumed to be `http` if not specified,
similar to other usages of the `appProtocol` field.

**Affected area**:

- New Functionality      [X]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [ ]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests                  [ ]
- CI System              [ ]
- Performance            [ ]
- Other                  [ ]

`No`

### DIFF
--- a/pkg/catalog/types.go
+++ b/pkg/catalog/types.go
@@ -116,8 +116,15 @@ type MeshCataloger interface {
 	// ListMonitoredNamespaces lists namespaces monitored by the control plane
 	ListMonitoredNamespaces() []string
 
-	// GetPortToProtocolMappingForService returns a mapping of the service's ports to their corresponding application protocol
+	// GetPortToProtocolMappingForService returns a mapping of the service's ports to their corresponding application protocol.
+	// The ports returned are the actual ports on which the application exposes the service derived from the service's endpoints,
+	// ie. 'spec.ports[].targetPort' instead of 'spec.ports[].port' for a Kubernetes service.
 	GetPortToProtocolMappingForService(service.MeshService) (map[uint32]string, error)
+
+	// GetPortToProtocolMappingForService returns a mapping of the service's ports to their corresponding application protocol,
+	// where the ports returned are the ones used by downstream clients in their requests. This can be different from the ports
+	// actually exposed by the application binary, ie. 'spec.ports[].port' instead of 'spec.ports[].targetPort' for a Kubernetes service.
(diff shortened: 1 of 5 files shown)


pkg/catalog/types.go
```diff
@@ -116,9 +116,16 @@ type MeshCataloger interface {
 	// ListMonitoredNamespaces lists namespaces monitored by the control plane
 	ListMonitoredNamespaces() []string
 
-	// GetPortToProtocolMappingForService returns a mapping of the service's ports to their corresponding application protocol
+	// GetPortToProtocolMappingForService returns a mapping of the service's ports to their corresponding application protocol.
+	// The ports returned are the actual ports on which the application exposes the service derived from the service's endpoints,
+	// ie. 'spec.ports[].targetPort' instead of 'spec.ports[].port' for a Kubernetes service.
 	GetPortToProtocolMappingForService(service.MeshService) (map[uint32]string, error)
```
snehachhabria, line 122: Should we rename this function to provide more clarity ? Given that we have two functions to return ports 
```suggestion
	GetTargetPortToProtocolMappingForService(service.MeshService) (map[uint32]string, error)
```

snehachhabria (approved):

pkg/catalog/types.go
shashankram: TargetPort is very specific to k8s, while the API is agnostic of providers if you look into it.

shashankram: Based on discussions with @snehachhabria, we will rename the two APIs for more clarity in a subsequent PR.

eduser25 (approved):

eduser25:
One more api tho... my catalog branch will suffer

PR #2304, opened 2021-01-14
Updated chart readme

ams0:
**Description**:

A more descriptive Readme for the Helm chart

**Affected area**:

- New Functionality      [ ]
- Documentation          [X]
- Install                [ ]
- Control Plane          [ ]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests                  [ ]
- CI System              [ ]
- Performance            [ ]
- Other                  [ ]

No


shashankram (approved):

nojnhuh (changes requested):
Thanks @ams0, these look like good changes.

I know it's a bit buried, but we generate the chart's README, primarily so it's easy to update the values section. Could you make sure the generated README still incorporates these changes? https://github.com/openservicemesh/osm/blob/de5bd8de1ab33987e69aefd0debc818656887cbc/Makefile#L58-L60

We've been using that tool's default configuration so far. I'm not very familiar with its customization options, but I know it's template-based and very flexible. Could you take a look and template what you have here and pass it to the tool in the Makefile?

ams0:
I see now @nojnhuh , will update the `values.yaml` to reflect the comments and will get done better. Looking into `helm-docs` now.

PR #2306, opened 2021-01-15
metrics : exposing metric for time take to issue xds certificates 

snehachhabria:
Description:

Adds a metric for the time take to issue of xds certificates.

**Affected area**:

- New Functionality      [ ]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [ ]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [X]
- SMI Policy             [ ]
- Security               [ ]
- Tests                  [ ]
- CI System              [ ]
- Performance            [ ]
- Other                  [ ]

`No`

### DIFF
--- a/pkg/injector/patch.go
+++ b/pkg/injector/patch.go
@@ -33,9 +33,10 @@ func (wh *mutatingWebhook) createPatch(pod *corev1.Pod, req *v1beta1.AdmissionRe
 		return nil, err
 	}
 	elapsed := time.Since(startTime)
-	certXdsIssueTimeTrack(elapsed, cn.String(), &certIssuanceSuccess)
 
-	metricsstore.DefaultMetricsStore.CertsXdsIssuedCounter.Inc()
+	metricsstore.DefaultMetricsStore.CertXdsIssuedCount.Inc()
+	metricsstore.DefaultMetricsStore.CertXdsIssuedTime.
+		WithLabelValues(cn.String(), fmt.Sprintf("%t", certIssuanceSuccess)).Observe(elapsed.Seconds())
 	originalHealthProbes := rewriteHealthProbes(pod)
 
 	wh.meshCatalog.ExpectProxy(cn)
--- a/pkg/injector/patch.go
+++ b/pkg/injector/patch.go
@@ -36,7 +36,7 @@ func (wh *mutatingWebhook) createPatch(pod *corev1.Pod, req *v1beta1.AdmissionRe
 
 	metricsstore.DefaultMetricsStore.CertXdsIssuedCount.Inc()
 	metricsstore.DefaultMetricsStore.CertXdsIssuedTime.
-		WithLabelValues(cn.String(), fmt.Sprintf("%t", certIssuanceSuccess)).Observe(elapsed.Seconds())
+		WithLabelValues(cn.String(), strconv.FormatBool(certIssuanceSuccess)).Observe(elapsed.Seconds())
 	originalHealthProbes := rewriteHealthProbes(pod)
 
 	wh.meshCatalog.ExpectProxy(cn)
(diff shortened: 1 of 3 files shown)


pkg/injector/patch.go
```diff
@@ -19,12 +20,16 @@ import (
 func (wh *mutatingWebhook) createPatch(pod *corev1.Pod, req *v1beta1.AdmissionRequest, proxyUUID uuid.UUID) ([]byte, error) {
 	namespace := req.Namespace
 
+	// Tracks the success of the current certificate issued for envoy to connect to XDS
+	success := true
```
shashankram, line 24: Could you rename this to `certIssuanceSuccess`, otherwise in the context of the the `createPatch` function, the existing variable name can be confusing.

pkg/injector/patch.go
```diff
@@ -19,12 +20,16 @@ import (
...
 	// Issue a certificate for the proxy sidecar - used for Envoy to connect to XDS (not Envoy-to-Envoy connections)
 	cn := catalog.NewCertCommonNameWithProxyID(proxyUUID, pod.Spec.ServiceAccountName, namespace)
+	defer certXdsIssueTimeTrack(time.Now(), cn.String(), &success)
```
shashankram, line 27: Isn't this going to lead to incorrect time tracking because of the defer statement? The time taken to issue the cert is not the same as the time taken for this function to complete. 

The webhook time tracking is done a bit differently because it calculates the entire time for the entrypoint function to execute, but in this case we should be calculating the time immediately before and after `IssueCertificate()` is invoked.

snehachhabria: That's right, I missed updating the logic. My latest commit has the necessary changes

pkg/injector/profiling.go
```diff
@@ -41,3 +41,13 @@ func webhookTimeTrack(start time.Time, timeout time.Duration, success *bool) {
 	metricsstore.DefaultMetricsStore.InjectorRqTime.
 		WithLabelValues(fmt.Sprintf("%t", *success)).Observe(elapsed.Seconds())
 }
+
+// Time tracking function for xds certificate issuance.
+// Will report the time taken to issue an xds certificate
+func certXdsIssueTimeTrack(elapsed time.Duration, commonName string, success *bool) {
+	log.Debug().Msgf("Time taken to issue xds certificate %s is %v",
```
shashankram, line 48: Lets remove this log, it seems a bit unnecessary if we have the metric being exported already. There is ongoing discussion on not leaking cert details such as CN. 
Can we remove this function and just update the counter in L36 of patch.go?

eduser25 (approved):

pkg/injector/profiling.go
eduser25: or `trace`, IMO

snehachhabria: updated

pkg/injector/profiling.go
```diff
@@ -41,3 +41,13 @@ func webhookTimeTrack(start time.Time, timeout time.Duration, success *bool) {
 	metricsstore.DefaultMetricsStore.InjectorRqTime.
 		WithLabelValues(fmt.Sprintf("%t", *success)).Observe(elapsed.Seconds())
 }
+
+// Time tracking function for xds certificate issuance.
+// Will report the time taken to issue an xds certificate
+func certXdsIssueTimeTrack(elapsed time.Duration, commonName string, success *bool) {
```
shashankram, line 47: We don't need this anymore right?

shashankram (approved):
Looks great!

pkg/injector/profiling.go
snehachhabria: I had a typ in my commit and then it didn't get pushed. Take care of

PR #2308, opened 2021-01-15
debugger: refactor httpserver and debugger

eduser25:
- Debugger specific code, logic and tests that were living in
`httpserver` have been entirely moved to `debugger` package.
- Some specific code related to metrics, probes and versioning that
was also living in `httpserver` has been moved out in order to make `httpserver`
reusable in both osm-controller and debugger HTTP server instances.
- `Debugger` package now manages its own HTTP server, specified and opened
on `constants.DebugPort` port.
- Updated E2E DebugServer to repeat few iterations of the on/off scenario.

**Affected area**:

- Control Plane          [X]

No


cmd/osm-controller/osm-controller.go
```diff
@@ -245,22 +246,34 @@ func main() {
...
+	// Health/Liveness probes
+	funcProbes := []health.Probes{xdsServer}
+	httpServer.AddHandlers(map[string]http.Handler{
+		"/health/ready": health.ReadinessHandler(funcProbes, getHTTPHealthProbes()),
+		"/health/alive": health.LivenessHandler(funcProbes, getHTTPHealthProbes()),
+	})
+	// Metrics
+	httpServer.AddHandler("/metrics", metricsstore.DefaultMetricsStore.Handler())
+	// Version
+	httpServer.AddHandler("/version", version.GetVersionHandler())
```
shashankram, line 265: Can we move this to a separate func `initializeHTTPServerHandlers()` or something to keep `main()` concise.

cmd/osm-controller/osm-controller.go
```diff
@@ -245,22 +246,34 @@ func main() {
...
+	if err != nil {
+		log.Fatal().Err(err).Msgf("Failed to start OSM metrics/probes HTTP server")
+	}
+
+	// Create DebugServer and start its config event listener.
+	// Listener takes care to start and stop the debug server as appropriate
 	debugConfig := debugger.NewDebugConfig(certDebugger, xdsServer, meshCatalog, kubeConfig, kubeClient, cfg, kubernetesClient)
-	debugServerInterface := httpserver.NewDebugHTTPServer(debugConfig, constants.DebugPort)
-	httpserver.RegisterDebugServer(debugServerInterface, cfg)
+	debugConfig.StartDebugServerConfigListener()
```
shashankram, line 276: To make this consistent with HttpServer init, I would redo this code block as follows:
```
debugServer := httpserver.NewHTTPServer(constants.DebugPort)
debugger.Start(debugServer, debugConfig)
```

pkg/debugger/types.go
```diff
@@ -60,9 +59,3 @@ type XDSDebugger interface {
 	// GetXDSLog returns a log of the XDS responses sent to Envoy proxies.
 	GetXDSLog() *map[certificate.CommonName]map[envoy.TypeURI][]time.Time
 }
-
-// DebugConfig is the interface of the debug config for debug HTTP server
-type DebugConfig interface {
-	// GetHandlers returns the HTTP handlers available for the debug server.
-	GetHandlers() map[string]http.Handler
-}
```
shashankram, line 68: Thanks for removing this, +1

pkg/httpserver/httpserver.go
```diff
@@ -24,71 +20,85 @@ var (
...
 		},
+		httpServeMux: serverMux,
+		port:         port,
+		stopSyncChan: make(chan struct{}),
 	}
 }
 
-// Start runs the Serve operations for the http.server on a separate go routine context
-func (s *HTTPServer) Start() {
+// AddHandler adds a handlers to the httpHandlers map
```
shashankram, line 46:
```suggestion
// AddHandler adds an HTTP handlers for the given path on the HTTPServer
```

pkg/httpserver/httpserver.go
```diff
@@ -24,71 +20,85 @@ var (
...
+// For changes to be effective, server requires restart
+func (s *HTTPServer) AddHandler(url string, handler http.Handler) {
```
shashankram, line 48: s/urs/path ?

tests/e2e/e2e_debug_server_test.go
```diff
@@ -57,36 +58,40 @@ var _ = OSMDescribe("Test Debug Server by toggling enableDebugServer",
 					Destination: controllerDest,
 				}
 
-				By("Ensuring debug server is available when enableDebugServer is enabled")
+				iterations := 3
```
shashankram, line 61: 2 should be enough here.

shashankram (approved):

cmd/osm-controller/osm-controller.go
eduser25: I originally did this but, but passing xdsServer was kind of ugly.  I can do it again though.

cmd/osm-controller/osm-controller.go
eduser25: only to make sure, this is not an httpserver. DebugConfig has the httpserver inside.
Starting or not the httpserver is up to DebugConfig, and should not really be done here, if I understood your suggestion properly.

shashankram: > `DebugConfig has the httpserver inside`

This logically does not make much sense, for a config to have a server inside? This is why I am suggesting to redo this part a bit differently.

cmd/osm-controller/osm-controller.go
shashankram: this can be fixed later as well.

cmd/osm-controller/osm-controller.go
eduser25: starting or not the server will depend on the config, you will need to have a handle to the server to issue the calls.
this call starts the `pkg/debugger/debugger_config_listener.go` function which monitors the configmap for changes on `enable_debug_server`

I'll listen to suggestions if you have other ideas, ofc

shashankram: I get that the config is required, but I think it would be more appropriate encapsulate the debugConfig within the debugServer instead of the other way around.

tests/e2e/e2e_debug_server_test.go
eduser25, line 61:
```suggestion
				iterations := 2
```

snehachhabria (approved):

PR #2324, opened 2021-01-19
metrics: add metrics for pod and namespace count

shashankram:
**Description**:
Adds metrics to report the number of monitored
namespaces and pods participating in the mesh.

Part of #902

**Affected area**:

- New Functionality      [ ]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [ ]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [X]
- SMI Policy             [ ]
- Security               [ ]
- Tests                  [ ]
- CI System              [ ]
- Performance            [ ]
- Other                  [ ]

`No`


draychev (approved):

ksubrmnn (approved):
This looks good.

In the case of pods, how do we consider if we delete the mesh and the pods still have their sidecars. Then a new mesh is created and the namespace is added again. How do those old pods get counted here?

shashankram:
> This looks good.
> 
> In the case of pods, how do we consider if we delete the mesh and the pods still have their sidecars. Then a new mesh is created and the namespace is added again. How do those old pods get counted here?

The metrics are pertaining to a particular instance of osm-controller. If we delete the mesh (osm-controller), and a new mesh (osm-controller) is created, the controller will receive the same events as the previous controller when its informer caches are initialized. This means metrics for namespaces and pods that were previously part of the mesh, will show the same count before and after they are associated with a different controller. This can be tested by restarting the osm-controller pod as an example.

snehachhabria (approved):

Issue #2151, opened 2020-12-07
Document how to upgrade OSM


nojnhuh:
Document the process for upgrading an OSM control plane in a new file under `docs/` called something like `upgrade_guide.md`.

It should contain:
- a description of what an upgrade does
- our policy for what upgrade paths are supported and tested
- steps to perform an upgrade

PR #2340, opened 2021-01-21
docs(upgrade): Add docs for helm upgrade

ksubrmnn:

**Description**:
This PR adds a script to clean up CRDs. This script is then referenced in upgrade documentation.

The docs only refer to the helm CLI for now. osm CLI upgrade support has yet to be implemented.

Partially addresses #2151

**Affected area**:

- New Functionality      [ ]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [X]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests                  [ ]
- CI System              [ ]
- Performance            [ ]
- Other                  [ ]

No


docs/upgrade_guide.md
```diff
@@ -0,0 +1,30 @@
...
+## Prerequisites
+- Kubernetes cluster with OSM installed
+- The [helm 3 CLI](https://helm.sh/docs/intro/install/)
+
+## Upgrade OSM using Helm
+Use the `helm` CLI to upgrade the OSM control plane from a Kubernetes cluster.
+
+Run `helm upgrade`. 
+```console
+# Add steps for helm upgrade
```
ksubrmnn, line 14: @michelleN would love to pair with you on this. I'm not super familiar with the helm commands to get the chart from openservicemesh.github.io/osm via helm CLI

docs/upgrade_guide.md
```diff
@@ -0,0 +1,30 @@
+# Upgrade Guide
+
+This guide describes how to upgrade Open Service Mesh (OSM).
```
michelleN, line 3: ` upgrade Open Service Mesh (OSM) control plane` might make it clearer

docs/upgrade_guide.md
```diff
@@ -0,0 +1,30 @@
...
+
+## Prerequisites
+- Kubernetes cluster with OSM installed
```
michelleN, line 6: `with OSM control plane` installed

docs/upgrade_guide.md
```diff
@@ -0,0 +1,30 @@
...
+- The [helm 3 CLI](https://helm.sh/docs/intro/install/)
```
michelleN, line 7: It might be worth explaining here why we're going through helm to do the upgrade instead of the `osm` CLI

docs/upgrade_guide.md
```diff
@@ -0,0 +1,30 @@
...
+
+Run `helm upgrade --help` for more options.
+
+## CRD Upgrades
+Please check the `CRD Updates` section of the [release notes](https://github.com/openservicemesh/osm/releases) to see if additional steps are required to update the CRDs used by OSM. If the new release does contain updates to the CRDs, it is required to first delete existing CRDs and the associated Custom Resources prior to upgrading.
+
+In the `./scripts/cleanup` directory we have included a helper script to delete those CRDs and Custom Resources: `./scripts/osm-cleanup.sh`
+
+## Upgrade support
+Upgrades are supported between consecutive minor versions. Skipping minor versions is not supported.
```
michelleN, line 25: I'm unclear on what supported means here since we don't yet have a CLI command that does the upgrade. Let's talk more about this. Will think about it for a bit.

docs/upgrade_guide.md
```diff
@@ -0,0 +1,30 @@
...
+
+Run `helm upgrade`. 
+```console
+# Add steps for helm upgrade
+```
+
+Run `helm upgrade --help` for more options.
+
+## CRD Upgrades
+Please check the `CRD Updates` section of the [release notes](https://github.com/openservicemesh/osm/releases) to see if additional steps are required to update the CRDs used by OSM. If the new release does contain updates to the CRDs, it is required to first delete existing CRDs and the associated Custom Resources prior to upgrading.
```
michelleN, line 20: 👍 
May want to also add that the end user will also have to recreate the CRDs and CRs once the control plane upgrade is complete

docs/upgrade_guide.md
```diff
@@ -0,0 +1,30 @@
+# Upgrade Guide
+
+This guide describes how to upgrade Open Service Mesh (OSM).
+
```
michelleN, line 4: We should add somewhere at the top that control plane upgrades are disruptive. It may make sense for an end user to mirror their environment on a different cluster, re-direct traffic to the mirrored environment before considering a control plane upgrade (especially one that involves CRDs)

docs/upgrade_guide.md
ksubrmnn: Yes I'd love to generate more of a discussion around this. To me, support means that we have validated that between minor releases, a helm upgrade will work smoothly (barring the CRD stuff). 

But I also believe we are kind of still in an "alpha" stage so can we really guarantee that? I'm not sure how to anticipate what kinds of change would result in a breaking release

docs/upgrade_guide.md
ksubrmnn: Moved that section up towards the top

docs/upgrade_guide.md
ksubrmnn: I've removed this section for now. We can add it back once we've had more time to discuss and when we add the osm cli upgrade command

docs/upgrade_guide.md
```diff
@@ -0,0 +1,40 @@
...
+Since upgrades include redeploying the osm-controller with the new version, there may be some downtime of the controller. This means that there will be a delay in processing new SMI resources. 
+
+However, already existing SMI resources will be unaffected (assuming [CRD Upgrades](#CRD-Upgrades) are not needed). This means that the data plane (which includes the Envoy sidecar configs) will also be unaffected by upgrading. 
+
+Data plane interruptions are expected if the upgrade includes CRD changes.
+
+## CRD Upgrades
+Please check the `CRD Updates` section of the [release notes](https://github.com/openservicemesh/osm/releases) to see if additional steps are required to update the CRDs used by OSM. If the new release does contain updates to the CRDs, it is required to first delete existing CRDs and the associated Custom Resources prior to upgrading.
+
+In the `./scripts/cleanup` directory we have included a helper script to delete those CRDs and Custom Resources: `./scripts/crd-cleanup.sh`
```
nojnhuh, line 20:
```suggestion
In the `./scripts/cleanup` directory we have included a helper script to delete those CRDs and Custom Resources: `./scripts/cleanup/crd-cleanup.sh`
```

docs/upgrade_guide.md
```diff
@@ -0,0 +1,40 @@
...
+## Upgrade OSM using Helm
+Use the `helm` CLI to upgrade the OSM control plane.
+
+### Preserve ConfigMap
+To preserve any edits you've made to the ConfigMap, use the `helm --set` flag. Find the corresponding field in the [values](https://github.com/openservicemesh/osm/blob/main/charts/osm/values.yaml) file and set the desired value.
+
+For example, to keep the `envoy_log_level` field in the ConfigMap set to `info`:
+
+```console
+$ helm upgrade <mesh name> osm --repo https://openservicemesh.github.io/osm --version <chart version> --namespace <osm namespace> --reuse-values --set OpenServiceMesh.envoyLogLevel=info
```
nojnhuh, line 36: Looks like `--reuse-values` will cause the control plane images not to update unless you explicitly `--set OpenServiceMesh.image.tag` to the newer version. `--reuse-values` would be nice to carry over all the other settings, but it seems like the images should upgrade by default. Maybe a safer option is what you had mentioned offline and have users create an override.yaml with only the values they want to override and pass that with `helm upgrade -f override.yaml` instead of relying on `--reuse-values`.

ksubrmnn: Thanks for catching that! I was just looking to see if the helm chart updated when I was testing. Using an override file is the safest way, but it is cumbersome for the user. However I think it's important that they don't make the same mistake I did. I'll update these docs to use an override file, instead of --set

docs/upgrade_guide.md
```diff
@@ -0,0 +1,46 @@
...
+
+After upgrading, the Custom Resources will need to be recreated using the updated CRDs installed by the upgrade.
+
+## OSM Configuration
+When upgrading, any custom settings used to install or run OSM may be reverted to the default. This includes any metrics deployments and any changes to the OSM ConfigMap. Please ensure that you carefully follow the guide to prevent these values from being overwritten.
+
+## Upgrade OSM using Helm
+Use the `helm` CLI to upgrade the OSM control plane.
+
+### Preserve OSM configuration
```
michelleN, line 30: Have you gotten a chance to play with the `--reuse-values` flag with `helm upgrade` ? It preserves the last release's values and merge in any overrides from the command line via --set and -f.

Making the command:
`helm upgrade osm osm --reuse-values --repo https://openservicemesh.github.io/osm --version <chart version> --namespace <osm namespace>

docs/upgrade_guide.md
ksubrmnn: @michelleN see this discussion regarding --reuse-values and how it's not super compatible since we use the image in the values file

docs/upgrade_guide.md
michelleN: sounds good

michelleN (approved):

nojnhuh (approved):

PR #2343, opened 2021-01-21
injector/init: generate init-container commands within osm-controller

shashankram:
**Description**:
This change generates the iptables commands within osm-controller
instead of running it as a script. The benefit in doing so is that
the controller can programmatically tweak iptables at the time
of sidecar injection, ex. to disable proxy admin port access,
or exclude certain control plane traffic from being intercepted
in managed environments.
Additionally, it prevents potential bugs due to code duplication
of port numbers used in iptable rules and within the controller.

Currently the iptable rules are less enough to be a part of the
init-container command spec. If the need arises, a separate
go binary can be used to program iptables.

Part of #2243

**Affected area**:

- New Functionality      [ ]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [ ]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [X]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests                  [ ]
- CI System              [ ]
- Performance            [ ]
- Other                  [ ]

`No`


Makefile
```diff
@@ -127,7 +127,7 @@ $(DOCKER_DEMO_TARGETS):
 	docker build -t $(CTR_REGISTRY)/$(NAME):$(CTR_TAG) -f dockerfiles/Dockerfile.$(NAME) demo/bin/$(NAME)
 
 docker-build-init:
-	docker build -t $(CTR_REGISTRY)/init:$(CTR_TAG) -f dockerfiles/Dockerfile.init init
+	docker build -t $(CTR_REGISTRY)/init:$(CTR_TAG) -f dockerfiles/Dockerfile.init .
```
nojnhuh, line 130: If no context is needed, you can do `docker build -` to signal that. That way no time is wasted copying the whole repo and any locally built binaries into the build context.
```suggestion
	docker build -t $(CTR_REGISTRY)/init:$(CTR_TAG) - < dockerfiles/Dockerfile.init
```
https://docs.docker.com/engine/reference/commandline/build/#build-with--

shashankram: Didn't know, thanks!

ksubrmnn (approved):

snehachhabria (approved):

PR #2446, opened 2021-02-03
docs(troubleshoot): Move upgrade troubleshooting guide

ksubrmnn:

**Description**:
Moves upgrade troubleshooting guide to the new docs directory

**Affected area**:

- New Functionality      [ ]
- Documentation          [X]
- Install                [ ]
- Control Plane          [ ]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests                  [ ]
- CI System              [ ]
- Performance            [ ]
- Other                  [ ]

No

### DIFF
--- a/docs/content/docs/troubleshooting/upgrade.md
+++ b/docs/content/docs/troubleshooting/upgrade.md
@@ -1,3 +1,9 @@
+---
+title: "Upgrade Troubleshooting"
+description: "OSM Upgrade Troubleshooting Guide"
+type: docs
+---
+
 # OSM Upgrade Troubleshooting Guide
 
 ## Server could not find requested resource
--- a/docs/content/docs/troubleshooting/_index.md
+++ b/docs/content/docs/troubleshooting/_index.md
@@ -2,4 +2,5 @@
 title: "Troubleshooting"
 description: "Docs related to troubleshooting issues."
 type: docs
+aliases: ["troubleshooting"]
 ---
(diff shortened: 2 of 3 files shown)


michelleN (approved):

docs/content/docs/troubleshooting/upgrade.md
```diff
@@ -1,3 +1,9 @@
+---
+title: "Upgrade Troubleshooting"
+description: "OSM Upgrade Troubleshooting Guide"
+type: docs
```
michelleN, line 4: Do you need `aliases: ["troubleshooting"]` in this section?

flynnduism: 👍 

Good idea to make the previous url point to the new one.

michelleN: @flynnduism - could you break down what aliases are for us? Are these the tags that folks can use for search purposes in the UI?

michelleN (approved):

docs/content/docs/troubleshooting/upgrade.md
flynnduism: Sure thing! [Aliases](https://gohugo.io/content-management/urls/#aliases) are Hugo's way to handle content redirects.

So here, the upgrade page exists at `/troubleshooting/upgrade`. If the page had previously had a diffferent url structure (like `troubleshooting-upgrades` or something) the addition of an alias would create additional pages in thge Hugo site that redirect to this page.

```
alias: ["troubleshooting-upgrades"]
```

Taking another look at this PR I see the page is being moved into the docs structure, but there isn't really a path change so I don't think an alias is needed in this instance

SanyaKochhar (approved):

PR #2478, opened 2021-02-08
fix(cds): return local cluster endpoints per  port

michelleN:
**Description**:
The `getLocalServiceCluster` function was returning an xds cluster with endpoints based on Kubernetes service endpoints when it should have been building local cluster endpoints based on the port specified in the service spec. i.e. 0.0.0.0:<port>
As a result, there were 18 extra, duplicate  entries in the envoy local clusters that were being programmed per Kubernetes service endpoints (or Pod replicas)

This PR fixes the getLocalServiceCluster to return a local cluster with endpoints based on the Kubernetes ports rather than the Kubernetes Service endpoints. The tests in the clusters_test.go have been converted from gingko to use the go testing framework and tests for getLocalServiceCluster has been added.

To see the issue exposed in a test, please pull down this branch and checkout the commit sha 9f952062e652ba66d4fb2cc46cb663217bef5088 and run the tests. You can also check out that commit, run the demo, play with the replica count on bookbuyer and inspect the clusters in the envoy admin UI to see the duplicate entries.

**Affected area**:

- New Functionality      [ ]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [X ]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests                  [ ]
- CI System              [ ]
- Performance            [X ]
- Other                  [ ]

### DIFF
--- a/pkg/envoy/cds/cluster.go
+++ b/pkg/envoy/cds/cluster.go
@@ -7,7 +7,6 @@ import (
 	xds_core "github.com/envoyproxy/go-control-plane/envoy/config/core/v3"
 	xds_endpoint "github.com/envoyproxy/go-control-plane/envoy/config/endpoint/v3"
 	"github.com/envoyproxy/go-control-plane/pkg/wellknown"
-
 	"github.com/golang/protobuf/ptypes"
 	"github.com/golang/protobuf/ptypes/wrappers"
 
@@ -73,6 +72,17 @@ func getOutboundPassthroughCluster() *xds_cluster.Cluster {
 	}
 }
 
+// getSyntheticCluster returns a static cluster with no endpoints
+func getSyntheticCluster(name string) *xds_cluster.Cluster {
+	return &xds_cluster.Cluster{
+		Name: name,
+		ClusterDiscoveryType: &xds_cluster.Cluster_Type{
+			Type: xds_cluster.Cluster_STATIC,
+		},
+		LbPolicy: xds_cluster.Cluster_ROUND_ROBIN,
+	}
+}
+
 // getLocalServiceCluster returns an Envoy Cluster corresponding to the local service
 func getLocalServiceCluster(catalog catalog.MeshCataloger, proxyServiceName service.MeshService, clusterName string) (*xds_cluster.Cluster, error) {
 	xdsCluster := xds_cluster.Cluster{
@@ -97,13 +107,13 @@ func getLocalServiceCluster(catalog catalog.MeshCataloger, proxyServiceName serv
 		Http2ProtocolOptions: &xds_core.Http2ProtocolOptions{},
 	}
 
-	endpoints, err := catalog.ListEndpointsForService(proxyServiceName)
+	ports, err := catalog.GetTargetPortToProtocolMappingForService(proxyServiceName)
 	if err != nil {
-		log.Error().Err(err).Msgf("Failed to get endpoints for service %s", proxyServiceName)
+		log.Error().Err(err).Msgf("Failed to get ports for service %s", proxyServiceName)
 		return nil, err
 	}
 
-	for _, ep := range endpoints {
+	for port := range ports {
 		localityEndpoint := &xds_endpoint.LocalityLbEndpoints{
 			Locality: &xds_core.Locality{
 				Zone: "zone",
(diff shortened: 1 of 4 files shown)


pkg/envoy/cds/cluster.go
```diff
@@ -97,21 +96,21 @@ func getLocalServiceCluster(catalog catalog.MeshCataloger, proxyServiceName serv
 		Http2ProtocolOptions: &xds_core.Http2ProtocolOptions{},
 	}
 
-	endpoints, err := catalog.ListEndpointsForService(proxyServiceName)
+	ports, err := catalog.GetPortToProtocolMappingForService(proxyServiceName)
```
shashankram, line 99: Ideally we need to be using `service.spec.ports[].targetPort` here. We can only use `service.spec.ports[].port` if it is the same as the `targetPort`. Since we are attempting to create local clusters for the root service in a traffic split.

I would change this to the following:
```
if cluster == root service in split:
  ports = GetPortToProtocolMappingForService
else:
  ports = GetTargetPortToProtocolMappingForService
```

This will ensure that we aren't setting up the local clusters with incorrect port numbers for cases where the targetPort is different from the port for a service.

shashankram (changes requested):
Thanks for the change.

While the existing usage of `GetEndpointsForService` is not is not correct, use `port` instead of `targetPort` to program local clusters introduces new issues.

shashankram (approved):
This looks good. I am guessing this will need to be changed again when there is a local cluster being created for the root service referenced in an SMI traffic split policy, which does not have endpoints (ex. bookstore in our demo).

pkg/envoy/cds/cluster.go
```diff
@@ -97,21 +96,25 @@ func getLocalServiceCluster(catalog catalog.MeshCataloger, proxyServiceName serv
...
+	ports, err := catalog.GetTargetPortToProtocolMappingForService(proxyServiceName)
 	if err != nil {
-		log.Error().Err(err).Msgf("Failed to get endpoints for service %s", proxyServiceName)
-		return nil, err
+		if proxyServiceName.SyntheticTargetPort != 0 {
+			ports = map[uint32]string{proxyServiceName.SyntheticTargetPort: ""}
+		} else {
+			log.Error().Err(err).Msgf("Failed to get ports for service %s", proxyServiceName)
+			return nil, err
+		}
```
shashankram, line 106: Instead of handling synthetic services, I don't see a reason why we need to create local clusters for services that do not really exist. 

Can we just check if the service name matches the syntax of a synthetic service and skip creating local clusters instead?

shashankram (dismissed):
I don't think we should be creating local clusters for synthetic services. Considering this is just an emulated service in the controller, we should be able to simply skip creating local clusters if the service name matches the naming syntax for the synthetic service.

pkg/envoy/cds/cluster.go
michelleN: If we skip creating a local cluster, the tcp e2e fails where there is no service for the client

pkg/catalog/xds_certificates.go
```diff
@@ -68,6 +68,14 @@ func makeSyntheticServiceForPod(pod *v1.Pod, proxyCommonName certificate.CommonN
...
 	}
 	syntheticService := svcAccount.GetSyntheticService()
+	for _, con := range pod.Spec.Containers {
+		if con.Name != constants.EnvoyContainerName {
+			if len(con.Ports) > 0 {
+				syntheticService.SyntheticTargetPort = uint32(con.Ports[0].ContainerPort)
+				break
+			}
+		}
+	}
```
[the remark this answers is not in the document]
michelleN: agreed. I didn't want to go this route but I did want to see what could potentially work.

pkg/envoy/cds/cluster.go
shashankram: Ah, the classic RDS issue which #2064 is blocked on.

shashankram (approved):
Looks great!

snehachhabria (approved):

Issue #984, opened 2020-07-02
Extend Envoy to get custom metrics needed to implement smi metrics


michelleN:

nojnhuh:
This is largely complete, but may need small changes as we continue with the SMI metrics implementation.

nojnhuh:
Just remembered @eduser25 is still working on the request duration metrics. Reopening to finish tracking that.

nojnhuh:
Now this is basically complete. Only waiting on upstream Envoy to fully support WASM filters before merging. All code is on my fork: https://github.com/nojnhuh/osm/tree/envoy-wasm

nojnhuh:
Reopening until this is merged.

michelleN:
Now that the wasm work has been merged into Envoy, the next steps should be:
1. Wait for Envoy 1.17 release 
2. Update Envoy to 1.17 in osm
3. merge the wasm filter work @nojnhuh has done into osm
4. Cut a release
5. Either create a new repo under openservicemesh to host the smi-metrics implementation OR merge smi-metrics implementation into the servicemeshinterface/smi-metrics repo (or both)
6. update docs

probably need to discuss step 3 a little bit more. What do u think about these next steps ^ @nojnhuh ?

nojnhuh:
@michelleN That's what I was thinking of at a high level. I agree that we should think through how we want to merge everything. I was planning on splitting my OSM branch into a reasonable number of smaller commits and making each one of those a PR to the main branch. There's a lot of junk commits on my branch currently from when I was still trying to figure things out. I haven't gone through the exercise of defining what each of those PRs should look like yet.

Side note: Since v1.17 of Envoy isn't going to be released until well after our v0.6.0, should we clear the milestone on this for now?

nojnhuh:
Taking this off the v0.6.0 milestone since WASM will likely not GA in Envoy until early 2021 with v1.17.0.

nojnhuh:
DONE with #2616 :tada:

PR #2488, opened 2021-02-08
feat(metrics): add WASM metrics source

nojnhuh:
**Description**:
This change adds the source and compiled binary for a WASM-based Envoy
extension that will generate additional statistics needed to implement
SMI metrics. Specific build instructions and docs will be added in a
subsequent change.

---

* (#2479) ~~ref(injector): pass pod object to getEnvoySidecarContainerSpec~~
* (YOU ARE HERE) **feat(metrics): add WASM metrics source**
* docs(metrics): Add docs for WASM stats
* feat(metrics): Add WASM feature flag
* feat(metrics): Add stats.wasm to workload pods
* feat(injector): Add name, workload name, workload kind to PodMetadata
* feat(metrics): Add source labels to WASM metrics
* feat(metrics): Add dest labels to WASM metrics
* feat(metrics): Add "unknown" for dest labels on local replies
* feat(metrics): clean up WASM metrics in Prometheus config
* feat(metrics): add flag to enable WASM metrics
* tests(e2e): add WASM metrics test

The full set of changes can be found here: https://github.com/openservicemesh/osm/compare/main...nojnhuh:wasm-metrics

**Affected area**:

- New Functionality      [ ]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [ ]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [X]
- SMI Policy             [ ]
- Security               [ ]
- Tests                  [ ]
- CI System              [ ]
- Performance            [ ]
- Other                  [ ]

No


wasm/stats.cc
```diff
@@ -0,0 +1,125 @@
...
+}
+
+using RqTotalCounter = Counter<std::string, std::string, std::string, std::string, std::string, std::string, std::string, std::string, std::string>;
+using RqDurationHist = Histogram<std::string, std::string, std::string, std::string, std::string, std::string, std::string, std::string>;
+
+class StatsContext : public Context
+{
+public:
+  explicit StatsContext(uint32_t id, RootContext *root) : Context(id, root),
+                                                          rq_total(RqTotalCounter::New("osm_request_total",
```
eduser25, line 20: I guess this will allow us to not tag the metrics at prometheus scrape time, as most of it will come from these.
Let me know when to test this, with so many labels with possible high cardinality values, i suspect this will choke prometheus to a rather fast yet painful death.

wasm/Makefile
```diff
@@ -0,0 +1,9 @@
+# Current build against rev 956f0d500c380cc1656a2d861b7ee12c2515a664 of https://github.com/proxy-wasm/proxy-wasm-cpp-sdk
+# Docker image: nojnhuh/proxy-wasm-cpp-sdk:956f0d500c380cc1656a2d861b7ee12c2515a664
```
shashankram, line 2: Should we be pushing this image to `openservicemesh` docker registry?

wasm/stats.cc
nojnhuh: I suppose we could add the `source_*` labels at scrape time since we're already only generating these metrics at the source. Not sure how we could work around the destination ones though, which are probably the bigger issue.

wasm/Makefile
nojnhuh: Makes sense. I'll re-tag that image.

wasm/stats.cc
```diff
@@ -0,0 +1,125 @@
+#include <string>
```
shashankram, line 1: I thought I left this comment but might have forgotten.

Does it make sense to move this file under `envoy/wasm/src/` and the precompiled binary under `envoy/wasm/bin/`?

nojnhuh: Right now it's easier to keep the source and binary in the same place. Once we have the build integrated with `make docker-build-osm-controller` then splitting up the source and binary would be easier. Until then, keeping them separate would just require more manual steps.

nojnhuh: Actually having the prebuilt docker image was the only blocker to integrating the build. I'll remove the compiled binary from this PR and have the build place it directly alongside the osm-controller binary in a future PR.

shashankram (approved):

eduser25 (approved):

Issue #2453, opened 2021-02-03
Docs contain links that are either broken in Github or the website


ksubrmnn:
**Please describe what should be documented
Various links in the repo are broken either in the website or on Github. We should standardize how we link to docs to ensure they work both in the repo and on the site.

Links should:

1. Maintain the tree when switching between docs in the Github site
2. Maintain the version when switching between docs on the website
3. Work for both the site and the Github repo

**Please suggest where in the repo the document should be located

ksubrmnn:
From @flynnduism:
"
# External Links 
Links to things outside of the Hugo Docs footprint need to be absolute - so those scripts, demo assets, various `gofoo.go` project files; these need absolute links so that they work when clicked outside of Github.com
 
 
# Internal Links

This is a tricky one - in past projects we've considered prioritizing one resource as the source of truth and ensuring content works in one at the expense of the other (for users of the hugo website or those reading on github.com). I still consider this to be the most concrete approach.

# Versioning
The release process should see new versions of the docs spun off (each git branch auto deploys as a subdomain) so there would be corresponding versions (going forward).
 
Currenty these exist:
 - docs.openservicemesh.io [Prod / main]	
- release-v0-7--osm-docs.netlify.app [0.7 branch]
 

Another approach was to run a rewrite task that would recreate the urls for Hugo during the site build, so that both scenarios worked without having to compromised - but that approach was complex, buggy and had regular pipeline issues.

flynnduism:
I've been looking through the approaches [Github](https://docs.github.com/en/github/creating-cloning-and-archiving-repositories/about-readmes) and [Hugo](https://gohugo.io/content-management/cross-references/) take to relative document links, to find a way for both to render working urls. 

## Testing 

Here I've spun off a test page with different linking methods, and deployed it for comparison.

:octocat: [Github.com result](https://github.com/flynnduism/osm/blob/docs-rel-links-testing/docs/content/docs/installation_guide.md)
🌐  [Hugo website result](https://vigorous-golick-cbf474.netlify.app/docs/installation_guide/)

```
* A [internal link](../development_guide) to the development guide
* B [internal link](../development_guide.md) to the development guide
* C [internal link](development_guide.md) to the development guide
* D [internal link](../patterns/ingress/) to ingress guide (in another directory)
* E [internal link]({{< ref "development_guide.md" >}} "development guide") to the development guide
* F [internal link]({{< ref "development_guide" >}} "development guide") to the development guide
[...]

PR #2497, opened 2021-02-09
docs: update website readme

flynnduism:
This PR adds notes to the `docs/Readme.md` regarding:

* best practices for creating links, in both Github and the Hugo website (as per #2453)  
* includes the v0.7 [release docs](https://release-v0-7--osm-docs.netlify.app/) as an example in the versioning section
* includes the installation of npm packages in steps to run site locally

---

**Affected area**:

- New Functionality      [ ]
- Documentation          [x]
- Install                [ ]
- Control Plane          [ ]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests                  [ ]
- CI System              [ ]
- Performance            [ ]
- Other                  [ ]

No

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -1,9 +1,8 @@
 # Open Service Mesh Docs
 
-> :book: OSM Docs are found within this [section](https://github.com/openservicemesh/osm/tree/main/docs/content)
-> :ship: Also the config to generate them as docs.openservicemesh.io
- 
-> :link: Looking for the main OSM website? Visit [osm-www](https://github.com/openservicemesh/osm-www).
+> :book: This section contains the [OSM Docs](https://github.com/openservicemesh/osm/tree/main/docs/content)  
+> :ship: Also the website config to generate [docs.openservicemesh.io](docs.openservicemesh.io)  
+> :link: Looking for the main OSM website? Visit [osm-www](https://github.com/openservicemesh/osm-www)  
 
 
 ## Editing Content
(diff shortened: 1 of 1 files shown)


nojnhuh (approved):

docs/README.md
```diff
@@ -27,6 +26,63 @@ type: docs
...
+Example:
+
+```
+Take a look at [the following test](../pkg/configurator/client_test.go)
+```
+
+Should use the a full, absolute path:
+
+```
+Take a look at [the following test](https://github.com/openservicemesh/osm/blob/release-v0.2/pkg/configurator/client_test.go)
```
shashankram, line 46: We need some guidance/script on updating URLs with every release. Is this captured in the readme?

flynnduism: > Note that the full path **can** also specify the release

I've worded the Readme in this PR to say `can` rather than `should`... 

My recommendation would be **to not do this** in the current documentation (aka on `main`). It's an extra step that would require regular edits to bump the version and I think would be an annoyance, so I suggest leaving `main` alone and instead to look at including version info later on where it becomes useful:

## Value of Versioned URLs

The advantage of versioned urls is to preserve external links - so that when a doc points to a different part of the OSM repo, it can be frozen in time to avoid causing confusion if that destination changes / moves / is deleted in subsequent releases. Versioned urls do nothing for the current, latest docs except add more maintenance overhead IMO. 

## Set a Version When Cutting a Release

I think versioning makes more sense when a new release is cut, say v0.8; that is when the [Process to Version the Docs Site](https://github.com/openservicemesh/osm/tree/main/docs#versioning-the-docs-site) would come into play.

At that point:

```
0.8 is now prod, aka docs.openservicemesh.io 
0.7 docs are spun off as an archive at https://release-v0-7--osm-docs.netlify.app
```

This is where versioned URLs become useful IMO - if someone is browsing the `release-v0-7` docs because they are still running that version, it's very likely they will hit broken links for anything that has changed on the live site. At the point the v0.7 docs are archival, so that's when urls should be updated:

```
example:

// the docs link to the helm chart as follows:
https://github.com/openservicemesh/osm/blob/main/charts/osm/Chart.yaml

// when an archival doc is being created, the ideal path would be 
https://github.com/openservicemesh/osm/blob/release-v0.7/charts/osm/Chart.yaml
```

i.e. add a commit to the `release-v0-7` branch which rewrites the urls as above.

---

This is just my opinion - what do you think? 

It's a workflow that needs input to find the cleanest process for the team. I plan to help with this process for the next release, so I can run through this and make sure it's carefully documented with the team.

shashankram (approved):

docs/README.md
shashankram: Thanks for the detailed info @flynnduism.

Issue #2415, opened 2021-01-29
Add doc for integrating with BYO Jaeger instance 


ritazh:
**Please describe what should be documented

Similar to the [BYO metrics doc](https://github.com/openservicemesh/osm/blob/v0.6.1/docs/patterns/observability/metrics.md#byo-bring-your-own), add a guide for how to point Envoy pods to your own Jaeger instance. 

**Please suggest where in the repo the document should be located

PR #2499, opened 2021-02-09
docs(tracing): add tracing.md for byo jaeger

SanyaKochhar:

**Description**:

**Affected area**:

- New Functionality      [ ]
- Documentation          [x]
- Install                [ ]
- Control Plane          [ ]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests                  [ ]
- CI System              [ ]
- Performance            [ ]
- Other                  [ ]

References language from https://www.jaegertracing.io/docs/1.21/#about

### DIFF
--- a/docs/content/docs/patterns/observability/tracing.md
+++ b/docs/content/docs/patterns/observability/tracing.md
@@ -1,0 +1,53 @@
+---
+title: "Tracing"
+description: "Tracing"
+type: docs
+---
+
+# Tracing
+Open Service Mesh (OSM) allows optional deployment of Jaeger for tracing. Similarly, tracing can be enabled and customized during installation (`tracing` section in `values.yaml`) or at runtime by editing the `osm-config` ConfigMap. Tracing can be enabled, disabled and configured at any time to support BYO scenarios.
+
+## Jaeger
+[Jaeger](https://www.jaegertracing.io/) is an open source tracing system used for monitoring and troubleshooting distributed systems. It can be deployed with OSM as a new instance or you may bring your own instance. 
+
+### Automatic bring up
+By default, Jaeger deployment and tracing as a whole is disabled.
+
+Jaeger can be automatically deployed by using the `--deploy-jaeger` OSM CLI flag at install time or by toggling the `deployJaeger` value in `values.yaml`.
+
+
+### BYO (Bring-your-own)
+The following section documents the additional steps needed to allow an already running instance of Jaeger to integrate with your OSM control plane.
+> NOTE: This guide outlines steps specifically for Jaeger but you may use your own tracing application instance with applicable values
+
+#### Prerequisites
+* A running Jaeger instance
+
+The sections below outline how to make required updates depending on whether you already have a running instance of OSM or are installing OSM for the first time. In either case, the following `tracing` values in `values.yaml` are being updated to point to your Jaeger instance:
+1. `enable`: set to `true` to tell the Envoy connection manager to send tracing data to a specific address (cluster)
+1. `address`: set to the destination cluster of your Jaeger instance
+1. `port`: set to the destination port for the listener that you intend to use
+1. `endpoint`: set to the destination's API or collector endpoint where the spans will be sent to
+
+
+#### Deploying Jaeger to a running instance of the OSM control plane
+
+If you already have OSM running, `tracing` values must be updated in the OSM ConfigMap using:
+
+```console
+osm mesh upgrade --enable-tracing --tracing-address <tracing server hostname> --tracing-port <tracing server port> --tracing-endpoint <tracing server endpoint>
+```
+> NOTE : This command upgrades an OSM control plane configuration by upgrading the underlying Helm release. Any user-provided configuration could be reverted during future changes made using the OSM CLI so make sure to persist these changes whenever you use OSM CLI for installs or upgrades
+
+You can verify these changes have been deployed by inspecting `osm-config`:
+```console
+kubectl get configmap osm-config -n osm-system -o yaml
+```
+
+#### Deploying Jaeger during OSM deployment
+
+To deploy _your own instance of Jaeger_ during OSM installation, you can use the `--set` flag as shown below to update the values:
+
+```console
+osm install --set OpenServiceMesh.tracing.enable=true,OpenServiceMesh.tracing.address=<tracing server hostname>,OpenServiceMesh.tracing.port=<tracing server port>,OpenServiceMesh.tracing.endpoint=<tracing server endpoint>
+```


docs/content/docs/patterns/observability/tracing.md
```diff
@@ -0,0 +1,51 @@
...
+---
+
+# Tracing
+Open Service Mesh (OSM) allows optional deployment of Jaeger for tracing. 
+
+## Jaeger
+[Jaeger](https://www.jaegertracing.io/) is an open source tracing system used for monitoring and troubleshooting distributed systems. It can be deployed with OSM as a new instance or you may point Envoy pods to your own Jaeger instance. 
+
+### Automatic bring up
+By default, Jaeger is disabled. However, when configured with the `--deploy-jaeger` flag, OSM will install a Jaeger pod and points all Envoy pods to it so that tracing data is port-forwarded to it.
```
eduser25, line 14:
```suggestion
By default, Jaeger deployment and tracing as a whole is disabled.
Jaeger can be automatically deployed through either `values.yaml` or `--deploy-jaeger` osm cli flags, at install time.
Similarly, tracing can be enabled during installation (`tracing` section in `values.yaml`) or at runtime by editting `osm-config-map`. Tracing can be enabled/disabled any time to support BYO scenarios.
```
Adding/clarifying few things

eduser25 (approved):
Good overall! suggested 2 changes to clarify some of the options.

docs/content/docs/patterns/observability/tracing.md
```diff
@@ -0,0 +1,55 @@
...
+
+## Jaeger
+[Jaeger](https://www.jaegertracing.io/) is an open source tracing system used for monitoring and troubleshooting distributed systems. It can be deployed with OSM as a new instance or you may bring your own instance. 
+
+### Automatic bring up
+By default, Jaeger deployment and tracing as a whole is disabled.
+
+Jaeger can be automatically deployed through either by using the `--deploy-jaeger` OSM CLI flag at install time or by toggling the `deployJaeger` value in `values.yaml`.
+
+Similarly, tracing can be enabled and customized during installation (`tracing` section in `values.yaml`) or at runtime by editing `osm-configmap`. Tracing can be enabled, disabled and configured any time to support BYO scenarios.
```
SanyaKochhar, line 18: @eduser25 I've made the edits you suggested and reworded this section a little for clarity. I'm a little confused about whether we're treating Jaeger and tracing as two separate things? It sounds like that in this line

docs/content/docs/patterns/observability/tracing.md
```diff
@@ -0,0 +1,55 @@
...
+# Tracing
+Open Service Mesh (OSM) allows optional deployment of Jaeger for tracing. 
+
+## Jaeger
+[Jaeger](https://www.jaegertracing.io/) is an open source tracing system used for monitoring and troubleshooting distributed systems. It can be deployed with OSM as a new instance or you may bring your own instance. 
+
+### Automatic bring up
+By default, Jaeger deployment and tracing as a whole is disabled.
+
+Jaeger can be automatically deployed through either by using the `--deploy-jaeger` OSM CLI flag at install time or by toggling the `deployJaeger` value in `values.yaml`.
```
nojnhuh, line 16:
```suggestion
Jaeger can be automatically deployed by using the `--deploy-jaeger` OSM CLI flag at install time or by toggling the `deployJaeger` value in `values.yaml`.
```

nojnhuh (changes requested):
This looks like all the right info to include. Left a couple nits and suggestions inline.

nojnhuh (approved):

SanyaKochhar:
Resolves #2415

docs/content/docs/patterns/observability/tracing.md
eduser25: they are separate things.
It is like so as it helps deploy a tracing application that are able to receive in zipkin protocol, not just Jaeger - can just be anything that can understand zipkin proto.

docs/content/docs/patterns/observability/tracing.md
```diff
@@ -0,0 +1,52 @@
...
+1. `port`: set to the destination port for the listener that you intend to use
+1. `endpoint`: set to the destination's API or collector endpoint where the spans will be sent to
+
+
+#### Deploying Jaeger to a running instance of the OSM control plane
+
+If you already have OSM running, `tracing` values must be updated in the OSM ConfigMap using:
+
+```console
+osm mesh upgrade --enable-tracing --tracing-address <tracing server hostname> --tracing-port <tracing server port> --tracing-endpoint <tracing server endpoint>
```
ritazh, line 38: I couldn't find too much docs on `osm mesh upgrade` since it's pretty new. It would be good to add:

> NOTE : This command upgrades an OSM control plane configuration by upgrading the
underlying Helm release.

Can this override user configured changes in the OSM configmap?

docs/content/docs/patterns/observability/tracing.md
```diff
@@ -0,0 +1,52 @@
...
+```console
+kubectl get configmap osm-config -n osm-system -o yaml
+```
+
+#### Deploying Jaeger during OSM deployment
+
+To deploy Jaeger during OSM installation, you can use the `--set` flag as shown below to update the values:
+
+```console
+osm install --set OpenServiceMesh.tracing.enable=true,OpenServiceMesh.tracing.address=<tracing server hostname>,OpenServiceMesh.tracing.port=<tracing server port>,OpenServiceMesh.tracing.endpoint=<tracing server endpoint>
```
ritazh, line 51: If the user is using the osm cli to install Jaeger, do they still need to provide the address, port, endpoint or will the default values in the values.yaml work as expected?

docs/content/docs/patterns/observability/tracing.md
ritazh: What if users don't want to upgrade their helm release? They could technically just `kubectl edit`.

nojnhuh: > Can this override user configured changes in the OSM configmap?

Yes, any values supplied to the `osm mesh upgrade` command will override the current state. When a value is not specified, it is left unchanged _from the previous Helm release_. So If a user does `osm install --enable-tracing`, then `kubectl edit` to disable it, then `osm mesh upgrade`, tracing will be enabled. If `osm mesh upgrade --enable-tracing=false` was used instead of `kubectl edit`, the subsequent `osm mesh upgrade` would keep it disabled.

> What if users don't want to upgrade their helm release? They could technically just `kubectl edit`.

Yes, that's correct. Users shouldn't expect those changes to persist after an upgrade though.

docs/content/docs/patterns/observability/tracing.md
SanyaKochhar: Hm so below where it references specifically Jaeger in the BYO section, is that wrong? Should I change it to say an already running instance of....something else?
```
The following section documents the additional steps needed to allow an already running instance of Jaeger to integrate with your OSM control plane.

Prerequisites
* A running Jaeger instance
```

docs/content/docs/patterns/observability/tracing.md
SanyaKochhar: If they're using the osm CLI with the `--deploy-jaeger` flag, wouldn't that fall under automatic bring up? I do see at least the default `port` value being used in [jaeger-service.yaml](https://github.com/openservicemesh/osm/blob/main/charts/osm/templates/jaeger-service.yaml) and [jaeger-deployment.yaml](https://github.com/openservicemesh/osm/blob/main/charts/osm/templates/jaeger-deployment.yaml). I can include a note for clarity if necessary?
cc: @eduser25 may know be better able to answer this

docs/content/docs/patterns/observability/tracing.md
SanyaKochhar: I've added the note. Is there anything I need to update about using `kubectl edit` or values persisting?

nojnhuh: For values persisting, if that's not clear between the upgrade guide and the CLI `--help` text then we should edit those.

ritazh: > For values persisting, if that's not clear between the upgrade guide and the CLI osm mesh upgrade --help text then we should edit those.

Yes that would be great. I just looked at `osm mesh upgrade -h` and `osm install -h`, there is nothing about value override for user-provided values. Would be good to add it. 

There is this for persisting flags provided by helm values:
> By default, if flags tied to Helm chart values are not specified for "osm mesh
upgrade", then their current value will be carried over to the new release.

So if these tracing flags are provided as part of `osm mesh upgrade` then it is safe to assume they will be persisted/carried over in future releases.

docs/content/docs/patterns/observability/tracing.md
ritazh: @SanyaKochhar Can you pls make this section more clear that these flags are only needed for BYO Jaeger/tracing solution, NOT for when --deploy-jaeger flag is used.

docs/content/docs/patterns/observability/tracing.md
nojnhuh: > Yes that would be great. I just looked at `osm mesh upgrade -h` and `osm install -h`, there is nothing about value override for user-provided values. Would be good to add it.

@ritazh Just opened #2518, PTAL

ritazh (approved):

PR #2512, opened 2021-02-11
demo: disable prometheus, grafana and jaegar by default

shashankram:
**Description**:
Makes the demo script consistent with the default
options in osm cli for flags related to Prometheus,
Grafana, and Jaegar deployments. Deploying these
by default does not seem to be testing anything
in the CI, and results in scheduling issues arising
from insufficient CPU available in the Kind CI cluster
environments sometimes.

These can be enabled by overriding the default values
in the `.env` file similar to other demo options.

**Affected area**:

- New Functionality      [ ]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [ ]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests                  [ ]
- CI System              [ ]
- Performance            [ ]
- Other                  [X]

`No`


snehachhabria:
I think it makes sense disabling this from the CI runs, but we decided to keep them enabled in the demo so that users would get to see all the functionality OSM supports

draychev (approved):

shashankram:
> I think it makes sense disabling this from the CI runs, but we decided to keep them enabled in the demo so that users would get to see all the functionality OSM supports

I don't think the automated demo is in a state where it is consumable by users to understand these features. There are way too many configurations and nuances to deal with, given the demo has been tweaked significantly to get it working with the `maestro CI`. This includes things like cleaning up existing testbeds via `clean-kubernetes.sh` etc. For this reason, I think it makes more sense to demo these via the manual demo, and let users enable this via the env file when running the demo script.

snehachhabria (approved):

PR #2503, opened 2021-02-10
docs(metrics): Add docs for WASM stats

nojnhuh:
**Description**:
This change adds documentation regarding the WASM-based Envoy extension
used to implement SMI metrics. User-facing docs have been added with
what metrics are generated and known gaps. Developer documentation
includes how to build the extension and a high-level description of how
it works.

---

This PR is part of several which together extend Envoy to generate custom 
metrics needed to implement the SMI metrics spec (#984).
Here is a map of the PRs as they're currently planned:

1. (#2479) ~~ref(injector): pass pod object to getEnvoySidecarContainerSpec~~
2. (#2488) ~~feat(metrics): add WASM metrics source~~
3. (YOU ARE HERE) **docs(metrics): Add docs for WASM stats**
4. feat(metrics): Add WASM feature flag
5. feat(metrics): Add stats.wasm to workload pods
6. feat(injector): Add name, workload name, workload kind to PodMetadata
7. feat(metrics): Add source labels to WASM metrics
8. feat(metrics): Add dest labels to WASM metrics
9. feat(metrics): Add "unknown" for dest labels on local replies
10. feat(metrics): clean up WASM metrics in Prometheus config
11. feat(metrics): add flag to enable WASM metrics
12. tests(e2e): add WASM metrics test

The full set of changes can be found here: https://github.com/openservicemesh/osm/compare/main...nojnhuh:wasm-metrics

**Affected area**:

- New Functionality      [ ]
- Documentation          [X]
- Install                [ ]
- Control Plane          [ ]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests                  [ ]
- CI System              [ ]
- Performance            [ ]
- Other                  [ ]

Please answer the following questions with yes/no.

- Does this change contain code from or inspired by another project? If so, did you notify the maintainers and provide attribution?
No

PR #2517, opened 2021-02-11
feat(metrics): Add WASM feature flag

nojnhuh:
**Description**:
This change adds a feature flag to encapsulate the additional logic
needed for the stats WASM extension needed for SMI metrics. This is
deliberately not wired up in the main package yet so it can't be enabled
until the stats are fully integrated.

---

1. (#2479) ~~ref(injector): pass pod object to getEnvoySidecarContainerSpec~~
2. (#2488) ~~feat(metrics): add WASM metrics source~~
3. (#2503) ~~docs(metrics): Add docs for WASM stats~~
4. (YOU ARE HERE) **feat(metrics): Add WASM feature flag**
5. feat(metrics): Add stats.wasm to workload pods
6. feat(injector): Add name, workload name, workload kind to PodMetadata
7. feat(metrics): Add source labels to WASM metrics
8. feat(metrics): Add dest labels to WASM metrics
9. feat(metrics): Add "unknown" for dest labels on local replies
10. feat(metrics): clean up WASM metrics in Prometheus config
11. feat(metrics): add flag to enable WASM metrics
12. tests(e2e): add WASM metrics test

The full set of changes can be found here: https://github.com/openservicemesh/osm/compare/main...nojnhuh:wasm-metrics

**Affected area**:

- New Functionality      [ ]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [ ]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests                  [ ]
- CI System              [ ]
- Performance            [ ]
- Other                  [ ]

No


pkg/featureflags/featureflags.go
```diff
@@ -11,6 +11,8 @@ type OptionalFeatures struct {
 
 	// RoutesV2
 	RoutesV2 bool
+
+	WASMStatsFilepath string
```
shashankram, line 15: This should ideally be the feature name instead of an attribute of the feature.

nojnhuh: So I was planning to set this directly to a command line flag for osm-controller, but it turns out I wasn't actually propagating that all the way through and was using basically a hardcoded value. I'll make this just a bool and keep the hardcoded path. I feel like if users want to swap out the .wasm or build their own container image, they know what they're doing enough to make that work themselves.

shashankram (approved):

snehachhabria (approved):

Issue #2134, opened 2020-12-02
cluster_not_found occurs when multi services existing of one application


addozhang:
**Bug description**:

Described in traffic split [workflow](https://github.com/servicemeshinterface/smi-spec/blob/master/apis/traffic-split/v1alpha3/traffic-split.md#workflow) in smi spec, there are two services form `foobar-v1` deployment: `foobar` and `foobar-v1`.

>   * Deployment named `foobar-v1`, with labels: `app: foobar` and `version: v1`.
>  * Service named `foobar`, with a selector of `app: foobar`.
>  * Service named `foobar-v1`, with selectors: `app:foobar` and `version: v1`.
>  * Clients use the FQDN of `foobar` to communicate.

Before create any TrafficSplit for `foobar` service, the root service will be not filtered out from service list, in logic of `catalog.GetServicesFromEnvoyCertificate`.

It will return two list, and in many invocations the logic will choose the index 0 one. But sometimes the one with 0 index are **not same**.

Such as in `RDS_Inbound` logic of `rds`, the one with 0 index is `foobar`, the local cluster name in route config is `foobar-local`.
And in logic of `cds`, the one with 0 index is `foobar-v1`, then the local cluster name is `foobar-v1-local`.

So, `cluster_not_found ` occurs.

**Affected area (please mark with X where applicable)**:
- Install                    [ ]
- SMI Traffic Access Policy  [ ]
- SMI Traffic Specs Policy   [ ]
- SMI Traffic Split Policy   [ ]
- Permissive Traffic Policy  [ ]
- Ingress                    [ ]
- Egress                     [ ]
- Envoy Control Plane        [X]
- CLI Tool                   [ ]
- Metrics                    [ ]
- Certificate Management     [ ]
- Sidecar Injection          [ ]
- Logging                    [ ]
- Debugging                  [ ]
- Tests                      [ ]
- CI System                  [ ]

**Expected behavior**:

the local cluster name in RDS_Inbound should be same the one return by cds

**Steps to reproduce the bug (as precisely as possible)**:

**How was OSM installed?**:

**Anything else we need to know?**:

**Environment**:
- OSM version (use `osm version`):
- Kubernetes version (use `kubectl version`):
- Size of cluster (number of worker nodes in the cluster):
- Others:

addozhang:
Solution, if the multi services founded, create local clusters with all services. The redundant one won't effect anything.

shashankram:
[...]

PR #2526, opened 2021-02-12
feat(*): implement traffic split v1alpha2 w/ routes refactor

### LINKED ISSUE #2400: add support for Traffic Split v1alpha2 for routes v2 feature

michelleN:
* resolves #2400

-->
**Affected area**:

- New Functionality      [ ]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [X ]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests                  [ ]
- CI System              [ ]
- Performance            [ ]
- Other                  [ ]

### DIFF
--- a/pkg/catalog/routes_test.go
+++ b/pkg/catalog/routes_test.go
@@ -25,6 +25,197 @@ import (
 	"github.com/openservicemesh/osm/pkg/utils"
 )
 
+var expectedBookbuyerOutbound []*trafficpolicy.OutboundTrafficPolicy = []*trafficpolicy.OutboundTrafficPolicy{
+	{
+		Name:      "bookstore-v1",
+		Hostnames: tests.BookstoreV1Hostnames,
+		Routes: []*trafficpolicy.RouteWeightedClusters{
+			{
+				HTTPRouteMatch:   wildCardRouteMatch,
+				WeightedClusters: mapset.NewSet(tests.BookstoreV1DefaultWeightedCluster),
+			},
+		},
+	},
+	{
+		Name:      "bookstore-v2",
+		Hostnames: tests.BookstoreV2Hostnames,
+		Routes: []*trafficpolicy.RouteWeightedClusters{
+			{
+				HTTPRouteMatch:   wildCardRouteMatch,
+				WeightedClusters: mapset.NewSet(tests.BookstoreV2DefaultWeightedCluster),
+			},
+		},
+	},
+	{
+		Name:      "bookstore-apex",
+		Hostnames: tests.BookstoreApexHostnames,
+		Routes: []*trafficpolicy.RouteWeightedClusters{
+			{
+				HTTPRouteMatch:   wildCardRouteMatch,
+				WeightedClusters: mapset.NewSet(tests.BookstoreApexDefaultWeightedCluster),
+			},
+		},
+	},
+}
+
+func TestListTrafficPoliciesForServiceAccount(t *testing.T) {
+	assert := tassert.New(t)
+
+	testCases := []struct {
+		name             string
+		serviceAccount   service.K8sServiceAccount
+		apexMeshServices []service.MeshService
+		trafficsplits    []*split.TrafficSplit
+		traffictargets   []*access.TrafficTarget
+		trafficspecs     []*specs.HTTPRouteGroup
+		expectedInbound  []*trafficpolicy.InboundTrafficPolicy
+		expectedOutbound []*trafficpolicy.OutboundTrafficPolicy
+		expectedErr      bool
+	}{
+		{
+			name:             "only traffic targets",
+			serviceAccount:   tests.BookbuyerServiceAccount,
+			apexMeshServices: []service.MeshService{},
+			trafficsplits:    []*split.TrafficSplit{},
+			traffictargets:   []*access.TrafficTarget{&tests.TrafficTarget},
+			trafficspecs:     []*specs.HTTPRouteGroup{&tests.HTTPRouteGroup},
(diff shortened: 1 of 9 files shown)


pkg/envoy/cds/response.go
```diff
@@ -62,6 +62,15 @@ func NewResponse(meshCatalog catalog.MeshCataloger, proxy *envoy.Proxy, _ *xds_d
 	}
 	clusters = append(clusters, localCluster)
 
+	if featureflags.IsRoutesV2Enabled() {
+		// if this service is associated with a trafficsplit service, add a local cluster for the apex service
+		for _, splitService := range meshCatalog.GetApexServicesForBackend(proxyServiceName) {
+			lcn := envoy.GetLocalClusterNameForService(splitService)
+			lc := getSyntheticCluster(lcn)
```
shashankram, line 69: I am not sure why a traffic split configuration should impact local clusters (on the uptream). Since traffic split is a client side configuration, why does this affect the configuration at the server side?

shashankram (commented):
Changes look good, left a comment to discuss. Would really like to see an e2e test for traffic split with routesV2 in a follow-up PR.

pkg/envoy/cds/response.go
michelleN: When a request comes, we need an inbound virtual host programmed for the match of on the domains corresponding to the apex service in the traffic split on upstream and that match needs a route action that points to a local cluster. I did think about having the route match for the traffic split apex service domains point to the existing local cluster instead of a new one but decided to go this simpler route for now. I can take another look at going that route.

shashankram: That makes sense. However, what if the apex service is not just a virtual service but an actual service accepting traffic from clients - in that case programming a synthetic service seems misleading? There could be clients trying to reach the apex service without a traffic split, in which case this cluster should not be synthetic?

michelleN: The scenario you're describing I believe is multiple services per proxy which I think we should handle with this issue #2134. At the moment, I was configuring v1alpha2 in the routes refactor work as it would behave in the legacy code path so we could switch over asap to the routes v2 code path. Was planning to solve multiple services pointing to a single proxy in #2134. The e2e scenario here would not have overlapping labels configured for the apex service and backend services.

shashankram: Got it, thanks for clarifying. When we do implement #2134 we can tweak this.

shashankram (approved):

pkg/trafficpolicy/trafficpolicy.go
```diff
@@ -158,15 +156,14 @@ func mergeRules(originalRules, latestRules []*Rule) []*Rule {
...
-	mergeErrors := []error{}
+func mergeRoutesWeightedClusters(originalRoutes, latestRoutes []*RouteWeightedClusters) []*RouteWeightedClusters {
 	for _, latest := range latestRoutes {
 		foundRoute := false
 		for _, original := range originalRoutes {
 			if reflect.DeepEqual(original.HTTPRouteMatch, latest.HTTPRouteMatch) {
 				foundRoute = true
 				if !reflect.DeepEqual(original.WeightedClusters, latest.WeightedClusters) {
-					mergeErrors = append(mergeErrors, errors.Errorf("Error merging RoutesWeightedClusters %v with %v", original, latest))
+					original.WeightedClusters = latest.WeightedClusters
```
ksubrmnn, line 166: Why do you take the latest and not take the union of both sets? Can you add a comment here?

pkg/catalog/routes_test.go
```diff
@@ -25,6 +25,197 @@ import (
...
+								service.WeightedCluster{ClusterName: "default/bookstore-v2", Weight: 10},
+							}),
+						},
+					},
+				},
+			},
+			expectedErr: false,
+		},
+		{
+			name:           "traffic targets and traffic splits",
```
ksubrmnn, line 137: Can you include in the name what makes this test case different from the one before it?

pkg/envoy/cds/response.go
ksubrmnn: Can you add issue number 2134 to a comment for this code?

ksubrmnn (changes requested):

pkg/trafficpolicy/trafficpolicy.go
michelleN: Each route match can only have one set of weighted clusters. 
Ex. 
**original.WeightedClusters**: {cluster1: 50, cluster2: 50}
**latest.WeightedClusters**: {cluster1: 20, cluster2: 80}

If you look at it like that, I think it makes more sense why we can't take the union. A route match is associated with an action. The action indicates the behavior on a route match. The original behavior (with the given example) is that each cluster gets 50 % of the traffic. In latest, cluster 1 gets 20% and cluster 2 gets 80%. When we merge RoutesWeightedClusters, we want to take the latest and replace the original. In practice, we first calculate outbound 
traffic policies using Traffic Targets. We can then separately calculate policies based on traffic split. When we merge the outbound policies, we want to ensure the traffic split related policies take precedence over traffic target if there is overlap.

pkg/catalog/routes_test.go
michelleN: yes good catch

ksubrmnn (approved):
Synced with Michelle offline. Tests/changes for inbound traffic policies to follow in separate PR

snehachhabria (approved):

Issue #1658, opened 2020-09-02
Destination endpoints for a service that maps to multiple ServiceAccounts doesn't honor TrafficTarget policy


shashankram:
**Bug description**:
Consider the following example:
- Service-X is backed by <Pod-1, SvcAccount-1> and <Pod-2, SvcAccount-2>.
-  Client-A (Service-A, SvcAccount-A) wants to access Service-X
- User configures single TrafficTarget policy to allow `src=SvcAccount-A, dst=SvcAccount-1`

_*Note: there is no traffic split configured_

OSM will translate the TrafficTarget policy to `src=Service-A, dst=Service-X`.

When EDS resolves the endpoints for the destination `Service-X`, it will receive the endpoints for both Pod-1 and Pod-2. However since `Pod-2, SvcAccount-2` is not allowed to be accessed by `Service-A, SvcAccount-A`, traffic from `Service-A, SvcAccount-A` to `Pod-2, Svc-Account-2` should be denied. Currently, the destination endpoints for a service do not take into consideration the traffic target policies.

**Affected area (please mark with X where applicable)**:
- Install                    [ ]
- SMI Traffic Access Policy  [X]
- SMI Traffic Specs Policy   [ ]
- SMI Traffic Split Policy   [ ]
- Permissive Traffic Policy  [ ]
- Ingress                    [ ]
- Egress                     [ ]
- Envoy Control Plane        [ ]
- CLI Tool                   [ ]
- Metrics                    [ ]
- Certificate Management     [ ]
- Sidecar Injection          [ ]
- Logging                    [ ]
- Debugging                  [ ]
- Tests                      [ ]
- CI System                  [ ]

**Expected behavior**:
Based on the above example:
- Service-A should be able to access <Pod-1, SvcAccount-1>
- Service-A should not be able to access <Pod-2, SvcAccount-2>

**Steps to reproduce the bug (as precisely as possible)**:
Use the above example.

**How was OSM installed?**: `osm install` 

**Anything else we need to know?**: n/a

**Environment**:
- OSM version (use `osm version`):
```
Version: dev; Commit: 4370f4c8e9036b77125da795eb565be5b7ec8b21; Date: 2020-09-01-16:55
```

- Kubernetes version (use `kubectl version`):
```
[...]

PR #2534, opened 2021-02-16
feat/fix(endpoints) : Ensure endpoints for a service honors traffic policy

snehachhabria:
**Description**:

This PR ensures that the endpoints built for a service in EDS honor
SMI traffic target policies. Only those destinations endpoints are
programmed on the envoy, if its destination pod has a service account
specified as a destination in any of of the applicable traffic targets.

Resolves issue #1658

**Affected area**:

- New Functionality      [ ]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [X]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests                  [X]
- CI System              [ ]
- Performance            [ ]
- Other                  [ ]


pkg/catalog/endpoint.go
```diff
@@ -36,3 +36,17 @@ func (mc *MeshCatalog) GetResolvableServiceEndpoints(svc service.MeshService) ([
 	}
 	return endpoints, nil
 }
+
+// ListEndpointsForPodFromServiceAccount retrieves the list of IP addresses for the given service account
+func (mc *MeshCatalog) ListEndpointsForPodFromServiceAccount(sa service.K8sServiceAccount) []endpoint.Endpoint {
```
shashankram, line 41: The signature of this function seems a bit confusing because endpoints are for a service and not service accounts, yet this function only takes a service account as input and returns all endpoints for pods in it.

I think it would be much cleaner to change this as follows and keep the logic in EDS simple.
In `catalog`:
`ListAllowedEndpointsForService(downstreamIdentity service.K8sServiceAccount, upstreamSvc service.MeshService) []endpoint.Endpoint`

In `EDS`, use `ListAllowedEndpointsForService(...)` directly without any need for filtering endpoints.

pkg/endpoint/types.go
```diff
@@ -12,6 +12,9 @@ type Provider interface {
 	// Retrieve the IP addresses comprising the given service.
 	ListEndpointsForService(service.MeshService) []Endpoint
 
+	// ListEndpointsForPodFromServiceAccount retrieves the list of IP addresses for the given service account
+	ListEndpointsForPodFromServiceAccount(sa service.K8sServiceAccount) []Endpoint
```
shashankram, line 16: Using the term `Pod` in the generic `Provider` interface doesn't look correct because this is meant to be generic regardless of the runtime (VMs, pods). Could you rename this to `ListEndpointsforServiceAccount` or `ListEndpointsForIdentity`.

shashankram (changes requested):
Thanks for the change. The idea looks fine overall, however I suggest decoupling traffic policy filtering logic from EDS and handling the filtering of endpoints to allowed upstreams within the `catalog`, for simplicity and because catalog is already traffic policy aware.

I have provided more details inline.

pkg/endpoint/types.go
snehachhabria: I have renamed it to `ListEndpointsForIdentity`

snehachhabria:
> Thanks for the change. The idea looks fine overall, however I suggest decoupling traffic policy filtering logic from EDS and handling the filtering of endpoints to allowed upstreams within the `catalog`, for simplicity and because catalog is already traffic policy aware.
> 
> I have provided more details inline.

@shashankram thank you for the suggestions. I have updated the PR to ensure that EDS isn't aware of traffic policy, all the business logic has been moved into catalog. PTAL

pkg/catalog/endpoint.go
```diff
@@ -36,3 +36,49 @@ func (mc *MeshCatalog) GetResolvableServiceEndpoints(svc service.MeshService) ([
...
 	return endpoints, nil
 }
+
+// ListAllowedEndpointsForService returns only those endpoints for a service that belong to the allowed outbound service accounts
+// for the given downstream identity
+func (mc *MeshCatalog) ListAllowedEndpointsForService(downstreamIdentity service.K8sServiceAccount, upstreamSvc service.MeshService) ([]endpoint.Endpoint, error) {
+	var outboundEndpoints []endpoint.Endpoint
+	outboundEndpoints, err := mc.ListEndpointsForService(upstreamSvc)
+	if err != nil {
+		log.Error().Err(err).Msgf("Error looking  up endpoints for service %s with downstream identity %s", upstreamSvc, downstreamIdentity)
```
shashankram, line 46:
```suggestion
		log.Error().Err(err).Msgf("Error looking  up endpoints for upstream service %s", upstreamSvc)
```

`service foo with downstream identity baz` can be misleading because the service might not have the identity passed to the function.

pkg/catalog/endpoint.go
```diff
@@ -36,3 +36,49 @@ func (mc *MeshCatalog) GetResolvableServiceEndpoints(svc service.MeshService) ([
...
+		return nil, err
+	}
+	log.Trace().Msgf("Outbound endpoints for service %s with downstream identity %s: %v", upstreamSvc, downstreamIdentity, outboundEndpoints)
```
shashankram, line 49: We can remove this log if it was just added for debugging during development and isn't required.

pkg/catalog/endpoint.go
```diff
@@ -36,3 +36,49 @@ func (mc *MeshCatalog) GetResolvableServiceEndpoints(svc service.MeshService) ([
 	}
 	return endpoints, nil
 }
+
+// ListAllowedEndpointsForService returns only those endpoints for a service that belong to the allowed outbound service accounts
+// for the given downstream identity
+func (mc *MeshCatalog) ListAllowedEndpointsForService(downstreamIdentity service.K8sServiceAccount, upstreamSvc service.MeshService) ([]endpoint.Endpoint, error) {
+	var outboundEndpoints []endpoint.Endpoint
```
shashankram, line 43: Declaration not required.

pkg/catalog/endpoint.go
```diff
@@ -36,3 +36,49 @@ func (mc *MeshCatalog) GetResolvableServiceEndpoints(svc service.MeshService) ([
...
+	}
+	log.Trace().Msgf("Outbound endpoints for service %s with downstream identity %s: %v", upstreamSvc, downstreamIdentity, outboundEndpoints)
+
+	destSvcAccounts, err := mc.ListAllowedOutboundServiceAccounts(downstreamIdentity)
+	if err != nil {
+		log.Error().Err(err).Msgf("Error looking up outbound service accounts for downstream identity %s", downstreamIdentity)
+		return nil, err
+	}
+
+	var allowedEndpoints []endpoint.Endpoint
```
shashankram, line 57: A short comment here regarding intersection of the 2 lists will improve readability.

pkg/catalog/endpoint.go
```diff
@@ -36,3 +36,49 @@ func (mc *MeshCatalog) GetResolvableServiceEndpoints(svc service.MeshService) ([
...
+	for _, destSvcAccount := range destSvcAccounts {
+		podEndpoints := mc.listEndpointsforIdentity(destSvcAccount)
+		for _, ep := range outboundEndpoints {
+			for _, podIP := range podEndpoints {
+				if ep.IP.Equal(podIP.IP) {
+					allowedEndpoints = append(allowedEndpoints, ep)
```
shashankram, line 63: Can we just use `set.NewSetFromSlice()` and do `set1.Intersect(set2)`?

pkg/catalog/endpoint.go
```diff
@@ -36,3 +36,49 @@ func (mc *MeshCatalog) GetResolvableServiceEndpoints(svc service.MeshService) ([
...
+				}
+			}
+		}
+	}
+	log.Trace().Msgf("Allowed endpoints for service %s with downstream identity %s: %v", upstreamSvc, downstreamIdentity, allowedEndpoints)
+	return allowedEndpoints, nil
+}
+
+// listEndpointsforIdentity retrieves the list of IP addresses for the given service account
```
shashankram, line 72:
```suggestion
// listEndpointsforIdentity retrieves the list of endpoints for the given service account
```

pkg/envoy/eds/response.go
```diff
@@ -52,3 +46,19 @@ func NewResponse(meshCatalog catalog.MeshCataloger, proxy *envoy.Proxy, _ *xds_d
...
+
+	for _, dstSvc := range meshCatalog.ListAllowedOutboundServicesForIdentity(proxyIdentity) {
+		endpoints, err := meshCatalog.ListAllowedEndpointsForService(proxyIdentity, dstSvc)
+		if err != nil {
+			log.Error().Err(err).Msgf("Failed listing allowed endpoints for service %s for proxy identity %s", dstSvc, proxyIdentity)
+			continue
+		}
+		allowedServicesEndpoints[dstSvc] = endpoints
+	}
+	log.Trace().Msgf("Allowed outbound service endpoints for proxy with identity %s: %v", proxyIdentity, allowedServicesEndpoints)
```
shashankram, line 62: Lets remove this if not required anymore and was only meant for debugging.

shashankram (commented):
Thanks for the rework, looks good overall. Left a few comments inline.
Do we have unit tests and e2e that verifies #1658?

pkg/catalog/endpoint.go
snehachhabria: We cannot use a set here cause `endpoint.Endpoint` has an unhashable type (`net.IP`)

snehachhabria:
> Thanks for the rework, looks good overall. Left a few comments inline.
> Do we have unit tests and e2e that verifies #1658?

We have two unit tests  `TestListAllowedEndpointsForService` and `TestGetEndpointsFromProxy` that have a number of scenarios which captures everything listed in #1658. I have not added an e2e test as I felt the unit tests give us sufficient coverage we need

shashankram:
> > Thanks for the rework, looks good overall. Left a few comments inline.
> > Do we have unit tests and e2e that verifies #1658?
> 
> We have two unit tests `TestListAllowedEndpointsForService` and `TestGetEndpointsFromProxy` that have a number of scenarios which captures everything listed in #1658. I have not added an e2e test as I felt the unit tests give us sufficient coverage we need

That's great, thanks for clarifying.

shashankram (approved):

eduser25 (approved):

PR #2538, opened 2021-02-16
ref(e2e): set pod termination grace period to zero

nojnhuh:
**Description**:
Because most of the client pods used for e2e tests call `sleep`, those
pods can't be terminated within the grace period, which causes each
namespace deletion to take about 30s longer than necessary. This change
sets the `spec.terminationGracePeriodSeconds` to 0 on pods created by
the framework helpers so they can be deleted immediately. Overall, this
speeds up the e2e suite by about 25%.

**Affected area**:

- New Functionality      [ ]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [ ]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests                  [X]
- CI System              [ ]
- Performance            [ ]
- Other                  [ ]

No


shashankram (approved):
25% improvements is very nice to see!

michelleN (approved):
🎉

PR #2539, opened 2021-02-16
Fix race on cert creation for Tresor

eduser25:
Tresor was suffering from a race when multi instance OSM is started.
As a consequence, global (for all cert providers) `createOrUpdate` logic
has been removed as (1) vault/cert-manager don't need it and (2) `createOrUpdate`
was causing races between instances.

- For Tresor, the first instance will create the CA when it has not
been created yet. This relies on Create atomicity on k8s API, and should
ensure only one is created and all (possible) instances synchronize on it
(this includes osm-injector).
- Vault and Cert-Manager both expect the CA to be already existing, so only
a read operation is needed (no sync required between instances) when
they are instantiated.
- Waiting for CA bundle has been removed from injector, since for all providers
synchronization is already taken care of when needed, if needed.

Tested e2e for cert-manager, vault and tresor.

- Control Plane          [X]

No


eduser25:
closing, will submit a different solution

shashankram (approved):

snehachhabria (approved):

Issue #2545, opened 2021-02-17
Traffic split does not work with routes-v2, no virtual_host for apex service in the inbound route at destination


shashankram:
**Bug description**:
Traffic split does not work with routesV2 when run with the default demo `./demo/run-osm-demo.sh --enable-routes-v2-experimental=true`.

There needs to be a virtual_host corresponding to the apex service in the inbound route configuration at the destination service, but this is not being programmed as expected.

**Affected area (please mark with X where applicable)**:
- Install                    [ ]
- SMI Traffic Access Policy  [ ]
- SMI Traffic Specs Policy   [ ]
- SMI Traffic Split Policy   [X]
- Permissive Traffic Policy  [ ]
- Ingress                    [ ]
- Egress                     [ ]
- Envoy Control Plane        [ ]
- CLI Tool                   [ ]
- Metrics                    [ ]
- Certificate Management     [ ]
- Sidecar Injection          [ ]
- Logging                    [ ]
- Debugging                  [ ]
- Tests                      [ ]
- CI System                  [ ]

**Expected behavior**:
Traffic split should work with routesV2.

**Steps to reproduce the bug (as precisely as possible)**:
1. Apply the following diff for spit policy to be processed correctly
```
--- a/demo/deploy-traffic-split.sh
+++ b/demo/deploy-traffic-split.sh
@@ -12,7 +12,7 @@ metadata:
   name: bookstore-split
   namespace: "$BOOKSTORE_NAMESPACE"
 spec:
-  service: "bookstore.$BOOKSTORE_NAMESPACE"
+  service: bookstore
   backends:
 
   - service: bookstore-v1
```

2. Run the demo:
./demo/run-osm-demo.sh --enable-routes-v2-experimental=true`

**Environment**:
- OSM version (use `osm version`):
`Version: dev; Commit: 1323707da19e7b8590dd06aaf91832c01a4ebb4d; Date: 2021-02-17-10:59`

- Kubernetes version (use `kubectl version`):
```
Client Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.3", GitCommit:"1e11e4a2108024935ecfcb2912226cedeafd99df", GitTreeState:"clean", BuildDate:"2020-10-14T12:50:19Z", GoVersion:"go1.15.2", Compiler:"gc", Platform:"linux/amd64"}
[...]

PR #2556, opened 2021-02-18
feat(routesv2/trafficsplit): Traffic split for routes v2

snehachhabria:
**Description**:

This PR has the final set of changes to integrate traffic split with
routes v2. The changes in the PR address issues #2545 and #1979.

This PR further incorporates the routes v2 feature flag as a part of our
automated demo, giving the users the ability to enable/disable routes v2
based on the value set in their .env file.

**Affected area**:

- New Functionality      [ ]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [X]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests                  [X]
- CI System              [ ]
- Performance            [ ]
- Other                  [ ]


pkg/catalog/types.go
```diff
@@ -143,9 +143,11 @@ type MeshCataloger interface {
...
 	ListInboundTrafficTargetsWithRoutes(service.K8sServiceAccount) ([]trafficpolicy.TrafficTargetWithRoutes, error)
 
-	// GetApexServicesForBackend returns a list of services that serve as the apex service in a traffic split where the
-	//	given service is a backend
-	GetApexServicesForBackend(service.MeshService) []service.MeshService
+	// IsApexService returns true if the given service is an apex service in any traffic split
+	IsApexService(service.MeshService) bool
+
+	// GetBackendServiceForApexService returns a list of backend services that serve the given apex service in a traffic split
+	GetBackendServiceForApexService(apexSvc service.MeshService) []service.MeshService
```
shashankram, line 150: This is internal to the `MeshCatalog` implementation, no reason to expose it in the interface.

snehachhabria: good catch! I have updated the code

pkg/catalog/routes.go
```diff
@@ -556,12 +556,23 @@ func (mc *MeshCatalog) buildInboundPolicies(t *access.TrafficTarget) []*trafficp
...
-
-		for _, sourceServiceAccount := range trafficTargetIdentitiesToSvcAccounts(t.Spec.Sources) {
-			for _, routeMatch := range routeMatches {
-				servicePolicy.AddRule(*trafficpolicy.NewRouteWeightedCluster(routeMatch, weightedCluster), sourceServiceAccount)
+		// while building inbound policies, check if the service is an apex service in traffic split
+		if mc.isApexService(destService) {
+			var localDestServices []service.MeshService
+			// build the inbound policy for an apex service to have its weighted clusters point to the backend services
+			backendServices := mc.getBackendServiceForApexService(destService)
+			// among all the backedn services, select only those that operlap with the destination services
```
shashankram, line 564: typo: s/backedn/backend, s/operlap/overlap

pkg/catalog/routes.go
```diff
@@ -556,12 +556,23 @@ func (mc *MeshCatalog) buildInboundPolicies(t *access.TrafficTarget) []*trafficp
...
+			// (i.e the destination service account)
```
shashankram, line 565: `destination services (i.e. destination service account)` is confusing. Maybe rephrase.

snehachhabria: removed `(i.e. destination service account)` . Hopefully that reduces the confusion

draychev (approved):
looks good to me - excited to start testing this

demo/deploy-bookstore.sh
```diff
@@ -25,10 +23,7 @@ spec:
   ports:
   - port: 80
     name: bookstore-port
-  selector:
```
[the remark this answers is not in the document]
snehachhabria: we don't need the selector on services that are used as apex services in a traffic split, that's the reason we got rid of it.

shashankram (approved):
Thanks!

pkg/catalog/routes.go
```diff
@@ -572,6 +582,18 @@ func (mc *MeshCatalog) buildInboundPolicies(t *access.TrafficTarget) []*trafficp
 	return inboundPolicies
 }
 
+func buildInboundPolicyRules(servicePolicy *trafficpolicy.InboundTrafficPolicy, routeMatches []trafficpolicy.HTTPRouteMatch, sourceSvcAccounts []service.K8sServiceAccount, services []service.MeshService) *trafficpolicy.InboundTrafficPolicy {
```
michelleN, line 585: What do you think about moving this func to pkg/trafficpolicy and hanging this as a method on *trafficpolicy.InboundTrafficPolicy?

pkg/catalog/routes.go
```diff
@@ -556,12 +556,22 @@ func (mc *MeshCatalog) buildInboundPolicies(t *access.TrafficTarget) []*trafficp
...
 		servicePolicy := trafficpolicy.NewInboundTrafficPolicy(buildPolicyName(destService, false), hostnames)
 
-		weightedCluster := getDefaultWeightedClusterForService(destService)
-
-		for _, sourceServiceAccount := range trafficTargetIdentitiesToSvcAccounts(t.Spec.Sources) {
-			for _, routeMatch := range routeMatches {
-				servicePolicy.AddRule(*trafficpolicy.NewRouteWeightedCluster(routeMatch, weightedCluster), sourceServiceAccount)
+		// while building inbound policies, check if the service is an apex service in traffic split
+		if mc.isApexService(destService) {
+			var localDestServices []service.MeshService
```
michelleN, line 561: Could you explain why we're checking if the service is an apex on inbound?

michelleN (changes requested):

pkg/catalog/routes.go
snehachhabria: I did think of that but `getDefaultWeightedClusterForService` is component of catalog and will not be available to use in pkg/trafficpolicy. Further catalog cannot be imported in trafficpolicy as it will cause circular dependency

pkg/catalog/routes.go
snehachhabria: In the case of building inbound routes for a particular service we need the check if its an apex service (with traffic split) cause the hostnames need to be the name of the apex service but its weighted cluster needs to be the backend service.

snehachhabria: We previously had it as the apex service itself but that didn't work cause we don't fully have support for multiple services on a single pod. This will require more work

michelleN (approved):

PR #2575, opened 2021-02-19
fix(injector): Increase memory for controller and injector

### LINKED ISSUE #2558: osm-injector memory request and limits are not sufficient for OpenShift
**Bug description**:
When installing on OpenShift, the osm-injector enters CrashLoopBackOff because it is OOMKilled. Note that the Pod is out of memory, not the Node.

The issue stems from the Namespaces Informer cache failing to sync for the Kubernetes Controller.

Bumping up the [injector resource request/limit](https://github.com/openservicemesh/osm/blob/f58bfb484b1da4f0f2e4a4b8692f47d4f055da11/charts/osm/values.yaml#L142) fixes the issue.

Although these values can be overridden both through the osm CLI and the helm CLI, the default values should still be sufficient for the injector pod to run on OpenShift.

I've observed that the injector often uses about 35M of memory

**Expected behavior**:
The osm-injector should run on a fresh OpenShift cluster without crashing.

Install OSM on OpenShift using `osm install`

**How was OSM installed?**:
run-osm-demo.sh

**Environment**:
- OSM version (use `osm version`): Built from main with last commit 1bec41f73b1bf43497fdcfe8a47281d1dc6f76b3
- Kubernetes version (use `kubectl version`): 1.18.3
- Size of cluster (number of worker nodes in the cluster): 3 workers
- Others: OpenShift cluster

ksubrmnn:

**Description**:
Fixes #2558

This increases the default memory limits of the injector and the controller so that they are less likely to crash.

**Affected area**:

- New Functionality      [ ]
- Documentation          [ ]
- Install                [X]
- Control Plane          [ ]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests                  [ ]
- CI System              [ ]
- Performance            [ ]
- Other                  [ ]

No

### DIFF
--- a/charts/osm/README.md
+++ b/charts/osm/README.md
@@ -93,13 +93,13 @@ The following table lists the configurable parameters of the osm chart and their
 | OpenServiceMesh.image.registry | string | `"openservicemesh"` | `osm-controller` image registry |
 | OpenServiceMesh.image.tag | string | `"v0.7.0"` | `osm-controller` image tag |
 | OpenServiceMesh.imagePullSecrets | list | `[]` | `osm-controller` image pull secret |
-| OpenServiceMesh.injector | object | `{"replicaCount":1,"resource":{"limits":{"cpu":"0.5","memory":"32M"},"requests":{"cpu":"0.3","memory":"32M"}}}` | Sidecar injector configuration |
+| OpenServiceMesh.injector | object | `{"replicaCount":1,"resource":{"limits":{"cpu":"0.5","memory":"64M"},"requests":{"cpu":"0.3","memory":"64M"}}}` | Sidecar injector configuration |
 | OpenServiceMesh.meshName | string | `"osm"` | Name for the new control plane instance |
 | OpenServiceMesh.osmNamespace | string | `""` | Optional parameter. If not specified, the release namespace is used to deploy the osm components. |
 | OpenServiceMesh.osmcontroller.resource.limits.cpu | string | `"1.5"` |  |
 | OpenServiceMesh.osmcontroller.resource.limits.memory | string | `"512M"` |  |
 | OpenServiceMesh.osmcontroller.resource.requests.cpu | string | `"0.5"` |  |
-| OpenServiceMesh.osmcontroller.resource.requests.memory | string | `"32M"` |  |
+| OpenServiceMesh.osmcontroller.resource.requests.memory | string | `"128M"` |  |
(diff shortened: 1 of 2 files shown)


shashankram (approved):

SanyaKochhar (approved):

charts/osm/README.md
```diff
@@ -93,13 +93,13 @@ The following table lists the configurable parameters of the osm chart and their
...
 | OpenServiceMesh.imagePullSecrets | list | `[]` | `osm-controller` image pull secret |
-| OpenServiceMesh.injector | object | `{"replicaCount":1,"resource":{"limits":{"cpu":"0.5","memory":"32M"},"requests":{"cpu":"0.3","memory":"32M"}}}` | Sidecar injector configuration |
+| OpenServiceMesh.injector | object | `{"replicaCount":1,"resource":{"limits":{"cpu":"0.5","memory":"64M"},"requests":{"cpu":"0.3","memory":"64M"}}}` | Sidecar injector configuration |
 | OpenServiceMesh.meshName | string | `"osm"` | Name for the new control plane instance |
 | OpenServiceMesh.osmNamespace | string | `""` | Optional parameter. If not specified, the release namespace is used to deploy the osm components. |
 | OpenServiceMesh.osmcontroller.resource.limits.cpu | string | `"1.5"` |  |
 | OpenServiceMesh.osmcontroller.resource.limits.memory | string | `"512M"` |  |
 | OpenServiceMesh.osmcontroller.resource.requests.cpu | string | `"0.5"` |  |
-| OpenServiceMesh.osmcontroller.resource.requests.memory | string | `"32M"` |  |
+| OpenServiceMesh.osmcontroller.resource.requests.memory | string | `"128M"` |  |
```
eduser25, line 102: wait, did you see this affect anyhow?

ksubrmnn: I didn't see this affected, but I noticed in the [scale doc](https://github.com/openservicemesh/osm/blob/3f0ea1977fb17ff41cb691a6dcbd81d25d4005b0/docs/content/docs/scale.md) that initial memory use was 45 MB so decided to bump this as well. What do you think?

eduser25: if nothing was observed, I'd think it's totally fine

PR #2580, opened 2021-02-20
feat(install): Enable optional privileged containers

### LINKED ISSUE #250: TrafficTarget does not work as expected sometimes
Some of us have observed that traffic target does not work as expected.

Here, only "/counter" is allowed based on the TrafficTarget spec but the demo is allowing 
"/incrementcounter" as well.

---[ 29 ]-----------------------------------------
Fetching http://bookstore.mesh/counter
Identity: bookstore-1--25412f8
Counter: 12
Server: envoy
Date: Tue, 03 Mar 2020 17:53:24 GMT
Status: 200 OK

Fetching http://bookstore.mesh/incrementcounter
Identity: bookstore-1--25412f8
Counter: 13
Server: envoy
Date: Tue, 03 Mar 2020 17:53:24 GMT
Status: 200 OK

---
kind: TrafficTarget
apiVersion: access.smi-spec.io/v1alpha1
metadata:
  name: bookstore-service
  namespace: "$K8S_NAMESPACE"
destination:
  # (todo): use service account
  kind: ServiceAccount
  name: bookstore-1-serviceaccount
  namespace: "$K8S_NAMESPACE"
specs:
- kind: HTTPRouteGroup
  name: bookstore-service-routes
  matches:
  - counter
sources:
# (todo): use service account
- kind: ServiceAccount
  name: bookbuyer-serviceaccount
  namespace: "$K8S_NAMESPACE"

ksubrmnn:

**Description**:
Fixes #250. A privileged init container is needed for init containers for OSM to work on openshift.

OpenShift specific documentation and e2es to follow

**Affected area**:

- New Functionality      [ ]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [X]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests                  [ ]
- CI System              [ ]
- Performance            [ ]
- Other                  [ ]

No


charts/osm/values.yaml
```diff
@@ -146,3 +146,6 @@ OpenServiceMesh:
       requests:
         cpu: "0.3"
         memory: "32M"
+  
+  # -- Deploy privileged init containers
```
shashankram, line 150:
```suggestion
  # -- Run init container in privileged mode
```

cmd/cli/install.go
```diff
@@ -189,6 +191,7 @@ func newInstallCmd(config *helm.Configuration, out io.Writer) *cobra.Command {
 	f.BoolVar(&inst.deployJaeger, "deploy-jaeger", defaultDeployJaeger, "Deploy Jaeger in the namespace of the OSM controller")
 	f.StringVar(&inst.envoyLogLevel, "envoy-log-level", defaultEnvoyLogLevel, "Envoy log level is used to specify the level of logs collected from envoy and needs to be one of these (trace, debug, info, warning, warn, error, critical, off)")
 	f.BoolVar(&inst.enforceSingleMesh, "enforce-single-mesh", defaultEnforceSingleMesh, "Enforce only deploying one mesh in the cluster")
+	f.BoolVar(&inst.enablePrivilegedInitContainer, "enable-privileged-init-container", defaultEnablePrivilegedInitContainer, "Deploy privileged init container for pods in mesh")
```
shashankram, line 194: We should simply be using the `--set` option to override chart values going forward to avoid the need to wrap every chart value using a cli option.

demo/run-osm-demo.sh
```diff
@@ -105,6 +105,8 @@ if [ "$CERT_MANAGER" = "vault" ]; then
       --deploy-prometheus="$DEPLOY_PROMETHEUS" \
       --enable-prometheus-scraping="$ENABLE_PROMETHEUS_SCRAPING" \
       --envoy-log-level "$ENVOY_LOG_LEVEL" \
+      --enable-privileged-init-container "$ENABLE_PRIVILEGED_INIT_CONTAINER" \
+      --enable-permissive-traffic-policy true \
```
shashankram, line 109: Why is this being set?

demo/run-osm-demo.sh
```diff
@@ -105,6 +105,8 @@ if [ "$CERT_MANAGER" = "vault" ]; then
       --deploy-prometheus="$DEPLOY_PROMETHEUS" \
       --enable-prometheus-scraping="$ENABLE_PROMETHEUS_SCRAPING" \
       --envoy-log-level "$ENVOY_LOG_LEVEL" \
+      --enable-privileged-init-container "$ENABLE_PRIVILEGED_INIT_CONTAINER" \
```
shashankram, line 108: If we use `--set` for this chart value (which I think we should), we shouldn't have to expose this flag via `.env` file.

pkg/configurator/types.go
```diff
@@ -65,4 +65,7 @@ type Configurator interface {
 
 	// GetOutboundIPRangeExclusionList returns the list of IP ranges of the form x.x.x.x/y to exclude from outbound sidecar interception
 	GetOutboundIPRangeExclusionList() []string
+
+	// GetEnablePrivilegedInitContainer determines whether init containers should be privileged
+	GetEnablePrivilegedInitContainer() bool
```
shashankram, line 70: This function name could read better:
```suggestion
	IsPrivilegedInitContainer() bool
```

shashankram (changes requested):
Thanks for the change, overall looks fine. I suggest using the `--set` cli option to override such chart values going forward instead of wrapping each chart value with its own cli flag. We have already started leveraging this for other chart values.

shashankram (approved):
Thanks, the changes look good. 
Gentle reminded to squash commits locally before merge.

pkg/configurator/validating_webhook.go
```diff
@@ -30,13 +30,13 @@ var (
 	deserializer = codecs.UniversalDeserializer()
 
 	// boolFields are the fields in osm-config that take in a boolean
-	boolFields = []string{"egress", "enable_debug_server", "permissive_traffic_policy_mode", "prometheus_scraping", "tracing_enable", "use_https_ingress"}
+	boolFields = []string{"egress", "enable_debug_server", "permissive_traffic_policy_mode", "prometheus_scraping", "tracing_enable", "use_https_ingress", "privileged_init_container"}
```
SanyaKochhar, line 33: Can we use is/use/enable_privileged_init_container / something in that vein? I know it makes the var name long but for the sake of consistency with the other boolean vars eg: `enable_debug_server`, `tracing_enable`, `use_https_ingress` and the way it's named in all the other files

SanyaKochhar (approved):

PR #2592, opened 2021-02-23
metrics: set dynamic metric registration

eduser25:
Initialization still happens for unrelated metrics to the componenets,
but this will at least prevent exposing unrelated metrics when scraping.

Additionally:
- `XdsCertCount/Times` have have had the `Xds` prefix stripped to be more generic for both Injector and OSM.

**Affected area**:

- Metrics                [X]

No

### DIFF
--- a/cmd/osm-injector/osm-injector.go
+++ b/cmd/osm-injector/osm-injector.go
@@ -127,7 +127,12 @@ func main() {
 	defer cancel()
 
 	// Start the default metrics store
-	metricsstore.DefaultMetricsStore.Start()
+	metricsstore.DefaultMetricsStore.Start(
+		metricsstore.DefaultMetricsStore.InjectorRqTime,
+		metricsstore.DefaultMetricsStore.InjectorSidecarCount,
+		metricsstore.DefaultMetricsStore.CertIssuedCount,
+		metricsstore.DefaultMetricsStore.CertIssuedTime,
+	)
 
 	// Initialize Configurator to watch osm-config ConfigMap
 	cfg := configurator.NewConfigurator(kubeClient, stop, osmNamespace, osmConfigMapName)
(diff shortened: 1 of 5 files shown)


snehachhabria (approved):

cmd/osm-injector/osm-injector.go
```diff
@@ -127,7 +127,7 @@ func main() {
 	defer cancel()
 
 	// Start the default metrics store
-	metricsstore.DefaultMetricsStore.Start()
+	startMetricsStore()
```
shashankram, line 130: Doesn't seem like we need a function that's a one liner.

shashankram (approved):

pkg/metricsstore/metricsstore.go
```diff
@@ -49,11 +49,11 @@ type MetricsStore struct {
...
 	 */
-	// CertXdsIssuedCount is the metric counter for the number of xds certificates issued
-	CertXdsIssuedCount prometheus.Counter
+	// CertIssuedCount is the metric counter for the number of certificates issued
+	CertIssuedCount prometheus.Counter
 
-	// CertXdsIssuedCounter the histogram to track the time to issue xds certificates
-	CertXdsIssuedTime *prometheus.HistogramVec
+	// CertXdsIssuedCounter the histogram to track the time to issue a certificates
+	CertIssuedTime *prometheus.HistogramVec
```
shashankram, line 56: These metrics only pertain to XDS certs, why are they being removed from the variable names?

eduser25: sorry, forgot to mention the change.
The current implementation actually only tracks the injector ones, the SDS part requires fixing (#2606). 
Since injector and osm can use the same metric, i just renamed it to something bit more generic (updated the doc too). For metric purposes, injector and xds cert creations will still be separate by labeling and origin, even when sharing the same metric.

PR #2594, opened 2021-02-23
[test] Change testing to multi replica by default

eduser25:
Set osm replica count to 2 by default.

- Tests                  [X]

No

### DIFF
--- a/charts/osm/values.yaml
+++ b/charts/osm/values.yaml
@@ -4,7 +4,7 @@
 
 OpenServiceMesh:
   # -- `osm-controller` replicas
-  replicaCount: 1
+  replicaCount: 2
   image:
     # -- `osm-controller` image registry
     registry: openservicemesh


charts/osm/values.yaml
```diff
@@ -4,7 +4,7 @@
 
 OpenServiceMesh:
   # -- `osm-controller` replicas
-  replicaCount: 1
+  replicaCount: 2
```
nojnhuh, line 7: If this is just for testing, could we hardcode `--set OpenServiceMesh.replicaCount=2` in the e2e framework vs. changing the default in values.yaml? Then regular installs will keep one by default.

shashankram: +1

eduser25: I think the argue last time was to ship the defaults we test (or at least, I remember saying that). I don't know if doing multi-replica would be a superset and catch all issues, including issues that would otherwise be more present in single replicaset. I don't have an answer.
@draychev ? ship HA by default or just change part/all testing to do multi-replica?

shashankram: Have you tested that all CLI commands and tooling work with multiple replicas? If not, let's set the default back to 1, to ensure we aren't breaking user facing features that currently do not handle multiple replicas of the controller.

eduser25: Good point. No I have not really tested any CLIs with multiple replica. I'll close this and maybe issue the change @nojnhuh suggested if the intention is to at least test multiple replica.

PR #2587, opened 2021-02-22
feat(metrics): Add "unknown" for dest labels on local replies

nojnhuh:
**Description**:
Because `destination_*` tags are generated at the request's destination,
requests that invoke a local response from Envoy have no `destination_*`
values. This change modifies Envoy to add the expected header keys with
the value "unknown" to stay consistent with the behavior for requests
that don't invoke a local reply.

---

This PR is part of several which together extend Envoy to generate custom 
metrics needed to implement the SMI metrics spec (#984).
Here is a map of the PRs as they're currently planned:

1. (#2479) ~~ref(injector): pass pod object to getEnvoySidecarContainerSpec~~
2. (#2488) ~~feat(metrics): add WASM metrics source~~
3. (#2503) ~~docs(metrics): Add docs for WASM stats~~
4. (#2517) ~~feat(metrics): Add WASM feature flag~~
5. (#2546) ~~feat(metrics): Add stats.wasm to proxy config~~
6. (#2554) ~~feat(injector): Add name, workload name, workload kind to PodMetadata~~
7. (#2571) ~~feat(metrics): Add source labels to WASM metrics~~
8. (#2574) ~~feat(metrics): Add dest labels to WASM metrics~~
9. (YOU ARE HERE) **feat(metrics): Add "unknown" for dest labels on local replies**
10. feat(metrics): clean up WASM metrics in Prometheus config
11. feat(metrics): add flag to enable WASM metrics
12. tests(e2e): add WASM metrics test

The remaining changes can be found here: https://github.com/openservicemesh/osm/compare/main...nojnhuh:wasm-metrics

**Affected area**:

- New Functionality      [ ]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [X]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [X]
- SMI Policy             [ ]
- Security               [ ]
- Tests                  [ ]
- CI System              [ ]
- Performance            [ ]
- Other                  [ ]

Please answer the following questions with yes/no.

- Does this change contain code from or inspired by another project? If so, did you notify the maintainers and provide attribution?
No

PR #2574, opened 2021-02-19
feat(metrics): Add dest labels to WASM metrics

nojnhuh:
**Description**:
This change configures RDS to add the destination metadata necessary for
SMI metrics to each response's headers. The headers are then read by the
WASM extension and used to tag the generated statistics. This change has
been integrated with both routes v2 and the existing routes
implementation.

---

This PR is part of several which together extend Envoy to generate custom 
metrics needed to implement the SMI metrics spec (#984).
Here is a map of the PRs as they're currently planned:

1. (#2479) ~~ref(injector): pass pod object to getEnvoySidecarContainerSpec~~
2. (#2488) ~~feat(metrics): add WASM metrics source~~
3. (#2503) ~~docs(metrics): Add docs for WASM stats~~
4. (#2517) ~~feat(metrics): Add WASM feature flag~~
5. (#2546) ~~feat(metrics): Add stats.wasm to proxy config~~
6. (#2554) ~~feat(injector): Add name, workload name, workload kind to PodMetadata~~
7. (#2571) ~~feat(metrics): Add source labels to WASM metrics~~
8. (YOU ARE HERE) **feat(metrics): Add dest labels to WASM metrics**
9. feat(metrics): Add "unknown" for dest labels on local replies
10. feat(metrics): clean up WASM metrics in Prometheus config
11. feat(metrics): add flag to enable WASM metrics
12. tests(e2e): add WASM metrics test

The remaining changes can be found here: https://github.com/openservicemesh/osm/compare/main...nojnhuh:wasm-metrics

**Affected area**:

- New Functionality      [ ]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [X]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [X]
- SMI Policy             [ ]
- Security               [ ]
- Tests                  [ ]
- CI System              [ ]
- Performance            [ ]
- Other                  [ ]

Please answer the following questions with yes/no.

- Does this change contain code from or inspired by another project? If so, did you notify the maintainers and provide attribution?
No

PR #2571, opened 2021-02-19
feat(metrics): Add source labels to WASM metrics

nojnhuh:
**Description**:
This change adds a simple Lua HTTP filter to add the necessary source metadata
for SMI metrics to each request's headers which are read in the WASM
extension and used to tag the generated metrics.

---

This PR is part of several which together extend Envoy to generate custom 
metrics needed to implement the SMI metrics spec (#984).
Here is a map of the PRs as they're currently planned:

1. (#2479) ~~ref(injector): pass pod object to getEnvoySidecarContainerSpec~~
2. (#2488) ~~feat(metrics): add WASM metrics source~~
3. (#2503) ~~docs(metrics): Add docs for WASM stats~~
4. (#2517) ~~feat(metrics): Add WASM feature flag~~
5. (#2546) ~~feat(metrics): Add stats.wasm to proxy config~~
6. (#2554) ~~feat(injector): Add name, workload name, workload kind to PodMetadata~~
7. (YOU ARE HERE) **feat(metrics): Add source labels to WASM metrics**
8. feat(metrics): Add dest labels to WASM metrics
9. feat(metrics): Add "unknown" for dest labels on local replies
10. feat(metrics): clean up WASM metrics in Prometheus config
11. feat(metrics): add flag to enable WASM metrics
12. tests(e2e): add WASM metrics test

The remaining changes can be found here: https://github.com/openservicemesh/osm/compare/main...nojnhuh:wasm-metrics

**Affected area**:

- New Functionality      [ ]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [X]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [X]
- SMI Policy             [ ]
- Security               [ ]
- Tests                  [ ]
- CI System              [ ]
- Performance            [ ]
- Other                  [ ]

Please answer the following questions with yes/no.

- Does this change contain code from or inspired by another project? If so, did you notify the maintainers and provide attribution?
No

PR #2546, opened 2021-02-17
feat(metrics): Add stats.wasm to proxy config

nojnhuh:
**Description**:
This change propagates the stats.wasm module to each sidecar proxy's
config. stats.wasm is built and embedded in the osm-controller binary
and configured inline via LDS.

Follow-up from #2531

---

This PR is part of several which together extend Envoy to generate custom 
metrics needed to implement the SMI metrics spec (#984).
Here is a map of the PRs as they're currently planned:

1. (#2479) ~~ref(injector): pass pod object to getEnvoySidecarContainerSpec~~
2. (#2488) ~~feat(metrics): add WASM metrics source~~
3. (#2503) ~~docs(metrics): Add docs for WASM stats~~
4. (#2517) ~~feat(metrics): Add WASM feature flag~~
5. (YOU ARE HERE) **feat(metrics): Add stats.wasm to proxy config**
6. feat(injector): Add name, workload name, workload kind to PodMetadata
7. feat(metrics): Add source labels to WASM metrics
8. feat(metrics): Add dest labels to WASM metrics
9. feat(metrics): Add "unknown" for dest labels on local replies
10. feat(metrics): clean up WASM metrics in Prometheus config
11. feat(metrics): add flag to enable WASM metrics
12. tests(e2e): add WASM metrics test

The remaining changes can be found here: https://github.com/openservicemesh/osm/compare/main...nojnhuh:wasm-metrics

**Affected area**:

- New Functionality      [ ]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [X]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [X]
- SMI Policy             [ ]
- Security               [ ]
- Tests                  [ ]
- CI System              [ ]
- Performance            [ ]
- Other                  [ ]

Please answer the following questions with yes/no.

- Does this change contain code from or inspired by another project? If so, did you notify the maintainers and provide attribution?
No

PR #2603, opened 2021-02-23
feat(metrics): clean up WASM metrics in Prometheus config

nojnhuh:
**Description**:
The proxy-wasm SDK doesn't directly tag the metrics it generates.
Instead, it encodes the tags in the metric name. This change modifies
the Prometheus config to modify the metrics and move the encoded tags to
actual Prometheus tags.

One limitation to note is that because Prometheus doesn't allow '-' or
'.' in metric names, proxy-wasm converts all instances of either to '_'.
This means a metric on a pod named 'abc-123' will have a pod name tag of
'abc_123' and will collide with the metrics on a pod named 'abc.123'.
This limitation has been documented in docs/content/docs/patterns/wasm.md.

For example,

    envoy_response_code_200_source_namespace_bookbuyer_source_kind_Deployment_source_name_bookbuyer_source_pod_bookbuyer_74ff55c5b_kcb8q_destination_namespace_bookstore_destination_kind_Deployment_destination_name_bookstore_v1_destination_pod_bookstore_v1_74ff55c5b_kcb8q_osm_request_total{}

is transformed into

    osm_request_total{response_code="200", source_namespace="bookbuyer", source_kind="Deployment", source_name="bookbuyer", source_pod="bookbuyer_74ff55c5b_kcb8q", destination_namespace="bookstore", destination_kind="Deployment", destination_name="bookstore_v1", destination_pod="bookstore_v1_74ff55c5b_kcb8q"}

---

1. (#2479) ~~ref(injector): pass pod object to getEnvoySidecarContainerSpec~~
2. (#2488) ~~feat(metrics): add WASM metrics source~~
3. (#2503) ~~docs(metrics): Add docs for WASM stats~~
4. (#2517) ~~feat(metrics): Add WASM feature flag~~
5. (#2546) ~~feat(metrics): Add stats.wasm to proxy config~~
6. (#2554) ~~feat(injector): Add name, workload name, workload kind to PodMetadata~~
7. (#2571) ~~feat(metrics): Add source labels to WASM metrics~~
8. (#2574) ~~feat(metrics): Add dest labels to WASM metrics~~
9. (#2587) ~~feat(metrics): Add "unknown" for dest labels on local replies~~
10. (YOU ARE HERE) **feat(metrics): clean up WASM metrics in Prometheus config**
11. feat(metrics): add flag to enable WASM metrics
12. tests(e2e): add WASM metrics test

The remaining changes can be found here: https://github.com/openservicemesh/osm/compare/main...nojnhuh:wasm-metrics

**Affected area**:

- New Functionality      [ ]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [ ]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [X]
- SMI Policy             [ ]
- Security               [ ]
- Tests                  [ ]
- CI System              [ ]
- Performance            [ ]
- Other                  [ ]

No

### DIFF
--- a/charts/osm/templates/prometheus-configmap.yaml
+++ b/charts/osm/templates/prometheus-configmap.yaml
@@ -111,6 +111,104 @@ data:
           regex: ^ReplicaSet;(.*)-[^-]+$
           target_label: source_workload_name    
 
+      - job_name: 'smi-metrics'
+        kubernetes_sd_configs:
+        - role: pod
+        relabel_configs:
+        - source_labels: [__meta_kubernetes_pod_annotation_prometheus_io_scrape]
+          action: keep
+          regex: true
+        - source_labels: [__meta_kubernetes_pod_annotation_prometheus_io_path]
+          action: replace
+          target_label: __metrics_path__
+          regex: (.+)
+        - source_labels: [__address__, __meta_kubernetes_pod_annotation_prometheus_io_port]
+          action: replace
+          regex: ([^:]+)(?::\d+)?;(\d+)
+          replacement: $1:$2
+          target_label: __address__
+        metric_relabel_configs:
+        - source_labels: [__name__]
+          regex: 'envoy_.*osm_request_(total|duration_ms_(bucket|count|sum))'
+          action: keep
+        - source_labels: [__name__]
+          action: replace
+          regex: envoy_response_code_(\d{3})_source_namespace_.*_source_kind_.*_source_name_.*_source_pod_.*_destination_namespace_.*_destination_kind_.*_destination_name_.*_destination_pod_.*_osm_request_total
+          target_label: response_code
+        - source_labels: [__name__]
+          action: replace
+          regex: envoy_response_code_\d{3}_source_namespace_(.*)_source_kind_.*_source_name_.*_source_pod_.*_destination_namespace_.*_destination_kind_.*_destination_name_.*_destination_pod_.*_osm_request_total
+          target_label: source_namespace
+        - source_labels: [__name__]
+          action: replace
+          regex: envoy_response_code_\d{3}_source_namespace_.*_source_kind_(.*)_source_name_.*_source_pod_.*_destination_namespace_.*_destination_kind_.*_destination_name_.*_destination_pod_.*_osm_request_total
+          target_label: source_kind
+        - source_labels: [__name__]
+          action: replace
+          regex: envoy_response_code_\d{3}_source_namespace_.*_source_kind_.*_source_name_(.*)_source_pod_.*_destination_namespace_.*_destination_kind_.*_destination_name_.*_destination_pod_.*_osm_request_total
+          target_label: source_name
+        - source_labels: [__name__]
+          action: replace
+          regex: envoy_response_code_\d{3}_source_namespace_.*_source_kind_.*_source_name_.*_source_pod_(.*)_destination_namespace_.*_destination_kind_.*_destination_name_.*_destination_pod_.*_osm_request_total
+          target_label: source_pod
+        - source_labels: [__name__]
+          action: replace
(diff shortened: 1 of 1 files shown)


charts/osm/templates/prometheus-configmap.yaml
```diff
@@ -111,6 +111,123 @@ data:
...
+          replacement: $9
+        - source_labels: [__name__]
+          action: replace
+          regex: .*osm_request_total
+          target_label: __name__
+          replacement: osm_request_total
+
+        - source_labels: [__name__]
+          action: replace
+          regex: envoy_source_namespace_(.*)_source_kind_(.*)_source_name_(.*)_source_pod_(.*)_destination_namespace_(.*)_destination_kind_(.*)_destination_name_(.*)_destination_pod_(.*)_osm_request_duration_ms_(bucket|sum|count)
```
eduser25, line 187: Wow. Is the relabeling necessary? I'd have expected the labeling should have worked out of the box with no relabeling needed. But now thinking about, it rings a bell of this issue.

If so:
`.*envoy_source_namespace_(.*)_source_kind_.*` would this work?
I think it will save quite some cycles if you avoided parsing all vars at each relabel.

At current conf, with some pods (400 or more), prometheus it reaches 0.6-0.8 cpu constant cpu usage just by parsing the envoys (probably also computing grafana requests since I'm looking). I was already thinking we will have to reduce the scraping interval at current rate; adding this will make this angle of matters worse.

We should see what can we remove from Line 57 also now that we have this. I honestly hope most of it.

nojnhuh: Yeah, it should work to only capture one group per regex. Originally I only did it this way because it was easier to copy-paste, didn't consider the performance aspect. I'll make that change.

charts/osm/templates/prometheus-configmap.yaml
```diff
@@ -111,6 +111,104 @@ data:
...
+          action: replace
+          regex: envoy_response_code_\d{3}_source_namespace_.*_source_kind_.*_source_name_.*_source_pod_.*_destination_namespace_.*_destination_kind_(.*)_destination_name_.*_destination_pod_.*_osm_request_total
+          target_label: destination_kind
+        - source_labels: [__name__]
+          action: replace
+          regex: envoy_response_code_\d{3}_source_namespace_.*_source_kind_.*_source_name_.*_source_pod_.*_destination_namespace_.*_destination_kind_.*_destination_name_(.*)_destination_pod_.*_osm_request_total
+          target_label: destination_name
+        - source_labels: [__name__]
+          action: replace
+          regex: envoy_response_code_\d{3}_source_namespace_.*_source_kind_.*_source_name_.*_source_pod_.*_destination_namespace_.*_destination_kind_.*_destination_name_.*_destination_pod_(.*)_osm_request_total
```
eduser25, line 168: you are still probably making it too hard on yourself?

if you are really only interested in one label, probably the unique tokens/word before and after it are enough, everything else on the string can just be obviated/`.*`. I this right?
ie. 
```
envoy_response_code_\d{3}_source_namespace_.*_source_kind_.*_source_name_.*_source_pod_.*_destination_namespace_.*_destination_kind_.*_destination_name_.*_destination_pod_(.*)_osm_request_total
```
vs
```
.*_destination_pod_(.*)_osm_request_total.*
```

eduser25 (approved):
I'm not stalling this for the comment I added

charts/osm/templates/prometheus-configmap.yaml
nojnhuh: Talked with @eduser25 offline

My main concern is making sure that the regexes handle weird cases like a pod named "source_namespace" and this was the easiest way to make sure those are handled properly.

I'll plan on keeping these as-is for now and then taking another pass at these later when I get a perf environment set up and can do more edge case testing.

snehachhabria (approved):

PR #2605, opened 2021-02-24
feat(routesv2) : Cut over OSM to build routes using routes v2

snehachhabria:
**Description**:

This PR makes routes v2 the default route building mechanism in OSM.

**Affected area**:

- New Functionality      [ ]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [X]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests                  [ ]
- CI System              [ ]
- Performance            [ ]
- Other                  [ ]


demo/deploy-bookstore.sh
```diff
@@ -23,6 +23,8 @@ spec:
   ports:
   - port: 80
     name: bookstore-port
+  selector:
```
snehachhabria, line 26: I have added the selector on apex service back @draychev, @shashankram

shashankram: FYI, this isn't needed but putting it here doesn't cause any harm for this demo.

tests/e2e/e2e_http_ingress_test.go
```diff
@@ -92,7 +92,8 @@ var _ = OSMDescribe("HTTP ingress",
 
 			ing := &v1beta1.Ingress{
 				ObjectMeta: metav1.ObjectMeta{
-					Name: svcDef.Name,
+					Name:      svcDef.Name,
+					Namespace: destNs,
```
shashankram, line 96: nit: not required because we are deploying it into the destination namespace.

shashankram (approved):

demo/deploy-bookstore.sh
snehachhabria: yes, I added it back since we had it previously

draychev (approved):
Thank you @snehachhabria

tests/scenarios/traffic_split_with_apex_service_test.go
```diff
@@ -129,9 +135,9 @@ var _ = Describe(``+
 
 				expectedWeightedCluster := &xds_route.WeightedCluster{
 					Clusters: []*xds_route.WeightedCluster_ClusterWeight{
-						weightedCluster("bookstore-v2", 10),
+						weightedCluster("bookstore-v2", 100),
```
michelleN, line 138: Curious to know why we changed these weights

michelleN (approved):

tests/scenarios/traffic_split_with_apex_service_test.go
snehachhabria: This is because in routes v2 we build the outbound policies with default weights : https://github.com/openservicemesh/osm/blob/8a19278719861d9adf13c1a1b82418592c4c040a/pkg/catalog/routes.go#L695

The weights will only get updated for outbound virtual hosts that is referenced in a traffic split, otherwise it always remains 100. This is different from routes v1 and hence the change in weights

PR #2617, opened 2021-02-24
go-mocks: generate mocks based on rules

shashankram:
**Description**:
Currently the mocks are not generated consistently
across the code due to inconsistencies arising from:
1. mocks not being generated using the same mode (reflect vs source)
2. manually edited mock files

Both of the above present problems that are not easy to catch. For
example, manually edited mock files don't reflect the interface
verbatim and will introduce bugs.

This change addresses both these problems by:
1. defining mockgen rules using a file to specify the parameters
2. generating mocks the same way for each of the rules
3. integrating checks for the mocks in the CI

**Affected area**:

- New Functionality      [ ]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [ ]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests                  [ ]
- CI System              [ ]
- Performance            [ ]
- Other                  [X]

`No`


eduser25 (approved):

snehachhabria (approved):

draychev (commented):
This is good stuff - thank you @shashankram 
Bummed that this is so resource intensive (takes about a minute) - but glad we have this check! (I'm guilty of not regenning when I should have - so this tool is much appreciated)

PR #2653, opened 2021-03-01
chore(*): Remove duplicate imports

snehachhabria:
**Description**:

This PR removes duplicate imports across the codebase. No functional changes

**Affected area**:

- New Functionality      [ ]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [ ]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests                  [ ]
- CI System              [ ]
- Performance            [ ]
- Other                  [X]

### DIFF
--- a/cmd/cli/mesh_upgrade_test.go
+++ b/cmd/cli/mesh_upgrade_test.go
@@ -6,19 +6,18 @@ import (
 
 	"github.com/stretchr/testify/assert"
 	"helm.sh/helm/v3/pkg/action"
-	helm "helm.sh/helm/v3/pkg/action"
 	"helm.sh/helm/v3/pkg/chartutil"
 	kubefake "helm.sh/helm/v3/pkg/kube/fake"
 	"helm.sh/helm/v3/pkg/storage"
 	"helm.sh/helm/v3/pkg/storage/driver"
 )
 
-func meshUpgradeConfig() *helm.Configuration {
+func meshUpgradeConfig() *action.Configuration {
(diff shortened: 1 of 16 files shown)


cmd/cli/mesh_upgrade_test.go
```diff
@@ -5,7 +5,6 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/assert"
-	"helm.sh/helm/v3/pkg/action"
 	helm "helm.sh/helm/v3/pkg/action"
```
shashankram, line 8: Generally I feel its not worth aliasing these imports if their package names are not versioned (`action` here vs `v1`). Could you remove the non aliased import instead and use `action.` in the code?

pkg/catalog/helpers_test.go
```diff
@@ -9,7 +9,6 @@ import (
 	specs "github.com/servicemeshinterface/smi-sdk-go/pkg/apis/specs/v1alpha4"
 	split "github.com/servicemeshinterface/smi-sdk-go/pkg/apis/split/v1alpha2"
 	corev1 "k8s.io/api/core/v1"
```
shashankram, line 11: This is a good example of why aliasing is helpful, because multiple imports packages could be named `v1` and changing the import alias later on is a bit annoying.

shashankram (approved):

eduser25 (approved):

Issue #2532, opened 2021-02-16
Document the requirement of apex service not having any label selectors matching a pod as an OSM requirement


michelleN:
+ document this

Gets around bug #1979 temporarily as we move forward to fix this

snehachhabria:
The PR #2556 removes the label selector, we just need to document this to caution users

michelleN:
@snehachhabria @shashankram - This issue I believe becomes obsolete since we now have support for multiple pods per service if I'm not mistaken.

michelleN:
Running a manual test now to confirm.

michelleN:
Closing as this is not really necessary and #2654 will take care of all the updates to the manual demo.

Issue #2654, opened 2021-03-01
Update manual demo documentation to reflect routesV2 behavior


shashankram:
**Please describe what should be documented
There are some behavior changes with routesV2 which means the existing manual demo documentation must be updated.

Following need to be updated:
- Due to existing routesV2 implementation, `spec.service` in a traffic split must be the service name without its namespace and not the FQDN. This behavior will be updated later on in routesV2 to accept an FQDN for a service.
- Since traffic split policy is no longer required to allow access to a service, document access policies without dependency on traffic split policy
- Provide clear guidance on how the root and leaf services and pods need to be deployed before traffic splitting and for traffic splitting to work.

**Please suggest where in the repo the document should be located
https://github.com/openservicemesh/osm/blob/main/docs/content/docs/install/manual_demo/_index.md

michelleN:
>Due to existing routesV2 implementation, spec.service in a traffic split must be the service name without its namespace and not the FQDN. This behavior will be updated later on in routesV2 to accept an FQDN for a service.

just an update: #2794 removes the constraint described in the first bullet.

PR #2658, opened 2021-03-01
ref(*): rm label selectors from example root service

michelleN:
* Should help prevent any confusion around label selectors
on root service #2532 

**Description**:

**Affected area**:

- New Functionality      [ ]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [ ]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests                  [ ]
- CI System              [ ]
- Performance            [ ]
- Other                  [ ]

### DIFF
--- a/docs/content/docs/install/manual_demo/_index.md
+++ b/docs/content/docs/install/manual_demo/_index.md
@@ -327,6 +327,8 @@ The counter in the `bookthief` window will start incrementing.
 
 We will now demonstrate how to balance traffic between two Kubernetes services, commonly known as a traffic split. We will be splitting the traffic directed to the root `bookstore` service between the backends `bookstore-v1` service and `bookstore-v2` service.
 
+_Note:_ The root Kubernetes service should not have any label selectors defined that match a/any pod on the mesh. This may lead to unintended behavior.
+
 #### Split Traffic to v2 of Bookstore
 
 Deploy an SMI TrafficTarget policy to allow `bookbuyer` and `bookthief` to access the `bookstore-v2` service:


docs/content/docs/install/manual_demo/_index.md
```diff
@@ -327,6 +327,8 @@ The counter in the `bookthief` window will start incrementing.
 
 We will now demonstrate how to balance traffic between two Kubernetes services, commonly known as a traffic split. We will be splitting the traffic directed to the root `bookstore` service between the backends `bookstore-v1` service and `bookstore-v2` service.
 
+_Note:_ The root Kubernetes service should not have any label selectors defined for it. This may lead to unintended behavior.
```
snehachhabria, line 330:
```suggestion
_Note:_ The root Kubernetes service should not have any label selectors defined that match a/any pod on the mesh. This may lead to unintended behavior.
```

docs/example/manifests/apps/bookstore.yaml
```diff
@@ -16,8 +16,6 @@ spec:
   ports:
   - port: 80
     name: bookstore-port
-  selector:
-    app: bookstore
```
shashankram, line 20: We need to retain this to demo permissive mode with the current manual demo documentation.

shashankram (changes requested):

docs/content/docs/install/manual_demo/_index.md
snehachhabria: I am not sure is we should say do not sure label selectors completely or we should say don't match them to a pod

docs/content/docs/install/manual_demo/_index.md
```diff
@@ -327,6 +327,8 @@ The counter in the `bookthief` window will start incrementing.
 
 We will now demonstrate how to balance traffic between two Kubernetes services, commonly known as a traffic split. We will be splitting the traffic directed to the root `bookstore` service between the backends `bookstore-v1` service and `bookstore-v2` service.
 
+_Note:_ The root Kubernetes service should not have any label selectors defined that match a/any pod on the mesh. This may lead to unintended behavior.
+
```
shashankram, line 331: @michelleN wouldn't this be the case if we split traffic to the root service to itself and another service, as we discussed offline?

michelleN: yes actually. Going to close out this PR and let #2654 take care of the demo updates.

Issue #2444, opened 2021-02-03
Docs broken link and typo


swartz-k:
a. a little typo in [ROADMAP.md](https://github.com/openservicemesh/osm/blob/main/ROADMAP.md#L14)
![image](https://user-images.githubusercontent.com/49771587/106757949-ec7bc400-666b-11eb-8234-3962e6d88c8d.png)

b. While reading [docs](https://docs.openservicemesh.io/docs/patterns/observability/metrics/), there are some images link broken. Like 

![image](https://user-images.githubusercontent.com/49771587/106756210-00bec180-666a-11eb-969b-4ca77b864fb0.png)

PR #2662, opened 2021-03-01
Correcting broken link

bridgetkromhout:
While this isn't how we probably want to store such videos, this is the shortest path to fixing the broken link. We also will want to decide:

- [ ] Does this video need to be updated?
- [ ] Do we want this video to have its path be under the docs folder?

Signed-off-by: Bridget Kromhout <bridget@kromhout.org>

### DIFF
--- a/docs/content/docs/tasks_usage/metrics.md
+++ b/docs/content/docs/tasks_usage/metrics.md
@@ -194,7 +194,7 @@ Sample result will be:
 # Grafana
 [Grafana][3] is an open source visualization and analytics software. It allows you to query, visualize, alert on, and explore your metrics.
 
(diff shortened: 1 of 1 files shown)


bridgetkromhout:
Issue noted in https://github.com/openservicemesh/osm/issues/2444

Was already working on https://github.com/openservicemesh/osm/blob/main/docs/content/docs/tasks_usage/metrics.md#grafana but broken on the website.

Fixed in https://deploy-preview-2662--osm-docs.netlify.app/docs/tasks_usage/metrics/#grafana and verified still working in https://github.com/bridgetkromhout/osm/blob/grafana-demo/docs/content/docs/tasks_usage/metrics.md#grafana.

shashankram (approved):

PR #2663, opened 2021-03-01
tests(fluent bit): add e2e to test fluent bit for correct output

### LINKED ISSUE #2159: Add e2e to test accuracy of logs forwarded via fluent bit

Currently, there is an e2e to check whether the Fluent Bit sidecar is running when enabled and not running when disabled. We need an additional e2e that checks whether all the expected logs are being correctly forwarded. This will ensure that the filters and output plugin are working correctly. 

- Logging                    [x]

**Possible use cases**

This test is important to be able to make sure no logs are being missed as well as that no extra logs are being sent.

SanyaKochhar:
**Description**: Tests that log output from Fluent Bit is filtered correctly:
* Modify filter for adding osm controller pod name
* Grep filter for log level

Resolves #2159 

**Affected area**:

- New Functionality      [ ]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [ ]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests                  [x]
- CI System              [ ]
- Performance            [ ]
- Other                  [ ]

no

### DIFF
--- a/tests/e2e/e2e_fluentbit_out_test.go
+++ b/tests/e2e/e2e_fluentbit_out_test.go
@@ -1,0 +1,98 @@
+package e2e
+
+import (
+	"bytes"
+	"context"
+	"strings"
+	"time"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/util/wait"
+
+	. "github.com/openservicemesh/osm/tests/framework"
+)
+
+var _ = OSMDescribe("Test deployment of Fluent Bit sidecar",
+	OSMDescribeInfo{
+		Tier:   2,
+		Bucket: 2,
+	},
+	func() {
+		Context("FluentbitOutput", func() {
+			It("Forwards correctly filtered logs to stdout", func() {
+				// Install OSM with Fluentbit
+				installOpts := Td.GetOSMInstallOpts()
+				installOpts.DeployFluentbit = true
+				installOpts.SetOverrides = []string{"OpenServiceMesh.fluentBit.logLevel=info"}
+				Expect(Td.InstallOSM(installOpts)).To(Succeed())
+
+				pods, err := Td.Client.CoreV1().Pods(Td.OsmNamespace).List(context.TODO(), metav1.ListOptions{
+					LabelSelector: labels.SelectorFromSet(map[string]string{"app": "osm-controller"}).String(),
+				})
+				Expect(err).NotTo(HaveOccurred())
+
+				// Query fluentbit-logger container logs to test if Fluent Bit filters are working
+				fluentBitContainerFound := false
+				for _, pod := range pods.Items {
+					// Wait until osm-controller has generated a log to check against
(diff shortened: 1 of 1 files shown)


michelleN (commented):
Looks like this code needs to be formatted with gofmt. Try running `gofmt -s -w` in the root directory and commit those formatting changes with your PR.

tests/e2e/e2e_fluentbit_out_test.go
```diff
@@ -0,0 +1,89 @@
...
+					LabelSelector: labels.SelectorFromSet(map[string]string{"app": "osm-controller"}).String(),
+				})
+				Expect(err).NotTo(HaveOccurred())
+
+				// Query fluentbit-logger container logs to test if Fluent Bit filters are working
+				for _, pod := range pods.Items {
+					// Wait until osm-controller has generated a log to check against
+					wait.Poll(time.Second*3, time.Second*180, isLogEmitted(pod.Namespace, pod.Name, "osm-controller", "\"level\":\"info\""))
+					for _, container := range pod.Spec.Containers {
+						if strings.Contains(container.Image, "fluent-bit") {
```
SanyaKochhar, line 43: @nojnhuh a little concerned if this `if` condition is not met, the test will pass since all the relevant expects are nested here

nojnhuh: That's true. I suppose you could have some sort of `fluentBitContainerFound` variable outside the loop, set it to true when you find it, then fail after the loop if it's false.

SanyaKochhar: ooh yeah sounds good

tests/e2e/e2e_fluentbit_out_test.go
```diff
@@ -0,0 +1,98 @@
...
+					LabelSelector: labels.SelectorFromSet(map[string]string{"app": "osm-controller"}).String(),
+				})
+				Expect(err).NotTo(HaveOccurred())
+
+				// Query fluentbit-logger container logs to test if Fluent Bit filters are working
+				fluentBitContainerFound := false
+				for _, pod := range pods.Items {
+					// Wait until osm-controller has generated a log to check against
+					logLevel := "\"level\":\"info\""
+					err := waitForLogEmission(pod.Namespace, pod.Name, "osm-controller", logLevel)
```
nojnhuh, line 43: This wait might actually not be needed. Since the osm-controller will definitely log some info messages before it becomes ready, and `Td.InstallOSM` already waits for the pod to be ready, by the time you get here the pod should already have logged some messages. Not a big deal though to check anyway.

This assumption is unique to info messages though, so if the test is ever changed to look for different levels, this will need to be reevaluated.

nojnhuh (approved):
One small, likely inconsequential comment. Looks good.

tests/e2e/e2e_fluentbit_out_test.go
SanyaKochhar: So I did see the test fail once because fluent bit took a bit longer to log to stdout than osm-controller. I think the poll gives it extra time to ensure that it won't be flakey. 

What you're saying is def a fair assumption, but I think it's also worth leaving in there in the case of changing log level. Ideally it would be useful to change this to test error level in the future. Since it is info rn, the wait should return true immediately based on what you're saying

michelleN (approved):

PR #2667, opened 2021-03-02
e2e: Add pod restart checks and Fail tests if seen by default

eduser25:
None of the tests we currently have should be seeing pod restarts.
This is put in place to rule out unwanted restart issues, but other
tests implementing restarts should be put in place.

**Affected area**:

- Control Plane          [x ]
- Tests                  [ x]

No

### DIFF
--- a/tests/framework/common.go
+++ b/tests/framework/common.go
@@ -157,6 +157,7 @@ type OsmTestData struct {
 
 	CleanupTest    bool // Cleanup test-related resources once finished
 	WaitForCleanup bool // Forces test to wait for effective deletion of resources upon cleanup
+	IgnoreRestarts bool // Ignore control plane processes restarts, if any
 
 	// OSM install-time variables
 	InstType          InstallType // Install type.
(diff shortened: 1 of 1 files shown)


SanyaKochhar (approved):

snehachhabria (approved):

shashankram (commented):
This will break for the test where the controller is intentionally restarted. Please fix this.

shashankram:
> This will break for the test where the controller is intentionally restarted. Please fix this.

Luckily, we don't restart the controller but rather delete it, and spin up a new one. Hope that doesn't mess with the verification checks!

eduser25:
good. I did check the restart increase part by issuing sigabort on the container application on the node itself. Did not remember we had a restart test...

shashankram:
> good. I did check the restart increase part by issuing sigabort on the container application on the node itself. Did not remember we had a restart test...

I didn't either, and its not even a restart test, but rather a new controller pod! If we had a way to force exit controller via an endpoint like `/forcequit`, we could test actual restart as well.

PR #2668, opened 2021-03-02
Fixing uninstall link

bridgetkromhout:
This may fix a broken link.

Signed-off-by: Bridget Kromhout <bridget@kromhout.org>


bridgetkromhout:
will experiment more later

PR #2677, opened 2021-03-02
docs/install: push additional images with latest_main tag

shashankram:
**Description**:
In addition to images pushed to the container registry
using the Git commit SHA on push events to the `main`
branch, this change pushes images with the `latest_main`
tag so that OSM cli built from latest source can use
the `latest_main` image tag. This simplifies the install
experience when using cli built from source. Previously
this would most likely fail due to version mismatches
between the cli built from source and the default
image versions (pointing to the previous release)
embedded in the cli.

**Affected area**:

- New Functionality      [ ]
- Documentation          [X]
- Install                [ ]
- Control Plane          [ ]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests                  [ ]
- CI System              [ ]
- Performance            [ ]
- Other                  [X]

`No`

### DIFF
--- a/docs/content/docs/install/_index.md
+++ b/docs/content/docs/install/_index.md
@@ -19,17 +19,31 @@ Unpack the `osm` binary and add it to `$PATH` to get started.
 
 ### From Source (Linux, MacOS)
 
-Building OSM from source requires more steps but is the best way to test the latest changes and useful in a development environment.
+Building OSM CLI from source requires more steps but is the best way to test the latest changes and useful in a development environment.
 
 You must have a working [Go](https://golang.org/doc/install) environment.
 
 ```console
 $ git clone git@github.com:openservicemesh/osm.git
 $ cd osm
+```
+
+When OSM CLI is built from source, it might not be compatible with the default control plane images derived from the latest released version. In such cases, use the `latest_main` image tag to specify that the CLI should use the latest control plane images built from the `main` branch. It is recommended to build the OSM CLI binary from the latest commit on the `main` branch while using the `latest_main` image tag for compatibility reasons.
+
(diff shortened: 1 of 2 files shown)


docs/content/docs/install/_index.md
```diff
@@ -19,18 +19,32 @@ Unpack the `osm` binary and add it to `$PATH` to get started.
...
+Building OSM CLI from source requires more steps but is the best way to test the latest changes and useful in a development environment.
 
 You must have a working [Go](https://golang.org/doc/install) environment.
 
 ```console
 $ git clone git@github.com:openservicemesh/osm.git
 $ cd osm
+```
+
+When OSM CLI is built from source, it might not be compatible with the default control plane images derived from the latest released version. In such cases, use the `latest_main` image tag to specify that the CLI should use the latest control plane images built from the `main` branch. It is recommended to build the OSM CLI binary from the latest commit on the `main` branch while using the `latest_main` image tag for compatibility reasons.
```
nojnhuh, line 31: Should we make the "default" recommendation when building the CLI from source to also build the images from the same source? That's guaranteed to work, or at least if issues are found then those are definitely bugs that need to be fixed. Then the `latest_main` images can be a convenient shortcut that will _usually_ work, but there's still some incompatibility risks.

shashankram: I thought about this. I think building controller images from source also means we would need to document information regarding container registry, tags, ImagePullSecrets etc. In the short term, `latest_main` could provide a reasonable alternative.

PR #2706, opened 2021-03-03
e2e/ci: add log collection

eduser25:
NOTE: Removed `forceCollectLogs`. To see the previous run, which was
getting all logs from all tests, see
https://github.com/openservicemesh/osm/actions/runs/619287544

Logs are only collected for tests which fail to avoid storage overuse.
(~20MB per e2e pipeline)

Adds log collection on CI steps for all workflows that run e2e tests.
By default, logs will only be collected for failing tests.

There's no good way to get test name to name the test folder.
For now it's using TestID, which is a generated number. When looking for
the logs of a failing test, TestID will have to be first looked up on the
test output, ie:

```
Tests HTTP traffic from Clients to the traffic split Cluster IP
  /home/eserra/src/osm/tests/scale/scale_trafficSplit_test.go:29
[BeforeEach] [Top Level]
  /home/eserra/src/osm/tests/framework/common.go:64
ID for test: 2446013194   <-------------------------------------
Test dir (abs): /home/eserra/src/osm/tests/scale/testResults
[It] Tests HTTP traffic from Clients to the traffic split Cluster IP
```

The log zip on GH pipelines will look like:
```
pr_test_logs.zip
  |--test-2446013194/
     |--logs/
        |--<pod_name>/
           |--[running|previous]_<container_name>.log
  |--test-845....
  ....
```
When using kind, also gets Kind logs:
```
kind export logs ...
/tmp/test-...
```

- Tests                  [X]
- CI System              [X]

No


.github/workflows/main.yml
```diff
@@ -185,18 +185,36 @@ jobs:
...
+          path: ./tests/e2e/test*
       - name: Run push tests
         if: ${{ github.event_name == 'push' }}
-        run: go test ./tests/e2e -test.v -ginkgo.v -ginkgo.progress -installType=KindCluster -test.timeout 0 -test.failfast -ginkgo.failFast -ginkgo.focus='\[Bucket ${{ matrix.bucket }}\]' -ginkgo.skip='Upgrade'
+        run: go test ./tests/e2e -test.v -ginkgo.v -ginkgo.progress -installType=KindCluster -test.timeout 0 -test.failfast -ginkgo.failFast -ginkgo.focus='\[Bucket ${{ matrix.bucket }}\]' -ginkgo.skip='Upgrade' -forceCollectLogs=true
+      - name: Upload push test logs
+        uses: actions/upload-artifact@v2
+        with:
+          name: push_test_logs
+          path: ./tests/e2e/test*
```
eduser25, line 208: ah... I need to clean the previous test logs or else they will also be swept in

shashankram (approved):

shalier (approved):

draychev (approved):
This is good stuff!
Thank you!

PR #2733, opened 2021-03-04
ref(docs): update manual guide for routes v2

### LINKED ISSUE #2654: Update manual demo documentation to reflect routesV2 behavior

michelleN:
+ updates examples so that traffic split no longer
required for traffic access demo
+ removes bookstore-v1 manifests and uses bookstore manifests for
fewer, simpler demo steps
+ add links to directly install apps from url so user
does not have to be in docs directory locally
+ some rewording for consistent language

resolves #2654

**Description**:

**Affected area**:

- New Functionality      [ ]
- Documentation          [ X]
- Install                [ ]
- Control Plane          [ ]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests                  [ ]
- CI System              [ ]
- Performance            [ ]
- Other                  [ ]


shashankram (commented):
There are more manifests in subsequent sections in the docs that are referencing local file paths. Seems as though some of them can be updated to point to URLs as well.

docs/example/manifests/split/traffic-split-50-50.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: split.smi-spec.io/v1alpha2
+kind: TrafficSplit
+metadata:
+  name: bookstore-split
+  namespace: bookstore
+spec:
+  service: bookstore.bookstore # <root-service>.<namespace>
```
shashankram, line 7: Since k8s service FQDN can be a short form as well (service, service.namespace etc.), we should clarify with a comment what are the possible allowed values.

shashankram (approved):
Looks good, will test!

docs/example/manifests/split/traffic-split-50-50.yaml
michelleN: added a comment in traffic-split-v1.yaml and also in the doc

nojnhuh (approved):
I ran through the whole demo top to bottom and didn't run into any issues as long as I was careful to build the CLI and images from source and use those.

The only comments I had look out of scope of these changes, so I'll open up separate issues/PRs for those.

PR #2750, opened 2021-03-05
tcp-client/serv: move all logging to log

eduser25:
And setting log level to Debug.

**Affected area**:

- Tests                  [X]
- CI System              [X]

No

### DIFF
--- a/demo/cmd/tcp-client/tcp-client.go
+++ b/demo/cmd/tcp-client/tcp-client.go
@@ -19,6 +19,7 @@ import (
 
 var (
 	log                   = logger.NewPretty("tcp-client")
+	logLevel              = flag.String("logLevel", "debug", "Log output level")
 	serverAddress         = flag.String("server-address", "", "address (ip:port) to connect to")
 	requestPrefix         = "request:"
 	connectRetryDelay     = 3 * time.Second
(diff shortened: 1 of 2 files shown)


demo/cmd/tcp-client/tcp-client.go
```diff
@@ -53,27 +58,25 @@ func main() {
 				log.Error().Err(writeErr).Msg("Write error")
 				continue
 			} else {
-				fmt.Printf("Wrote %d bytes, msg written: [%s]\n", bytesWritten, requestMsg)
+				log.Debug().Msgf("Wrote %d bytes, msg written: [%s]\n", bytesWritten, requestMsg)
```
shashankram, line 61: This should be info, because it logs what message is being sent so it can be compared with the echo response.

demo/cmd/tcp-echo-server/tcp-echo-server.go
```diff
@@ -53,21 +58,22 @@ func echoResponse(conn net.Conn) {
 		switch err {
 		case nil:
 			// respond to the request, prepend the prefix to the response
-			fmt.Printf("Received request: [%s]\n", requestMsg)
+			log.Debug().Msgf("Received request: [%s]", requestMsg)
```
shashankram, line 61: This should be Info().

draychev (approved):
Much needed!
Thank you @eduser25

PR #2752, opened 2021-03-05
fix(chart): template jaeger address to use osm.namespace value

### LINKED ISSUE #2585: Tracing address should not have namespace hardcoded

**Possible use cases**

The tracing address [here](https://github.com/openservicemesh/osm/blob/main/charts/osm/values.yaml#L125) is hardcoded to osm-system. This should be changed to release namespace as the default value or a user-provided value if provided. 

Will likely require a new template that uses the `osm.namespace` [template](https://github.com/openservicemesh/osm/blob/main/charts/osm/templates/_helpers.tpl)

SanyaKochhar:

**Description**: Resolves #2585 

**Affected area**:

- New Functionality      [ ]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [ ]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests                  [ ]
- CI System              [ ]
- Performance            [ ]
- Other                  [x]

no

### DIFF
--- a/charts/osm/templates/_helpers.tpl
+++ b/charts/osm/templates/_helpers.tpl
@@ -2,3 +2,9 @@
 {{- define "osm.namespace" -}} 
 {{ default .Release.Namespace .Values.OpenServiceMesh.osmNamespace}} 
 {{- end -}}
+
+{{/* Default tracing address */}}
+{{- define "osm.tracingAddress" -}}
+{{- $address := printf "jaeger.%s.svc.cluster.local" (include "osm.namespace" .) -}}
+{{ default $address .Values.OpenServiceMesh.tracing.address}} 
+{{- end -}}
--- a/charts/osm/templates/osm-configmap.yaml
+++ b/charts/osm/templates/osm-configmap.yaml
@@ -13,7 +13,7 @@ data:
 
   tracing_enable: {{ .Values.OpenServiceMesh.tracing.enable | quote }}
 {{- if .Values.OpenServiceMesh.tracing.enable }}
-  tracing_address: {{ .Values.OpenServiceMesh.tracing.address | quote }}
+  tracing_address: {{ include "osm.tracingAddress" . | quote }}
   tracing_port: {{ .Values.OpenServiceMesh.tracing.port | quote }}
   tracing_endpoint: {{ .Values.OpenServiceMesh.tracing.endpoint | quote }}
 {{- end }}
--- a/charts/osm/README.md
+++ b/charts/osm/README.md
@@ -106,7 +106,7 @@ The following table lists the configurable parameters of the osm chart and their
 | OpenServiceMesh.replicaCount | int | `1` | `osm-controller` replicas |
 | OpenServiceMesh.serviceCertValidityDuration | string | `"24h"` | Sets the service certificatevalidity duration |
 | OpenServiceMesh.sidecarImage | string | `"envoyproxy/envoy-alpine:v1.17.1"` | Envoy sidecar image |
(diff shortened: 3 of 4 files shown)


snehachhabria (approved):

charts/osm/templates/_helpers.tpl
```diff
@@ -2,3 +2,9 @@
 {{- define "osm.namespace" -}} 
 {{ default .Release.Namespace .Values.OpenServiceMesh.osmNamespace}} 
 {{- end -}}
+
+{{/* Default tracing address */}}
+{{- define "osm.tracingAddress" -}}
+{{- $address := printf "jaeger.%s.svc.cluster.local" (include "osm.namespace" .) -}}
+{{ default $address .Values.OpenServiceMesh.tracing.address}}
```
shashankram, line 9: We don't need a default if we are always going to use `jaegar.<osm-namespace>.svc.cluster.local`. We could just remove the default from `values.yaml`?

charts/osm/templates/osm-configmap.yaml
```diff
@@ -13,7 +13,7 @@ data:
 
   tracing_enable: {{ .Values.OpenServiceMesh.tracing.enable | quote }}
 {{- if .Values.OpenServiceMesh.tracing.enable }}
-  tracing_address: {{ .Values.OpenServiceMesh.tracing.address | quote }}
+  tracing_address: {{ include "osm.tracingAddress" . | quote }}
```
nojnhuh, line 16: Doesn't this break BYO Jaeger? e.g. If I already have Jaeger installed behind a service called "my-jaeger" outside the OSM namespace, how would I point to that? I think you could workaround it with the `tpl` function: https://helm.sh/docs/howto/charts_tips_and_tricks/#using-the-tpl-function

shashankram: Can we just make the entire address configurable instead of the namespace?

nojnhuh: That's how it is today where you just change the `tracing.address` value directly. The problem is the default isn't responsive to when you also set `osmNamespace`, so anytime you override `osmNamespace` you also have to override `tracing.address` if you intend to point to the Jaeger instance installed by the chart.

SanyaKochhar: I'm confused as to why this would break BYO Jaeger? The default function makes it so that it only uses the templated value if the value is left empty in `values.yaml` . The assumption is that users will replace the address if they bring up their own instance, which is what is documented under BYO tracing in [tracing.md](https://github.com/openservicemesh/osm/blob/main/docs/content/docs/patterns/observability/tracing.md)

SanyaKochhar: After discussing with @eduser25, it seemed like it would make sense to use this template but also set the `values.yaml` value to empty + add a comment in values.yaml and in tracing.md explaining the default. The rationale: this value is only used when `--enable-tracing` is used for the BYO experience but even so, OSM will still not bring up the tracing instance. So there is a default but we expect users to override it

nojnhuh: Ah, yep. I missed the last `{{ default }}` line in the template. That should work for BYO. Should the default in values.yaml be an empty string then?

SanyaKochhar: yes that's what we were thinking of changing it to

nojnhuh: Yep, sounds good to me. I should've hit refresh before asking that 😄.

nojnhuh (approved):
LGTM pending rebase

eduser25 (approved):

PR #2774, opened 2021-03-08
envoy: wait indefinitely on TLS secrets

shashankram:
**Description**:
When configuring transport sockets for clusters and
listeners, it is possible for the SDS secret fetch
to timeout (defaults to 15s). This timeout poses problems
because after timeout, the secret is never fetched,
leaving the proxy with broken TLS connections. This breaks
the eventual consistency model. The side effect of using
a timeout is that clusters and listeners that are warming
become active after the timeout, leading to connection
resets and rejections.

This change makes the proxy indefinitely wait till
it can retrieve the SDS secrets referenced in the
transport socket configurations of clusters and
listeners.

Potentially addresses #2749

**Affected area**:

- New Functionality      [ ]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [X]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [X]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests                  [ ]
- CI System              [ ]
- Performance            [ ]
- Other                  [ ]

`No`

### DIFF
--- a/pkg/envoy/xdsutil.go
+++ b/pkg/envoy/xdsutil.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"fmt"
 	"strings"
+	"time"
 
 	xds_accesslog_filter "github.com/envoyproxy/go-control-plane/envoy/config/accesslog/v3"
 	xds_core "github.com/envoyproxy/go-control-plane/envoy/config/core/v3"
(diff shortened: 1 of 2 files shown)


eduser25 (approved):
We'll see how it goes

nojnhuh (approved):

snehachhabria (approved):

PR #2791, opened 2021-03-09
test(pkg/service): Add tests for service package

snehachhabria:
**Description**:

Add tests for service package

Fixes: #2688 

**Affected area**:

- New Functionality      [ ]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [ ]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests                  [X]
- CI System              [ ]
- Performance            [ ]
- Other                  [ ]

### DIFF
--- a/pkg/service/service_test.go
+++ b/pkg/service/service_test.go
@@ -86,3 +86,80 @@ func TestServerName(t *testing.T) {
 	actual := namespacedService.ServerName()
 	assert.Equal("service-name-here.namespace-here.svc.cluster.local", actual)
 }
+
+func TestEquals(t *testing.T) {
+	assert := tassert.New(t)
+
+	testCases := []struct {
+		name           string
+		service        MeshService
+		anotherService MeshService
+		isEqual        bool
+	}{
+		{
+			name: "services are equal",
+			service: MeshService{
+				Namespace: "default",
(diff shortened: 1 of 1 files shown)


nojnhuh (approved):

ksubrmnn (approved):

pkg/service/service_test.go
```diff
@@ -86,3 +86,80 @@ func TestServerName(t *testing.T) {
...
+				Namespace: "default",
+				Name:      "bookbuyer",
+			},
+			anotherService: MeshService{
+				Namespace: "default",
+				Name:      "bookbuyer",
+			},
+			isEqual: true,
+		},
+		{
```
ksubrmnn, line 111: Why are there two tests here? Aren't they functionally the same since the Namespace is a hardcoded string in either case

snehachhabria: The two test cases is for two scenarios: one when the two services are the same ( expected test result is true) and the other when they are not the same ( expected test result is false). Namespace has nothing to do with it, so I am not 100% sure what you mean.

PR #2792, opened 2021-03-10
workerpool: introduce workerpool concept and unit tests

eduser25:
Workerpool is an implementation of the thread-pool paradigm
in Go. The benefits of it in Go however, can be quite different
from any other language able to schedule themselves on system threads.

By using a workpool model, the main focus and intention is to limit the
number of goroutines that can do busy-work and get scheduled concurrently
at any point in time.

Too many goroutines being scheduled at the same time will cause other
goroutines (maybe more critical ones) to be scheduled less often, thus
incurring in resource starvation on those and potentially triggering other
issues.

By being able to queue up work, we should be able to run a more deterministic
runtime (despite Go's nature, this we will not be able to help), less dependant
on the scheduler and more accurate in terms of time, as now the number of routines
doing busy work can be made constant at the expense of serialization, as opposed to
having O(N) routines attempting to run at the same time.

- New Functionality      [x]
- Performance            [x]

No

### DIFF
--- a/pkg/workerpool/workerpool.go
+++ b/pkg/workerpool/workerpool.go
@@ -1,0 +1,125 @@
+package workerpool
+
+import (
+	"runtime"
+	"sync"
+	"sync/atomic"
+	"time"
+
+	"github.com/openservicemesh/osm/pkg/logger"
+)
+
+const (
+	// Size of the job queue per worker
+	maxJobPerWorker = 4096
+)
+
+var (
+	log = logger.New("workerpool")
+)
+
+// worker context for a worker routine
+type worker struct {
+	id            int
+	jobs          chan Job        // Job queue
+	stop          chan struct{}   // Stop channel
+	wg            *sync.WaitGroup // Pointer to WorkerPool wg
+	jobsProcessed uint64          // Jobs processed by this worker
+}
+
+// WorkerPool object representation
+type WorkerPool struct {
+	wg            sync.WaitGroup // Sync group, to stop workers if needed
+	workerContext []*worker      // Worker contexts
+	nWorkers      uint64         // Number of workers. Uint64 for easier mod hash later
+	rRobinCounter uint64         // Used only by the round robin api. Modified atomically on API.
+}
+
+// Job is a runnable interface to queue jobs on a WorkerPool
+type Job interface {
+	JobName() string // Returns name for a job
+	Hash() uint64    // Returns a uint64 hash for a job.
+	Run()            // Executes the job
+}
+
+// NewWorkerPool creates a new work group.
+// If nWorkers is 0, will poll goMaxProcs to get the number of routines to spawn.
+// Reminder: routines are never pinned to system threads, it's up to the go scheduler to decide
+// when and where these will be scheduled.
+func NewWorkerPool(nWorkers int) *WorkerPool {
+	if nWorkers == 0 {
+		// read GOMAXPROCS, -1 to avoid changing it
+		nWorkers = runtime.GOMAXPROCS(-1)
+	}
+
+	log.Info().Msgf("New worker pool setting up %d workers", nWorkers)
+
+	var workPool WorkerPool
+	for i := 0; i < nWorkers; i++ {
+		workPool.workerContext = append(workPool.workerContext,
+			&worker{
+				id:            i,
+				jobs:          make(chan Job, maxJobPerWorker),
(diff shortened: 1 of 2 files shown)


pkg/workerpool/workerpool.go
```diff
@@ -0,0 +1,130 @@
...
+	// Size of the job queue per worker
+	maxJobPerWorker = 4096
+)
+
+var (
+	log = logger.New("workerpool")
+)
+
+// Worker context for a worker routine
+type Worker struct {
```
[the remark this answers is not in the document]
eduser25: yes it can I think. will do

pkg/workerpool/workerpool_test.go
```diff
@@ -0,0 +1,102 @@
...
+
+		wp.AddJob(&joblist[i])
+	}
+
+	// Verifiy all jobs ran through the workers
+	for i := 0; i < njobs; i++ {
+		<-joblist[i].jobDone
+	}
+
+	wp.Stop()
```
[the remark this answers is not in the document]
eduser25: This one's tricky but doable. I thought of a [quiescing](https://en.wikipedia.org/wiki/Quiesce) mechanism, which could be ie. putting a job in each queue and waiting for them to return us something, to ensure all queues are flushed to that point.

pkg/workerpool/workerpool.go
```diff
@@ -0,0 +1,130 @@
...
+	"runtime"
+	"sync"
+	"sync/atomic"
+	"time"
+
+	"github.com/openservicemesh/osm/pkg/logger"
+)
+
+const (
+	// Size of the job queue per worker
```
shashankram, line 13: A comment on how this number should be determined will help.

pkg/workerpool/workerpool.go
```diff
@@ -0,0 +1,130 @@
...
+// Job is a runnable interface to queue jobs on a WorkerPool
+type Job interface {
+	JobName() string // Returns name for a job
+	Hash() uint64    // Returns a uint64 hash for a job.
+	Run()            // Executes the job
+}
+
+// NewWorkerPool creates a new work group.
+// If nWorkers is 0, will poll goMaxProcs to get the number of routines to spawn.
+// Reminder routines are never pinned to system threads, it's up to the go scheduler to decide
```
shashankram, line 47: s/reminder/remainder ?

pkg/workerpool/workerpool.go
```diff
@@ -0,0 +1,130 @@
...
+	}
+
+	log.Info().Msgf("New worker pool setting up %d workers", nWorkers)
+
+	workPool := WorkerPool{
+		wg:            sync.WaitGroup{},
+		workerContext: []*Worker{},
+		rRobinCounter: 0,
+		nWorkers:      0,
+	}
```
shashankram, line 62: `var workpool WorkerPool` doesn't suffice?

pkg/workerpool/workerpool.go
```diff
@@ -0,0 +1,130 @@
...
+		case j := <-workContext.jobs:
+			t := time.Now()
+			log.Debug().Msgf("work[%d]: Starting %v", workContext.id, j.JobName())
+
+			// Run current job
+			j.Run()
+
+			log.Debug().Msgf("work[%d][%s] : took %v", workContext.id, j.JobName(), time.Since(t))
+			workContext.JobsProcessed++
+		case <-workContext.stop:
```
shashankram, line 125: nit: leave line between cases

shashankram (approved):

pkg/workerpool/workerpool.go
eduser25: I don't have measures for it yet. This happens to be roughly 8-to-10 times the current (expected) number of jobs we will have for what I believe will be its primary use case in short term, but this is complete ballparking.
As soon as we have integration changes, I plan to put in place some Prom counters to understand how the queues do under stress, which will later help determine better this number.

pkg/workerpool/workerpool.go
eduser25: oh. I thought slices in a struct require init (either copy/move, or `make()`)

shashankram: Since its a slice, it will allocate on append if current capacity is not enough. I'd be surprised if default initialization doesn't work.

nojnhuh (approved):

PR #2794, opened 2021-03-10
fix(pkg/catalog): resolve root service hostnames

michelleN:
fix(pkg/catalog): resolve root service hostnames

the routes v2 logic was assuming that the root service was
a kubernetes service with no namespace suffix. So specifying
"bookstore.bookstore" as the root service didn't work.
There is more work to be done here to make sure that we follow
the SMI spec which states that a root service should be an
fqdn but for the time being, this fix allows us to run the manual
demo successfully.

Followup: #2797 

**Description**:

**Affected area**:

- New Functionality      [ ]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [ ]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests                  [ ]
- CI System              [ ]
- Performance            [ ]
- Other                  [ ]

### DIFF
--- a/pkg/catalog/routes.go
+++ b/pkg/catalog/routes.go
@@ -61,8 +61,8 @@ func (mc *MeshCatalog) listOutboundTrafficPoliciesForTrafficSplits(sourceNamespa
 	apexServices := mapset.NewSet()
 	for _, split := range mc.meshSpec.ListTrafficSplits() {
 		svc := service.MeshService{
-			Name:      split.Spec.Service,
-			Namespace: split.ObjectMeta.Namespace,
+			Name:      kubernetes.GetServiceFromHostname(split.Spec.Service),
+			Namespace: split.Namespace,
 		}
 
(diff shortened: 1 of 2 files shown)


pkg/catalog/routes.go
```diff
@@ -60,10 +62,7 @@ func (mc *MeshCatalog) listOutboundTrafficPoliciesForTrafficSplits(sourceNamespa
 
 	apexServices := mapset.NewSet()
 	for _, split := range mc.meshSpec.ListTrafficSplits() {
-		svc := service.MeshService{
-			Name:      split.Spec.Service,
-			Namespace: split.ObjectMeta.Namespace,
-		}
+		svc := getRootMeshServiceFromTrafficSplit(split)
```
shashankram, line 65:
```suggestion
		svc := kubernetes.GetServiceFromHostname(split.Spec.Service)
```

PR #2809, opened 2021-03-10
fix(*): rm filterTrafficSplit logic as it is no longer needed

michelleN:
Keeping this prevents us from referencing just a service name
without the namespace suffic as the traffic split root service

**Description**:

**Affected area**:

- New Functionality      [ ]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [X ]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests                  [ ]
- CI System              [ ]
- Performance            [ ]
- Other                  [ ]

### DIFF
--- a/tests/e2e/e2e_trafficsplit_recursive_split.go
+++ b/tests/e2e/e2e_trafficsplit_recursive_split.go
@@ -1,0 +1,323 @@
+package e2e
+
+import (
+	"context"
+	"fmt"
+	"sync"
+	"time"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	. "github.com/openservicemesh/osm/tests/framework"
+)
+
+var _ = OSMDescribe("Test traffic split where root service is same as backend service",
+	OSMDescribeInfo{
+		Tier:   1,
+		Bucket: 2,
+	},
+	func() {
+		Context("HTTP traffic splitting with SMI", func() {
+			testRecursiveTrafficSplit(AppProtocolHTTP)
+		})
+
+		Context("TCP traffic splitting with SMI", func() {
+			testRecursiveTrafficSplit(AppProtocolTCP)
+		})
+	})
+
+func testRecursiveTrafficSplit(appProtocol string) {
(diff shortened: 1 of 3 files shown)


shashankram (approved):

snehachhabria (approved):

snehachhabria (approved):
Thanks a lot for getting rid of this!

michelleN:
I added an e2e. Would ya'll mind re-reviewing @snehachhabria @shashankram

tests/e2e/e2e_trafficsplit_recursive_split.go
```diff
@@ -0,0 +1,323 @@
...
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	. "github.com/openservicemesh/osm/tests/framework"
+)
+
+var _ = OSMDescribe("Test traffic split where root service is same as backend service",
+	OSMDescribeInfo{
+		Tier:   1,
+		Bucket: 1,
```
shashankram, line 21: Does this need to be Tier 1 (run pre-merge)? If not, move this to Tier 2 so it doesn't increase pre-merge CI times.

Additionally, we need to bucket the test such that the buckets are more or less balanced, and update e2e/README to list this test. Currently:

image

michelleN: I do think this should be tier 1 since it is testing a core feature's functionality not an add on or plugin or extra, non required feature.

snehachhabria: if its needed in Tier 1, lets move it to bucket 2 to balance the time

shashankram (approved):

tests/e2e/e2e_trafficsplit_recursive_split.go
shashankram: > I do think this should be tier 1 since it is testing a core feature's functionality not an add on or plugin or extra, non required feature.

We have a lot of core features running in Tier 2 :-)
Typically, I tend to put variations of a feature (such as this one where we are testing a particular scenario of traffic split) in Tier 2.

snehachhabria (approved):

PR #2890, opened 2021-03-16
Skip http_ingress_test if not kind cluster

nshankar13:

**Description**: The http_ingress_test sends a request to localhost, which is not compatible for non-kind clusters. Thus, I added a condition to skip this for other install types. 

**Affected area**:

- New Functionality      [ ]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [ ]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests                  [X]
- CI System              [X]
- Demo                   [ ]
- Performance            [ ]
- Other                  [ ]

No

### DIFF
--- a/tests/e2e/e2e_http_ingress_test.go
+++ b/tests/e2e/e2e_http_ingress_test.go
@@ -27,6 +27,11 @@ var _ = OSMDescribe("HTTP ingress",
 		const destNs = "server"
 
 		It("allows HTTP ingress traffic", func() {
+
+			if Td.InstType != KindCluster {
+				Skip("test requires Kind Cluster")
+			}
+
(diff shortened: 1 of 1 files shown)


shashankram (commented):
Can we instead get this test working in non Kind clusters?

nojnhuh:
I think I have a fix. Will sync with @nshankar13 to confirm it works.

michelleN (commented):
Looks like https://github.com/openservicemesh/osm/pull/2896 would resolve this 👍  good catch

PR #2897, opened 2021-03-16
cli, Azure Monitor: add configmap update flag and logic

eduser25:
[Tentative]

Adds CLI flag `--azmon` to also update the required config map upon
enabling/disabling metrics on a namespace for Azure Monitor backend.

- CLI Tool               [X]

No


michelleN (changes requested):
What command are we adding the `--azmon` flag to here?

I'd strongly object to us adding vendor specific flags to the project. In Helm, Draft, and other projects, we created a plugin system for the CLI so that vendors could easily build plugins for their vendor specific needs and easily integrate them into the project without having vendor specific logic in the core codebase. See Helm's plugin model [here](https://helm.sh/docs/topics/plugins/) and Draft's plugin model [here](https://github.com/Azure/draft/blob/3206e97af58cb9bb656f0df5cb5ae0808b87cbe1/docs/reference/dep-005.md).

Added a proposal [here](https://github.com/openservicemesh/osm/issues/2899).

eduser25:
Sounds good @michelleN, thanks for the feedback.

eduser25:
closing for now.

PR #2918, opened 2021-03-17
docs(release): Update version dropdown for docs site

SanyaKochhar:

**Description**:

**Affected area**:

- New Functionality      [ ]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [ ]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests                  [ ]
- CI System              [ ]
- Demo                   [ ]
- Performance            [ ]
- Other                  [ ]

### DIFF
--- a/docs/config.toml
+++ b/docs/config.toml
@@ -74,8 +74,12 @@ latexDashes = true
 
 # Add your release versions here
 [[params.versions]]
-  version = "v0.7.0"
+  version = "v0.8.0"
   url = "https://docs.openservicemesh.io"
+
+[[params.versions]]
+  version = "v0.7.0"
+  url = "https://release-v0.7--osm-docs.netlify.app"
 
 
 # Theme Additional params


docs/config.toml
```diff
@@ -74,9 +74,13 @@ latexDashes = true
 
 # Add your release versions here
 [[params.versions]]
-  version = "v0.7.0"
+  version = "v0.8.0"
   url = "https://docs.openservicemesh.io"
```
nojnhuh, line 78: Should this point to the release-v0.8 branch? For docs tracking `main` it seems like those should be versioned separately as either v0.9 or `dev` since they may include docs for features that will never be in v0.8.

SanyaKochhar: Hm that makes sense. I made this change based on number 3 here https://github.com/openservicemesh/osm/tree/main/docs#release-and-versioning-process . @bridgetkromhout wdyt?

michelleN (approved):

docs/config.toml
bridgetkromhout: Well, if something gets updated and affects all extant branches, will it be backported to all? That's the problem with versioning docs.

nojnhuh: I think it would be reasonable to support a version of the docs as long as its corresponding release of OSM is supported. That would mean removing the v0.7 docs from the site now that v0.8.0 is published and future fixes to docs will only be made on the `main` and `release-v0.8` branches.

I'm not married to that exact policy, but having the docs support roughly match the code support makes sense to me.

nojnhuh (approved):

PR #2937, opened 2021-03-19
pkg/ticker: reintroduce ticker concept

eduser25:
New package contains a ticker implementation that is fully
interacted through pubsub (this will allow us to decouple
from whichever way we want to configure it with, configmap,
pflag, debugServer endpoint, etc.)

It starts turned off by default and is run after dispatcher
starts. Next commit will introduce configurability.

No


draychev (approved):
Thanks for bringing this back!
Look forward to the second part where we initialize this w/ a optional OSM Controller CLI --param.

shashankram (approved):

snehachhabria (approved):

PR #2938, opened 2021-03-19
xds: handle wildcard TypeURI

eduser25:
Handle wildcard xDS proto.

No

### DIFF
--- a/pkg/envoy/ads/response.go
+++ b/pkg/envoy/ads/response.go
@@ -65,7 +65,7 @@ func (s *Server) sendResponse(typeURIsToSend mapset.Set,
 
 		// Handle request when is not provided, and the SDS case
 		var finalReq *xds_discovery.DiscoveryRequest
-		if request == nil {
+		if request == nil || request.TypeUrl == envoy.TypeWildcard.String() {
 			if typeURI == envoy.TypeSDS {
 				finalReq = makeRequestForAllSecrets(proxy, s.catalog)
 				if finalReq == nil {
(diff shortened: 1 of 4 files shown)


pkg/envoy/ads/response.go
```diff
@@ -65,7 +65,7 @@ func (s *Server) sendResponse(typeURIsToSend mapset.Set,
 
 		// Handle request when is not provided, and the SDS case
 		var finalReq *xds_discovery.DiscoveryRequest
-		if request == nil {
+		if request == nil || request.TypeUrl == string(envoy.TypeWildcard) {
```
shashankram, line 68: Would change `string(envy.TypeWildcard)` to `envoy.TypeWildcard.String()` and add `String()` method receiver for the type.

pkg/envoy/ads/stream.go
```diff
@@ -172,7 +172,20 @@ func (s *Server) StreamAggregatedResources(server xds_discovery.AggregatedDiscov
...
-			err := s.sendResponse(mapset.NewSetWith(typeURL), proxy, &server, &discoveryRequest, s.cfg)
+			var xdsUpdatePaths mapset.Set
+			switch typeURL {
+			case envoy.TypeWildcard:
+				xdsUpdatePaths = mapset.NewSetWith(
+					envoy.TypeCDS,
+					envoy.TypeEDS,
+					envoy.TypeLDS,
+					envoy.TypeRDS,
+					envoy.TypeSDS)
```
shashankram, line 183: Does order matter here? If it does, does mapset gaurantee ordering (I guess not)?

shashankram (approved):

pkg/envoy/ads/response.go
eduser25: can do. Not that I remember, I think that was the VC autocompletion

pkg/envoy/ads/stream.go
eduser25: (1) i think order is not guaranteed (i think it uses internal map?) and (2) order here doesn't matter

Order is however guaranteed when calling `sendResponse` over [here](https://github.com/openservicemesh/osm/pull/2938/files#diff-7a1f450fbabb0088828083ed90ad3ccc49ed4edd9ae6f58ad5bdcdd158b2dda8R62)

eduser25:
scale test was run ffor this commit https://github.com/eduser25/osm/actions/runs/668859481

snehachhabria (approved):

PR #2950, opened 2021-03-20
envoy: correctly set downstream root cert names

shashankram:
**Description**:
This change fixes a bug related to the root validation
cert names when the controller crafts them as a part
of the workflow to create all secrets associated with
the proxy.

The root validation certs referenced in the
DownstreamTlsContext in the service's inbound
filter chain transport socket must correspond
to the service, but they were incorrectly being
derived from the proxy's service account. The buggy
behavior works only when the proxy service name is the
same as the service account name, which is the
case in existing e2e tests. When the service account
name is different than the service name, this bug
manifests often.

The fix is simple - similar to how the service cert
created corresponding to the service referenced in
the DownstreamTlsContext, root validation certs
are also created based on the services associated
with a proxy.

Newly added unit tests cover the different possible
scenarios:
1. when the proxy does not have services
2. When the proxy has one or more services
3. When the proxy is both - downstream and upstream
4. When the proxy does not have allowed upstreams

Required for #2749 and #2930

**Affected area**:

- New Functionality      [ ]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [X]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests                  [ ]
- CI System              [ ]
- Demo                   [ ]
- Performance            [ ]
- Other                  [ ]

`No`


eduser25 (approved):

draychev (approved):
LGTM

Small suggestion: It'd be easier for the reviewer to split PRs like this in 2 pieces 1) move the function in a different file (no functional code changes; easy to review)  2) changing what the function does.

shashankram:
> LGTM
> 
> Small suggestion: It'd be easier for the reviewer to split PRs like this in 2 pieces 1) move the function in a different file (no functional code changes; easy to review) 2) changing what the function does.

I agree. Sorry I moved the function out in this PR to test all SDS related responses using `go testing + mocks`. The only change to this function from the previous implementation is moving the secret resources associated with the inbound root validation certs into the loop.

PR #2980, opened 2021-03-23
docs: document potential container startup issues

shashankram:
-->
**Description**:

**Affected area**:

- New Functionality      [ ]
- Documentation          [X]
- Install                [ ]
- Control Plane          [ ]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests                  [ ]
- CI System              [ ]
- Demo                   [ ]
- Performance            [ ]
- Other                  [ ]

`No`

### DIFF
--- a/docs/content/docs/tasks_usage/container_startup.md
+++ b/docs/content/docs/tasks_usage/container_startup.md
@@ -1,0 +1,27 @@
+---
+title: "Application Container Startup"
+description: "Application Container Startup"
+type: docs
+---
+
+# Container Startup
+
+Since OSM injects application pods that are a part of the service mesh with a sidecar proxy and sets up traffic redirection rules to route all traffic to/from pods via the sidecar proxy, in some circumstances existing application containers might not startup as expected.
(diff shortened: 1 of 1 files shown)


draychev (approved):

eduser25 (approved):
looks good, wondering if there should be a threshold of what to expect at startup given reasonable assumptions (as in, when does connectivity not establishing become an issue)

snehachhabria (approved):

shashankram:
> looks good, wondering if there should be a threshold of what to expect at startup given reasonable assumptions (as in, when does connectivity not establishing become an issue)

This threshold won't be deterministic because it depends on the type of traffic (egress vs in-mesh), type of traffic policy mode (permissive vs SMI), time taken for the config to be realized, time taken for proxy container to come up etc. Some of these depend on k8s (container scheduling, resource being available).

PR #2991, opened 2021-03-24
rds: use consistent naming style for config name

shashankram:
**Description**:
Use `resource-name` style for naming RDS configs to be
consistent with other config names (listener, filter chains,
virtual host, stats etc.).

No functional change.

**Affected area**:

- New Functionality      [ ]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [X]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests                  [ ]
- CI System              [ ]
- Demo                   [ ]
- Performance            [ ]
- Other                  [ ]

`No`

### DIFF
--- a/docs/content/docs/tasks_usage/traffic_management/permissive_traffic_policy_mode.md
+++ b/docs/content/docs/tasks_usage/traffic_management/permissive_traffic_policy_mode.md
@@ -168,7 +168,7 @@ In permissive mode, OSM controller programs wildcard routes for client applicati
           "ads": {},
           "resource_api_version": "V3"
          },
-         "route_config_name": "RDS_Outbound"
+         "route_config_name": "rds-outbound"
         },
         "http_filters": [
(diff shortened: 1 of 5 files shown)


nojnhuh (approved):

eduser25 (approved):
I would even argue the rds prefix is redundant... but since resource name could be anything on wildcard I guess makes sense to keep

shashankram:
> I would even argue the rds prefix is redundant... but since resource name could be anything on wildcard I guess makes sense to keep

Yes it might not be necessary to prefix `rds` here, but left this to be sort of consistent with other resources, ex. `inbound-mesh-http-filter-chain`.

michelleN (commented):
👍

PR #3005, opened 2021-03-25
e2e: Make a namespace for ingress nginx

eduser25:
For NoInstall, nginx could install in `kube-system` if `OsmNamespace` is there.
It's not a major problem, but it makes cleanup better in case something
fails, as later the namespace can be just removed, as opposed to
deleting individual leftovers from `kube-system`.

- Tests                  [x]

No

### DIFF
--- a/tests/e2e/e2e_http_ingress_test.go
+++ b/tests/e2e/e2e_http_ingress_test.go
@@ -76,12 +76,15 @@ var _ = OSMDescribe("HTTP ingress",
 				}
 			}
 
+			ingressNs := "ingress-ns"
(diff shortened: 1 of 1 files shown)


shashankram (approved):

tests/e2e/e2e_http_ingress_test.go
```diff
@@ -76,12 +76,15 @@ var _ = OSMDescribe("HTTP ingress",
 				}
 			}
 
+			inressNs := "ingress-ns"
```
nojnhuh, line 79:
```suggestion
			ingressNs := "ingress-ns"
```

nojnhuh (approved):
LGTM, just one typo.

PR #3014, opened 2021-03-26
Manual guide for Prometheus and Grafana deployment

eduser25:
Separate guide for individual bring up and configuration of the
OSM observability components.

- Uses helm deployments to avoid unsavory batches of `apply -f`s
- Instructs to replace scraping configuration with OSM's one on
Prometheus.
- Instructs how to load Prometheus as Data source and load OSM
dashboards from Grafana.com

**Affected area**:

- Documentation          [ ]

No


draychev (approved):

eduser25:
I see that on the `netlify/osm-docs/deploy-preview` preview images are not rendered, but I doubled checked the paths are correct.
I wonder if it's because I'm using html notation instead of markdown. The relative paths are correct nonetheless, and other guides do use relative paths with no issues.

In case it were a preview problem, I can fix it to move the images to markdown post-merge.

shashankram (approved):

shashankram:
> I see that on the `netlify/osm-docs/deploy-preview` preview images are not rendered, but I doubled checked the paths are correct.
> I wonder if it's because I'm using html notation instead of markdown. The relative paths are correct nonetheless, and other guides do use relative paths with no issues. Github markdown rendering does show them just fine.
> 
> In case it were a preview problem, I can fix it to move the images to markdown post-merge.

Let's follow this up post merge.

nojnhuh:
> I see that on the `netlify/osm-docs/deploy-preview` preview images are not rendered, but I doubled checked the paths are correct.
> I wonder if it's because I'm using html notation instead of markdown. The relative paths are correct nonetheless, and other guides do use relative paths with no issues. Github markdown rendering does show them just fine.
> 
> In case it were a preview problem, I can fix it to move the images to markdown post-merge.

Looks like adding an extra `../` to the image paths on the site makes the images work there.

shashankram:
> > I see that on the `netlify/osm-docs/deploy-preview` preview images are not rendered, but I doubled checked the paths are correct.
> > I wonder if it's because I'm using html notation instead of markdown. The relative paths are correct nonetheless, and other guides do use relative paths with no issues. Github markdown rendering does show them just fine.
> > In case it were a preview problem, I can fix it to move the images to markdown post-merge.
> 
> Looks like adding an extra `../` to the image paths on the site makes the images work there.

Yes, for non index.md files an extra `../` is necessary

eduser25:
working now.

Issue #2798, opened 2021-03-10
Make ingress compatible with v1 of Kubernetes ingress


shashankram:
**Please describe the Improvement and/or Feature Request**
Kubernetes ingress became GA in 1.19, and there were several improvements to the ingress spec.

One very noticeable change OSM must support is honoring the `pathType` attribute for an HTTP path in the ingress spec:
https://kubernetes.io/docs/concepts/services-networking/ingress/#path-types

At the very minimum, OSM must support 3 types of path matching semantics:
1. Exact - exact match on paths (`/foo matches /foo`)
2. Prefix - prefix matching on paths (`/foo matches /foo/bar and not /foobar`)
3. Regex - Regex matching on paths (`/?(.*) matches /foo and /foobar`)

Without this support, ingress in OSM will be of limited use for the following reasons:
1. Most ingress controllers support some form of prefix based matching (should be the default in OSM)
2. Limiting path matching to regex has multiple issues: 
        - Regex engine used by the ingress controller could be different than the one required by Envoy (Google RE2), ex. PCRE2 vs Google RE2
        - Regex matching is error prone. A good example of this is the Cloudflare outage described in https://blog.cloudflare.com/details-of-the-cloudflare-outage-on-july-2-2019/  that was using a very complex regex match that broke in production
3. In some cases exact matching is required

This issue proposes to tweak the existing ingress implementation and support `pathType` based path matching for k8s >= 1.19. 

`ImplementationSpecific` pathType will defer to regex matching if the path looks like a regex, otherwise use string based prefix matching. Note that string based prefix matching is different than the `Prefix` path type as defined by k8s.

**Scope (please mark with X where applicable)**
- New Functionality          [ ]
- Install                    [ ]
- SMI Traffic Access Policy  [ ]
- SMI Traffic Specs Policy   [ ]
- SMI Traffic Split Policy   [ ]
- Permissive Traffic Policy  [ ]
- Ingress                    [X]
- Egress                     [ ]
- Envoy Control Plane        [ ]
- CLI Tool                   [ ]
- Metrics                    [ ]
- Certificate Management     [ ]
- Sidecar Injection          [ ]
- Logging                    [ ]
- Debugging                  [ ]
- Tests                      [ ]
- CI System                  [ ]
- Project Release            [ ]

**Possible use cases**
Necessary for ingress to be usable.

shashankram:
[...]

PR #3064, opened 2021-04-01
trafficpolicy: extend HTTPRouteMatch to prefix and exact path matching

shashankram:
**Description**:
Currently, OSM only supports regex based path matching for
HTTP routes. Additional path matching semantics are required
for ingress, specifically `exact` and `prefix` based matching.

This change updates the `HTTPRouteMatch` type to make HTTP
path matching generic enough to match based on regex, exact,
and prefix matching. Functionally, since all existing code is
based on regex based path matching, all HTTP routes specify
the path match type as `regex`.

Required for #2798

**Affected area**:

- New Functionality      [ ]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [X]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests                  [ ]
- CI System              [ ]
- Demo                   [ ]
- Performance            [ ]
- Other                  [ ]

`No`


michelleN (approved):
Thanks @shashankram. If I'm understanding this correctly the prefix and exact matches will be implemented and tested with a follow up PR for ingress support which is 👍. If that weren't the case, I'd recommend we add some comments to the code mentioning those types are unimplemented.

shashankram:
> Thanks @shashankram. If I'm understanding this correctly the prefix and exact matches will be implemented and tested with a follow up PR for ingress support which is 👍. If that weren't the case, I'd recommend we add some comments to the code mentioning those types are unimplemented.

Thanks for the review @michelleN. Yes, I will be implementing the prefix and exact matching on HTTP paths in a follow-up, but dropped the additional types in there to build up the context.

eduser25 (approved):
lgtm

shashankram:
> > Thanks @shashankram. If I'm understanding this correctly the prefix and exact matches will be implemented and tested with a follow up PR for ingress support which is 👍. If that weren't the case, I'd recommend we add some comments to the code mentioning those types are unimplemented.
> 
> Thanks for the review @michelleN. Yes, I will be implementing the prefix and exact matching on HTTP paths in a follow-up, but dropped the additional types in there to build up the context.

https://github.com/openservicemesh/osm/pull/3066 leverages the newly added patch match types.

PR #3079, opened 2021-04-05
fix(pkg/configurator): fix VWH validateFields() test

### LINKED ISSUE #3078: TestValidateFields is not testing the function properly
**Bug description**:
Unit test for TestValidatefields does not actually check that each reason for denial is in the response's StatusReason

- Tests                      [X ]

**Expected behavior**:
Should check that each reason for denial is found in the resp.StatusReason 

**How was OSM installed?**:

**Environment**:
- Others:

shalier:
**Description**:
Resolves #3078

**Affected area**:

- New Functionality      [ ]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [ ]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests                  [ ]
- CI System              [ ]
- Demo                   [ ]
- Performance            [ ]
- Other                  [ X]

No

### DIFF
--- a/pkg/configurator/validating_webhook_test.go
+++ b/pkg/configurator/validating_webhook_test.go
@@ -191,9 +191,9 @@ func TestCheckDefaultFields(t *testing.T) {
 	}
 
 	testCases := []struct {
-		testName  string
-		configMap corev1.ConfigMap
-		expRes    *admissionv1.AdmissionResponse
+		testName         string
+		configMap        corev1.ConfigMap
+		expectedResponse *admissionv1.AdmissionResponse
 	}{
 		{
 			testName: "Contains all default fields",
@@ -213,7 +213,7 @@ func TestCheckDefaultFields(t *testing.T) {
 					"enable_privileged_init_container": "",
 				},
 			},
-			expRes: &admissionv1.AdmissionResponse{
+			expectedResponse: &admissionv1.AdmissionResponse{
 				Allowed: true,
 				Result:  &metav1.Status{Reason: ""},
 			},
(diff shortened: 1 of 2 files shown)


pkg/configurator/validating_webhook_test.go
```diff
@@ -258,9 +258,10 @@ func TestValidateFields(t *testing.T) {
 	assert.Nil(err)
 
 	testCases := []struct {
-		testName  string
-		configMap corev1.ConfigMap
-		expRes    *admissionv1.AdmissionResponse
+		testName   string
+		configMap  corev1.ConfigMap
+		expAllowed bool
```
shashankram, line 263: Would you mind renaming `exp` to `expect` or `expected`.

pkg/configurator/validating_webhook_test.go
```diff
@@ -258,9 +258,10 @@ func TestValidateFields(t *testing.T) {
 	assert.Nil(err)
 
 	testCases := []struct {
-		testName  string
-		configMap corev1.ConfigMap
-		expRes    *admissionv1.AdmissionResponse
```
shashankram, line 263: We should continue testing the admission response.

shashankram (changes requested):

pkg/configurator/validating_webhook_test.go
shalier: Do you mean checking if the AdmissionResponse is created? I removed it because before I was using the AdmissionResponse to check the ReasonForDenial, but that check wasn't actually checking that.

shashankram: I am suggesting to keep the `&admissionv1.AdmissionResponse` object in the test case expectation and compare it with the result returned by `validateFields()`. That way, the entire response will be matched for equality which includes the reason as well.

shalier: Ok, separated the test cases so that the `&admissionv1.AdmissionResponse` were comparable

pkg/configurator/validating_webhook_test.go
```diff
@@ -358,8 +401,9 @@ func TestValidateFields(t *testing.T) {
 				Result:  &metav1.Status{Reason: ""},
 			}
 			res := whc.validateFields(tc.configMap, resp)
-			assert.Contains(tc.expRes.Result.Status, res.Result.Status)
-			assert.Equal(tc.expRes.Allowed, res.Allowed, res.Result.Reason)
+
+			assert.Equal(tc.expectedResponse.Allowed, res.Allowed)
+			assert.Equal(res.Result.Reason, tc.expectedResponse.Result.Reason)
```
shashankram, line 406:
```suggestion
			assert.Equal(tc.expectedResponse, res)
```

shashankram (approved):

snehachhabria (approved):

PR #3108, opened 2021-04-08
feat(osm/Chart.yaml): Pre-install check for k8s version

nshankar13:
Description: Add a pre-install check to ensure that the kubernetes cluster is a supported version for the current version of OSM. We were initially considering limiting this to downstream charts, but this change may be relevant upstream as well. 

**Affected area**:

- New Functionality      [ ]
- Documentation          [ ]
- Install                [X]
- Control Plane          [ ]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests                  [ ]
- CI System              [ ]
- Demo                   [ ]
- Performance            [ ]
- Other                  [ ]

### DIFF
--- a/charts/osm/Chart.yaml
+++ b/charts/osm/Chart.yaml
@@ -19,3 +19,6 @@ version: 0.8.2
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application.
 appVersion: v0.8.2
+
+# This specifies a particular range of k8s versions that the application is compatible with. 
+kubeVersion: ">= 1.18.0"
--- a/tests/e2e/e2e_k8s_version_test.go
+++ b/tests/e2e/e2e_k8s_version_test.go
@@ -21,11 +21,8 @@ var _ = OSMDescribe("Test HTTP traffic for different k8s versions",
 		Context("Version v1.19.7", func() {
 			testK8sVersion("v1.19.7")
 		})
-		Context("Version v1.18.15", func() {
-			testK8sVersion("v1.18.15")
-		})
-		Context("Version v1.17.17", func() {
-			testK8sVersion("v1.17.17")
+		Context("Version v1.18.0", func() {
+			testK8sVersion("v1.18.0")
 		})
 	})
 


charts/osm/Chart.yaml
```diff
@@ -19,3 +19,6 @@ version: 0.8.2
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application.
 appVersion: v0.8.2
+
+# This specifies a particular range of k8s versions that the application is compatible with. 
+kubeVersion: >= 1.18.14 <= 1.20.2
```
shashankram, line 24: Can we remove the upper bound of 1.20.2 so when a new patch is released, CI with Kind based k8s cluster automatically picks it up?

nshankar13: Should we get rid of the upper bound altogether, or set it to the default version for kind used in the CI?

shashankram: We likely want to always support latest supported k8s version including patches, and find issues early, so yes let's get rid of the upper bound. If we cannot support the latest release for whatever reason, we can revisit adding upper bound then.

Also shouldn't the lower bound be 1.18.0.

nshankar13: I set it to 1.18.14 because that is currently the default version of k8s on AKS that was used to validate OSM v0.8.2 in our CI in osm-azure (and the kind version in this upstream repo is 1.19.0 I believe) It likely does work for 1.18.0, but since we haven't tested it for v0.8.2 I'm not sure if we should make that the lower bound. 

This does bring up a good question in general of how the lower bound should be determined.

nshankar13: nvm, the e2e_k8s_version test validates on multiple versions.

snehachhabria (approved):

tests/e2e/e2e_k8s_version_test.go
```diff
@@ -24,8 +24,8 @@ var _ = OSMDescribe("Test HTTP traffic for different k8s versions",
 		Context("Version v1.18.15", func() {
 			testK8sVersion("v1.18.15")
 		})
-		Context("Version v1.17.17", func() {
-			testK8sVersion("v1.17.17")
+		Context("Version v1.18.0", func() {
```
shashankram, line 27: We can remove v1.18.15 above, we just wanted to run against 1 minor version here. We anyway aren't running this test against every patch version.

shashankram (approved):

PR #3131, opened 2021-04-12
envoy/ingress: use a separate route config for ingress

### LINKED ISSUE #3130: Separate ingress route configuration for inbound mesh route configuration for hardening security

There is an existing problem where an ingress route
can open up routes to the service within the mesh
for clients that are part of the mesh even if an
SMI policy does not exist for in-mesh clients to
access the service over the routes specified in
the ingress resource.

This is because both in-mesh routes based on SMI
policies and ingress routes based on k8s ingress
use the same inbound route configuration. This is
problematic and even a security concern because
using ingress also allows in-mesh clients to access
the service over the ingress routes when the client
has access to some but not all routes on the service.

Consider the following example where `client` and
`server` are part of the mesh.
- `server` exposes 2 paths: `/foo` and `/bar`
- `client` should only access `/foo` on `server` based on SMI
- `/bar` on `server` is exposed to the internet using ingress

With the existing implementation, `client` will be able to
access /bar even though not allowed to by the mesh SMI policy,
but because there exists an ingress route for /bar on `server`.

This issue proposes to use a separate route configuration for ingress so that in-mesh policies are not bypassed.

- Ingress                    [X]

**Possible use cases**
Ingress routes should not allow in-mesh clients to bypass in-mesh SMI traffic policies.

shashankram:
**Description**:
There is an existing problem where an ingress route
can open up routes to the service within the mesh
for clients that are part of the mesh even if an
SMI policy does not exist for in-mesh clients to
access the service over the routes specified in
the ingress resource.

This is because both in-mesh routes based on SMI
policies and ingress routes based on k8s ingress
use the same inbound route configuration. This is
problematic and even a security concern because
using ingress also allows in-mesh clients to access
the service over the ingress routes when the client
has access to some but not all routes on the service.

Consider the following example where `client` and
`server` are part of the mesh.
- `server` exposes 2 paths: `/foo` and `/bar`
- `client` should only access `/foo` on `server` based on SMI
- `/bar` on `server` is exposed to the internet using ingress

With the existing implementation, `client` will be able to
access `/bar` even though not allowed to by the mesh SMI policy,
but because there exists an ingress route for `/bar` on `server`.

This change prevents such a problem by using a separate
route configuration for ingress, that the ingress filter
chains reference.

Resolves #3130

**Affected area**:

- New Functionality      [ ]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [X]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests                  [ ]
- CI System              [ ]
- Demo                   [ ]
- Performance            [ ]
- Other                  [ ]

`No`

### DIFF
--- a/pkg/envoy/lds/ingress.go
+++ b/pkg/envoy/lds/ingress.go
@@ -38,10 +38,10 @@ func (lb *listenerBuilder) newIngressHTTPFilterChain(cfg configurator.Configurat
 		return nil
 	}
 
-	inboundConnManager := getHTTPConnectionManager(route.InboundRouteConfigName, cfg, nil)
-	marshalledInboundConnManager, err := ptypes.MarshalAny(inboundConnManager)
+	ingressConnManager := getHTTPConnectionManager(route.IngressRouteConfigName, cfg, nil)
+	marshalledIngressConnManager, err := ptypes.MarshalAny(ingressConnManager)
 	if err != nil {
-		log.Error().Err(err).Msgf("Error marshalling inbound HttpConnectionManager object for proxy %s", svc)
+		log.Error().Err(err).Msgf("Error marshalling ingress HttpConnectionManager object for proxy %s", svc)
 		return nil
 	}
 
@@ -58,7 +58,7 @@ func (lb *listenerBuilder) newIngressHTTPFilterChain(cfg configurator.Configurat
 			{
 				Name: wellknown.HTTPConnectionManager,
 				ConfigType: &xds_listener.Filter_TypedConfig{
-					TypedConfig: marshalledInboundConnManager,
+					TypedConfig: marshalledIngressConnManager,
 				},
 			},
 		},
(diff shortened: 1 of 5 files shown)


eduser25 (approved):
lgtm, talked offline

snehachhabria (approved):
Thanks for this change!

michelleN (commented):
How do you ensure that no two RouteConfigurations have the same set of domains?

shashankram:
> How do you ensure that no two RouteConfigurations have the same set of domains?

If I understand your question correctly, that doesn't need to be ensured here because `rds-ingress` (used for ingress) and `rds-inbound` (used for in-mesh) are 2 separate configurations meant for 2 distinct types of traffic. This is similar to how we don't need to be concerned about the same set of domains present in `rds-inbound` and `rds-outbound`.

michelleN:
gotchya thanks for the explanation @shashankram

michelleN (commented):
👍

Issue #3128, opened 2021-04-10
Envoy validation fails when method/path is not specified by user in Traffic Spec


snehachhabria:
**Bug description**:

I modified the demo to remove the host header from the traffic  spec for bookwarehouse : https://github.com/openservicemesh/osm/blob/71b619c6ccedc6701f235a060786371d9a4b422a/demo/deploy-traffic-specs.sh#L47

This resulted in OSM demo failing and the following error message was observed on the osm controller pod : 
[...]

PR #3139, opened 2021-04-12
bugfix(pkg/catalog) : wildcard path and/or method in http route when not specified

snehachhabria:
**Description**:

This change wildcards the path and/or method for a HTTPRouteGroup when
not specified. This is necessary to prevent an envoy configuration error
as described in issue #3128.

The mock blocks for all the unit tests in files related to this change
have been moved inside the loop to prevent the tests from being flaky.

Fixes : #3128 

**Affected area**:

- New Functionality      [ ]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [X]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [X]
- Security               [ ]
- Tests                  [ ]
- CI System              [ ]
- Demo                   [ ]
- Performance            [ ]
- Other                  [ ]

### DIFF
--- a/pkg/catalog/inbound_traffic_policies.go
+++ b/pkg/catalog/inbound_traffic_policies.go
@@ -203,14 +203,12 @@ func (mc *MeshCatalog) getHTTPPathsPerRoute() (map[trafficpolicy.TrafficSpecName
 				Headers:       trafficSpecsMatches.Headers,
 			}
 
-			if len(serviceRoute.Headers) != 0 {
-				// When pathRegex and methods are not defined, the header filters are applied to any path and all HTTP methods
-				if serviceRoute.Path == "" {
-					serviceRoute.Path = constants.RegexMatchAll
-				}
-				if serviceRoute.Methods == nil {
-					serviceRoute.Methods = []string{constants.WildcardHTTPMethod}
-				}
+			// When pathRegex or/and methods are not defined, they will be wildcarded
+			if serviceRoute.Path == "" {
+				serviceRoute.Path = constants.RegexMatchAll
+			}
+			if len(serviceRoute.Methods) == 0 {
+				serviceRoute.Methods = []string{constants.WildcardHTTPMethod}
 			}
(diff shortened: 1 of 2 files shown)


pkg/catalog/inbound_traffic_policies.go
```diff
@@ -203,14 +203,12 @@ func (mc *MeshCatalog) getHTTPPathsPerRoute() (map[trafficpolicy.TrafficSpecName
...
-					serviceRoute.Path = constants.RegexMatchAll
-				}
-				if serviceRoute.Methods == nil {
-					serviceRoute.Methods = []string{constants.WildcardHTTPMethod}
-				}
+			// When pathRegex or/and methods are not defined, they will be wildcarded
+			if serviceRoute.Path == "" {
+				serviceRoute.Path = constants.RegexMatchAll
+			}
+			if serviceRoute.Methods == nil || len(serviceRoute.Methods) == 0 {
```
shashankram, line 210:
```suggestion
			if len(serviceRoute.Methods) == 0 {
```

pkg/catalog/inbound_traffic_policies_test.go
```diff
@@ -683,6 +671,19 @@ func TestListInboundPoliciesForTrafficSplits(t *testing.T) {
 	}
 	for _, tc := range testCases {
 		t.Run(tc.name, func(t *testing.T) {
+			mockCtrl := gomock.NewController(t)
+			defer mockCtrl.Finish()
```
shashankram, line 675: This can be outside the loop as before.

shashankram (approved):

pkg/catalog/inbound_traffic_policies_test.go
snehachhabria: As discussed offline, having having these outside the loop was taking the tests to be falky

snehachhabria: I have updated the PR description and commit to explain the same

michelleN (approved):

PR #3181, opened 2021-04-15
prometheus: add deployment resource values to values.yaml

eduser25:
This commit adds prometheus resource configurability through
`values.yaml`.

Based on https://github.com/eduser25/osm/actions/runs/720587234:
- Bumped requested Mem to 512M, which is the highest of the lowest we can go with
(should support up to 100pods consistently without having prom get killed if the node under
memory pressure form other pods).
- Also adjusted CPU requested. 100m is too low for even 100 pods.
Half vcpu should be a more solid value to consistently support the values we advertise.

- Metrics                [x]

No

### DIFF
--- a/charts/osm/README.md
+++ b/charts/osm/README.md
@@ -104,6 +104,7 @@ The following table lists the configurable parameters of the osm chart and their
 | OpenServiceMesh.osmcontroller.resource.requests.memory | string | `"128M"` |  |
 | OpenServiceMesh.outboundIPRangeExclusionList | list | `[]` | Optional parameter to specify a global list of IP ranges to exclude from outbound traffic interception by the sidecar proxy. If specified, must be a list of IP ranges of the form a.b.c.d/x. |
 | OpenServiceMesh.prometheus.port | int | `7070` | Prometheus port |
+| OpenServiceMesh.prometheus.resources | object | `{"limits":{"cpu":1,"memory":"2G"},"requests":{"cpu":0.5,"memory":"512M"}}` | Resource limits for prometheus instance |
 | OpenServiceMesh.prometheus.retention.time | string | `"15d"` | Prometheus retention time |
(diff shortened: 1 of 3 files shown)


charts/osm/README.md
```diff
@@ -104,6 +104,7 @@ The following table lists the configurable parameters of the osm chart and their
 | OpenServiceMesh.osmcontroller.resource.requests.memory | string | `"128M"` |  |
 | OpenServiceMesh.outboundIPRangeExclusionList | list | `[]` | Optional parameter to specify a global list of IP ranges to exclude from outbound traffic interception by the sidecar proxy. If specified, must be a list of IP ranges of the form a.b.c.d/x. |
 | OpenServiceMesh.prometheus.port | int | `7070` | Prometheus port |
+| OpenServiceMesh.prometheus.resources | object | `{"limits":{"cpu":1,"memory":"2G"},"requests":{"cpu":0.5,"memory":"512M"}}` | Resource limits for prometheus instance |
```
eduser25, line 107: Not sure why this is looking like this.... I ran `make chart-readme`, not sure i'm missing something

allenlsy (approved):

charts/osm/README.md
shashankram: It picks up the default values defined in `values.yaml`. It looks fine, were you expecting something else?

shashankram (approved):

draychev (approved):
Thank you @eduser25

charts/osm/README.md
eduser25: At line [101](https://github.com/openservicemesh/osm/pull/3181/files/491e6e77e0c7e99456ce7ce842b1e5a41c87177c#diff-5b88a1fc4bc9f14c08553477ec45b3628356876507a5cbe7a597abfc103db7daR101) seems to unroll the entire structure for `osmcontroller`, but not for prometheus here. Just wondered.

PR #3202, opened 2021-04-20
e2e: Fix resource consumption of Prometheus

eduser25:
CI is failing due to low resorce provision on CI.

Making sure limits are set to allow Prometheus to run
in restricted resource environments.

- Tests                  [x]
- CI System              [x]

No

### DIFF
--- a/tests/e2e/e2e_metrics_test.go
+++ b/tests/e2e/e2e_metrics_test.go
@@ -28,7 +28,12 @@ var _ = OSMDescribe("Custom WASM metrics between one client pod and one server",
 			// Install OSM
 			installOpts := Td.GetOSMInstallOpts()
 			installOpts.DeployPrometheus = true
-			installOpts.SetOverrides = []string{"OpenServiceMesh.enableWASMStatsExperimental=true"}
+			installOpts.SetOverrides = []string{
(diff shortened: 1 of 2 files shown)


nojnhuh (approved):

tests/e2e/e2e_metrics_test.go
```diff
@@ -28,7 +28,11 @@ var _ = OSMDescribe("Custom WASM metrics between one client pod and one server",
 			// Install OSM
 			installOpts := Td.GetOSMInstallOpts()
 			installOpts.DeployPrometheus = true
-			installOpts.SetOverrides = []string{"OpenServiceMesh.enableWASMStatsExperimental=true"}
+			installOpts.SetOverrides = []string{
+				"OpenServiceMesh.enableWASMStatsExperimental=true",
```
shashankram, line 32: Could you add a comment on why these values.

shashankram (approved):

tests/e2e/e2e_metrics_test.go
eduser25: they are just the previous ones we had, where things used to work. I don't know the exact values accepted by CI tbh

shashankram: Even that in itself can be documented.

PR #3206, opened 2021-04-20
egress/api: add ipAddresses field to match TCP based traffic

shashankram:
**Description**:
Adds an optional `ipAddresses` field to the egress API to specify
a list of IP ranges the egress policy applies to. This is specifically
required when more than 1 external TCP endpoints use the same port
because just port alone will not be sufficient to route traffic
differently for the different endpoints.

Part of #3045 

**Affected area**:

- New Functionality      [X]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [ ]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests                  [ ]
- CI System              [ ]
- Demo                   [ ]
- Performance            [ ]
- Other                  [ ]

`No`

### DIFF
--- a/pkg/apis/policy/v1alpha1/egress.go
+++ b/pkg/apis/policy/v1alpha1/egress.go
@@ -34,6 +34,10 @@ type EgressSpec struct {
 	// +optional
 	Hosts []string `json:"hosts,omitempty"`
 
+	// IPAddresses defines the list of external IP addresses the Egress policy is applicable to
+	// +optional
+	IPAddresses []string `json:"ipAddresses,omitempty"`
+
 	// Ports defines the list of ports the Egress policy is applicable to
 	Ports []PortSpec `json:"ports"`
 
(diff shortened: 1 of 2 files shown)


pkg/apis/policy/v1alpha1/egress.go
```diff
@@ -34,6 +34,10 @@ type EgressSpec struct {
 	// +optional
 	Hosts []string `json:"hosts,omitempty"`
 
+	// IPAddresses defines the list of external IP addresses the Egress policy is applicable to
```
eduser25, line 37: suggesting?
```suggestion
	// IPAddresses defines the list of external destination IP addresses the Egress policy is applicable to
```

eduser25 (approved):

snehachhabria (approved):

pkg/apis/policy/v1alpha1/egress.go
shashankram: Doesn't external imply destination?

eduser25: you can totally skip, no big deal
Reading the line in isolation, one could think the egress policy is applicable or filters source instead

PR #3222, opened 2021-04-23
Osm controller init container

allenlsy:
**Description**:

Create a initContainer named `init-osm-controller` to osm-controller pod.

The purpose of this initContainer is to create MeshConfig resource in k8s if it does not exist. In the long term, initContainer can be a central place for all control plane bootstrap tasks. We are also leveraging the nature of initContainer that it is guaranteed executed before normal containers.

This PR is the first step of issue #3039. The next step would be remove configmap and use MeshConfig CRD to configure osm-controller.

**Affected area**:

- New Functionality      [x]
- Install                [x]
- Control Plane          [x]

`No`


cmd/init-osm-controller/init-osm-controller.go
```diff
@@ -0,0 +1,110 @@
...
+
+const meshConfigName = "osm-mesh-config"
+
+var settings = cli.New()
+
+var (
+	flags = pflag.NewFlagSet("init-osm-controller", pflag.ExitOnError)
+	log   = logger.New("osm-controller/main")
+
+	meshName           string // An ID that uniquely identifies an OSM instance
```
shashankram, line 30: This does not identify an OSM instance, but rather an instance of a mesh. Within the same cluster, multiple OSM instances could exist as a part of the same logical mesh. Would you mind rewording this statement to reflect this.

cmd/init-osm-controller/init-osm-controller.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+
+	configClient, err := configClientset.NewForConfig(kubeConfig)
+	if err != nil {
+		log.Fatal().Err(err).Msgf("Could not access Kubernetes cluster, check kubeconfig.")
+		return
+	}
+
+	var retry int
+	for retry = 0; retry < fetchMeshConfigMaxRetry; retry++ {
```
shashankram, line 76: This seems a bit hacky. Why is the retry required to create the default config? The CRD should always be present prior to the init container starting up, so the creation of the default config should never fail. Is there something I am missing?

cmd/init-osm-controller/init-osm-controller.go
```diff
@@ -0,0 +1,110 @@
...
+		if _, err := configClient.ConfigV1alpha1().MeshConfigs(settings.Namespace()).Get(context.TODO(), meshConfigName, v1.GetOptions{}); err != nil {
+			// MeshConfig does not exist. Create a default one.
+			log.Info().Msgf("MeshConfig not exist. Trying create the default MeshConfig named %s into namespace %s", meshConfigName, osmNamespace)
+
+			if err := createDefaultMeshConfig(configClient); err != nil {
```
shashankram, line 81: There could be a race here with multiple controller pods starting up attempting to create the default MeshConfig CRD.

Could you follow a similar pattern here as done for loading the OSM CA bundle from a k8s secret here:
https://github.com/openservicemesh/osm/blob/77531d59eb10fc2cbf9a74fab3a7a5389445fce7/pkg/certificate/providers/config.go#L181-L188

cmd/init-osm-controller/init-osm-controller.go
```diff
@@ -0,0 +1,110 @@
...
+		log.Fatal().Msgf("Max retry (%d) of fetching or creating MeshConfig reached. Exit", fetchMeshConfigMaxRetry)
+	}
+}
+
+func createDefaultMeshConfig(configClient *configClientset.Clientset) error {
+	// create meshConfig from struct
+	meshConfig := &v1alpha1.MeshConfig{}
+	meshConfig.Kind = "MeshConfig"
+	meshConfig.APIVersion = "config.openservicemesh.io/v1alpha1"
+	meshConfig.ObjectMeta.Name = "osm-mesh-config"
```
shashankram, line 106: This can be filled in L103 with the struct initialization. Other defaults should go here as well.

cmd/init-osm-controller/init-osm-controller.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+
+	if retry == fetchMeshConfigMaxRetry {
+		log.Fatal().Msgf("Max retry (%d) of fetching or creating MeshConfig reached. Exit", fetchMeshConfigMaxRetry)
+	}
+}
+
+func createDefaultMeshConfig(configClient *configClientset.Clientset) error {
+	// create meshConfig from struct
+	meshConfig := &v1alpha1.MeshConfig{}
```
shashankram, line 103: Where are the defaults for the other options being set? Currently, `enablePrometheusScraping` is set to true in `values.yaml`.

shashankram (changes requested):
Thanks for the change. Looks good overall, left some comment to get further clarifications.

cmd/init-osm-controller/init-osm-controller.go
```diff
@@ -0,0 +1,110 @@
...
+	"k8s.io/client-go/tools/clientcmd"
+)
+
+const meshConfigName = "osm-mesh-config"
+
+var settings = cli.New()
+
+var (
+	flags = pflag.NewFlagSet("init-osm-controller", pflag.ExitOnError)
+	log   = logger.New("osm-controller/main")
```
nojnhuh, line 28: Nit:
```suggestion
	log   = logger.New("init-osm-controller")
```

cmd/init-osm-controller/init-osm-controller.go
```diff
@@ -0,0 +1,110 @@
...
+
+func parseFlags() error {
+	if err := flags.Parse(os.Args); err != nil {
+		return err
+	}
+	_ = flag.CommandLine.Parse([]string{})
+	return nil
+}
+
+func main() {
```
nojnhuh, line 56: I would try to pull as much of this out into its own function as possible so it can be unit tested.

nojnhuh (commented):
What's the rationale for doing this in a separate container vs. adding this step to osm-controller's regular startup tasks?

cmd/init-osm-controller/init-osm-controller.go
allenlsy: In this design, the purpose of init container is to create the default config, by itself. Because the default MeshConfig is not stored in helm charts.

The init container here will create the default MeshConfig if it does not exist.

Thus, if multiple osm-controller pod exist, they will race to create this default one. The retry strategy is trying to ensure only one pod creates it, and others wait for it to come up, and not overload the api server.

One assumption I made here, which I'm not sure, is the api-server will provide the mutex or stuff to ensure only one MeshConfig will be processed at a time. But even if it's not, since the MeshConfig will have the same name and content, I think it should be fine.

cmd/init-osm-controller/init-osm-controller.go
allenlsy: They are defined here: https://github.com/openservicemesh/osm/blob/c0de988e82c5de275afb55876266019c38b1b461/pkg/apis/config/v1alpha1/mesh_config.go#L43

cmd/init-osm-controller/init-osm-controller.go
shashankram: I would suggest tweaking the implementation and make it similar to the code described in https://github.com/openservicemesh/osm/pull/3222#discussion_r620641915

cmd/init-osm-controller/init-osm-controller.go
shashankram: I don't think we should be defining implementation defaults in the API definition. It seems odd to me this is the case. Wouldn't it make it easier define it in the structs at the time of default CR creation? The current encoding of the defaults in the API definition does not make this obvious.

cmd/init-osm-controller/init-osm-controller.go
allenlsy: I will explore whether that really solves the race issue. But for the coding style, sure I will update.

cmd/init-osm-controller/init-osm-controller.go
allenlsy: What if we need to create another default MeshConfig object later? Then we will need to provide the default value again.

I think, if the default values make sense to be the nature of the definition, then put it in the struct definition is fine. E.g. here, if user does not specify `prometheusScraping` value, we want it to be true rather than false.

What's your opinion?

shashankram: > What if we need to create another default MeshConfig object later? Then we will need to provide the default value again.

I don't see a use case for this.
To me, it doesn't seem correct to tie OSM install defaults to the API definition defaults. There is a more subtle reason for this. At some point when our API is mature and there is a change to the API, the API version will need to be updated (ex. from v1alpha1 to v1alpha2). So if we decide to update the install time defaults, then the API version will also require to be updated which is not desirable. This seems to not be a problem because users are not using the MeshConfig CR, but once this is released it will be better to control install time defaults via the application code rather than encoding it in the API definition.

cmd/init-osm-controller/init-osm-controller.go
shashankram: It does solve the race issue between multiple osm-controller's, which is why I am suggesting to change it :-)

cmd/init-osm-controller/init-osm-controller.go
allenlsy: Ok, make sense to me

allenlsy: FYI, what I mentioned was not accurate. The code `meshConfig` was later overwritten by k8s api creation. Thus the default value is from the CRD definition.

allenlsy:
@nojnhuh I can think of two reasons.

1. we have to ensure MeshConfig exists before osm-controller starts. A normal container does not guarantee containers start order.
2. It's not necessary to have a long running container for this task.

nojnhuh:
> 1. we have to ensure MeshConfig exists before osm-controller starts. A normal container does not guarantee containers start order.
> 2. It's not necessary to have a long running container for this task.

But does it need to be in a separate container at all? Is there any reason we couldn't add this same functionality to the osm-controller binary to create the MeshConfig before it enters its main loop? I'm wondering why this should be in a separate container but other short-lived tasks like creating the root CA or patching the webhooks can be built into osm-controller.

cmd/init-osm-controller/init-osm-controller.go
shashankram: I didn't get the last comment. What is the final decision?

shashankram:
> > 1. we have to ensure MeshConfig exists before osm-controller starts. A normal container does not guarantee containers start order.
> > 2. It's not necessary to have a long running container for this task.
> 
> But does it need to be in a separate container at all? Is there any reason we couldn't add this same functionality to the osm-controller binary to create the MeshConfig before it enters its main loop? I'm wondering why this should be in a separate container but other short-lived tasks like creating the root CA or patching the webhooks can be built into osm-controller.

@nojnhuh That's a reasonable question. I think partly the rationale was to create a bootstrapping mechanism for osm-controller in a generic enough way that could be extendable to future needs of similar sorts. The CA bundle secret creation could move into an init-container as well, although that requires some additional effort since it gets created/read by both osm-controller and osm-injector. In general, I think for one-time tasks, it makes sense to rely on an init-container, so if the controller container crashes for some reason we don't have to worry about issues with re-initializing resources that were previously initialized.

cmd/init-osm-controller/init-osm-controller.go
```diff
@@ -0,0 +1,89 @@
...
+		log.Fatal().Err(err).Msgf("Could not access Kubernetes cluster, check kubeconfig.")
+		return
+	}
+
+	meshConfig := &v1alpha1.MeshConfig{
+		TypeMeta: v1.TypeMeta{
+			Kind:       "MeshConfig",
+			APIVersion: "config.openservicemesh.io/v1alpha1",
+		},
+		ObjectMeta: v1.ObjectMeta{
```
shashankram, line 77: Am I correct to assume we want to be defining defaults here instead of the API definition as discussed previously?

shashankram: In that case, why aren't defaults being set here? Is that revision pending?

allenlsy: Update. Default values are all explicitly set in Golang code in init-osm-controller, `createDefaultMeshConfig()`

allenlsy:
@shashankram @nojnhuh  updated PR description to reflected the intention behind a initContainer approach.

shashankram (approved):

snehachhabria (approved):
Changes look good, thank you! Coudl you create another github issue : `Push init-osm-controller image to MCR` for us to push this new image to MCR ? 

You can add the issue to the current milestone and assign it to me. I will take care of the changes needed

snehachhabria:
I've created an issue for MCR : https://github.com/openservicemesh/osm/issues/3245

PR #3228, opened 2021-04-23
ads/rds: send always inbound/outbound resources

eduser25:
LDS refers to `rds-inbound` whenever a pod is backed by a service,
however RDS does not necessarily send `rds-inbound` unless there is
inbound routes for the proxy.

The net result is things work, but envoy will still perpetually try to request
`rds-inbound` as it will be kept referenced on LDS (now that we handle proto
resource diffs, this is a problem).

This change makes `rds-inbound` and `rds-outbound` to always be sent, even
when there are no routes in them. If envoy does not request for one or either of them,
they will just be ignored, but otherwise this is a guarantee to be consistent with a potential
LDS reference, when present.

- Control Plane          [x]

No

### DIFF
--- a/pkg/envoy/route/route_config.go
+++ b/pkg/envoy/route/route_config.go
@@ -58,37 +58,36 @@ const (
 func BuildRouteConfiguration(inbound []*trafficpolicy.InboundTrafficPolicy, outbound []*trafficpolicy.OutboundTrafficPolicy, proxy *envoy.Proxy) []*xds_route.RouteConfiguration {
 	var routeConfiguration []*xds_route.RouteConfiguration
 
-	if len(inbound) > 0 {
-		inboundRouteConfig := NewRouteConfigurationStub(InboundRouteConfigName)
-		for _, in := range inbound {
-			virtualHost := buildVirtualHostStub(inboundVirtualHost, in.Name, in.Hostnames)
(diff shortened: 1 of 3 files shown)


pkg/envoy/route/route_config.go
```diff
@@ -58,37 +58,33 @@ const (
...
 	var routeConfiguration []*xds_route.RouteConfiguration
 
-	if len(inbound) > 0 {
-		inboundRouteConfig := NewRouteConfigurationStub(InboundRouteConfigName)
-		for _, in := range inbound {
-			virtualHost := buildVirtualHostStub(inboundVirtualHost, in.Name, in.Hostnames)
-			virtualHost.Routes = buildInboundRoutes(in.Rules)
-			inboundRouteConfig.VirtualHosts = append(inboundRouteConfig.VirtualHosts, virtualHost)
-		}
+	inboundRouteConfig := NewRouteConfigurationStub(InboundRouteConfigName)
```
shashankram, line 61: A few lines of comment regarding why an empty route-config could be sent down would be helpful.

shashankram (approved):

pkg/envoy/route/route_config.go
eduser25: done

eduser25:
fixing unit tests

snehachhabria (approved):

PR #3231, opened 2021-04-23
fix(install): Adds hostaccess SCC when fluentbit is enabled on OpenShift

### LINKED ISSUE #3230: Osm-controller needs hostaccess SCC on OpenShift when fluentbit is enabled
**Bug description**:
Osm-controller is not deployed on OpenShift with fluentbit enabled because fluentbit requires hostPaths. This is blocked by OpenShift by default without the hostaccess SCC enabled.

**Expected behavior**:
osm-controller deploys properly

deploy osm on openshift with fluentbit enabled
**How was OSM installed?**:

**Environment**:
- Others:

ksubrmnn:

**Description**:
Fixes #3230

Adds hostaccess SCC when fluentbit is deployed on OpenShift so that osm-controller can be scheduled without manually adding the scc.

**Affected area**:

- New Functionality      [ ]
- Documentation          [ ]
- Install                [X]
- Control Plane          [ ]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests                  [ ]
- CI System              [ ]
- Demo                   [ ]
- Performance            [ ]
- Other                  [ ]

No

### DIFF
--- a/charts/osm/templates/osm-rbac.yaml
+++ b/charts/osm/templates/osm-rbac.yaml
@@ -48,6 +48,13 @@ rules:
   - apiGroups: ["cert-manager.io"]
     resources: ["certificaterequests"]
     verbs: ["list", "get", "watch", "create", "delete"]
+
+  {{- if and (.Capabilities.APIVersions.Has "security.openshift.io/v1") .Values.OpenServiceMesh.enableFluentbit }}
+  - apiGroups: ["security.openshift.io"]
+    resourceNames: ["hostaccess"]
+    resources: ["securitycontextconstraints"]
+    verbs: ["use"]
+  {{- end }}
 ---
 
 apiVersion: v1


charts/osm/values.yaml
```diff
@@ -159,3 +159,6 @@ OpenServiceMesh:
 
   # -- Run init container in privileged mode
   enablePrivilegedInitContainer: false
+
+  # -- Optional parameter to which Kubernetes distribution OSM is running on.
+  distribution: ""
```
nojnhuh, line 164: Purely a suggestion: For OpenShift, this is something we could detect automatically with something like `{{ Capabilities.APIVersions.Has "security.openshift.io/v1" }}`. I imagine this trick wouldn't work for all distributions, but maybe we could wrap this in a helper like we do for OSM namespace to default it to "openshift" when we can detect it.

charts/osm/values.schema.json
```diff
@@ -596,6 +596,15 @@
...
                     "additionalProperties": true
+                },
+                "distribution": {
+                    "$id": "#/properties/OpenServiceMesh/properties/distribution",
+                    "type": "string",
+                    "title": "Distribution",
+                    "description": "Indicates which Kubernetes distribution OSM is running on",
+                    "examples": [
+                        false
+                    ]
```
nojnhuh, line 607: Let's make this example realistic or take it out entirely.

charts/osm/values.schema.json
```diff
@@ -596,6 +596,15 @@
                         }
                     },
                     "additionalProperties": true
+                },
+                "distribution": {
+                    "$id": "#/properties/OpenServiceMesh/properties/distribution",
+                    "type": "string",
+                    "title": "Distribution",
+                    "description": "Indicates which Kubernetes distribution OSM is running on",
```
nojnhuh, line 604: We should add validation for this here.

nojnhuh (changes requested):
Left a couple nits about the schema for the value. Besides those, LGTM.

charts/osm/values.yaml
ksubrmnn: Using this suggestion. We can revisit having an explicit parameter if customers require it in the future.

nojnhuh (approved):

SanyaKochhar (approved):

PR #3232, opened 2021-04-26
ads: Add resource response validation

eduser25:
Commit introduces a validation function to track that resources responded
to requests are indeed fully responded to, or otherwise a `warn` log is issued
on logs.

This is to make sure we keep track of situations where our logic is not being
consistent, and not returning a superset of the actual resources requested by
envoy.

**Affected area**:

- Control Plane          [x]

No


pkg/envoy/ads/profile.go
```diff
@@ -45,3 +49,34 @@ func (s *Server) trackXDSLog(cn certificate.CommonName, typeURL envoy.TypeURI) {
...
+	// Get resources being responded by name
+	for _, res := range respResources {
+		resourcesToSend.Add(cache.GetResourceName(res))
+	}
+
+	// Compute difference from request resources' perspective
+	resDifference := resourcesRequested.Difference(resourcesToSend)
+	diffCardinality := resDifference.Cardinality()
+	if diffCardinality != 0 {
+		log.Warn().Msgf("Proxy SerialNumber=%s PodUID=%s: not all request resources for type %s are being responded to req [%v] resp [%v] diff [%v]",
```
shashankram, line 76: Is it possible due to timing scenarios where a requested resource is not responded to because the requested resource is no longer relevant as computed by the controller?

eduser25: yes, though if the resource is no longer available/relevant, that should get reflected also on some next update which will/should remove all references to it from all verticals (this is an assumption given our way of doing things).

So, in general, this case is possible but I'd say unlikely. I suggested/committed it as `warn` precisely because it does not automatically imply error, but it's very relevant information to know when debugging imo. wdyt?

shashankram: Thanks for clarifying, that makes sense.

shashankram (approved):

snehachhabria (approved):

PR #3239, opened 2021-04-27
ads: Fix/address wildacard comments and documentation in ADS code

eduser25:
Both CDS and LDS use at all times wildacard mode (not just at init time).
https://www.envoyproxy.io/docs/envoy/latest/api-docs/xds_protocol#client-behavior

This makes envoy always ACK whatever it is sent at face-value for LDS and CDS, and the ACK
itself is also wildcarded (it doesn't specify ACKd resources for LDS and CDS).

Current logic for us still works, as the `len` for resource difference for an empty request resources
is 0, hence is treated rightfully as an ACK.

- Control Plane          [x]

No


eduser25:
ACK from CDS as empty in the following diagram, for ADS cases
https://www.envoyproxy.io/docs/envoy/latest/api-docs/xds_protocol#aggregated-discovery-service

shashankram (approved):

snehachhabria (approved):

PR #3250, opened 2021-04-28
(feat) add initContainer to osm-controller

allenlsy:
The initContainer creates a MeshConfig resource in k8s if it does not
exist

**Description**:

This PR is the continuation from #3222. 

The initial PR failed the CI test. The cause is that, the MeshConfig CRD spelled the field `prometheusScrapping`, while the Go struct spelled the field `prometheusScraping`.

**Affected area**:

- New Functionality      [x]
- Install                [x]
- Control Plane          [x]

`No`


tests/e2e/e2e_controller_test.go
```diff
@@ -0,0 +1,39 @@
+package e2e
```
shashankram, line 1: Could you rename this test to `e2e_init_controller_test.go` or `e2e_controller_init_test.go`.

Also this test needs to be added to the list of tests in `e2e/README.md`

shashankram (approved):

snehachhabria (approved):

PR #3255, opened 2021-04-28
feat(cli/install): remove install cmd cli flags

C123R:
Remove install command CLI flags in favor of using --set flag for overriding helm chart values and update docs referring to osm install cmd.

Reference: #3124

Signed-off-by: Cizer Pereira <mail@cizer.dev>

**Description**:

As mentioned in #3124, if this PR merges - it will remove install cmd flags used specifically to override the default chart values defined in `charts/osm/values.yaml`, in favor of using the `--set` flag.

Removes few flag validation from `install.go` for instance validation of envoy and controller log level type.( which is already implemented using JSON schema(`values.schema.json`) validation)

**Affected area**:

- New Functionality      [ ]
- Documentation          [X]
- Install                [ ]
- Control Plane          [ ]
- CLI Tool               [X]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests                  [ ]
- CI System              [ ]
- Demo                   [ ]
- Performance            [ ]
- Other                  [ ]

  No

### DIFF
--- a/cmd/cli/install.go
+++ b/cmd/cli/install.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"fmt"
 	"io"
-	"strings"
 	"time"
 
 	"github.com/pkg/errors"
@@ -21,7 +20,6 @@ import (
 
 	"github.com/openservicemesh/osm/pkg/cli"
 	"github.com/openservicemesh/osm/pkg/constants"
-	"github.com/openservicemesh/osm/pkg/logger"
 )
 
 const installDesc = `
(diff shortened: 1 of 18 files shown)


nojnhuh:
Hi @C123R, thanks for all the work you've done on this so far!

I'm still taking a look, but I can tell the e2e tests are going to fail because the tests are still using the flags in this function:
https://github.com/openservicemesh/osm/blob/888c268c26fb19f11e6631057748cd360927c825/tests/framework/common.go#L525

nojnhuh:
To clarify the needed e2e changes, I think it's fine for the tests to still have distinct flags for individual values for now, but that function should still translate the flags to the new `--set` equivalent.

e.g.
https://github.com/openservicemesh/osm/blob/888c268c26fb19f11e6631057748cd360927c825/tests/framework/common.go#L614

should be changed to something like

```go
args = append(args, fmt.Sprintf("--set OpenServiceMesh.deployGrafana=%v", instOpts.DeployGrafana))
```

If we want to make a similar change to the tests where we remove fields for individual values, we can do that later.

C123R:
> To clarify the needed e2e changes, I think it's fine for the tests to still have distinct flags for individual values for now, but that function should still translate the flags to the new `--set` equivalent.
> 
> e.g.
> https://github.com/openservicemesh/osm/blob/888c268c26fb19f11e6631057748cd360927c825/tests/framework/common.go#L614
> 
> should be changed to something like
> 
> ```go
> args = append(args, fmt.Sprintf("--set OpenServiceMesh.deployGrafana=%v", instOpts.DeployGrafana))
> ```
> 
> If we want to make a similar change to the tests where we remove fields for individual values, we can do that later.

Thanks, @nojnhuh, let me check this!

demo/run-osm-demo.sh
```diff
@@ -14,7 +14,7 @@ fi
 source .env
 
 # Set meaningful defaults for env vars we expect from .env
-CI="${CI:-false}"  # This is set to true by Github Actions
+CI="${CI:-false}" # This is set to true by Github Actions
```
shashankram, line 17: Hey @C123R, would you mind carving out changes unrelated to this PR, such as this one in a separate PR instead. It would make it easier to review this PR. This includes other changes to indentation, as well.

docs/content/docs/tasks_usage/certificates.md
```diff
@@ -240,8 +248,8 @@ default).
 
 Once ready, it is **required** to store the root CA certificate of your issuer
 as a Kubernetes secret in the OSM namespace (`osm-system` by default) at the
-`ca.crt` key. The target CA secret name can be configured on OSM with the flag
-`--ca-bundle-secret-name` (typically `osm-ca-bundle`).
+`ca.crt` key. The target CA secret name can be configured on OSM using `osm install` set option:
+`OpenServiceMesh.caBundleSecretName` (typically `osm-ca-bundle`).
```
nojnhuh, line 252: Another minor wording nit:
```suggestion
`ca.crt` key. The target CA secret name can be configured on OSM using
`osm install --set OpenServiceMesh.caBundleSecretName=my-secret-name` (typically `osm-ca-bundle`).
```

nojnhuh (changes requested):
Leaving my comments so far. I'll continue looking through the docs once you've reverted the extra changes like Shashank mentioned in https://github.com/openservicemesh/osm/pull/3255#discussion_r623189250.

cmd/cli/mesh_upgrade_test.go
```diff
@@ -51,7 +51,9 @@ func TestMeshUpgradeDefault(t *testing.T) {
 
 	i := getDefaultInstallCmd(ioutil.Discard)
 	i.chartPath = testChartPath
-
+	i.setOptions = []string{
+		"OpenServiceMesh.image.tag=v0.8.2",
+	}
```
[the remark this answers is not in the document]
C123R: Thanks for the hint :)

demo/run-osm-demo.sh
C123R: Sure, will do so.

cmd/cli/install.go
```diff
@@ -243,50 +154,17 @@ func (i *installCmd) loadOSMChart() error {
...
-	if i.containerRegistrySecret != "" {
-		valuesConfig = append(valuesConfig, fmt.Sprintf("OpenServiceMesh.imagePullSecrets[0].name=%s", i.containerRegistrySecret))
-	}
-
-	for _, val := range valuesConfig {
-		// parses Helm strvals line and merges into a map for the final overrides for values.yaml
-		if err := strvals.ParseInto(val, finalValues); err != nil {
-			return nil, err
-		}
+	if err := parseVal(valuesConfig, finalValues); err != nil {
```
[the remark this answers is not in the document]
C123R: > The changes to `resolveValues()` do introduce a slight change in behavior. Before, all the values that could be set by a flag were "overridden" in the Helm release, whether or not the flag for a value was explicitly set. This change will only keep that behavior for `OpenServiceMesh.meshName` and `OpenServiceMesh.enforceSingleMesh`. All other values will only be stored as overridden values for the release if it was passed to the command line with `--set`.
> 
Correct, but I believe this is an improvement that eliminates the need for default values to be duplicated in two places(`install.go` and `values.yml`).
> This shouldn't impact install at all, but does force us to consider how we're handling them for upgrade. [94b3ffa](https://github.com/openservicemesh/osm/commit/94b3ffa8fd9284039045a6d3dc31c5784846863e) describes the precedence order of values as:
> 
> 1. New chart defaults, overridden by
> 2. Old chart defaults, overridden by
> 3. Old release overrides, overridden by
> 4. New release overrides
> 
> Since `osm mesh upgrade` already only sets values when they've explicitly been set on the command line, these changes shouldn't affect its existing behavior.
> 
> Overall, the Helm release will look a bit different, but users probably won't see any changes in behavior. We could emulate the old behavior while still using the rest of these changes if we find it's an issue.

Sorry if I misunderstood, but I don't think this change would affect `mesh upgrade` or its behavior. Since the update is only removing explicit override which is anyways overridden by chart's `values.yaml`.

cmd/cli/install.go
```diff
@@ -55,86 +53,24 @@ well as for adding a Kubernetes Namespace to the list of Namespaces a control
...
 const (
-	defaultCertificateManager            = "tresor"
-	defaultCertManagerIssuerGroup        = "cert-manager.io"
-	defaultCertManagerIssuerKind         = "Issuer"
-	defaultCertManagerIssuerName         = "osm-ca"
-	defaultChartPath                     = ""
-	defaultContainerRegistry             = "openservicemesh"
-	defaultContainerRegistrySecret       = ""
-	defaultMeshName                      = "osm"
-	defaultOsmImagePullPolicy            = "IfNotPresent"
```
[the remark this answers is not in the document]
C123R: good to resolve?

tests/framework/common.go
```diff
@@ -566,27 +566,34 @@ func (td *OsmTestData) InstallOSM(instOpts InstallOSMOpts) error {
...
-		"--enable-debug-server="+strconv.FormatBool(instOpts.EnableDebugServer),
-		"--envoy-log-level="+instOpts.EnvoyLogLevel,
+		fmt.Sprintf("--timeout=%v", 90*time.Second),
+	)
+
+	instOpts.SetOverrides = append(instOpts.SetOverrides,
+		fmt.Sprintf("OpenServiceMesh.image.registry=%s", instOpts.ContainerRegistryLoc),
+		fmt.Sprintf("OpenServiceMesh.image.tag=%s", instOpts.OsmImagetag),
+		fmt.Sprintf("OpenServiceMesh.certificateManager=%s", instOpts.CertManager),
+		fmt.Sprintf("OpenServiceMesh.enableEgress=%s", strconv.FormatBool(instOpts.EgressEnabled)),
```
nojnhuh, line 577: Nit: I would either use concatenation with `+` and not `fmt.Sprintf` or use `%v/%t` in the format string so then the `strconv.FormatBool` isn't needed.

So either

```go
"OpenServiceMesh.someBool=" + strconv.FormatBool(someBool),
```

or

```go
fmt.Sprintf("OpenServiceMesh.someBool=%v", someBool),
```

shashankram (commented):
Hey @C123R, this change looks really good. There seem to be some unit test failures (Go test GH action is failing). Almost looking good to go.

tests/framework/common.go
C123R: Ahh, good catch - it's fixed now.

C123R:
> Hey @C123R, this change looks really good. There seem to be some unit test failures (Go test GH action is failing). Almost looking good to go.

Should be fine now.  🤞

shashankram (approved):

nojnhuh (approved):
Looks good! Thanks!

PR #3261, opened 2021-04-30
ads: do not handle empty URI

### LINKED ISSUE #3258: XDS state machine results in continuous looping when updated listeners are yet to drain

eduser25:
- Renames `Wildcard` to `EmptyURI` to avoid naming overlap with xDS
"wildcard" mode for a given URI.
- Avoid handling Empty URI as some of the logic needs some work for
it to be properly handled.

We will revisit this after further exploration.

Fixes #3258

- Control Plane          [x]

No

### DIFF
--- a/pkg/envoy/types.go
+++ b/pkg/envoy/types.go
@@ -4,7 +4,6 @@ package envoy
 
 import (
 	"github.com/openservicemesh/osm/pkg/logger"
-	"github.com/openservicemesh/osm/pkg/utils"
 )
 
 var (
@@ -24,20 +23,12 @@ func (t TypeURI) String() string {
 
 // Short returns an abbreviated version of the TypeURI, which is easier to spot in logs and metrics.
 func (t TypeURI) Short() string {
-	if t == TypeWildcard {
-		return Wildcard
-	}
-	return utils.GetLastChunkOfSlashed(t.String())
-}
-
-// IsWildcard returns true when the TypeURI is the one requesting ALL discovery types.
-func (t TypeURI) IsWildcard() bool {
-	return t == TypeWildcard
+	return XDSShortURINames[t]
 }
 
 // ValidURI defines valid URIs
 var ValidURI = map[string]TypeURI{
-	string(TypeWildcard):           TypeWildcard,
+	string(TypeEmptyURI):           TypeEmptyURI,
 	string(TypeSDS):                TypeSDS,
 	string(TypeCDS):                TypeCDS,
 	string(TypeLDS):                TypeLDS,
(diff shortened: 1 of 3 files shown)


pkg/envoy/types.go
```diff
@@ -24,20 +23,16 @@ func (t TypeURI) String() string {
...
 func (t TypeURI) Short() string {
-	if t == TypeWildcard {
-		return Wildcard
-	}
-	return utils.GetLastChunkOfSlashed(t.String())
+	return XDSShortURINames[t]
 }
 
 // IsWildcard returns true when the TypeURI is the one requesting ALL discovery types.
 func (t TypeURI) IsWildcard() bool {
```
shashankram, line 30:
```suggestion
func (t TypeURI) IsEmptyTypeURI() bool {
```

pkg/envoy/types.go
```diff
@@ -24,20 +23,16 @@ func (t TypeURI) String() string {
...
-	return t == TypeWildcard
+	return t == TypeEmptyURI
 }
 
 // ValidURI defines valid URIs
 var ValidURI = map[string]TypeURI{
-	string(TypeWildcard):           TypeWildcard,
```
shashankram, line 40: Removing empty typeURI results in error logs, which is not correct per these changes:
```
{"level":"error","component":"envoy/ads","time":"2021-04-30T16:44:57Z","file":"stream.go:174","message":"Proxy SerialNumber=136867002654723736450590910962478278124 PodUID=a9d0fb42-e2ce-4596-a3aa-92631baeb4ce: Unknown/Unsupported URI: "}
```

pkg/envoy/types.go
eduser25: :+1:

pkg/envoy/types.go
eduser25: I will instead, allow `emptyUrl` as valid, but add a check just right after to skip empty url with debug log instead

shashankram (approved):

snehachhabria (approved):

PR #3287, opened 2021-05-04
fix(cli): check mesh exists before namespace add

### LINKED ISSUE #2494: osm CLI should check if mesh exists before adding namespace

`osm namespace add` should check if the mesh exists before adding the namespace. Otherwise the user may be mislead into thinking that the namespace is part of a mesh despite forgetting to run the install command.

**Possible use cases**

johnsonshi:
The command osm namespace add should check if
the mesh exists before adding the namespace.
Otherwise the user may be mislead into thinking
that the namespace is part of a mesh despite
forgetting to run the install command.

Resolves #2494

**Affected area**:

- New Functionality      [ ]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [ ]
- CLI Tool               [X]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests                  [ ]
- CI System              [ ]
- Demo                   [ ]
- Performance            [ ]
- Other                  [ ]

- Does this change contain code from or inspired by another project? No. If so, did you notify the maintainers and provide attribution? N/A.

### DIFF
--- a/cmd/cli/namespace_add.go
+++ b/cmd/cli/namespace_add.go
@@ -88,6 +88,14 @@ func (a *namespaceAddCmd) run() error {
 		ctx, cancel := context.WithCancel(context.Background())
 		defer cancel()
 
+		exists, err := meshExists(a.clientSet, a.meshName)
+		if err != nil {
+			return err
+		}
+		if !exists {
+			return errors.Errorf("Mesh [%s] does not exist. Please specify another mesh using --mesh-name or create a new mesh.", a.meshName)
+		}
+
 		deploymentsClient := a.clientSet.AppsV1().Deployments(ns)
 		labelSelector := metav1.LabelSelector{MatchLabels: map[string]string{"app": constants.OSMControllerName}}
 
@@ -132,7 +140,7 @@ func (a *namespaceAddCmd) run() error {
 }`, constants.OSMKubeResourceMonitorAnnotation, a.meshName, constants.SidecarInjectionAnnotation)
 		}
 
-		_, err := a.clientSet.CoreV1().Namespaces().Patch(ctx, ns, types.StrategicMergePatchType, []byte(patch), metav1.PatchOptions{}, "")
+		_, err = a.clientSet.CoreV1().Namespaces().Patch(ctx, ns, types.StrategicMergePatchType, []byte(patch), metav1.PatchOptions{}, "")
 		if err != nil {
 			return errors.Errorf("Could not add namespace [%s] to mesh [%s]: %v", ns, a.meshName, err)
 		}
@@ -141,4 +149,21 @@ func (a *namespaceAddCmd) run() error {
 	}
 
 	return nil
+}
+
+// meshExists determines if a mesh with meshName exists within the cluster
+func meshExists(clientSet kubernetes.Interface, meshName string) (bool, error) {
+	// search for the mesh across all namespaces
+	deploymentsClient := clientSet.AppsV1().Deployments("")
+	// search and match using the mesh name provided
+	labelSelector := metav1.LabelSelector{MatchLabels: map[string]string{"meshName": meshName}}
+	listOptions := metav1.ListOptions{
+		LabelSelector: labels.Set(labelSelector.MatchLabels).String(),
+	}
+	osmControllerDeployments, err := deploymentsClient.List(context.TODO(), listOptions)
+	if err != nil {
+		return false, errors.Errorf("Cannot obtain information about the mesh [%s]: [%v]", meshName, err)
+	}
+	// the mesh is present if there are osm controllers for the mesh
+	return len(osmControllerDeployments.Items) != 0, nil
 }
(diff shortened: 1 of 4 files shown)


cmd/cli/namespace_add.go
```diff
@@ -142,3 +151,17 @@ func (a *namespaceAddCmd) run() error {
 
 	return nil
 }
+
+func getMeshes(clientSet kubernetes.Interface, meshName string, namespace string) ([]v1.Deployment, error) {
```
allenlsy, line 155: Is it possible that we deployed multiple meshes in one namespace? What's our assumption here?

If getMeshes() returns multiple meshes, why the `meshName` is given here?

shashankram: There cannot be multiple meshes in the same control plane namespace. Multiple instances of the control plane can exist as replicas within a single mesh though.

johnsonshi: @allenlsy and @shashankram sorry the name wasn't clear. What this does is check if the mesh with `meshName` is present in the cluster by getting `meshName`'s controllers. If the controllers for `meshName` exists, then that means `meshName` exists. This means we can proceed to add `namespace` to be managed by the mesh `meshName`.

If there are no controllers, then that means the mesh `meshName` does not exist.

image
image

cmd/cli/namespace_add.go
```diff
@@ -88,6 +90,14 @@ func (a *namespaceAddCmd) run() error {
 		ctx, cancel := context.WithCancel(context.Background())
 		defer cancel()
 
+		controllers, err := getControllerDeployementsForMesh(a.clientSet, a.meshName, "")
```
shashankram, line 93: I think it might be more succinct to have a `func meshExists(name, namespace) bool` and wrap the logic of controllers being present within the function.

cmd/cli/namespace_add.go
```diff
@@ -142,3 +152,18 @@ func (a *namespaceAddCmd) run() error {
...
+	listOptions := metav1.ListOptions{
+		LabelSelector: labels.Set(labelSelector.MatchLabels).String(),
+	}
+	osmControllerDeployments, err := deploymentsClient.List(context.TODO(), listOptions)
+	return osmControllerDeployments.Items, err
+}
+
+func errMeshNotFound(name string) error {
+	return errors.Errorf("Mesh [%s] does not exist. Please specify another mesh using --mesh-name or create a new mesh.", name)
+}
```
shashankram, line 169: why not `errors.Errorf()` directly instead of this intermediary function?

cmd/cli/util.go
```diff
@@ -61,3 +68,28 @@ func annotateErrorMessageWithActionableMessage(actionableMessage string, errMsgF
 
 	return errors.Errorf(errMsgFormat+actionableMessage, args...)
 }
+
+// helper function for tests that adds deployment to the clientset
+func addDeployment(fakeClientSet kubernetes.Interface, depName string, meshName string, namespace string, osmVersion string, isMesh bool) (*v1.Deployment, error) {
```
shashankram, line 73: Since this is only used by test code, kindly move it to a `_test.go` file so the actual binary doesn't use it.

cmd/cli/namespace_add.go
johnsonshi: Addressed in latest changes.

cmd/cli/namespace_add.go
johnsonshi: Addressed in latest changes.

cmd/cli/util.go
johnsonshi: Addressed in latest changes.

cmd/cli/namespace_add.go
```diff
@@ -142,3 +151,20 @@ func (a *namespaceAddCmd) run() error {
...
+	listOptions := metav1.ListOptions{
+		LabelSelector: labels.Set(labelSelector.MatchLabels).String(),
+	}
+	osmControllerDeployments, err := deploymentsClient.List(context.TODO(), listOptions)
+	if err != nil {
+		return false, errors.Errorf("Cannot obtain information about the mesh [%s]: [%v]", meshName, err)
+	}
+	// the mesh is present if there are osm controllers for the mesh
+	return len(osmControllerDeployments.Items) != 0, nil
+}
```
johnsonshi, line 170: @shashankram & @jaellio: As per both of your suggestions, I have simplified + clarified the code that determines whether the mesh exists within the cluster.

shashankram (approved):

cmd/cli/namespace_add.go
jaellio: For documentation: I suggested removing the namespace parameter from getMeshController since this value would only be "" to search all namespaces for an existing mesh.

jaellio (approved):

PR #3307, opened 2021-05-06
Create preset meshconfig

allenlsy:
**Description**:

We still need to support setting mesh configurations from CLI during installation. This was removed when migrating from osm-config configmap to MeshConfig CRD.

In this PR, `osm install` will create a MeshConfig resource `preset-mesh-config` using Helm. `init-osm-controller` container will check it and create a `osm-mesh-config` MeshConfig if does not exist. In this case, it actually means that OSM is in the fresh installation process.. However if `osm-mesh-config` exists, `init-osm-controller` respects the existing MeshConfig, and ignores the preset. This implies that OSM is in upgrade process.

Note that, in this PR, we are not fixing the issue that `osm upgrade` will also take arguments to update MeshConfig. 

Issue #3301 

**Affected area**:

- New Functionality      [ ]
- Documentation          [ ]
- Install                [X]
- Control Plane          [ ]
- CLI Tool               [X]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests                  [ ]
- CI System              [ ]
- Demo                   [ ]
- Performance            [ ]
- Other                  [ ]

`No`

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -193,7 +193,7 @@ jobs:
         if: ${{ github.event_name == 'pull_request' }}
         env:
           K8S_NAMESPACE: "osm-system"
-        run: go test ./tests/e2e -test.v -ginkgo.v -ginkgo.progress -installType=KindCluster -test.timeout 0 -test.failfast -ginkgo.failFast -ginkgo.focus='\[Tier 1\]\[Bucket ${{ matrix.bucket }}\]' -ginkgo.skip='Upgrade'
+        run: go test ./tests/e2e -test.v -ginkgo.v -ginkgo.progress -installType=KindCluster -test.timeout 0 -test.failfast -ginkgo.failFast -ginkgo.skip='Upgrade'
         continue-on-error: true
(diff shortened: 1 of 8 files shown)


allenlsy:
I will be OOF for a week. Other members might take over this PR. 

Here is the progress of the current PR:

- [x] preset-mesh-config template
- [x] logic that `init-osm-controller` checks preset and create actual MeshConfig if it does not exist

TODO:

Test cases:

- [ ] in a fresh OSM install, created `osm-mesh-config` should have the values set in values.yaml. `preset-mesh-config` should be deleted after installation
- [ ] in an upgrade, existing `osm-mesh-config` should not be changed at all, regardless of the the values.yaml or arguments in the `osm upgrade`

There is an open question discussion regarding `osm upgrade` command capability in #3301

snehachhabria:
Closing this as I have created an alternative PR : https://github.com/openservicemesh/osm/pull/3331

PR #3311, opened 2021-05-07
fix(portExclusion) : Fix port exclusion to array of int

### LINKED ISSUE #3308: Change outboundPortExclusionList type to array of int
**Bug description**:

The current definition of `outboundPortExclusionList` in MeshConfig is an array of strings. 

**Expected behavior**:

This should be changed to an array of integers.

**How was OSM installed?**:

**Environment**:
- Others:

snehachhabria:
**Description**:

This commit updates the type of portExclusion from []string to []int

Fixes #3308

**Affected area**:

- New Functionality      [ ]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [ ]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests                  [ ]
- CI System              [ ]
- Demo                   [ ]
- Performance            [ ]
- Other                  [ ]

### DIFF
--- a/charts/osm/crds/meshconfig.yaml
+++ b/charts/osm/crds/meshconfig.yaml
@@ -91,7 +91,7 @@ spec:
                       description: Global list of ports to exclude from outbound traffic interception by the sidecar proxy.
                       type: array
                       items:
-                        type: string
+                        type: integer
                         pattern: ^[1-9]\d*$
                     useHTTPSIngress:
                       description: Enable HTTPS ingress on the mesh
--- a/pkg/configurator/methods.go
+++ b/pkg/configurator/methods.go
@@ -155,18 +155,8 @@ func (c *Client) GetOutboundIPRangeExclusionList() []string {
 }
 
 // GetOutboundPortExclusionList returns the list of ports (positive integers) to exclude from outbound sidecar interception
-func (c *Client) GetOutboundPortExclusionList() []string {
-	portsStr := c.getMeshConfig().OutboundPortExclusionList
-	if portsStr == "" {
-		return nil
-	}
-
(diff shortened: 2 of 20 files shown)


charts/osm/crds/meshconfig.yaml
```diff
@@ -91,7 +91,7 @@ spec:
                       description: Global list of ports to exclude from outbound traffic interception by the sidecar proxy.
                       type: array
                       items:
-                        type: string
+                        type: integer
                         pattern: ^[1-9]\d*$
```
allenlsy, line 95: integer does not have a pattern I think?

pkg/configurator/methods.go
```diff
@@ -155,17 +155,8 @@ func (c *Client) GetOutboundIPRangeExclusionList() []string {
...
-		return nil
-	}
-
-	portExclusionList := strings.Split(portsStr, ",")
-	for i := range portExclusionList {
-		portExclusionList[i] = strings.TrimSpace(portExclusionList[i])
-	}
-
+func (c *Client) GetOutboundPortExclusionList() []int {
+	portExclusionList := c.getMeshConfig().OutboundPortExclusionList
```
allenlsy, line 159: nit: prefer make it one line.

```
return c.getMeshConfig().OutboundPortExclusionList
```

allenlsy (approved):

shashankram (approved):

charts/osm/crds/meshconfig.yaml
snehachhabria: it is allowed, and I would want a check for positive integers only

PR #3326, opened 2021-05-11
meshconfig: add resource limits to MeshConfig sidecar spec

eduser25:
Commit adds resource limits to meshconfig, in the sidecar spec. It follows and uses
upstream resource limits definition for a resource from corev1.

- Control Plane          [x]

No


eduser25:
small test:
```
meshConfig.Spec.Sidecar.Resources = v1.ResourceRequirements{
	Limits: v1.ResourceList{
		v1.ResourceCPU: resource.MustParse("1"),
	},
}

Td.ConfigClient.ConfigV1alpha1().MeshConfigs(Td.OsmNamespace).Update(context.TODO(), meshConfig, metav1.UpdateOptions{})
Expect(err).ShouldNot(HaveOccurred())
```

kubectl get MeshConfig osm-mesh-config -o yaml
```
  sidecar:
    enablePrivilegedInitContainer: false
    envoyImage: envoyproxy/envoy-alpine:v1.17.2
    initContainerImage: openservicemesh/init:v0.8.3
    logLevel: error
    maxDataPlaneConnections: 0
    resources:
      limits:
        cpu: "1"
```

charts/osm/crds/meshconfig.yaml
```diff
@@ -73,6 +73,17 @@ spec:
                       description: Image for the init container
                       type: string
                       default: "openservicemesh/init:v0.8.3"
+                    resources:
+                      type: object
+                      properties:
+                        limits:
+                          description: "Limits describes the maximum amount of compute resources allowed. More info: https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/"
+                          type: object
```
shashankram, line 81: Should we define the sub types `cpu and memory`?

snehachhabria: +1

charts/osm/crds/meshconfig.yaml
```diff
@@ -73,6 +73,17 @@ spec:
                       description: Image for the init container
                       type: string
                       default: "openservicemesh/init:v0.8.3"
+                    resources:
```
snehachhabria, line 76: What would be the use case of this, further would this be configurable at the time of install ?

snehachhabria: Also would there be any defaults, if its not specified ?

eduser25: I didn't do that to be compliant and be able to reuse corev1's `resourceRequirements`. Consider cpu and memory are not the only possible values here (ie. hugepage limits are also available).

The internal client implementation of corev1's `resourceRequirements` for `limits` and `requests` are actually `map[string]limit`, and the way to support this is the one I used (map type doesn't exist in crd).

charts/osm/crds/meshconfig.yaml
eduser25: I didn't do that to be compliant and be able to reuse corev1's `resourceRequirements`. Consider cpu and memory are not the only possible values here (ie. hugepage limits are also available).

The internal client implementation of corev1's `resourceRequirements` for `limits` and `requests` are actually `map[string]limit`, and the way to support this is the one I used (map type doesn't exist in crd).

eduser25: more in detail...
```
// ResourceRequirements describes the compute resource requirements.
type ResourceRequirements struct {
	// Limits describes the maximum amount of compute resources allowed.
	// More info: https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/
	// +optional
	Limits ResourceList `json:"limits,omitempty" protobuf:"bytes,1,rep,name=limits,casttype=ResourceList,castkey=ResourceName"`
	// Requests describes the minimum amount of compute resources required.
	// If Requests is omitted for a container, it defaults to Limits if that is explicitly specified,
	// otherwise to an implementation-defined value.
	// More info: https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/
	// +optional
	Requests ResourceList `json:"requests,omitempty" protobuf:"bytes,2,rep,name=requests,casttype=ResourceList,castkey=ResourceName"`
}
```
```
// ResourceList is a set of (resource name, quantity) pairs.
type ResourceList map[ResourceName]resource.Quantity
```
```
type ResourceName string
```
So it's basically a map[string]string, only the value gets parsed as an actual object (`mustParse`)

shashankram: Reasonable if this is of type `map[string]string`

shashankram (approved):

snehachhabria (approved):

PR #3329, opened 2021-05-12
egress: add support for server-first TCP protocols

shashankram:
**Description**:
This change adds support to enable egress for server-first
protocols such as MySQL, PostgreSQL etc. by bypassing the
listener filters that wait on client data,  which would
otherwise time out for such protocols.
An egress policy for TCP based server-first protocols must
set the protocol to `tcp-server-first` to indicate that
protocol detection must be skipped for such a port.

Verified this change with mySQL based egress and confirmed
that such traffic no longer times out the listener filter
that wait on data from the client.

Part of #3045

**Affected area**:

- New Functionality      [X]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [X]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests                  [ ]
- CI System              [ ]
- Demo                   [ ]
- Performance            [ ]
- Other                  [ ]

`No`

### DIFF
--- a/pkg/catalog/egress.go
+++ b/pkg/catalog/egress.go
@@ -44,7 +44,7 @@ func (mc *MeshCatalog) GetEgressTrafficPolicy(serviceIdentity identity.ServiceId
 					DestinationProtocol: portSpec.Protocol,
 				})
 
-			case constants.ProtocolTCP:
+			case constants.ProtocolTCP, constants.ProtocolTCPServerFirst:
 				// ---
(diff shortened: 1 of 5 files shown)


draychev (approved):
🎆

snehachhabria (approved):

PR #3335, opened 2021-05-13
egress: update comments for types and add field validation

shashankram:
**Description**:
Updates the comments for the API types and adds validation
for the IPAddresses and Port fields.

Part of #3045

**Affected area**:

- New Functionality      [ ]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [ ]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests                  [ ]
- CI System              [ ]
- Demo                   [ ]
- Performance            [ ]
- Other                  [X]

`No`

### DIFF
--- a/charts/osm/crds/policy.yaml
+++ b/charts/osm/crds/policy.yaml
@@ -73,6 +73,7 @@ spec:
                   type: array
                   items:
                     type: string
+                    pattern: ((?:\d{1,3}\.){3}\d{1,3})\/(\d{1,2})$
                 ports:
                   description: Ports that the sources are allowed to direct external traffic to.
                   type: array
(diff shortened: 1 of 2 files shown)


charts/osm/crds/policy.yaml
```diff
@@ -85,6 +86,7 @@ spec:
                       number:
                         description: Port number of this port.
                         type: integer
+                        pattern: ^[1-9]\d*$
```
eduser25, line 89: would be good to prevent non-uint16 ranges. guess this could also be done by validating webhook

eduser25 (approved):

snehachhabria (approved):

charts/osm/crds/policy.yaml
```diff
@@ -73,6 +73,7 @@ spec:
                   type: array
                   items:
                     type: string
+                    pattern: ((?:\d{1,3}.){3}\d{1,3})\/(\d{1,2})$
```
jaellio, line 76: Should the "." be escaped?

jaellio (approved):

charts/osm/crds/policy.yaml
shashankram: Done!

charts/osm/crds/policy.yaml
shashankram: It should, good catch!

Issue #3268, opened 2021-04-30
Remove intermediate data structure `osm-config`


allenlsy:
**Please describe the Improvement and/or Feature Request**

`osm-config` is a intermediate data structure used for the configmap. Since configuration with configmap is deprecated, `osm-config` is no longer needed. 

Related file: https://github.com/openservicemesh/osm/blob/0e9429bb9c44bb6b026719d895449e534c66a1ad/pkg/configurator/client.go#L204

**Scope (please mark with X where applicable)**
- New Functionality          [ ]
- Install                    [ ]
- SMI Traffic Access Policy  [ ]
- SMI Traffic Specs Policy   [ ]
- SMI Traffic Split Policy   [ ]
- Permissive Traffic Policy  [ ]
- Ingress                    [ ]
- Egress                     [ ]
- Envoy Control Plane        [ ]
- CLI Tool                   [ ]
- Metrics                    [ ]
- Certificate Management     [ ]
- Sidecar Injection          [ ]
- Logging                    [ ]
- Debugging                  [ ]
- Tests                      [ ]
- CI System                  [ ]
- Demo                       [ ]
- Project Release            [ ]
- Code refactor             [x] 

**Possible use cases**

Related issues:

#2738

PR #3344, opened 2021-05-14
chore(MeshConfig): Removing reduntant configMap code

snehachhabria:
**Description**:

This commit removes references to configMap

Part of #3040

**Affected area**:

- New Functionality      [ ]
- Documentation          [ ]
- Install                [ ]
- Control Plane          [ ]
- CLI Tool               [ ]
- Certificate Management [ ]
- Networking             [ ]
- Metrics                [ ]
- SMI Policy             [ ]
- Security               [ ]
- Tests                  [ ]
- CI System              [ ]
- Demo                   [ ]
- Performance            [ ]
- Other                  [ ]


charts/osm/templates/cleanup-hook.yaml
```diff
@@ -1,28 +0,0 @@
...
-      restartPolicy: Never
-      containers:
-        - name: garbage-collector
-          image: bitnami/kubectl
-          imagePullPolicy: IfNotPresent
-          command:
-            - sh
-            - -c
-            - >
-             kubectl delete --ignore-not-found configmap -n '{{ include "osm.namespace" . }}' osm-config;
```
shashankram, line 28: This should delete the default mesh config going instead.

pkg/configurator/client.go
```diff
@@ -24,46 +24,46 @@ const (
...
-	// PermissiveTrafficPolicyModeKey is the key name used for permissive mode in the ConfigMap
+	// PermissiveTrafficPolicyModeKey is the key name used for permissive mode in the MeshConfig
 	PermissiveTrafficPolicyModeKey = "permissive_traffic_policy_mode"
 
-	// egressKey is the key name used for egress in the ConfigMap
+	// egressKey is the key name used for egress in the MeshConfig
 	egressKey = "egress"
 
-	// enableDebugServer is the key name used for the debug server in the ConfigMap
+	// enableDebugServer is the key name used for the debug server in the MeshConfig
```
shashankram, line 33: These comments are not accurate. The key names are still referring to the older ConfigMap keys. Ex. `permissive_traffic_policy_mode -> enablePermissiveTrafficPolicyMode`

pkg/configurator/client.go
```diff
@@ -226,7 +226,7 @@ func meshConfigDeletedMessageHandler(psubMsg *events.PubSubMessage) {
 }
 
 func meshConfigUpdatedMessageHandler(psubMsg *events.PubSubMessage) {
-	// Get config map
+	// Get mesh config
```
shashankram, line 229:
```suggestion
	// Get the MeshConfig resource
```

pkg/constants/constants.go
```diff
@@ -147,6 +147,9 @@ const (
 
 	// OSMConfigMap is the name of the OSM ConfigMap
 	OSMConfigMap = "osm-config"
```
shashankram, line 149: remove.

pkg/kubernetes/events/event_pubsub_test.go
```diff
@@ -21,15 +21,6 @@ func TestPubSubEvents(t *testing.T) {
 		publish       PubSubMessage
 		expectMessage bool
 	}{
-		{
```
shashankram, line 24: Is the CRD equivalent already there for this test or does it need to be added while this is being removed?

pkg/configurator/client.go
eduser25: all these keys are only used in testing (client_test only). I think it would be in line to also get rid of them in a separate commit, by a caritative soul that wants to handle this unholy burden.

pkg/kubernetes/events/event_pubsub_test.go
eduser25: `client_test.go` -> `TestMeshConfigEventTriggers`

eduser25 (approved):

pkg/configurator/client.go
snehachhabria: Yes, we can get rid of keys all together which will be handled in a separate commit, it is tracked by this issue : https://github.com/openservicemesh/osm/issues/3268

charts/osm/templates/cleanup-hook.yaml
snehachhabria: good catch, I added it back and updated the file to delete the mesh config instead

pkg/constants/constants.go
snehachhabria: This cannot be removed as its being referenced in the validating webhook code, once that code is removed we can remove this

pkg/kubernetes/events/event_pubsub_test.go
snehachhabria: Thanks Edu! Since the equivalent already exists, I will go ahead and delete this

shashankram (approved):

PR #3351, opened 2021-05-14
fix(MeshConfig): adding rbac for deleting meshConfig

snehachhabria:
**Description**:

The PR adds rbac to delete meshconfig resources in osm, while using the helm cleanup hooks. This was causing issues in the CI

Signed-off-by: Sneha Chhabria snchh@microsoft.com

**Affected area**:

### DIFF
--- a/charts/osm/templates/osm-rbac.yaml
+++ b/charts/osm/templates/osm-rbac.yaml
@@ -90,7 +90,7 @@ rules:
     verbs: ["list"]
   - apiGroups: ["config.openservicemesh.io"]
     resources: ["meshconfigs"]
-    verbs: ["get", "list", "watch", "create", "update", "patch"]
+    verbs: ["get", "list", "watch", "create", "update", "patch", "delete"]
   - apiGroups: ["split.smi-spec.io"]
     resources: ["trafficsplits"]
     verbs: ["list", "get", "watch"]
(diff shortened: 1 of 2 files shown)


charts/osm/templates/osm-rbac.yaml
```diff
@@ -90,7 +90,7 @@ rules:
     verbs: ["list"]
   - apiGroups: ["config.openservicemesh.io"]
     resources: ["meshconfigs"]
-    verbs: ["get", "list", "watch", "create", "update", "patch"]
+    verbs: ["get", "list", "watch", "create", "update", "patch", "delete"]
```
shashankram, line 93: Duplicate present in L104

snehachhabria: removed duplicate

nojnhuh (approved):

charts/osm/templates/cleanup-hook.yaml
```diff
@@ -26,3 +26,4 @@ spec:
             - -c
             - >
              kubectl delete --ignore-not-found meshconfig -n '{{ include "osm.namespace" . }}' osm-mesh-config;
+             kubectl delete --ignore-not-found meshconfig -n '{{ include "osm.namespace" . }}' preset-mesh-config;
```
shashankram, line 29: Why is this required? Shouldn't this be automatically deleted by Helm?

snehachhabria: the configmap also previously was installed by helm but this cleanup hook handled it, hence I added the preset-mesh-config

cc @nojnhuh

nojnhuh: We added the configmap deletion here in anticipation of removing it from the chart. Since the preset-mesh-config is part of the chart, we don't need to delete it here.

shashankram: Hmm, unsure why this would be necessary. @nojnhuh was there a reason to explicitly remove this?

shashankram (approved):

charts/osm/templates/cleanup-hook.yaml
snehachhabria: thanks for the clarification @nojnhuh , I have removed the preset-mesh-config

PR #3359, opened 2021-05-17
feat(extAuthz): Add external authorization to meshconfig

eduser25:
- Adds a rather agnostic ExternalAuthorization type object in
meshconfig, for generic definition of an external authorization entity.
- Adds `InboundExternalAuthorization`, which will allow configuring
inbound and egress external authorization, under `Traffic`.

No
1. Is this a breaking change?
No


snehachhabria (approved):

charts/osm/crds/meshconfig.yaml
```diff
@@ -117,6 +117,34 @@ spec:
                       description: True for allowing traffic to flow between client and service pods within the mesh without SMI traffic policies, i.e. no traffic policy enforcement in the mesh. If set to false, enables deny-all traffic policy in mesh i.e. an SMI Traffic Target is necessary for services to communicate.
                       type: boolean
                       default: false
+                    inboundExternalAuthorization:
+                      description: Configures additional external authorization for inbound and ingress connections.
```
shashankram, line 121: What does additional imply here?

charts/osm/crds/meshconfig.yaml
```diff
@@ -117,6 +117,34 @@ spec:
...
+                          maximum: 65535
+                        statPrefix:
+                          description: String prefix for inbound external authorization related metrics.
+                          type: string
+                          default: "inboundExtAuthz"
+                        timeout:
+                          description: Defines the timeout to consider for the remote endpoint to reply in time.
+                          type: string
+                          default: "1s"
+                        failureModeAllow:
```
shashankram, line 144: All of these seem specific to what OPA needs. Might as well call the top level key something along the lines of `inboundExternalOPAAuth`

pkg/apis/config/v1alpha1/mesh_config.go
```diff
@@ -77,6 +77,10 @@ type TrafficSpec struct {
 
 	// EnablePermissiveTrafficPolicyMode defines a boolean indicating if permissive traffic policy mode is enabled mesh-wide.
 	EnablePermissiveTrafficPolicyMode bool `json:"enablePermissiveTrafficPolicyMode,omitempty"`
+
+	// InboundExternalAuthorization defines a ruleset that, if enabled, will configure a remote external authorization endpoint
+	// for all inbound and ingress traffic towards the mesh.
```
shashankram, line 82:
```suggestion
	// for all inbound and ingress traffic in the mesh.
```

pkg/apis/config/v1alpha1/mesh_config.go
```diff
@@ -106,6 +110,29 @@ type TracingSpec struct {
 	Endpoint string `json:"endpoint,omitempty"`
 }
 
+// ExternalAuthorizationSpec is a type to represent external authorization configuration
+type ExternalAuthzSpec struct {
```
shashankram, line 114: Let's make this struct specific to OPA? Don't think this needs to be made generic just in its name.

charts/osm/crds/meshconfig.yaml
eduser25: i think I had in mind RBAC or other things we might do, if a user wants to. I'll remove the `additional`.

charts/osm/crds/meshconfig.yaml
eduser25: The config for envoy proto is agnostic, and I think we should remain as such. Multiple external authorization services already exists, by just being in line with envoy we can support any of them (or at least subset, as long we can translate config)
https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/ext_authz/v3/ext_authz.proto.html

pkg/apis/config/v1alpha1/mesh_config.go
eduser25: :+1:

charts/osm/crds/meshconfig.yaml
shashankram: Thanks, that makes sense.

charts/osm/crds/meshconfig.yaml
```diff
@@ -117,6 +117,34 @@ spec:
                       description: True for allowing traffic to flow between client and service pods within the mesh without SMI traffic policies, i.e. no traffic policy enforcement in the mesh. If set to false, enables deny-all traffic policy in mesh i.e. an SMI Traffic Target is necessary for services to communicate.
                       type: boolean
                       default: false
+                    inboundExternalAuthorization:
+                      description: Configures additional external authorization for inbound and ingress connections.
+                      type: object
+                      properties:
+                        enable:
+                          description: Enables/disables the inbound external authorization spec if present.
```
shashankram, line 125:
```suggestion
                          description: Enables/disables the inbound external authorization policy if present.
```

pkg/apis/config/v1alpha1/mesh_config.go
```diff
@@ -106,6 +110,29 @@ type TracingSpec struct {
...
+	// Enable defines a boolean indicating if this external authorization spec is to be active.
```
shashankram, line 115:
```suggestion
	// Enable defines a boolean indicating if the external authorization policy is enabled.
```

pkg/apis/config/v1alpha1/mesh_config.go
```diff
@@ -106,6 +110,29 @@ type TracingSpec struct {
...
+	Enable bool `json:"enable,omitempty"`
+
+	// Address defines the remote address of the external authorization endpoint
+	Address string `json:"address,omitempty"`
+
+	// Port defines the destination port to talk to the remote external authorization endpoint
```
shashankram, line 121:
```suggestion
	// Port defines the destination port of the remote external authorization endpoint
```

pkg/apis/config/v1alpha1/mesh_config.go
```diff
@@ -106,6 +110,29 @@ type TracingSpec struct {
...
+
+	// StatPrefix defines a prefix for the stats sink for this external authorization spec
+	StatPrefix string `json:"statPrefix,omitempty"`
+
+	// Timeout defines the timeout in which a response from the external authorization endpoint
+	// is expected to execute.
+	Timeout string `json:"timeout,omitempty"`
+
+	// FailureModeAllow defines if to allow traffic if failure to get a response against the external
+	// authorization endpoint is observed.
```
shashankram, line 132:
```suggestion
	// FailureModeAllow defines a boolean indicating if traffic should be allowed on a failure to get a response against the external
	// authorization endpoint.
```

charts/osm/crds/meshconfig.yaml
eduser25: :+1:

shashankram (approved):

PR #3365, opened 2021-05-18
feat(extAuthz): Add inbound external authorization

eduser25:
Commit adds external authorization.

Current version of the feature allows configuring external authorization
through MeshConfig's `inboundExternalAuthorization`.

When `enabled`, external authorization is enforced for all inbound and
ingress for all pods in the mesh.

patchset[1]:
    - moved auth-only configuration definition to standalone `auth` package.
    - Moved doc out, will push to osm-docs.
    - s/incoming/inbound s/outgoing/outbound

no

1. Is this a breaking change?
no

### DIFF
--- a/pkg/auth/types.go
+++ b/pkg/auth/types.go
@@ -1,0 +1,26 @@
+package auth
+
+import (
+	"time"
+)
+
+// ExtAuthConfig implements a generic subset of External Authz to configure external authorization through HttpFilters
+type ExtAuthConfig struct {
+	// Enable enables/disables the inbound external authorization policy if present.
+	Enable bool
+
+	// Address is the target destination endpoint that will handle external authorization.
+	Address string
+
+	// Port is the remote destination port for the external authorization endpoint.
+	Port uint16
+
+	// StatPrefix is a prefix for inbound external authorization related metrics.
+	StatPrefix string
+
+	// AuthzTimeout defines the timeout to consider for the remote endpoint to reply in time.
+	AuthzTimeout time.Duration
+
+	// FailureModeAllow allows specifying if traffic should succeed or fail if the external authorization endpoint fails to respond.
+	FailureModeAllow bool
+}
--- a/pkg/configurator/client.go
+++ b/pkg/configurator/client.go
@@ -80,6 +80,9 @@ const (
 
 	// proxyResources is the key used to configure proxy resources
 	proxyResourcesKey = "proxy_resources"
+
+	// inboundExtAuthz is the key used to enable the inbound external authorization filter
+	inboundExtAuthz = "inbound_ext_authz"
 )
 
 // NewConfigurator implements configurator.Configurator and creates the Kubernetes client to manage namespaces.
@@ -201,6 +204,10 @@ func parseOSMMeshConfig(meshConfig *v1alpha1.MeshConfig) *osmConfig {
 		osmConfig.TracingPort = int(spec.Observability.Tracing.Port)
 	}
 
+	if spec.Traffic.InboundExternalAuthorization.Enable {
+		osmConfig.inboundExternalAuth = spec.Traffic.InboundExternalAuthorization
+	}
+
 	return &osmConfig
 }
 
@@ -249,6 +256,22 @@ func meshConfigUpdatedMessageHandler(psubMsg *events.PubSubMessage) {
 	triggerGlobalBroadcast = triggerGlobalBroadcast || (prevMeshConfig.TracingAddress != newMeshConfig.TracingAddress)
 	triggerGlobalBroadcast = triggerGlobalBroadcast || (prevMeshConfig.TracingEndpoint != newMeshConfig.TracingEndpoint)
 	triggerGlobalBroadcast = triggerGlobalBroadcast || (prevMeshConfig.TracingPort != newMeshConfig.TracingPort)
+
+	triggerGlobalBroadcast = triggerGlobalBroadcast || (prevMeshConfig.inboundExternalAuth.Enable != newMeshConfig.inboundExternalAuth.Enable)
+	// Do not trigger updates on the inner configuration changes of ExtAuthz if disabled,
+	// or otherwise skip checking if the update is to be scheduled anyway
+	if newMeshConfig.inboundExternalAuth.Enable && !triggerGlobalBroadcast {
+		triggerGlobalBroadcast = triggerGlobalBroadcast ||
+			(prevMeshConfig.inboundExternalAuth.Address != newMeshConfig.inboundExternalAuth.Address)
+		triggerGlobalBroadcast = triggerGlobalBroadcast ||
+			(prevMeshConfig.inboundExternalAuth.Port != newMeshConfig.inboundExternalAuth.Port)
+		triggerGlobalBroadcast = triggerGlobalBroadcast ||
+			(prevMeshConfig.inboundExternalAuth.StatPrefix != newMeshConfig.inboundExternalAuth.StatPrefix)
+		triggerGlobalBroadcast = triggerGlobalBroadcast ||
+			(prevMeshConfig.inboundExternalAuth.Timeout != newMeshConfig.inboundExternalAuth.Timeout)
+		triggerGlobalBroadcast = triggerGlobalBroadcast ||
+			(prevMeshConfig.inboundExternalAuth.FailureModeAllow != newMeshConfig.inboundExternalAuth.FailureModeAllow)
+	}
 
 	if triggerGlobalBroadcast {
 		log.Debug().Msgf("[%s] OSM MeshConfig update triggered global proxy broadcast",
@@ -351,4 +374,7 @@ type osmConfig struct {
 
 	// proxyResources are the proxy resources speficied for a proxy, if any
 	proxyResources corev1.ResourceRequirements
+
+	// inboundExternalAuth is the external authorization configuration for inbound and ingress
+	inboundExternalAuth v1alpha1.ExternalAuthzSpec
 }
--- a/pkg/configurator/methods.go
+++ b/pkg/configurator/methods.go
@@ -8,6 +8,7 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 
+	"github.com/openservicemesh/osm/pkg/auth"
 	"github.com/openservicemesh/osm/pkg/constants"
 )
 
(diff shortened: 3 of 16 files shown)


pkg/configurator/client.go
```diff
@@ -80,6 +80,9 @@ const (
 
 	// proxyResources is the key used to configure proxy resources
 	proxyResourcesKey = "proxy_resources"
+
+	// inboundExtauthzEnable is the key used to enable the inbound external authorization filter
+	inboundExtauthz = "inbound_extauthz"
```
shashankram, line 85: s/inboundExtauthz/inboundExtAuthz?
s/inbound_extauthz/inbound_ext_authz?

pkg/configurator/client.go
```diff
@@ -250,6 +257,20 @@ func meshConfigUpdatedMessageHandler(psubMsg *events.PubSubMessage) {
 	triggerGlobalBroadcast = triggerGlobalBroadcast || (prevMeshConfig.TracingEndpoint != newMeshConfig.TracingEndpoint)
 	triggerGlobalBroadcast = triggerGlobalBroadcast || (prevMeshConfig.TracingPort != newMeshConfig.TracingPort)
 
+	triggerGlobalBroadcast = triggerGlobalBroadcast || (prevMeshConfig.inboundExternalAuth.Enable != newMeshConfig.inboundExternalAuth.Enable)
+	if newMeshConfig.inboundExternalAuth.Enable && !triggerGlobalBroadcast {
```
shashankram, line 261: Why is this conditional needed? Why can't we update if the previous config does not match the new one as evaluated in L260?

At the very least, some comments are required here since the code looks a bit convoluted.

pkg/configurator/methods.go
```diff
@@ -182,3 +182,24 @@ func (c *Client) GetConfigResyncInterval() time.Duration {
...
+	extAuthConfig.Enable = inboundExtAuthzMeshConfig.Enable
+	extAuthConfig.Address = inboundExtAuthzMeshConfig.Address
+	extAuthConfig.Port = uint16(inboundExtAuthzMeshConfig.Port)
+	extAuthConfig.StatPrefix = inboundExtAuthzMeshConfig.StatPrefix
+	extAuthConfig.FailureModeAllow = inboundExtAuthzMeshConfig.FailureModeAllow
+
+	duration, err := time.ParseDuration(inboundExtAuthzMeshConfig.Timeout)
+	if err != nil {
+		log.Debug().Err(err).Msgf("ExternAuthzTimeout: Not a valid duration %s. defaulting to 1s.", duration)
+		duration = 1 * time.Second
```
shashankram, line 200: Why is a default being used here? Should this config be rejected instead so the user is aware?

pkg/configurator/types.go
```diff
@@ -89,4 +89,17 @@ type Configurator interface {
...
+}
+
+// ExternAuthConfig implements a generic subset of External Authz to configure external authorization in envoy's format
+type ExternAuthConfig struct {
+	Enable           bool
+	Address          string
+	Port             uint16
+	StatPrefix       string
+	AuthzTimeout     time.Duration
+	FailureModeAllow bool
```
shashankram, line 104: I'd create a separate pkg for this `pkg/auth` and put this definition there so it can be reused by configurator and XDS.

pkg/envoy/lds/connection_manager.go
```diff
@@ -17,13 +17,22 @@ import (
 	"github.com/openservicemesh/osm/pkg/featureflags"
 )
 
+// connectionDirection defines, for filter terms, the direction of the traffic from an application
+// perspective, in which the connection manager filters will be applied
+type connectionDirection string
```
shashankram, line 22: since the comment refers to the direction of traffic, maybe just call this `trafficDirection`.

pkg/envoy/lds/connection_manager.go
```diff
@@ -17,13 +17,22 @@ import (
...
+
 const (
 	meshHTTPConnManagerStatPrefix       = "mesh-http-conn-manager"
 	prometheusHTTPConnManagerStatPrefix = "prometheus-http-conn-manager"
 	prometheusInboundVirtualHostName    = "prometheus-inbound-virtual-host"
+
+	// incoming defines in-mesh inbound and ingress traffic driections
+	incoming = "incoming"
+	// outgoing defines in-mesh outbound and egress traffic directions
+	outgoing = "outgoing"
```
shashankram, line 32: For consistency with LDS naming conventions, s/incoming/inbound, s/outgoing/outbound.

pkg/envoy/lds/connection_manager.go
```diff
@@ -47,6 +51,16 @@ func getHTTPConnectionManager(routeName string, cfg configurator.Configurator, h
 		AccessLog: envoy.GetAccessLog(),
 	}
 
+	incomingExtAuthCfg := cfg.GetInboundExternalAuthConfig()
```
shashankram, line 54: Make the call to `GetInboundExternalAuthConfig()` only if the direction is inbound (minor perf optimization).

pkg/envoy/lds/egress_test.go
```diff
@@ -222,6 +225,9 @@ func TestGetEgressFilterChainsForMatches(t *testing.T) {
 				cfg: mockConfigurator,
 			}
 			mockConfigurator.EXPECT().IsTracingEnabled().Return(false).AnyTimes()
+			mockConfigurator.EXPECT().GetInboundExternalAuthConfig().Return(configurator.ExternAuthConfig{
+				Enable: false,
+			}).AnyTimes()
```
shashankram, line 230: Let's avoid this by only making this function call for the inbound direction.

shashankram (changes requested):
Change looks good, left some comments I prefer we address in this revision.

docs/external_auth_opa.md
```diff
@@ -0,0 +1,119 @@
+# OPA-plugin-enabled OSM POC
```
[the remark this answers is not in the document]
eduser25: ill take it out

pkg/configurator/client.go
eduser25: :+1:

pkg/configurator/client.go
eduser25: I'll add a comment. It's just a small optimization, if extAuthz remains disabled, there's no point in checking or triggering updates because the inner config changed. Same thing should be done with the tracing section.

pkg/configurator/types.go
eduser25: So, Q for you. Here I added the extra structure in config, but we could have just propagated the `v1alpha.ExternalAuthzSpec` that comes directly from meshconfig (after all, they both hold the exact same things).
Though this structure does not bring anything new, I thought we might want a different structure to represent segment what inner parts of OSM do not interact directly with MeshSpec, and only configurator does.

wdyt? would you skip this intermediate structure?

pkg/configurator/client.go
shashankram: That optimization makes sense. Carving these into checks might facilitate easier unit testing.

pkg/envoy/lds/connection_manager.go
eduser25: :+1:

pkg/configurator/types.go
shashankram: That's a good question. It depends on if you want to tie API definitions to internal implementations. For example, if there is ever a case where an internal struct does not need to be the same as the API Go struct, or if the internal struct can be extended with additional metadata, then it makes sense to keep it separate so you don't have to change the API (new API version rollout). Updating a CRD in theory should be a version update once users start using the API.

However if you don't forsee such a scenario, it makes sense to avoid the intermediate struct mapping.

pkg/envoy/lds/connection_manager.go
eduser25: I thought we talked about this. I thought `inbound` is for in-mesh only, but here I needed a way to both identify `inbound` and `ingress`, hence the `incoming` tag. Lmk if you have other thoughts.

pkg/envoy/lds/connection_manager.go
eduser25: the `enable` is in the config :), can't skip calling this guy

pkg/envoy/lds/egress_test.go
eduser25: I guess you mean outbound, and pair it with your previous suggestion to only call `GetInboundExternalAuthConfig` when is inbound.

Good idea. this will remove a lot of `mockConfigurator` test code.

pkg/envoy/lds/connection_manager.go
shashankram: I consider `inbound == mesh inbound || ingress`

pkg/configurator/methods.go
eduser25: Not sure we can reject or consider the entire config nil if timeout fails to parse? what's your take on this?
IMO enforcing a correct parseable timeout string should be part of the Mesh CRD spec somehow. We can return empty config here if timeout fails to parse, but sounds like overkill to me.

pkg/envoy/lds/connection_manager.go
eduser25: oh... ok, I'll change in this case

pkg/configurator/methods.go
shashankram: We will have a validator for this soon, so invalid values will be prevented.

pkg/envoy/lds/connection_manager.go
eduser25: done

pkg/auth/auth.go
```diff
@@ -0,0 +1,54 @@
...
+
+	"github.com/openservicemesh/osm/pkg/logger"
+)
+
+var (
+	log = logger.New("auth")
+)
+
+// GetExtAuthzHTTPFilter returns an envoy HttpFilter given an ExternAuthConfig configuration
+func GetExtAuthzHTTPFilter(extAuthConfig ExternAuthConfig) *xds_hcm.HttpFilter {
```
shashankram, line 20: Sorry this is not what I meant when I suggested moving the struct type to pkg/auth.

This function should very much be under `pkg/envoy/lds`

eduser25: I wonderd about that. moving back.

shashankram (approved):

eduser25:
tnx for patience and reviews

pkg/auth/types.go
```diff
@@ -0,0 +1,26 @@
+package auth
+
+import (
+	"time"
+)
+
+// ExternAuthConfig implements a generic subset of External Authz to configure external authorization through HttpFilters
+type ExternAuthConfig struct {
```
snehachhabria, line 8:
```suggestion
type ExternalAuthConfig struct {
```

snehachhabria, line 8:
```suggestion
type ExtAuthConfig struct {
```

snehachhabria (approved):

pkg/auth/types.go
eduser25: :+1:

eduser25: :+1:

PR #3380, opened 2021-05-19
RFC: ref(build): refer to compatible images in CLI

### LINKED ISSUE #3371: Ensure CLI references compatible images
## Problem

Currently, the `main` branch references the latest tagged version for the default images used by the OSM chart (and therefore `osm install`):
https://github.com/openservicemesh/osm/blob/3fde0f12166148558c14bc5f1e05a16593954c68/charts/osm/values.yaml#L14

This causes version skew when the `main` branch is updated but the default image tag stays the same. This becomes an issue when following documentation corresponding to the `main` branch which refers to new features since the last release and instructions use `osm install` without overriding the images to images newer than the latest release that implement those features.

## Goal

The goal is to come up with a solution such that `make build-osm && bin/osm install` "just works" in all cases.

To do this, `make build-osm` would have to become dependent on building and pushing the images used by the control plane, then the CLI would refer to those images by default. The image tag must be unique to maintain a 1:1 mapping of CLI to images.

## Solution
Here's one possible implementation I can think of:

1. list `docker-push-{init,init-osm-controller,osm-controller,osm-injector}` as dependencies of `build-osm` in the Makefile:
    https://github.com/openservicemesh/osm/blob/b14458f9dd15d2a967e1ef8e4a0a0317ba753a6b/Makefile#L108
    - Ensure the `CTR_TAG` is set to some deterministic value. The current Git SHA might be close enough, though ideally it would differentiate between dirty and clean worktrees. `CTR_REGISTRY` can be set to 
2. create new package-scoped variables (likely in cmd/cli/install.go) for the tag and registry location of the built images and pass the values to them in the Makefile using the same method as we do for version info:
    https://github.com/openservicemesh/osm/blob/3fde0f12166148558c14bc5f1e05a16593954c68/Makefile#L21
3. In the install command, if `OpenServiceMesh.image.registry` or `OpenServiceMesh.image.tag` are not overridden via `--set`, override them with the values of the package-scoped variables defined above. This could be done in this function:
    https://github.com/openservicemesh/osm/blob/3fde0f12166148558c14bc5f1e05a16593954c68/cmd/cli/install.go#L154

    This would still allow users to override the images the standard way per-`osm install` invocation.

## Other things I haven't fully thought through yet
- Installing with `helm` instead of `osm`
- images stored in password-protected registries
- Upgrades
- Release builds and maintenance

/cc @shashankram

nojnhuh:
**Description**:
Currently, the `osm install` command will use the default image tag
listed in the chart's values.yaml which corresponds to the latest
release. Using the default images can cause problems when the main
branch introduces changes to the images that are not included by default
when `osm install` is invoked. This can include missing features

This change synchronizes the images referenced by default for `osm install`
with published images built from source with a unique tag, ensuring that
each CLI build references images that are compatible to it and will not
go out of date.

Fixes #3371

**Affected area**:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,7 +18,7 @@ BUILD_DATE_VAR := github.com/openservicemesh/osm/pkg/version.BuildDate
 BUILD_VERSION_VAR := github.com/openservicemesh/osm/pkg/version.Version
 BUILD_GITCOMMIT_VAR := github.com/openservicemesh/osm/pkg/version.GitCommit
 
-LDFLAGS ?= "-X $(BUILD_DATE_VAR)=$(BUILD_DATE) -X $(BUILD_VERSION_VAR)=$(VERSION) -X $(BUILD_GITCOMMIT_VAR)=$(GIT_SHA) -X main.chartTGZSource=$$(cat -) -s -w"
+LDFLAGS ?= "-X $(BUILD_DATE_VAR)=$(BUILD_DATE) -X $(BUILD_VERSION_VAR)=$(VERSION) -X $(BUILD_GITCOMMIT_VAR)=$(GIT_SHA) -X main.chartTGZSource=$$(cat -) -X main.defaultImageTag=$(CTR_TAG) -X main.defaultImageRegistry=$(CTR_REGISTRY) -s -w"
 
 # These two values are combined and passed to go test
 E2E_FLAGS ?= -installType=KindCluster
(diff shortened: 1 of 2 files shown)


nojnhuh:
For anyone that wants to try this out, `make build-osm` will now also build and push the OSM control plane images, so make sure your `CTR_REGISTRY` and `CTR_TAG` Makefile variables are set appropriately. Then, `bin/osm install` should work and reference the images that were just built.

Makefile
```diff
@@ -105,7 +105,8 @@ build-osm-injector: check-go-version clean-osm-injector
 	CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -v -o ./bin/osm-injector/osm-injector -ldflags "-X $(BUILD_DATE_VAR)=$(BUILD_DATE) -X $(BUILD_VERSION_VAR)=$(VERSION) -X $(BUILD_GITCOMMIT_VAR)=$(GIT_SHA) -s -w" ./cmd/osm-injector
 
 .PHONY: build-osm
-build-osm: check-go-version
+build-osm: CTR_TAG=$(GIT_SHA)
+build-osm: check-go-version $(addprefix docker-push-, init init-osm-controller osm-controller osm-injector)
```
shashankram, line 109: Can we avoid docker-push when just building the binary? Since docker-push on the targets can take time, it would increase the time to check if the CLI compiles using `make build-osm`.

nojnhuh: I'm not sure how we would avoid pushing the images if we want to reference them by default for `osm install`. e.g. If I add a new feature in osm-controller that I want to `osm install`, then a new osm-controller image has to get built at some point before the CLI is invoked.

I suppose I could add a new Makefile target that's CLI+images and have the CLI built before the images so those errors are surfaced first. Then the CLI-only target could be invoked by itself too, but at that point we've basically come full circle back to where we are now.

Overall it seems like this is going to make CLI builds too heavyweight to be worth it, especially since it's solely for convenience and doesn't really solve any problems.

nojnhuh:
We already have a good-enough solution to this issue, so I'm not planning on taking this any further.

PR #3382, opened 2021-05-19
ref(controller): remove osmConfig struct

allenlsy:

**Description**:

Remove intermediate struct `osmConfig`. It was previously used to represent the data of osm-config configmap. As OSM migrates to MeshConfig, `osmConfig` is not useful.

**Affected area**:

`No`

1. Is this a breaking change?

`No`

### DIFF
--- a/pkg/configurator/client.go
+++ b/pkg/configurator/client.go
@@ -2,9 +2,7 @@ package configurator
 
 import (
 	"fmt"
-	"strings"
 
-	corev1 "k8s.io/api/core/v1"
 	"k8s.io/client-go/tools/cache"
 
 	"github.com/openservicemesh/osm/pkg/announcements"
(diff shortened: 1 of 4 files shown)


pkg/configurator/client_test.go
```diff
@@ -27,78 +24,6 @@ const (
 	osmMeshConfigName = "-test-osm-mesh-config-"
 )
 
-var _ = Describe("Test OSM MeshConfig parsing", func() {
```
allenlsy, line 30: This test case is removed, since there is no parsing any more.

allenlsy:
https://github.com/openservicemesh/osm/blob/8da72874d95b3258c6d04cceb253e613f07527d4/pkg/configurator/methods_test.go#L216

@snehachhabria  @shashankram  is this a hard requirement that after tracing.enabled set to false, other tracing fields will be reset to default?

shashankram:
> https://github.com/openservicemesh/osm/blob/8da72874d95b3258c6d04cceb253e613f07527d4/pkg/configurator/methods_test.go#L216
> 
> @snehachhabria @shashankram is this a hard requirement that after tracing.enabled set to false, other tracing fields will be reset to default?

No, it shouldn't. If tracing is disabled other tracing parameters can be ignored.

allenlsy:
@shashankram  cool. So that the changes here can be removed.

https://github.com/openservicemesh/osm/pull/3382/files#diff-09ca33777a9661a71336ae83cf3400056ece37a5c5287200c868b685d281a6f5R164

eduser25 (approved):
👍

pkg/configurator/client_test.go
```diff
@@ -255,7 +180,7 @@ func TestMeshConfigEventTriggers(t *testing.T) {
 				meshConfig.Spec.Observability.Tracing.Port = int16(port)
 			case envoyLogLevelKey:
 				meshConfig.Spec.Sidecar.LogLevel = mapVal
-			case enableDebugServer:
+			case enableDebugServerKey:
```
snehachhabria, line 183: deltaMeshConfigContents no longer needs to be a map with the keys, you can use the meshConfig object in the test case. Please update this test to remove reference to keys and map

pkg/configurator/client.go
```diff
@@ -31,10 +29,7 @@ const (
 	egressKey = "egress"
 
 	// enableDebugServer is the key name used for the debug server in the MeshConfig
-	enableDebugServer = "enable_debug_server"
-
-	// prometheusScrapingKey is the key name used for prometheus scraping in the MeshConfig
-	prometheusScrapingKey = "prometheus_scraping"
+	enableDebugServerKey = "enable_debug_server"
```
snehachhabria, line 32: Once the test is updated these constants wont be needed

jaellio (approved):

pkg/configurator/client_test.go
allenlsy: Updated test cases. Can you check again?

allenlsy:
Close issue #3268

PR #3391, opened 2021-05-21
Makes unit test pass on Windows

davinci26:
Make checkout of yaml files to be the same on both platforms and adds a small delay to test to accommodate for Windows which has lower accuracy in `time.Now()`

### DIFF
--- a/pkg/envoy/proxy_test.go
+++ b/pkg/envoy/proxy_test.go
@@ -3,7 +3,9 @@ package envoy
 import (
 	"context"
 	"fmt"
+	"runtime"
 	"testing"
+	"time"
 
 	"github.com/golang/mock/gomock"
 	"github.com/google/uuid"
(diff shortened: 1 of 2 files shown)


snehachhabria (approved):

pkg/envoy/proxy_test.go
```diff
@@ -63,7 +64,8 @@ var _ = Describe("Test proxy methods", func() {
 
 			firstNonce := proxy.GetLastSentNonce(TypeCDS)
 			Expect(firstNonce).ToNot(Equal(uint64(0)))
-
+			// Accomodate platforms where time now has lower accuracy.
```
shashankram, line 67: Would make this comment more specific to name the platform so readers have context.
`// XXX(Windows): accommodate for slowness where time has lower accuracy...`

davinci26: WDYT `// Platform(<name>) as a a comment pattern for platform specific issues. I can add any windows specific comments there myself and it might be useful in the future for `arm` or specific quirks in POSIX OS

shashankram: I'd also suggest to put this under a conditional so we don't mask bugs arising from timing issues:
```
if runtime.GOOS == "windows" {
    time.Sleep(x)
}
```

shashankram (approved):

PR #3393, opened 2021-05-21
Update Envoy config to comply with 1.18.3 api

davinci26:
This patch makes the following updates:

1. Use `type.googleapis.com/envoy.extensions.access_loggers.stream.v3.StdoutAccessLog` in admin
2. Use `type.googleapis.com/envoy.extensions.access_loggers.stream.v3.StdoutAccessLog` in listeners
3. Use explicit http2 config

### DIFF
--- a/pkg/envoy/cds/cluster.go
+++ b/pkg/envoy/cds/cluster.go
@@ -35,6 +35,11 @@ func getUpstreamServiceCluster(downstreamIdentity identity.ServiceIdentity, upst
 		return nil, err
 	}
 
+	HTTP2ProtocolOptions, err := envoy.GetHTTP2ProtocolOptions()
+	if err != nil {
+		return nil, err
+	}
+
 	remoteCluster := &xds_cluster.Cluster{
 		Name:           clusterName,
 		ConnectTimeout: ptypes.DurationProto(clusterConnectTimeout),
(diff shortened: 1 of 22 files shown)


pkg/envoy/xdsutil.go
```diff
@@ -216,6 +217,26 @@ func GetUpstreamTLSContext(downstreamIdentity identity.ServiceIdentity, upstream
 	return tlsConfig
 }
 
+// GetHTTP2ProtocolOptions creates an Envoy http configuration that matches the downstream protocol
+func GetHTTP2ProtocolOptions() map[string]*any.Any {
```
shashankram, line 221: Would return an error and let the caller handle it instead of silently filling a nil value in a struct field.

shashankram (approved):

davinci26:
@shashankram do you want to do another round of review

pkg/envoy/cds/cluster.go
```diff
@@ -158,8 +166,12 @@ func getEgressClusters(clusterConfigs []*trafficpolicy.EgressClusterConfig) []*x
...
 			// Cluster config does not have a Host specified, route it to its original destination.
 			// Used for TCP based clusters
-			egressClusters = append(egressClusters, getOriginalDestinationEgressCluster(config.Name))
-
+			originalDestinationEgressCluster, err := getOriginalDestinationEgressCluster(config.Name)
+			if err != nil {
+				log.Error().Err(err).Msg("Error building the original destination cluster for the given egress cluster config")
+			} else {
+				egressClusters = append(egressClusters, originalDestinationEgressCluster)
+			}
```
shashankram, line 174: nit: collapse conditional:
```suggestion
			if originalDestinationEgressCluster, err := getOriginalDestinationEgressCluster(config.Name);  err != nil {
				log.Error().Err(err).Msg("Error building the original destination cluster for the given egress cluster config")
			} else {
				egressClusters = append(egressClusters, originalDestinationEgressCluster)
			}
```

shashankram (approved):

pkg/envoy/cds/cluster.go
davinci26: neat golang feature

draychev (approved):

PR #3475, opened 2021-05-28
ref(pkg/injector): use XDS structs for envoy bootstrap config

jaellio:
**Description**:

Updates the envoy bootstrap config to be constructed using XDS
structs directly instead of maps. The XDS objects are marshaled
to JSON and then to YAML. Using the XDS structs directly will help
ensure the objects are configured correctly and all required
fields are set. Additionally, this change will make it easier to
enforce the use of the same API version across the injector and
controller.

**Affected area**:
| Sidecar Injection          | [ X ] |

### DIFF
--- a/pkg/injector/envoy_config.go
+++ b/pkg/injector/envoy_config.go
@@ -2,13 +2,27 @@ package injector
 
 import (
 	"context"
-	"encoding/base64"
 	"fmt"
-	"strconv"
+	"time"
 
+	"google.golang.org/protobuf/encoding/protojson"
+	"google.golang.org/protobuf/reflect/protoreflect"
+	"google.golang.org/protobuf/types/known/durationpb"
 	"gopkg.in/yaml.v2"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	xds_access "github.com/envoyproxy/go-control-plane/envoy/config/accesslog/v3"
+	xds_bootstrap "github.com/envoyproxy/go-control-plane/envoy/config/bootstrap/v3"
+	xds_cluster "github.com/envoyproxy/go-control-plane/envoy/config/cluster/v3"
+	xds_core "github.com/envoyproxy/go-control-plane/envoy/config/core/v3"
+	xds_endpoint "github.com/envoyproxy/go-control-plane/envoy/config/endpoint/v3"
+	xds_listener "github.com/envoyproxy/go-control-plane/envoy/config/listener/v3"
+	xds_transport_sockets "github.com/envoyproxy/go-control-plane/envoy/extensions/transport_sockets/tls/v3"
+	xds_upstream_http "github.com/envoyproxy/go-control-plane/envoy/extensions/upstreams/http/v3"
+
+	"github.com/golang/protobuf/ptypes"
+	"github.com/golang/protobuf/ptypes/any"
 
 	"github.com/openservicemesh/osm/pkg/certificate"
 	"github.com/openservicemesh/osm/pkg/configurator"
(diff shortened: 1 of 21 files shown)


pkg/injector/envoy_config.go
```diff
@@ -214,5 +278,37 @@ func getXdsCluster(config envoyBootstrapConfigMeta) map[string]interface{} {
...
+	var jsonObj interface{}
+	// We are using yaml.Unmarshal here (instead of json.Unmarshal) because the
+	// Go JSON library doesn't try to pick the right number type (int, float,
+	// etc.) when unmarshalling to interface{}, it just picks float64
+	// universally. go-yaml does go through the effort of picking the right
+	// number type, so we can preserve number type throughout this process.
+	err = yaml.Unmarshal([]byte(configJSON), &jsonObj)
+	if err != nil {
+		return nil, err
+	}
```
shashankram, line 304: This can be a separate function that references https://github.com/ghodss/yaml/blob/v1.0.0/yaml.go#L89 (references code for a release tag instead of master).

pkg/injector/envoy_config_health_probes.go
```diff
@@ -63,100 +86,229 @@ func getProbeCluster(clusterName string, port int32) map[string]interface{} {
...
 				},
-				"http_filters": []map[string]interface{}{
-					{
-						"name": "envoy.filters.http.router",
+			},
+		}
+		pbHTTPConnectionManager, err := ptypes.MarshalAny(httpConnectionManager)
+		if err != nil {
+			log.Error().Err(err).Msgf("Error marshaling HttpConnectionManager struct into an anypb.Any message")
+			return nil
```
shashankram, line 132: nit: have this function return an error else health probes functionality could be broken since an `append()` with `nil` value is valid.

pkg/injector/envoy_config.go
```diff
@@ -214,5 +278,37 @@ func getXdsCluster(config envoyBootstrapConfigMeta) map[string]interface{} {
 				},
 			},
 		},
+	}, nil
+}
+
+func xdsToYaml(m protoreflect.ProtoMessage) ([]byte, error) {
```
shashankram, line 284:
```suggestion
func protoToYaml(m protoreflect.ProtoMessage) ([]byte, error) {
```

pkg/injector/envoy_config_health_probes.go
```diff
@@ -63,100 +86,229 @@ func getProbeCluster(clusterName string, port int32) map[string]interface{} {
...
+						},
 					},
 				},
 			},
 		},
 	}
 }
 
+// GetAccessLog creates an Envoy AccessLog struct.
+func GetAccessLog() []*xds_accesslog_filter.AccessLog {
```
shashankram, line 216: un-export this function.

pkg/injector/envoy_config_health_probes.go
```diff
@@ -63,100 +86,229 @@ func getProbeCluster(clusterName string, port int32) map[string]interface{} {
...
+	return []*xds_accesslog_filter.AccessLog{{
+		Name: wellknown.FileAccessLog,
+		ConfigType: &xds_accesslog_filter.AccessLog_TypedConfig{
+			TypedConfig: accessLog,
+		}},
+	}
+}
+
+// GetTCPAccessLog creates an Envoy AccessLog struct.
+func GetTCPAccessLog() []*xds_accesslog_filter.AccessLog {
```
shashankram, line 231: un-export this.

pkg/injector/test_fixtures/expected_envoy_bootstrap_config.yaml
```diff
@@ -23,7 +23,7 @@ dynamic_resources:
     resource_api_version: V3
 static_resources:
   clusters:
-  - connect_timeout: 0.25s
+  - connect_timeout: 0.250s
```
shashankram, line 26: Was this change intentional?

pkg/injector/envoy_config.go
```diff
@@ -152,60 +183,93 @@ func (wh *mutatingWebhook) createEnvoyBootstrapConfig(name, namespace, osmNamesp
...
+	}
+	pbUpstreamTLSContext, err := ptypes.MarshalAny(upstreamTLSContext)
+	if err != nil {
+		log.Error().Err(err).Msgf("Error marshaling UpstreamTlsContext struct into an anypb.Any message")
+		return nil, err
+	}
+
+	return &xds_cluster.Cluster{
+		Name:           config.XDSClusterName,
+		ConnectTimeout: durationpb.New(time.Millisecond * 250),
```
shashankram, line 242: This was previously 25ms.

pkg/injector/test_fixtures/expected_envoy_bootstrap_config.yaml
```diff
@@ -72,7 +71,6 @@ static_resources:
     name: liveness_cluster
     type: STATIC
   - connect_timeout: 1s
-    lb_policy: ROUND_ROBIN
```
shashankram, line 75: We should be setting this in the XDS struct.

shashankram (commented):
Looks almost good to go. Noticed some discrepancies in values of certain object fields as noted in the comments inline.

pkg/injector/envoy_config_health_probes.go
jaellio: Done

pkg/injector/envoy_config.go
jaellio: I think it was .25 seconds which is 250 ms

pkg/injector/test_fixtures/expected_envoy_bootstrap_config.yaml
jaellio: When the xds structs are marshaled to YAML the 0 after .25 remains

pkg/injector/envoy_config.go
jaellio: Updated

pkg/injector/test_fixtures/expected_envoy_bootstrap_config.yaml
jaellio: Round robin is the default lb_policy. I am currently setting the lb_policy 'LbPolicy: xds_cluster.Cluster_ROUND_ROBIN', but the value of 'xds_cluster.Cluster_ROUND_ROBIN' is 0. The protojson marshaler considers 0 an empty value and it is therefore omitted. I have not found a way to keep default values. I have the same issue with setting 'codec_type: AUTO' for a filter.

pkg/injector/envoy_config_health_probes.go
jaellio: Should the getters also return an error to be handled in `getStaticResources` in `envoy_config.go`?

pkg/injector/test_fixtures/expected_envoy_bootstrap_config.yaml
jaellio: Here is the issue discussing protojson support for masking what fields are omitted https://github.com/golang/protobuf/issues/1183

pkg/injector/envoy_config.go
```diff
@@ -2,81 +2,117 @@ package injector
...
+	"github.com/golang/protobuf/ptypes"
+	"github.com/golang/protobuf/ptypes/any"
+
 	"github.com/openservicemesh/osm/pkg/certificate"
 	"github.com/openservicemesh/osm/pkg/configurator"
 	"github.com/openservicemesh/osm/pkg/constants"
 	"github.com/openservicemesh/osm/pkg/version"
 )
 
 func getEnvoyConfigYAML(config envoyBootstrapConfigMeta, cfg configurator.Configurator) ([]byte, error) {
```
allenlsy, line 33: We encountered similar issue before that, people assumes this methods returns a YAML format string, while it is a byte array.

Do you want to make it return string instead?

E.g. https://github.com/openservicemesh/osm/pull/3388

pkg/injector/envoy_config_health_probes.go
shashankram: I would suggest returning an error from functions that can generate errors. In some cases such as this one where `MarshalAny` won't result in an error (because of correctness in the http connection manager struct), it is more about correctness.

pkg/injector/test_fixtures/expected_envoy_bootstrap_config.yaml
shashankram: My bad, this looks good!

pkg/injector/envoy_config.go
shashankram: My bad, looks correct.

shashankram (approved):

snehachhabria (approved):

pkg/injector/envoy_config.go
jaellio: Good idea. I will fix this in a follow-up PR

PR #3479, opened 2021-06-01
charts/osm: allow additional root properties

shashankram:
**Description**:
Currently, the schema prevents additional
Helm chart properties other than OpenServiceMesh
which is the root property for all values.

This change enforces strict validation only
for the OpenServiceMesh root property. Other
root properties that might be leveraged by
Helm charts using the OSM chart as a dependency
will not be subjected to schema validations.

**Affected area**:

### DIFF
--- a/charts/osm/values.schema.json
+++ b/charts/osm/values.schema.json
@@ -913,6 +913,5 @@
             },
             "additionalProperties": false
         }
-    },
-    "additionalProperties": false
+    }
 }


eduser25 (approved):

snehachhabria (approved):

shashankram:
@Mergifyio backport release-v0.9

PR #3485, opened 2021-06-01
test: add test for init-osm-controller methods

### LINKED ISSUE #3404: Tests: add test cases for cmd/init-osm-controller validate CLI params

Add test cases for cmd/init-osm-controller `validateCLIParams()`, when `osmNamespace` is not specified.

image

**Possible use cases**

allenlsy:

**Description**:

Add test for init-osm-controller internal methods

Resolve #3404 

**Affected area**:
| Tests                      | ✅ |

No

1. Is this a breaking change?
No

### DIFF
--- a/cmd/init-osm-controller/init-osm-controller_test.go
+++ b/cmd/init-osm-controller/init-osm-controller_test.go
@@ -59,3 +59,45 @@ func TestCreateDefaultMeshConfig(t *testing.T) {
 	assert.Equal(meshConfig.Spec.Observability.EnableDebugServer, false)
 	assert.Equal(meshConfig.Spec.Certificate.ServiceCertValidityDuration, "24h")
 }
+
+func TestValidateCLIParams(t *testing.T) {
+	assert := tassert.New(t)
+
+	// save original global values
+	prevOsmNamespace := osmNamespace
+
+	tests := []struct {
+		caseName string
+		setup    func()
+		verify   func(error)
+	}{
+		{
+			caseName: "osm-namespace is empty",
+			setup: func() {
+				osmNamespace = ""
+			},
+			verify: func(err error) {
+				assert.NotNil(err)
+				assert.Contains(err.Error(), "--osm-namespace")
+			},
+		},
+		{
+			caseName: "osm-namespace is valid",
+			setup: func() {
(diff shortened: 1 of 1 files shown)


cmd/init-osm-controller/init-osm-controller_test.go
```diff
@@ -59,3 +59,11 @@ func TestCreateDefaultMeshConfig(t *testing.T) {
 	assert.Equal(meshConfig.Spec.Observability.EnableDebugServer, false)
 	assert.Equal(meshConfig.Spec.Certificate.ServiceCertValidityDuration, "24h")
 }
+
+func TestValidateCLIParams(t *testing.T) {
+	assert := tassert.New(t)
+
+	err := validateCLIParams()
```
nojnhuh, line 66: This seems fine as long as other tests in this package are careful not to mess too much with the global state this test is reading.

In general though, I think tests should be setting the global state they need, even if that state is the default. Otherwise it's not at all clear what the test is verifying. And if it's manipulating global state, then it should save the old state and reset it after the test so it doesn't affect other tests. That of course is a lot to consider, so refactoring this in the future to eliminate global state where we can may be worthwhile.

allenlsy: Issues created: https://github.com/openservicemesh/osm/issues/3547

shalier (commented):
Would it be possible to also add a check for the `return nil` in L55?
https://github.com/openservicemesh/osm/blob/main/cmd/init-osm-controller/init-osm-controller.go#L51-L56

cmd/init-osm-controller/init-osm-controller_test.go
```diff
@@ -59,3 +59,43 @@ func TestCreateDefaultMeshConfig(t *testing.T) {
...
+		},
+	}
+
+	for _, tc := range tests {
+		tc.setup()
+		err := validateCLIParams()
+		tc.verify(err)
+	}
+
+	osmNamespace = prevOsmNamespace
```
shashankram, line 100: nit: add a comment regarding this preventing test pollution.

shashankram (approved):

PR #3488, opened 2021-06-01
PodMetadata: Use kubecontroller to gather PodMetadata

eduser25:
Moves using Proxy UUID, and instead uses available kubecontroller
interface to gather information about the pod for a given proxy.

This is needed to be able to reuse the Node ID for snapshot cache.

No
1. Is this a breaking change?
No


pkg/envoy/ads/stream.go
```diff
@@ -293,3 +299,50 @@ func isCNforProxy(proxy *envoy.Proxy, cn certificate.CommonName) bool {
 	identityForCN := identity.K8sServiceAccount{Name: chunks[0], Namespace: chunks[1]}
 	return identityForCN == proxyIdentity
 }
+
+// recordPodMetadata records pod metadata and verifies the certificate issued for this pod
+// is for the same service account as seen on the pod's service account
+func (s *Server) recordPodMetadata(p *envoy.Proxy) error {
```
allenlsy, line 305: Suggest split recording and verifying into two methods

allenlsy (approved):

pkg/envoy/ads/stream.go
eduser25: dealing with this on a commit on top

PR #3499, opened 2021-06-03
feat(*): Add hpa to osm controller and injector

shalier:
**Description**:
Adds horizontal pod autoscaler for osm-controller and osm-injector

Refer to the [HA doc](https://docs.google.com/document/d/1gpNbsghOTmKx0rVh4TL1kzpH6UnlWL3Hkw7lce7WlYM/edit#heading=h.ap5u1c8i7rvb) for more reference.
Part of #3390

**Affected area**:
| New Functionality          | [ X ] |

1. Is this a breaking change?
no

### DIFF
--- a/charts/osm/values.schema.json
+++ b/charts/osm/values.schema.json
@@ -67,6 +67,59 @@
                     "additionalProperties": false
                 }
             }
+        },
+        "autoScale": {
+            "$id": "#/properties/definitions/properties/autoScale",
+            "type": "object",
+            "title": "The autoScale schema",
+            "description": "Autoscale configuration parameters",
+            "required": [
+                "enable"
+            ],
+            "properties": {
+                "enable": {
+                    "$id": "#/properties/definitions/properties/autoScale/properties/enable",
+                    "type": "boolean",
+                    "title": "Autoscale enable",
+                    "description": "Indicates whether autoscale should be enabled or not.",
+                    "examples": [
+                        false
+                    ]
+                },
+                "minReplicas": {
+                    "$id": "#/properties/definitions/properties/autoScale/properties/minReplicas",
+                    "type": "integer",
+                    "title": "Autoscale minimum replicas",
+                    "description": "Indicates the minimum replicas for autoscale.",
+                    "minimum": 1,
+                    "maximum": 10,
+                    "examples": [
+                        1
+                    ]
+                },
+                "maxReplicas": {
+                    "$id": "#/properties/definitions/properties/autoScale/properties/maxReplicas",
+                    "type": "integer",
+                    "title": "Autoscale maximum replicase",
+                    "description": "Indicates the maximum replicas for autoscale.",
+                    "minimum": 1,
+                    "maximum": 10,
+                    "examples": [
+                        5
+                    ]
+                },
+                "targetAverageUtilization": {
+                    "$id": "#/properties/definitions/properties/autoScale/properties/targetAverageUtilization",
+                    "type": "integer",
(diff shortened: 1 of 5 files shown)


charts/osm/values.schema.json
```diff
@@ -606,6 +659,59 @@
...
+                                        false
+                                    ]
+                                },
+                                "minReplicas": {
+                                    "$id": "#/properties/OpenServiceMesh/properties/osmController/properties/autoScale/properties/minReplicas",
+                                    "type": "integer",
+                                    "title": "Autoscale minimum replicas",
+                                    "description": "Indicates the minimum replicas for autoscale.",
+                                    "minimum": 1,
+                                    "maximum": 1000,
```
shashankram, line 687:
```suggestion
                                    "maximum": 10,
```

charts/osm/values.schema.json
```diff
@@ -606,6 +659,59 @@
...
+                                        1
+                                    ]
+                                },
+                                "maxReplicas": {
+                                    "$id": "#/properties/OpenServiceMesh/properties/osmController/properties/autoScale/properties/maxReplicas",
+                                    "type": "integer",
+                                    "title": "Autoscale maximum replicase",
+                                    "description": "Indicates the maximum replicas for autoscale.",
+                                    "minimum": 1,
+                                    "maximum": 1000,
```
shashankram, line 698:
```suggestion
                                    "maximum": 10,
```

charts/osm/values.yaml
```diff
@@ -36,6 +36,12 @@ OpenServiceMesh:
     podLabels: {}
     # -- Enable Pod Disruption Budget
     enablePodDisruptionBudget: false
+    # -- Auto scale configuration
+    autoScale:
+      enable: false
+      minReplicas: 1
+      maxReplicas: 5
+      targetAverageUtilization: 80
```
shashankram, line 44: Need comments for each of the fields similar to other objects defined here.

shashankram (approved):
Looks great

charts/osm/values.schema.json
```diff
@@ -67,6 +67,59 @@
...
+                        false
+                    ]
+                },
+                "minReplicas": {
+                    "$id": "#/properties/definitions/properties/autoScale/properties/minReplicas",
+                    "type": "integer",
+                    "title": "Autoscale minimum replicas",
+                    "description": "Indicates the minimum replicas for autoscale.",
+                    "minimum": 1,
+                    "maximum": 10,
```
allenlsy, line 95: What if user requires more than 10 mimReplicas? I intention here is trying to avoid come back and update here in the future

shashankram: @allenlsy Does JSON schema allow using other values to define minimum/maximum? If not, such arithmetic conditions can't be enforced for 100% correctness when a combination of values are involved. If user specifies min > max, that's a user error.

allenlsy: No, I don't find JSON schema allows that. My actual concern is, what if 10 replicas is not enough, and should we remove the `maximum` constraint.

shashankram: I get your point now. I don't think we need beyond 10 replicas, if we do we can update it then. There is an argument to be made to also remove the max, although it opens the door for expoits where someone can potentially overwhelm the cluster with a large number of replicas and bring it down if there is a way to constantly increase CPU load (potentially through an API endpoint that results in it).

allenlsy: Ok cool. Let's use 10 as the max for now

shalier:
@Mergifyio refresh

PR #3505, opened 2021-06-03
ref(prometheusScraping): Remove prometheus scraping flag

snehachhabria:
**Description**:

This commit removes the prometheus_scraping flag and adds the envoy
prometheus scraping configs when a pod has the required prometheus annotation.

Part of #3065

**Affected area**:

1. Is this a breaking change? `no`. However this PR removes a flag from install and  the meshConfig

### DIFF
--- a/pkg/catalog/catalog.go
+++ b/pkg/catalog/catalog.go
@@ -36,3 +36,8 @@ func NewMeshCatalog(kubeController k8s.Controller, kubeClient kubernetes.Interfa
 
 	return &mc
 }
+
+// GetKubeController returns the kube controller instance handling the current cluster
+func (mc *MeshCatalog) GetKubeController() k8s.Controller {
+	return mc.kubeController
+}
--- a/pkg/envoy/cds/response.go
+++ b/pkg/envoy/cds/response.go
@@ -75,7 +75,9 @@ func NewResponse(meshCatalog catalog.MeshCataloger, proxy *envoy.Proxy, _ *xds_d
 	}
 
 	// Add an inbound prometheus cluster (from Prometheus to localhost)
-	if cfg.IsPrometheusScrapingEnabled() {
+	if pod, err := envoy.GetPodFromCertificate(proxy.GetCertificateCommonName(), meshCatalog.GetKubeController()); err != nil {
+		log.Warn().Msgf("Could not find pod for connecting proxy %s. No metadata was recorded.", proxy.GetCertificateSerialNumber())
+	} else if meshCatalog.GetKubeController().IsMetricsEnabled(pod) {
 		clusters = append(clusters, getPrometheusCluster())
 	}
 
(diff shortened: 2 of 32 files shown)


pkg/catalog/catalog.go
```diff
@@ -36,3 +36,8 @@ func NewMeshCatalog(kubeController k8s.Controller, kubeClient kubernetes.Interfa
 
 	return &mc
 }
+
+// GetKubecontroller returns the kubecontroller instance handling the current cluster
+func (mc *MeshCatalog) GetKubecontroller() k8s.Controller {
```
allenlsy, line 41: nit: `GetKubeController()`

allenlsy (approved):

pkg/envoy/xdsutil.go
```diff
@@ -396,3 +398,21 @@ func GetServiceAccountFromProxyCertificate(cn certificate.CommonName) (identity.
 
 	return svcAccount, nil
 }
+
+// IsMetricsEnabled returns true if the pod in the mesh is correctly annotated for prometheus scrapping
+func IsMetricsEnabled(proxy *Proxy, meshCatalog catalog.MeshCataloger) bool {
+	isScrapingEnabled := false
+	pod, err := GetPodFromCertificate(proxy.GetCertificateCommonName(), meshCatalog.GetKubecontroller())
```
shashankram, line 405: Could you instead implement this function in the `kubernetes.Controller` interface given this is very specific to k8s and now leaks k8s stuff even further into pkg/envoy?

Code in pkg/envoy can use:
`meshCatalog.kubeController.IsMetricsEnabled(pod)`

shashankram (commented):
The change looks good in itself, however I suggest moving the implementation of `IsMetricsEnabled()` to `pkg/kubernetes` instead.

pkg/envoy/cds/response.go
```diff
@@ -75,7 +75,7 @@ func NewResponse(meshCatalog catalog.MeshCataloger, proxy *envoy.Proxy, _ *xds_d
 	}
 
 	// Add an inbound prometheus cluster (from Prometheus to localhost)
-	if cfg.IsPrometheusScrapingEnabled() {
+	if envoy.IsMetricsEnabled(proxy, meshCatalog) {
```
eduser25, line 78: neat

eduser25 (approved):

pkg/envoy/cds/response.go
eduser25: neat the fact that it now depends on pod basis*, not the api call itself

pkg/envoy/xdsutil.go
snehachhabria: Thanks for the suggestion, I moved the code into `pkg/kubernetes`

PR #3510, opened 2021-06-04
Bump cert manager

irbekrm:
**Description**:
This PR bumps the API version of  cert-manager (Jetstack cert-manager) from `v1beta1` and `v1alpha2` to `v1` everywhere. The beta and alpha APIs are getting deprecated in cert-manager `v1.4` and will be removed in `v1.6` (see [cert-manager#4021](https://github.com/jetstack/cert-manager/pull/4021)). It also bumps the version of cert-manager library and cert-manager installed in tests and demo to the latest stable version `v1.3.1`.
I tested the change by running `make go-test`.

**Affected area**:
| Tests                      | [ x] |
| Demo                       | [x ] |

1. Is this a breaking change? It could break folks who would currently have `cert-manager` older than `v1` installed (as the `v1` API only becomes available in `v1`). Perhaps I should also PR somewhere a minimum required version of `cert-manager` - I guess docs could be a suitable place?


charts/osm/values.yaml
```diff
@@ -78,7 +78,7 @@ OpenServiceMesh:
     # -- cert-manager issuer kind
     issuerKind: Issuer
     # -- cert-manager issuer group
-    issuerGroup: cert-manager
+    issuerGroup: cert-manager.io
```
shashankram, line 81: @irbekrm could you run `make chart-readme` and commit the diff to this PR. You also need to squash your commits into a single commit for this PR to be able to merge.

shashankram (commented):
Thanks @irbekrm. Just 1 comment, looks good to merge otherwise.

charts/osm/values.yaml
irbekrm: Thank you! I've done that.

shashankram (approved):
Thanks @irbekrm!

nojnhuh (approved):

Issue #3501, opened 2021-06-03
Integrate Contour's Ingress gateway


shashankram:
**Please describe the Improvement and/or Feature Request**
Integrate [Contour's Ingress gateway](https://projectcontour.io/) with OSM:
 
Tasks:
- [X] Design
- [X] POC
- [x] Implementation
- [x] Documentation

**Scope (please mark with X where applicable)**
- New Functionality          [X]

PR #3514, opened 2021-06-04
envoy: add additional metadata for the proxy's kind

shashankram:
**Description**:

Adds additional metadata in the proxy's XDS cert
to identify the proxy's kind.

Currently, osm-controller only supports programming proxies
that run as a sidecar on app pods. As a part of the effort
to implement a proxy gateway, osm-controller will also
require to identify and program a proxy gateway. The
config for a gateway proxy will be completely different
compared to that of a sidecar proxy, and this change will
allow identifying the kind of the proxy to do so.

Existing code and tests have been updated to set the
proxy's kind to `sidecar`.

Part of #3501

**Affected area**:

### DIFF
--- a/pkg/envoy/types.go
+++ b/pkg/envoy/types.go
@@ -83,3 +83,14 @@ const (
 	// The local cluster refers to the cluster corresponding to the service the proxy is fronting, accessible over localhost by the proxy.
 	localClusterSuffix = "-local"
 )
+
+// ProxyKind is the type used to define the proxy's kind
+type ProxyKind string
+
+const (
+	// KindSidecar implies the proxy is a sidecar
+	KindSidecar ProxyKind = "sidecar"
+
+	// KindGateway implies the proxy is a gateway
+	KindGateway ProxyKind = "gateway"
+)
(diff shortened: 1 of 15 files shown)


eduser25 (approved):
for a moment i thought this was podmetadata, was about to table flip

pkg/envoy/types.go
```diff
@@ -83,3 +83,14 @@ const (
 	// The local cluster refers to the cluster corresponding to the service the proxy is fronting, accessible over localhost by the proxy.
 	localClusterSuffix = "-local"
 )
+
+// ProxyKind is the type used to define the proxy's kind
+type ProxyKind string
+
+const (
+	// ProxyKindSidecar implies the proxy is a sidecar
+	ProxyKindSidecar ProxyKind = "sidecar"
```
snehachhabria, line 92:
```suggestion
	Sidecar ProxyKind = "sidecar"
```

Since the type is already ProxyKind

snehachhabria (approved):

pkg/envoy/types.go
shashankram: Maybe `KindSidecar` works best!

Issue #3408, opened 2021-05-25
Tests: add tests for pkg/envoy/ads


allenlsy:
**Please describe the Improvement and/or Feature Request**

Add test cases for pkg/envoy/ads.

Coverage required:

* `Server.sendResponse()`
* `Server.newAggregatedDiscoveryResponse()`
* `Server.StreamAggregatedResources()`
* `respondToRequest()`: function is not called in test suite.
* `shouldPushUpdate()`
* `parseRequestVersion()`
* `proxyResponseJob.JobName()`

**Scope (please mark with X where applicable)**
- New Functionality          [ ]
- Install                    [ ]
- SMI Traffic Access Policy  [ ]
- SMI Traffic Specs Policy   [ ]
- SMI Traffic Split Policy   [ ]
- Permissive Traffic Policy  [ ]
- Ingress                    [ ]
- Egress                     [ ]
- Envoy Control Plane        [ ]
- CLI Tool                   [ ]
- Metrics                    [ ]
- Certificate Management     [ ]
- Sidecar Injection          [ ]
- Logging                    [ ]
- Debugging                  [ ]
- Tests                      [x]
- CI System                  [ ]
- Demo                       [ ]
- Project Release            [ ]

**Possible use cases**

allenlsy:
Ref #1489

allenlsy:
Reference to my comment here: https://github.com/openservicemesh/osm/pull/3515#issuecomment-875058594

PR #3515, opened 2021-06-04
[HOLD] test: add tests to pkg/envoy/ads

allenlsy:

**Description**:

Add tests to pkg/envoy/ads.

Partially resolve  issue: #3408 

Coverage mainly increased for:

* `StreamAggregatedResources()`: cases not handled in ADS server event loop: https://github.com/openservicemesh/osm/blob/3e62fd03d34c1a816e414131780fa364d9916af6/pkg/envoy/ads/stream.go#L98
* `recordEnvoyPodMetadata()`

Screen Shot 2021-06-07 at 5 37 31 PM

Screen Shot 2021-06-07 at 5 36 57 PM

**Affected area**:
| Tests                      | ✅ |

No

1. Is this a breaking change?

No

### DIFF
--- a/pkg/envoy/ads/stream_test.go
+++ b/pkg/envoy/ads/stream_test.go
@@ -1,15 +1,45 @@
 package ads
 
 import (
+	"context"
+	"crypto/x509"
 	"fmt"
 	"testing"
+	"time"
 
+	"github.com/golang/mock/gomock"
 	"github.com/google/uuid"
 	tassert "github.com/stretchr/testify/assert"
+	"google.golang.org/grpc/peer"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	testclient "k8s.io/client-go/kubernetes/fake"
 
+	discovery "github.com/envoyproxy/go-control-plane/envoy/service/discovery/v3"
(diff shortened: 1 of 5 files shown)


pkg/envoy/ads/stream_test.go
```diff
@@ -51,3 +80,156 @@ func TestIsCNForProxy(t *testing.T) {
...
+
+	tests := []struct {
+		caseName               string
+		triggerEvent           func() bool
+		expectedEventTriggered bool
+	}{
+		{
+			caseName: "trigger ProxyBroadcast event",
+			triggerEvent: func() bool {
+				proxyBroadcastChannel <- true
```
eduser25, line 191:
```
proxyBroadcastChannel <- true
...
case <-proxyBroadcastChannel:
```

is... this correct? or am I imagining things

allenlsy:
Testing on ADS server listening to broadcast channel messages is not possible at this moment.

In this PR, test case can send message to the broadcast channel, and ADS stream server can receive the message. In order to verify the test, we plan to check if a new response job is created. This requires us to force update the last nonce of the proxy in the test. In the current test setup, there is no way to get the original proxy object. `GetProxyFromPod()` method returns a partial duplicate of the proxy, while it's not the original one. Thus we cannot complete the test verification.

Closing the PR for now.

Issue #3432, opened 2021-05-25
Tests: add test to pkg/injector


allenlsy:
In `pkg/injector`, some areas need improved test coverage:

a) Add tests to pkg/injector `mutatingWebhook.createPatch`. Test should check if pod annotations are updated when `enableMetrics=true`.

![image](https://user-images.githubusercontent.com/872876/119568093-f6801a80-bd61-11eb-9202-eaab0ed0d822.png)

b) Add tests to pkg/injector `getPort()`. 

Scenario not covered:

* Container does not declare any port
* No matching port to the given named port

![image](https://user-images.githubusercontent.com/872876/119567506-43172600-bd61-11eb-8e40-fc949c743b81.png)

Ref #1489

SanyaKochhar:
@shuheiktgw I see you have two PRs open for this issue, do you mind if I assign it to you?

PR #3519, opened 2021-06-05
tests: add unit tests for pkg/injector/patch

shuheiktgw:
**Description**:

Relates to https://github.com/openservicemesh/osm/issues/3432
I've added missing unit tests for pkg/injector/patch. Thank you for your review!

**Affected area**:

No

1. Is this a breaking change?

No

### DIFF
--- a/pkg/injector/patch_test.go
+++ b/pkg/injector/patch_test.go
@@ -2,11 +2,9 @@ package injector
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"testing"
-
-	. "github.com/onsi/ginkgo"
-	. "github.com/onsi/gomega"
 
 	mapset "github.com/deckarep/golang-set"
 	"github.com/golang/mock/gomock"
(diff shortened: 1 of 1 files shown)


pkg/injector/patch_test.go
```diff
@@ -65,50 +147,45 @@ var _ = Describe("Test all patch operations", func() {
...
-				`]`
+				for _, expectedPatch := range expectedPatches {
+					expectedPatchPath := expectedPatch["path"].(string)
+					if path == expectedPatchPath {
+						assert.Equal(expectedPatch, patch)
+						continue Outer
+					}
+				}
 
-			Expect(string(jsonPatches)).ToNot(Equal(expectedJSONPatches),
```
shuheiktgw, line 107: The original assertion was asserting `not to equal` to the expected patches so it actually did nothing...

pkg/injector/patch_test.go
```diff
@@ -2,50 +2,134 @@ package injector
...
 	"fmt"
 	"testing"
 
-	. "github.com/onsi/ginkgo"
-	. "github.com/onsi/gomega"
-
 	mapset "github.com/deckarep/golang-set"
 	"github.com/golang/mock/gomock"
 	"github.com/google/uuid"
+	. "github.com/onsi/ginkgo"
```
allenlsy, line 12: Is this import still necessary? I don't see we using describe any more here.

shuheiktgw: It's actually used here[ ](https://github.com/openservicemesh/osm/pull/3519/files#diff-3a53dad499d16baa10ff7e0fc76ceb1a7805ca8abcc83e308fef48c5ece1f45aR127) but I can remove the dependency 👍

shashankram (commented):
@shuheiktgw Thanks for the fix. Would you mind squashing your commits into a single commit as described in https://github.com/openservicemesh/osm/blob/main/CONTRIBUTING.md#4-commit

pkg/injector/patch_test.go
```diff
@@ -2,50 +2,133 @@ package injector
...
+	}{
+		{
+			name: "creates a patch",
+			namespace: &corev1.Namespace{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: namespace,
+				},
+			},
+			rawExpectedPatches: fmt.Sprintf(`[`+
+				// Add Envoy UID Label
```
allenlsy, line 49: I feel using this extremely long string to verification is over-killing, and easy to break by other changes (e.g. what if we change the ENV feed to envoy container?).

I suggest, doing substring search in the `createPatch()` response (`rawPatch` in this case) should be sufficient.

shashankram (approved):

pkg/injector/patch_test.go
```diff
@@ -65,50 +146,44 @@ var _ = Describe("Test all patch operations", func() {
...
+			// The patch order may change so assertions should be independent from it
+			for i, patch := range patches {
+				path := patch["path"].(string)
 
-				`]`
+				for _, expectedPatch := range expectedPatches {
+					expectedPatchPath := expectedPatch["path"].(string)
+					if path == expectedPatchPath {
+						assert.Equal(expectedPatch, patch)
+						continue Outer
```
allenlsy, line 178: Just curious, why not just use `break` keyword? Or is this a preference?

pkg/injector/patch_test.go
allenlsy: Looks good now.

shuheiktgw:
@shashankram That makes sense. Thank you for your clarification!

pkg/injector/patch_test.go
allenlsy: Yes, I actually mean why we use `continue Outer` instead of `break`. Although line label is supported, I personally don't find it widely used.

allenlsy (approved):

shuheiktgw:
Thank you for your review. I'll squash the commits now 👍

PR #3534, opened 2021-06-07
utils/proto: consolidate proto helpers

shashankram:
**Description**:
Consolidates helpers into a separate file in the
utils pkg for reusability. The proto marshalling
to YAML will be required in a subsequent change
that builds the envoy bootstrap config in a generic
way for code reuse between the injector and OSM
gateway.

**Affected area**:

### DIFF
--- a/pkg/utils/proto.go
+++ b/pkg/utils/proto.go
@@ -1,0 +1,44 @@
+package utils
+
+import (
+	"google.golang.org/protobuf/encoding/protojson"
+	"google.golang.org/protobuf/reflect/protoreflect"
+	yaml "gopkg.in/yaml.v2"
+)
+
+// ProtoToYAML converts a Proto message to it's YAML representation in bytes
+func ProtoToYAML(m protoreflect.ProtoMessage) ([]byte, error) {
+	marshalOptions := protojson.MarshalOptions{
+		UseProtoNames: true,
+	}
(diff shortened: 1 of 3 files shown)


snehachhabria (approved):

pkg/utils/proto.go
```diff
@@ -0,0 +1,44 @@
+package utils
+
+import (
+	"google.golang.org/protobuf/encoding/protojson"
+	"google.golang.org/protobuf/reflect/protoreflect"
+	yaml "gopkg.in/yaml.v2"
+)
+
+// ProtoToYAML converts a Proto message to it's YAML representation in bytes
+func ProtoToYAML(m protoreflect.ProtoMessage) ([]byte, error) {
```
allenlsy, line 10: Can we add tests for proto.go?

nojnhuh (approved):

pkg/utils/proto.go
shashankram: It gets implicitly tested in `envoy_config_test.go`, though I agree there should be tests for this file. I consider it outside the scope of this change because this change only moves the function to a different package, which had no test for these functions.

shashankram: Added a test for the function in https://github.com/openservicemesh/osm/pull/3535.

PR #3537, opened 2021-06-07
Feature/provider

whitneygriffith:
**Description**:

A provider should also know how to handle services, and therefore the providers should be in its own package

**Affected area**:
| Documentation              | [x ] |

No
1. Is this a breaking change?
No


DESIGN.md
```diff
@@ -87,8 +87,8 @@ The Proxy Control Plane's availability is of foremost importance when it comes t
 Certificate Manager is a component that provides each service participating in the service mesh with a TLS certificate.
 These service certificates are used to establish and encrypt connections between services using mTLS.
 
-### (3) Endpoints Providers
-Endpoints Providers are one or more components that communicate with the compute platforms (Kubernetes clusters, on-prem machines, or cloud-providers' VMs) participating in the service mesh. Endpoints providers resolve service names into lists of IP addresses. The Endpoints Providers understand the specific primitives of the compute provider they are implemented for, such as virtual machines, virtual machine scale sets, and Kubernetes clusters.
+### (3) Providers
```
shashankram, line 90: Hey @whitneygriffith, I am unsure what the intent of this change is. `Endpoint Providers` is a component within OSM that provides information regarding service endpoints. Why is this being changed?

shashankram (changes requested):

DESIGN.md
steeling: Hey @shashankram, I can answer that! For the multicluster work, we are planning on folding all of the service discovery mechanisms from the MeshCatalog into the provider. This has the added benefit of reducing the MeshCatalog's reliance on running on k8s. By implementing this change we can remove the KubeController from the MeshCatalog all together.

This means we will be moving the following functions to the provider:
```go
ListMonitoredNamespaces() ([]string, error)
ListServices() []service.MeshService
ListServiceIdentitiesForService(svc service.MeshService) ([]identity.ServiceIdentity, error)
GetPortToProtocolMappingForService(svc service.MeshService) (map[uint32]string, error)
GetHostnames(svc service.MeshService) ([]string, error)
```

We figured we'd break up the changes and simply move the provider to a new dir first, without changing or adding any functionality.

shashankram: Thanks @steeling for the info. It seems like what you propose above is a package to perform service discovery, which makes sense. However, this change seems to be altering the endpoints provider package, which does not seem to be related to service discovery.

Would you mind drafting a brief design doc outlining the rationale for this change so this can be reviewed, given this is a fundamental change that has a large impact area?

steeling: https://docs.google.com/document/d/1sA03IwqC6tuJpCleGMakQpHzwFJ5ulfxzCq7iSXSqZc/edit#

Issue #3403, opened 2021-05-25
Create osm multicluster CRD


snehachhabria:
**Please describe the Improvement and/or Feature Request**

The design doc :  [RFC: Multicluster osm](https://docs.google.com/document/d/1O3XcG27uryBxCHX719qYUUzAo913GoFv3LOwdnfNudQ/edit#heading=h.frj1aeltd2ge), specifies the need of a crd to support multicluster osm. 

This issue is to track the following : 
- [ ] Propose/Design a multicluster CRD (yaml only) based on the multicluster design doc 
- [ ] Create code gen helper for generation of Go Client code for the new multicluster CRD 
- [ ] Generate Go Client code for the new multicluster CRD 
- [ ] Create informers for this CRD
- [ ] creation of the CRD, deploying it via helm and setting up of the informers should be triggered only when the multicluster feature is enabled in osm #3402 

**Scope (please mark with X where applicable)**
- New Functionality          [ ]
- Install                    [ ]
- SMI Traffic Access Policy  [ ]
- SMI Traffic Specs Policy   [ ]
- SMI Traffic Split Policy   [ ]
- Permissive Traffic Policy  [ ]
- Ingress                    [ ]
- Egress                     [ ]
- Envoy Control Plane        [ ]
- CLI Tool                   [ ]
- Metrics                    [ ]
- Certificate Management     [ ]
- Sidecar Injection          [ ]
- Logging                    [ ]
- Debugging                  [ ]
- Tests                      [ ]
- CI System                  [ ]
- Demo                       [ ]
- Project Release            [ ]

**Possible use cases**

steeling:
From https://docs.google.com/document/d/1sA03IwqC6tuJpCleGMakQpHzwFJ5ulfxzCq7iSXSqZc/edit#, we plan on augmenting the previous OSM CRD to break up the config such that there is 1 config per service per remote cluster. The new CRD will look like:

```
apiVersion: v1alpha1
kind: RemoteService
metadata:
     	name: store-clusterX
     	namespace: osm-system 
spec:
      cluster:
     	     	name: clusterX
     	     	gateway: 10.10.10.10:90         # a routable IP + port
     	     	certificate: ==               # mTLS certificates (optional)
      equivalence:     # allows multiple equivalence schemes in the future
       	serviceName: store
      	serviceNamespace: store
      	serviceAccount: store
```

PR #3540, opened 2021-06-08
feat(multicluster): crd only

timmyreilly:
Signed-off-by: timmyreilly <tireilly@microsoft.com>

Adding crd for MulticlusterService

Part of : https://github.com/openservicemesh/osm/issues/3403

**Affected area**:

No
1. Is this a breaking change?
No

### DIFF
--- a/charts/osm/crds/multiclusterservice.yaml
+++ b/charts/osm/crds/multiclusterservice.yaml
@@ -1,0 +1,63 @@
+# Custom Resource Definition (CRD) for OSM's multicluster config specification.
+#
+# Copyright Open Service Mesh authors.
+#
+#    Licensed under the Apache License, Version 2.0 (the "License");
+#    you may not use this file except in compliance with the License.
+#    You may obtain a copy of the License at
+#
+#        http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS,
+#    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#    See the License for the specific language governing permissions and
+#    limitations under the License.
+---
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  name: multiclusterservices.config.openservicemesh.io
+spec:
+  group: config.openservicemesh.io
+  scope: Namespaced # osm-system is the required namespace
+  names:
+    kind: MultiClusterService
+    shortNames:
+      - mcs
+    plural: multiclusterservices
+    singular: multiclusterservice
+  versions:
+    - name: v1alpha1
+      served: true
+      storage: true
+      schema:
+        openAPIV3Schema:
+          type: object
+          properties:
+            spec:
+              type: object
+              required:
+                - serviceAccount
+                - clusters
+              properties:
+                serviceAccount:
+                  type: string
+                clusters:
+                  description: The clusters the service accounts are hosted on.
+                  type: array
+                  items:
+                    type: object
+                    required:
+                      - address
+                      - name
+                    properties:
+                      address:
+                        description: a routable IP + port
+                        type: string
+                      name:
+                        description: Name of the remote cluster
+                        type: string
+                      certificate:
+                        description: mTLS certificates (optional)
+                        type: string


charts/osm/crds/multiclusterservice.yaml
```diff
@@ -0,0 +1,47 @@
+apiVersion: apiextensions.k8s.io/v1
```
snehachhabria, line 1: Could we add comments similar to this : https://github.com/openservicemesh/osm/blob/b223d9ed3fb50ce6fcc09842d1bba5fba6238482/charts/osm/crds/policy.yaml#L1

snehachhabria (approved):

allenlsy (approved):

charts/osm/crds/multiclusterservice.yaml
```diff
@@ -0,0 +1,63 @@
...
+---
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  name: multiclusterservices.config.openservicemesh.io
+spec:
+  group: config.openservicemesh.io
+  scope: Namespaced # osm-system is the required namespace
+  names:
+    kind: MultiClusterService
```
snehachhabria, line 25: I am not sure if the right name for this crd would be `MultiClusterService`, since there is no mention of a service anywhere in the crd, we might have to come up with a different name that is a better representation what the crd entails

snehachhabria (changes requested):
Lets come up with a better name for the crd

charts/osm/crds/multiclusterservice.yaml
timmyreilly: @steeling maybe something like `MultiClusterConfiguration`?

steeling: I think MultiClusterService is fitting. It really is closer to a service.

FWIW `Service` has no mention of service either, it's just in the name, and generates a service in K8s. Similarly, the MultiClusterService generates (multiple) services, but in OSM (service.MeshService).

From a user perspective, a MultiClusterService of name `bookstore` with namespace `bookstore`, and clusters `cluster-x` and `cluster-y`, will generate 3 services:

`bookstore.bookstore.svc.cluster.global`
`bookstore.bookstore.svc.cluster.cluster-x`
`bookstore.bookstore.svc.cluster.cluster-y`

snehachhabria: This is true only if the service and service account name is the same, which might not always be the case. I think I prefer `MultiClusterConfiguration`

steeling: That's actually not how this will work out. Since this refers to multi cluster services (where it doesn't have the ability to extract the identity from the pods, or services), it infers the service account from this object.

It *does* create several service.MeshService. Moreover, a MultiClusterConfiguration sounds like a singleton for a multicluster, whereas there will be 1 of these *per service*.

snehachhabria: Post an offline discussion we decided to keep the name as is, let's add a comment to mention that the same of this resource needs to match the name of the service this resource intends to target

charts/osm/crds/multiclusterservice.yaml
```diff
@@ -0,0 +1,63 @@
...
+        openAPIV3Schema:
+          type: object
+          properties:
+            spec:
+              type: object
+              required:
+                - serviceAccount
+                - clusters
+              properties:
+                serviceAccount:
```
timmyreilly, line 44: Needs to match backing service account in other cluster

charts/osm/crds/multiclusterservice.yaml
```diff
@@ -0,0 +1,64 @@
+# Custom Resource Definition (CRD) for OSM's multicluster config specification.
+# MultiClusterService needs to match the name of the service backing the pods in each cluster
```
timmyreilly, line 2: @snehachhabria @steeling any other comments you'd like added?

snehachhabria (approved):
Looks good!

shashankram:
@timmyreilly Could you kindly squash your commits, should be okay to merge after that.

charts/osm/crds/multiclusterservice.yaml
shashankram, line 2: This comment seems out of place here. I would add this to the Go types that are used for autogenerating the client libs, from which we can autogenerate API docs:
https://github.com/openservicemesh/osm/blob/main/docs/apidocs/README.md

timmyreilly: Yes, I can add this comment to autogen api as well. @snehachhabria should I remove the comment from here?

snehachhabria: sure that works

PR #3543, opened 2021-06-08
Add support for ClusterName in MeshService

brunoterkaly:
This pull request represents the changes needed to support multicluster Kubernetes. An extra field has been added to the MeshService, called ClusterName.

**Affected area**:

No.

          Not applicable.

1. Is this a breaking change?

Yes. More work needs to be done before the merge.


pkg/catalog/outbound_traffic_policies.go
```diff
@@ -62,8 +62,9 @@ func (mc *MeshCatalog) listOutboundTrafficPoliciesForTrafficSplits(sourceNamespa
 	apexServices := mapset.NewSet()
 	for _, split := range mc.meshSpec.ListTrafficSplits() {
 		svc := service.MeshService{
-			Name:      kubernetes.GetServiceFromHostname(split.Spec.Service),
-			Namespace: split.Namespace,
+			Name:        kubernetes.GetServiceFromHostname(split.Spec.Service),
+			Namespace:   split.Namespace,
+			ClusterName: "local",
```
steeling, line 67: can we add this to the constants file, and say constants.LocalClusterID?

pkg/catalog/outbound_traffic_policies.go
```diff
@@ -75,9 +76,10 @@ func (mc *MeshCatalog) listOutboundTrafficPoliciesForTrafficSplits(sourceNamespa
 
 		var weightedClusters []service.WeightedCluster
 		for _, backend := range split.Spec.Backends {
-			ms := service.MeshService{Name: backend.Service, Namespace: split.ObjectMeta.Namespace}
+			//ms := service.MeshService{Name: backend.Service, Namespace: split.ObjectMeta.Namespace, ClusterName: "local"}
 			wc := service.WeightedCluster{
-				ClusterName: service.ClusterName(ms.String()),
```
steeling, line 80: This is different.. this is an envoy cluster, and needs to be set from the meshservice.

pkg/catalog/outbound_traffic_policies.go
```diff
@@ -75,9 +76,10 @@ func (mc *MeshCatalog) listOutboundTrafficPoliciesForTrafficSplits(sourceNamespa
 
 		var weightedClusters []service.WeightedCluster
 		for _, backend := range split.Spec.Backends {
-			ms := service.MeshService{Name: backend.Service, Namespace: split.ObjectMeta.Namespace}
+			//ms := service.MeshService{Name: backend.Service, Namespace: split.ObjectMeta.Namespace, ClusterName: "local"}
```
steeling, line 79: Only need to change this line in this file

pkg/catalog/outbound_traffic_policies.go
```diff
@@ -301,8 +303,9 @@ func (mc *MeshCatalog) ListMeshServicesForIdentity(identity identity.ServiceIden
 				if backend.Service == upstreamSvc.Name {
 					rootServiceName := kubernetes.GetServiceFromHostname(split.Spec.Service)
 					rootMeshService := service.MeshService{
-						Namespace: split.Namespace,
-						Name:      rootServiceName,
+						Namespace:   split.Namespace,
+						Name:        rootServiceName,
+						ClusterName: "local",
```
steeling, line 308: same here and everywhere for referencing a constant

pkg/service/types.go
```diff
@@ -16,6 +16,9 @@ type MeshService struct {
 
 	// The name of the service
 	Name string
+
+	// The cluster name in which the service runs
```
steeling, line 20: Add a note to the comment, that says it's "local", for the local cluster

pkg/catalog/outbound_traffic_policies.go
shashankram: This doesn't seem like the cluster name, rather the cluster's trust domain suffix. If this is a special notation for denoting local clusters without their actual name, should move this to `pkg/constants`. If this is not meant to represent the name of the cluster, the field needs to be renamed as well.

steeling:
Let's not rename those providers files in this PR.

pkg/endpoint/provider/providers/kube/client.go
```diff
@@ -122,6 +122,8 @@ func (c Client) GetServicesForServiceAccount(svcAccount identity.K8sServiceAccou
 			services.Add(service.MeshService{
 				Namespace: pod.Namespace,
 				Name:      svc.Name,
+				ClusterName: "local",
+
```
steeling, line 126: nit: remove empty line

pkg/catalog/service.go
```diff
@@ -21,6 +21,8 @@ func (mc *MeshCatalog) isTrafficSplitBackendService(svc service.MeshService) boo
 			backendService := service.MeshService{
 				Name:      backend.Service,
 				Namespace: split.ObjectMeta.Namespace,
+				ClusterName: "local",
+
```
steeling, line 25: nit: remove empty line

steeling:
I also don't think this is catching every example. If you do a full-code search for "MeshService{", you'll find every instance.

And lastly, let's add a function to the MeshService, "IsLocal() bool"

pkg/catalog/outbound_traffic_policies.go
steeling: ya that's true. Thoughts on naming in that case? Alternatively, We could optionally put the cluster domain, ie: cluster.local, and call it ClusterDomain.

shashankram: If `ClusterDomain` works with the proposed approach, that would be the best way forward, given we are using a similar concept for defining identities (search `ClusterLocalTrustDomain`)

pkg/catalog/outbound_traffic_policies.go
brunoterkaly: Fixed:
```
	apexServices := mapset.NewSet()
	for _, split := range mc.meshSpec.ListTrafficSplits() {
		svc := service.MeshService{
			Name:        kubernetes.GetServiceFromHostname(split.Spec.Service),
			Namespace:   split.Namespace,
			ClusterDomain: "cluster.local",
		}
```

pkg/catalog/outbound_traffic_policies.go
steeling: thinking about this a bit more, it would be cleaner if we can set this to just "local". There's places where we will call ms.Name(), where we want it to only have the "local".

@shashankram  is there a term we can use to refer to this? Conceptually it is the name of a cluster in an OSM MultiCluster setup. That is to say, it is not stricly tied to kubernetes. In fact a name of "cluster-x" might not be "cluster.cluster-x" in a multi cluster setup on VM's or bare metal.

Cluster is a super overloaded term... but what do we want to call a cluster in an OSM multi cluster scenario? Whatever we call that, should be the name of this field.

pkg/endpoint/provider/providers/kube/client.go
brunoterkaly: Fixed.

pkg/catalog/service.go
brunoterkaly: Fixed.

pkg/catalog/outbound_traffic_policies.go
shashankram: @steeling I don't have a strong opinion on this. My comment was merely pointing out that `Cluster == local` is not the cluster's name, and if we end up using the trust domain instead then we should name the field a such. 

I am with whatever you and @snehachhabria decide.

snehachhabria: I agree with @shashankram, calling that field the `ClusterName` seems inappropriate since it is representing the domain

pkg/service/types.go
```diff
@@ -16,6 +16,9 @@ type MeshService struct {
 
 	// The name of the service
 	Name string
+
+	// The cluster domain in which the service runs
+	ClusterDomain string
 }
```
steeling, line 23: can you create the IsLocal function?

pkg/service/types.go
```diff
@@ -16,6 +16,9 @@ type MeshService struct {
...
 
 func (ms MeshService) String() string {
```
steeling, line 24: We'll also need this function to take into consideration the cluster name.

And we'll want to create a new one that returns this value, ie: UnscopedCluster()

steeling: check out https://github.com/steeling/osm/commit/b0dede3453d857439a7e3a8739b2d455037ea184 for `UnscopedName` in every place we want to use that

brunoterkaly:
Redoing PR due to git conflicts

PR #3565, opened 2021-06-10
fix(e2e): Add default meshconfig

nshankar13:

**Description**: Add default meshconfig, and update the meshconfig with default for the noInstall option in e2e tests. 

**Affected area**:

1. Is this a breaking change?

### DIFF
--- a/tests/framework/common.go
+++ b/tests/framework/common.go
@@ -545,6 +545,25 @@ func (td *OsmTestData) LoadImagesToKind(imageNames []string) error {
 	return nil
 }
 
+func setMeshConfigToDefault(instOpts InstallOSMOpts, meshConfig *v1alpha1.MeshConfig) (defaultConfig *v1alpha1.MeshConfig) {
+	meshConfig.Spec.Traffic.EnableEgress = instOpts.EgressEnabled
+	meshConfig.Spec.Traffic.EnablePermissiveTrafficPolicyMode = instOpts.EnablePermissiveMode
+	meshConfig.Spec.Traffic.OutboundPortExclusionList = []int{}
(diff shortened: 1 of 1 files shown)


pkg/gen/client/config/clientset/versioned/typed/config/v1alpha1/meshconfig.go
```diff
@@ -60,6 +63,57 @@ func newMeshConfigs(c *ConfigV1alpha1Client, namespace string) *meshConfigs {
 	}
 }
 
+func (c *meshConfigs) GetDefaultConfig(resourceVersion string) (defaultConfig *v1alpha1.MeshConfig) {
```
snehachhabria, line 66: Let please create this in the common.go of the e2e, since it will be used only for the e2e tests and has hard coded values

snehachhabria (changes requested):

pkg/gen/client/config/clientset/versioned/typed/config/v1alpha1/meshconfig.go
shashankram: +1

This should not be updating the API.

nshankar13: @shashankram @snehachhabria instead of creating a new meshconfig object, I will just override the fields I need to in the existing meshconfig. Otherwise, I am seeing that the envoy / init images that initially point to another registry (ex MCR) are being overridden and pointing to the osm dockerhub images instead every time the default is applied. Also, we don't necessarily want to hardcode values like EnablePermissiveTrafficPolicy in the tests,

snehachhabria (approved):

PR #3567, opened 2021-06-10
feat(pkg/*): Lay some groundwork for multi cluster gateways.

steeling:
Multi cluster gateway configurations will have multiple routes
with the same service name. This PR introduces a cluster field
to the MeshService, which currently always returns true for local.
We also change TrafficPolicy names to *always* include the cluster id
and the namespace, export a few functions, and introduce a Locality
type to  determine if a service is being reached from which locality.

#3456 

**Affected area**:

No.

1. Is this a breaking change?
No -- it does change policy names, but that should have no impact.

### DIFF
--- a/pkg/catalog/outbound_traffic_policies.go
+++ b/pkg/catalog/outbound_traffic_policies.go
@@ -66,12 +66,16 @@ func (mc *MeshCatalog) listOutboundTrafficPoliciesForTrafficSplits(sourceNamespa
 			Namespace: split.Namespace,
 		}
 
-		hostnames, err := mc.getServiceHostnames(svc, svc.Namespace == sourceNamespace)
+		locality := service.LocalCluster
+		if svc.Namespace == sourceNamespace {
+			locality = service.LocalNS
+		}
+		hostnames, err := mc.GetServiceHostnames(svc, locality)
 		if err != nil {
 			log.Error().Err(err).Msgf("Error getting service hostnames for apex service %v", svc)
 			continue
 		}
-		policy := trafficpolicy.NewOutboundTrafficPolicy(buildPolicyName(svc, sourceNamespace == svc.Namespace), hostnames)
+		policy := trafficpolicy.NewOutboundTrafficPolicy(svc.FQDN(), hostnames)
 
 		var weightedClusters []service.WeightedCluster
 		for _, backend := range split.Spec.Backends {
@@ -141,14 +145,15 @@ func (mc *MeshCatalog) buildOutboundPermissiveModePolicies() []*trafficpolicy.Ou
 	}
 
 	for _, destService := range destServices {
-		hostnames, err := mc.getServiceHostnames(destService, false)
+		// TODO(steeling): shouldn't this check the source namespace.... not relevant to this PR though.
+		hostnames, err := mc.GetServiceHostnames(destService, service.LocalCluster)
 		if err != nil {
 			log.Error().Err(err).Msgf("Error getting service hostnames for service %s", destService)
 			continue
 		}
 
 		weightedCluster := getDefaultWeightedClusterForService(destService)
-		policy := trafficpolicy.NewOutboundTrafficPolicy(buildPolicyName(destService, false), hostnames)
+		policy := trafficpolicy.NewOutboundTrafficPolicy(destService.FQDN(), hostnames)
 		if err := policy.AddRoute(trafficpolicy.WildCardRouteMatch, weightedCluster); err != nil {
 			log.Error().Err(err).Msgf("Error adding route to outbound policy in permissive mode for destination %s(%s)", destService.Name, destService.Namespace)
 			continue
@@ -182,14 +187,18 @@ func (mc *MeshCatalog) buildOutboundPolicies(sourceServiceIdentity identity.Serv
 		// Do not build an outbound policy if the destination service is an apex service in a traffic target
 		// this will be handled while building policies from traffic split (with the backend services as weighted clusters)
 		if !mc.isTrafficSplitApexService(destService) {
-			hostnames, err := mc.getServiceHostnames(destService, source.Namespace == destService.Namespace)
+			locality := service.LocalCluster
+			if destService.Namespace == source.Namespace {
+				locality = service.LocalNS
+			}
+			hostnames, err := mc.GetServiceHostnames(destService, locality)
 			if err != nil {
 				log.Error().Err(err).Msgf("Error getting service hostnames for service %s", destService)
 				continue
 			}
 			weightedCluster := getDefaultWeightedClusterForService(destService)
 
-			policy := trafficpolicy.NewOutboundTrafficPolicy(buildPolicyName(destService, source.Namespace == destService.Namespace), hostnames)
+			policy := trafficpolicy.NewOutboundTrafficPolicy(destService.FQDN(), hostnames)
 			needWildCardRoute := false
 			for _, routeMatch := range routeMatches {
 				// If the traffic target has a route with host headers
@@ -292,6 +301,10 @@ func (mc *MeshCatalog) ListMeshServicesForIdentity(identity identity.ServiceIden
 	splitPolicy := mc.meshSpec.ListTrafficSplits()
 
 	for upstreamSvc := range dstServicesSet {
+		// Traffic Splits aren't yet supported for non-local services.
+		if !upstreamSvc.Local() {
+			continue
+		}
 		for _, split := range splitPolicy {
 			// Split policy must be in the same namespace as the upstream service that is a backend
 			if split.Namespace != upstreamSvc.Namespace {
--- a/pkg/kubernetes/util.go
+++ b/pkg/kubernetes/util.go
@@ -10,6 +10,7 @@ import (
 	"k8s.io/client-go/kubernetes"
 
 	"github.com/openservicemesh/osm/pkg/constants"
+	"github.com/openservicemesh/osm/pkg/service"
 )
 
 const (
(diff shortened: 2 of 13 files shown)


pkg/service/types.go
```diff
@@ -1,27 +1,59 @@
...
+
 // MeshService is the struct defining a service (Kubernetes or otherwise) within a service mesh.
 type MeshService struct {
 	// If the service resides on a Kubernetes service, this would be the Kubernetes namespace.
 	Namespace string
 
 	// The name of the service
 	Name string
+
+	Cluster string
```
snehachhabria, line 36: Wasn't it decide to have this as `ClusterDomain`

pkg/kubernetes/util.go
```diff
@@ -18,16 +19,16 @@ const (
 
 // GetHostnamesForService returns a list of hostnames over which the service can be accessed within the local cluster.
 // If 'sameNamespace' is set to true, then the shorthand hostnames service and service:port are also returned.
-func GetHostnamesForService(service *corev1.Service, sameNamespace bool) []string {
+func GetHostnamesForService(svc *corev1.Service, locality service.Locality) []string {
```
snehachhabria, line 22: Lets avoid the name change from service to svc

pkg/catalog/outbound_traffic_policies.go
```diff
@@ -292,6 +301,10 @@ func (mc *MeshCatalog) ListMeshServicesForIdentity(identity identity.ServiceIden
 	splitPolicy := mc.meshSpec.ListTrafficSplits()
 
 	for upstreamSvc := range dstServicesSet {
+		// Traffic Splits aren't yet supported for non-local services.
+		if !upstreamSvc.Local() {
```
snehachhabria, line 305: I am not sure we would need this. If the service mentioned in the traffic split isn't present on the local cluster but there are remote clusters that have the service ( we could get this from the MultiClusterService) then the traffic should be split to all those remote services in the round robin manner with the max weight being the weight specific on the traffic split.

Example: if bookstore-v1 is only on cluster x and cluster y and the weight is 50 then we would split the traffic to 25 on each of the clusters

pkg/service/types.go
```diff
@@ -1,27 +1,59 @@
...
 }
 
 func (ms MeshService) String() string {
 	return fmt.Sprintf("%s%s%s", ms.Namespace, namespaceNameSeparator, ms.Name)
 }
 
+// FullName is similar to String(), but uses a dot separator and is in a different order.
+func (ms MeshService) FullName() string {
```
shashankram, line 44:
```suggestion
func (ms MeshService) FQDN() string {
```

pkg/catalog/outbound_traffic_policies.go
```diff
@@ -182,14 +187,18 @@ func (mc *MeshCatalog) buildOutboundPolicies(sourceServiceIdentity identity.Serv
 		// Do not build an outbound policy if the destination service is an apex service in a traffic target
 		// this will be handled while building policies from traffic split (with the backend services as weighted clusters)
 		if !mc.isTrafficSplitApexService(destService) {
-			hostnames, err := mc.getServiceHostnames(destService, source.Namespace == destService.Namespace)
+			locality := service.LocalCluster
+			if destService.Namespace == source.Namespace {
+				locality = service.LocalNS
+			}
```
snehachhabria, line 193: I am not sure I understand the logic here, the source and destination can be in the same namespace but on different clusters. I don't think having locality represent the cluster and namespace is the right way forward, they need to be separate.

Example: bookbuyer and bookstore can be the the same ns but can be on two different clusters.

pkg/catalog/outbound_traffic_policies.go
```diff
@@ -66,12 +66,16 @@ func (mc *MeshCatalog) listOutboundTrafficPoliciesForTrafficSplits(sourceNamespa
 			Namespace: split.Namespace,
 		}
 
-		hostnames, err := mc.getServiceHostnames(svc, svc.Namespace == sourceNamespace)
+		locality := service.LocalCluster
+		if svc.Namespace == sourceNamespace {
+			locality = service.LocalNS
+		}
+		hostnames, err := mc.GetServiceHostnames(svc, locality)
```
snehachhabria, line 73: Same here, I don't think this logic is right. Refer to the comment below

snehachhabria (changes requested):

pkg/service/types.go
```diff
@@ -1,27 +1,59 @@
...
 )
 
+// Locality is the relative locality of a service. ie: if a service is being accessed from the same namespace or a
+// remote cluster.
+type Locality int
+
+const (
+	// LocalNS refers to the local namespace within the local cluster.
+	LocalNS Locality = iota
+	// LocalCluster refers to access within the cluster, but not within the same namespace.
```
shashankram, line 22: nit: leave line between the constants

pkg/service/types.go
```diff
@@ -1,27 +1,59 @@
...
 }
 
 func (ms MeshService) String() string {
 	return fmt.Sprintf("%s%s%s", ms.Namespace, namespaceNameSeparator, ms.Name)
 }
 
+// FullName is similar to String(), but uses a dot separator and is in a different order.
+func (ms MeshService) FullName() string {
+	if ms.Cluster == "" {
+		ms.Cluster = "local"
```
shashankram, line 46: nit: define a constant for `local` within this file.

pkg/service/types.go
```diff
@@ -1,27 +1,59 @@
...
+	}
+	return strings.Join([]string{ms.Name, ms.Namespace, ms.Cluster}, ".")
+}
+
+// Local returns whether or not this is service is in the local cluster.
+func (ms MeshService) Local() bool {
+	// TODO(steeling): if it's unset consider it local for now.
+	return ms.Cluster == "local" || ms.Cluster == ""
```
shashankram, line 54: nit: compare with a constant

pkg/catalog/outbound_traffic_policies.go
steeling: That's actually goes against the intent of the design, and is counterintuitive to users. If I reference `bookstore-v1` in namespace `bookstore`, then that gets resolved to `bookstore-v1.bookstore.svc.cluster.local`. It should never be the case that `.cluster.local` sends traffic to the non local cluster.

pkg/catalog/outbound_traffic_policies.go
steeling: We think of this as concentric circles. A namespace within a namespace is reachable without specifying the namespace subdomain. A service within the same cluster is reachable without specifying the FQDN. Outside of the cluster, the FQDN is required, *even* if they are in the same namespace

shashankram: > We think of this as concentric circles. A namespace within a namespace is reachable without specifying the namespace subdomain. A service within the same cluster is reachable without specifying the FQDN. Outside of the cluster, the FQDN is required, _even_ if they are in the same namespace

@steeling what you mention above makes sense, though clients are allowed to use full or partial FQDNs depending on where the service is located. Ex. client in ns `ns1` should be able to access service `foo` in `ns1` using `foo.ns1.svc.cluster.local`, `foo.ns1.svc.cluster`, `foo.ns1.svc`, `foo.ns1`, or just `foo`. I hope we aren't breaking this use case.

pkg/kubernetes/util.go
steeling: it ends up conflicting with the service package

pkg/catalog/outbound_traffic_policies.go
steeling: correct we are maintaining this. I think I misused the term FQDN in referring strictly full FQDN vs partial, but yes we are keeping this case.

pkg/catalog/outbound_traffic_policies.go
steeling: same comment as above

pkg/service/types.go
steeling: done!

pkg/service/types.go
steeling: done!

pkg/service/types.go
steeling: done!

pkg/service/types.go
steeling: done!

pkg/catalog/outbound_traffic_policies.go
steeling: to clarify, any partial FQDN will resolve to local, whereas to access the remote it needs the full FQDN

pkg/service/types.go
steeling: done!

pkg/catalog/outbound_traffic_policies.go
snehachhabria: makes sense, I assume that the logic to build the FQDN will change then once its identified to be a local or remote cluster ?

steeling: Yup! That will come in a PR early next week

snehachhabria: got it, thanks for clarifying

pkg/catalog/outbound_traffic_policies.go
snehachhabria: post our offline discussion we will let go ahead with your proposal

snehachhabria (approved):

shashankram (approved):

Issue #3441, opened 2021-05-25
Resolve multicluster gateway IP


snehachhabria:
**Please describe the Improvement and/or Feature Request**

Per the design doc [RFC: Multicluster osm](https://docs.google.com/document/d/1O3XcG27uryBxCHX719qYUUzAo913GoFv3LOwdnfNudQ/edit#heading=h.frj1aeltd2ge), the CRD containing the remote cluster details will contain the fqdn of the gateway. This issue is to figure out how to : 

- [ ] resolve the the fqdn to an ip address 
- [ ] ensure the ip address is reachable, as one cluster needs to send traffic to another cluster and they may or may not be on the same vent

**Scope (please mark with X where applicable)**
- New Functionality          [ ]
- Install                    [ ]
- SMI Traffic Access Policy  [ ]
- SMI Traffic Specs Policy   [ ]
- SMI Traffic Split Policy   [ ]
- Permissive Traffic Policy  [ ]
- Ingress                    [ ]
- Egress                     [ ]
- Envoy Control Plane        [ ]
- CLI Tool                   [ ]
- Metrics                    [ ]
- Certificate Management     [ ]
- Sidecar Injection          [ ]
- Logging                    [ ]
- Debugging                  [ ]
- Tests                      [ ]
- CI System                  [ ]
- Demo                       [ ]
- Project Release            [ ]

**Possible use cases**

PR #3568, opened 2021-06-10
adding guard clauses to endpoints provider for non-local condition

timmyreilly:
Signed-off-by: timmyreilly <tireilly@microsoft.com>

Handling non-local services in endpoints provider with guard clauses. 
#3456 and #3441 

**Affected area**:

No
1. Is this a breaking change?
No

### DIFF
--- a/pkg/endpoint/providers/kube/client.go
+++ b/pkg/endpoint/providers/kube/client.go
@@ -36,6 +36,11 @@ func (c Client) ListEndpointsForService(svc service.MeshService) []endpoint.Endp
 	log.Trace().Msgf("[%s] Getting Endpoints for service %s on Kubernetes", c.providerIdent, svc)
 	var endpoints []endpoint.Endpoint
 
+	if !svc.IsLocal() {
+		log.Trace().Msg("Not a local service and we can't handle that scenario yet.")
+		return nil
+	}
+
 	kubernetesEndpoints, err := c.kubeController.GetEndpoints(svc)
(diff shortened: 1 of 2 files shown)


pkg/endpoint/providers/kube/client.go
```diff
@@ -36,6 +36,11 @@ func (c Client) ListEndpointsForService(svc service.MeshService) []endpoint.Endp
 	log.Trace().Msgf("[%s] Getting Endpoints for service %s on Kubernetes", c.providerIdent, svc)
 	var endpoints []endpoint.Endpoint
 
+	if !svc.IsLocal() {
+		log.Trace().Msg("Not a local service and we can't handle that scenario yet.")
```
shashankram, line 40:
```suggestion
		log.Warn().Msgf("%s is not a local service, cannot retrieve endpoints", svc)
```

pkg/service/types.go
```diff
@@ -18,6 +18,10 @@ type MeshService struct {
 	Name string
 }
 
+func (ms MeshService) IsLocal() bool {
+	return true
```
snehachhabria, line 22: Why is defaulting to true, shouldn't there be some logic to figure it out from mesh service

timmyreilly: Placeholder until we implement TODO: Implement this. Do an exact match for local on .ClusterDomain

steeling: tim can you add a TODO comment on this?

PR #3580, opened 2021-06-11
osm-gateway: add support for gateway bootstrapping

shashankram:
**Description**:
Adds support to bootstrap osm-gateway.
This change allows osm-gateway to connect to
osm-controller over XDS gRPC while returning nil
LDS and CDS configs so that further processing
of XDS requests are not exercised.

Part of #3501

**Affected area**:

### DIFF
--- a/charts/osm/templates/osm-gateway-deployment.yaml
+++ b/charts/osm/templates/osm-gateway-deployment.yaml
@@ -1,0 +1,54 @@
+{{- if .Values.OpenServiceMesh.featureFlags.enableOSMGateway }}
+---
+kind: Deployment
+apiVersion: apps/v1
+metadata:
+  name: osm-gateway
+  namespace: {{ include "osm.namespace" . }}
+  labels:
+    app: osm-gateway
+spec:
+  selector:
+    matchLabels:
+      app: osm-gateway
+  template:
+    metadata:
+      labels:
+        app: osm-gateway
+      name: osm-gateway
+    spec:
+      initContainers:
+        - name: osm-gateway-init
+          image: curlimages/curl
+          args:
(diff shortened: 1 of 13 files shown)


charts/osm/templates/osm-gateway-deployment.yaml
```diff
@@ -0,0 +1,54 @@
...
+          - /bin/sh
+          - -c
+          - >
+            set -x;
+            while [ $(curl -sw '%{http_code}' "http://osm-controller.{{ include "osm.namespace" . }}.svc.cluster.local:9091/health/ready" -o /dev/null) -ne 200 ]; do
+              sleep 10;
+            done
+      containers:
+        - name: envoy
+          image: envoyproxy/envoy-alpine:v1.18.3
```
snehachhabria, line 33: Can we not use the image version from the vaules.yaml ?

snehachhabria (approved):

allenlsy (approved):

charts/osm/templates/osm-gateway-deployment.yaml
shashankram: We should, my bad will fix this.

charts/osm/templates/osm-service.yaml
```diff
@@ -14,5 +14,8 @@ spec:
     - name: debug-port
       port: 9092
       targetPort: 9092
+    - name: healthz
```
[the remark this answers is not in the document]
shashankram: Sort of a convention in k8s: https://stackoverflow.com/questions/43380939/where-does-the-convention-of-using-healthz-for-application-health-checks-come-f

PR #3593, opened 2021-06-14
feat(pkg/catalog):  gateway is configured for SSL passthrough.

steeling:
The gateway is configured for SSL passthrough, and discovers every service in the local cluster.

**Affected area**:

No

1. Is this a breaking change?
No.

### DIFF
--- a/pkg/catalog/outbound_traffic_policies.go
+++ b/pkg/catalog/outbound_traffic_policies.go
@@ -104,6 +104,15 @@ func (mc *MeshCatalog) listOutboundTrafficPoliciesForTrafficSplits(sourceNamespa
 // Note: ServiceIdentity must be in the format "name.namespace" [https://github.com/openservicemesh/osm/issues/3188]
 func (mc *MeshCatalog) ListAllowedOutboundServicesForIdentity(serviceIdentity identity.ServiceIdentity) []service.MeshService {
 	ident := serviceIdentity.ToK8sServiceAccount()
+	if mc.IsMultiClusterGateway(serviceIdentity) {
+		var svcs []service.MeshService
+		for _, s := range mc.listMeshServices() {
+			if s.Local() {
+				svcs = append(svcs, s)
+			}
+		}
+		return svcs
+	}
 	if mc.configurator.IsPermissiveTrafficPolicyMode() {
 		return mc.listMeshServices()
 	}
--- a/pkg/envoy/cds/cluster.go
+++ b/pkg/envoy/cds/cluster.go
@@ -25,14 +25,30 @@ const (
 	clusterConnectTimeout = 1 * time.Second
 )
 
+type options struct {
+	noPermissive bool
+	noTLS        bool
+}
+
+// Option is type of function that edits the defaults of the options struct.
+type Option func(o options)
+
+// NoTLS is an option to disable TLS.
+func NoTLS(o options) {
+	o.noTLS = true
+}
+
+// NoPermissive is an option to not generate permissive endpoints discovery for clusters.
+func NoPermissive(o options) {
+	o.noPermissive = true
+}
+
 // getUpstreamServiceCluster returns an Envoy Cluster corresponding to the given upstream service
 // Note: ServiceIdentity must be in the format "name.namespace" [https://github.com/openservicemesh/osm/issues/3188]
-func getUpstreamServiceCluster(downstreamIdentity identity.ServiceIdentity, upstreamSvc service.MeshService, cfg configurator.Configurator) (*xds_cluster.Cluster, error) {
-	clusterName := upstreamSvc.String()
-	marshalledUpstreamTLSContext, err := ptypes.MarshalAny(
-		envoy.GetUpstreamTLSContext(downstreamIdentity, upstreamSvc))
-	if err != nil {
-		return nil, err
+func getUpstreamServiceCluster(downstreamIdentity identity.ServiceIdentity, upstreamSvc service.MeshService, cfg configurator.Configurator, opts ...Option) (*xds_cluster.Cluster, error) {
+	o := options{}
+	for _, opt := range opts {
+		opt(o)
 	}
 
 	HTTP2ProtocolOptions, err := envoy.GetHTTP2ProtocolOptions()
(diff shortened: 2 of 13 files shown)


pkg/catalog/outbound_traffic_policies.go
```diff
@@ -135,6 +150,36 @@ func (mc *MeshCatalog) ListAllowedOutboundServicesForIdentity(serviceIdentity id
...
+			continue
+		}
+
+		weightedCluster := getDefaultWeightedClusterForService(destService)
+		policy := trafficpolicy.NewOutboundTrafficPolicy(destService.FQDN(), hostnames)
+		if err := policy.AddRoute(trafficpolicy.WildCardRouteMatch, weightedCluster); err != nil {
+			log.Error().Err(err).Msgf("Error adding route to outbound policy in permissive mode for destination %s(%s)", destService.Name, destService.Namespace)
+			continue
+		}
+		outPolicies = append(outPolicies, policy)
```
shashankram, line 178: Why is an RDS route required for SNI passthrough clusters?

pkg/envoy/cds/cluster.go
```diff
@@ -25,6 +25,22 @@ const (
...
+
+	return &xds_cluster.Cluster{
+		Name:                          upstreamSvc.String(),
+		ConnectTimeout:                ptypes.DurationProto(clusterConnectTimeout),
+		TypedExtensionProtocolOptions: HTTP2ProtocolOptions,
+		ClusterDiscoveryType:          &xds_cluster.Cluster_Type{Type: xds_cluster.Cluster_EDS},
+		EdsClusterConfig:              &xds_cluster.Cluster_EdsClusterConfig{EdsConfig: envoy.GetADSConfigSource()},
+		LbPolicy:                      xds_cluster.Cluster_ROUND_ROBIN,
+	}, nil
+}
```
shashankram, line 42: We have been referring to passthrough clusters as clusters that use original destination type cluster discovery (ex. with global egress).  Can you not use `getUpstreamServiceCluster()` for your use case?

pkg/envoy/cds/response.go
```diff
@@ -34,6 +29,33 @@ func NewResponse(meshCatalog catalog.MeshCataloger, proxy *envoy.Proxy, _ *xds_d
...
+			cluster, err := getPassthroughCluster(dstService)
+			if err != nil {
+				log.Error().Err(err).Msgf("Failed to construct service cluster for service %s for proxy with XDS Certificate SerialNumber=%s on Pod with UID=%s",
+					dstService.Name, proxy.GetCertificateSerialNumber(), proxy.String())
+				return nil, err
+			}
+
+			clusters = append(clusters, cluster)
+		}
+		return makeResources(proxy, clusters), nil
```
shashankram, line 56: Should leverage the `getUpstreamServiceCluster()` here if cluster discovery is happening through EDS.

shashankram (changes requested):
Hey @steeling, left a few comments that I'd like to be addressed. Would also prefer breaking up the changes as follows:
1. Introduce catalog API + tests
2. Introduce CDS integration + tests
3. Introduce LDS intergration + tests

pkg/envoy/cds/cluster.go
steeling: `getUpstreamServiceCluster`, sets the TLS context, which we don't want to do for SSL Passthrough.

Renamed to getSSLPasshthrough, but open to other suggestions.

In permissive mode, we also don't necessarily want to use the ORIGINAL_DST, since that will look for a cluster with a name where we don't have a DNS entry (ie: with the .global suffix)

pkg/envoy/cds/response.go
steeling: same commment as above

pkg/catalog/identity.go
```diff
@@ -0,0 +1,14 @@
+package catalog
```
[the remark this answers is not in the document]
steeling: done!

pkg/catalog/outbound_traffic_policies.go
steeling: oops, that was an old change, removed now.

pkg/envoy/cds/cluster.go
shashankram: Could you instead reuse this function and pass a bool for `useTLS bool`?

steeling: added variadic options, IMO the other way is more clean though.

pkg/envoy/cds/cluster.go
```diff
@@ -25,14 +25,30 @@ const (
...
+
 // getUpstreamServiceCluster returns an Envoy Cluster corresponding to the given upstream service
 // Note: ServiceIdentity must be in the format "name.namespace" [https://github.com/openservicemesh/osm/issues/3188]
-func getUpstreamServiceCluster(downstreamIdentity identity.ServiceIdentity, upstreamSvc service.MeshService, cfg configurator.Configurator) (*xds_cluster.Cluster, error) {
-	clusterName := upstreamSvc.String()
-	marshalledUpstreamTLSContext, err := ptypes.MarshalAny(
-		envoy.GetUpstreamTLSContext(downstreamIdentity, upstreamSvc))
-	if err != nil {
-		return nil, err
+func getUpstreamServiceCluster(downstreamIdentity identity.ServiceIdentity, upstreamSvc service.MeshService, cfg configurator.Configurator, opts ...Option) (*xds_cluster.Cluster, error) {
```
shashankram, line 48: Hey @steeling, I didn't mean to suggest using variadic param here, but rather an additional param to use mTLS.

The goal here is to have a generic way to build clusters for different use cases, ex.
```
getUpstreamServiceCluster(clusterConfig []someStruct) (*xds_cluster.Cluster, error)
```

Since this is outside the scope of your PR, I suggested to simply pass a bool to not initialize the upstream TLS context when using SNI passthrough.

shashankram (commented):
Hey @steeling my original request to split up the PR into individual units would still be beneficial to reviews and ensure a faster merge cycle.

pkg/envoy/cds/cluster.go
steeling: Sure, but we also need to disable permissive discovery. So there needs to be an additional bool option, at which variation Params would be more clean than tacking on multiple bool options

Issue #3444, opened 2021-05-25
OSM Controller changes to support multicluster mode


snehachhabria:
**Please describe the Improvement and/or Feature Request**

In order to support multicluster in oms, we will have to make a number of changes in osm-controller to ensure that the correct configurations are programmed on multicluster gateway and  sidecars.

Some of the changes needed for this are as follows: 

- [ ] Extend endpoint discovery to identify and include the multicluster gateway
- [ ] Program the multicluster gateway 's envoy to include all the services on the cluster it belongs to - This can be scoped/filtered tot the services mentioned in the multicluster crd 
- [ ] Watch for updates to the multicluster crd and trigger envoy updates

 **Scope (please mark with X where applicable)**
- New Functionality          [ ]
- Install                    [ ]
- SMI Traffic Access Policy  [ ]
- SMI Traffic Specs Policy   [ ]
- SMI Traffic Split Policy   [ ]
- Permissive Traffic Policy  [ ]
- Ingress                    [ ]
- Egress                     [ ]
- Envoy Control Plane        [ ]
- CLI Tool                   [ ]
- Metrics                    [ ]
- Certificate Management     [ ]
- Sidecar Injection          [ ]
- Logging                    [ ]
- Debugging                  [ ]
- Tests                      [ ]
- CI System                  [ ]
- Demo                       [ ]
- Project Release            [ ]

**Possible use cases**

snehachhabria:
All the required XDS and controler changes based on the doc : https://docs.google.com/document/d/1O3XcG27uryBxCHX719qYUUzAo913GoFv3LOwdnfNudQ/edit#heading=h.4ze0rn9oj80n have been added to the repo

PR #3599, opened 2021-06-15
feat(pkg/envoy/cds):  add CDS for multi cluster gateway

steeling:
Add cluster SNI passthrough for the multicluster gateway, which should
have every upstream cluster in the local mesh.

part of #3444

-->
**Affected area**:
| Control Plane              | [ x] |

No

1. Is this a breaking change?
No

### DIFF
--- a/pkg/envoy/cds/cluster.go
+++ b/pkg/envoy/cds/cluster.go
@@ -25,14 +25,31 @@ const (
 	clusterConnectTimeout = 1 * time.Second
 )
 
+type clusterOptions struct {
+	permissive bool
+	withTLS    bool
+}
+
+// clusterOption is type of function that edits the defaults of the options struct.
+type clusterOption func(o *clusterOptions)
+
+// withTLS is an option to disable TLS.
+func withTLS(o *clusterOptions) {
+	o.withTLS = true
+}
+
+// permissive is an option to not generate permissive endpoints discovery for clusters.
+func permissive(o *clusterOptions) {
+	o.permissive = true
+}
+
 // getUpstreamServiceCluster returns an Envoy Cluster corresponding to the given upstream service
 // Note: ServiceIdentity must be in the format "name.namespace" [https://github.com/openservicemesh/osm/issues/3188]
-func getUpstreamServiceCluster(downstreamIdentity identity.ServiceIdentity, upstreamSvc service.MeshService, cfg configurator.Configurator) (*xds_cluster.Cluster, error) {
-	clusterName := upstreamSvc.String()
-	marshalledUpstreamTLSContext, err := ptypes.MarshalAny(
-		envoy.GetUpstreamTLSContext(downstreamIdentity, upstreamSvc))
-	if err != nil {
-		return nil, err
+// The defaults are non-permissive, and *no tls*.
+func getUpstreamServiceCluster(downstreamIdentity identity.ServiceIdentity, upstreamSvc service.MeshService, cfg configurator.Configurator, opts ...clusterOption) (*xds_cluster.Cluster, error) {
+	o := &clusterOptions{}
+	for _, opt := range opts {
+		opt(o)
 	}
 
 	HTTP2ProtocolOptions, err := envoy.GetHTTP2ProtocolOptions()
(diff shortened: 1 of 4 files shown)


pkg/envoy/cds/cluster.go
```diff
@@ -25,14 +25,30 @@ const (
 	clusterConnectTimeout = 1 * time.Second
 )
 
+type options struct {
```
shashankram, line 28: nit: s/options/clusterOptions

pkg/envoy/cds/cluster.go
```diff
@@ -25,14 +25,30 @@ const (
...
+// NoTLS is an option to disable TLS.
+func NoTLS(o options) {
+	o.noTLS = true
+}
+
+// NoPermissive is an option to not generate permissive endpoints discovery for clusters.
+func NoPermissive(o options) {
+	o.noPermissive = true
+}
+
```
shashankram, line 45: Un-export local types.

pkg/envoy/cds/cluster.go
```diff
@@ -25,14 +25,30 @@ const (
 	clusterConnectTimeout = 1 * time.Second
 )
 
+type options struct {
+	noPermissive bool
+	noTLS        bool
```
shashankram, line 30: Let's invert these to `permissive, withTLS`

pkg/envoy/cds/cluster.go
```diff
@@ -41,19 +57,29 @@ func getUpstreamServiceCluster(downstreamIdentity identity.ServiceIdentity, upst
...
 			ConfigType: &xds_core.TransportSocket_TypedConfig{
 				TypedConfig: marshalledUpstreamTLSContext,
 			},
-		},
-		TypedExtensionProtocolOptions: HTTP2ProtocolOptions,
+		}
 	}
 
-	if cfg.IsPermissiveTrafficPolicyMode() {
+	if cfg.IsPermissiveTrafficPolicyMode() && !o.noPermissive {
```
shashankram, line 79: would let the caller set the permissive option instead based on `cfg.IsPermissiveTrafficPolicyMode()`, so we don't have similar conflicting conditionals around permissive mode here.

This logic should be as simple as 
```
if o.permissive {
 ...
}
```

pkg/envoy/cds/response.go
```diff
@@ -34,6 +23,21 @@ func NewResponse(meshCatalog catalog.MeshCataloger, proxy *envoy.Proxy, _ *xds_d
...
+			cluster, err := getUpstreamServiceCluster(proxyIdentity.ToServiceIdentity(), dstService, cfg, NoPermissive, NoTLS)
+			if err != nil {
+				log.Error().Err(err).Msgf("Failed to construct service cluster for service %s for proxy with XDS Certificate SerialNumber=%s on Pod with UID=%s",
+					dstService.Name, proxy.GetCertificateSerialNumber(), proxy.String())
+				return nil, err
+			}
+
+			clusters = append(clusters, cluster)
+		}
+		return removeDups(proxy, clusters), nil
```
shashankram, line 38: Can we just use a mapset or map[string]struct{} and deduplicate by cluster.Name inline?

shashankram (changes requested):

pkg/envoy/cds/cluster.go
steeling: that changes the default values to false, is that something we want?

pkg/envoy/cds/response.go
steeling: this was just re-using logic that already existed. It's done in 2 separate places so figured i'd pull it out into a separate function.

shashankram: It makes sense now that I see why you did it this way (early return from function for gateway)

pkg/envoy/cds/response.go
```diff
@@ -89,6 +99,10 @@ func NewResponse(meshCatalog catalog.MeshCataloger, proxy *envoy.Proxy, _ *xds_d
 		clusters = append(clusters, getTracingCluster(cfg))
 	}
 
+	return removeDups(proxy, clusters), nil
+}
+
+func removeDups(proxy *envoy.Proxy, clusters []*xds_cluster.Cluster) []types.Resource {
```
shashankram, line 105: needs a unit test.

shashankram: why is a proxy object needed here to dedup?

pkg/envoy/cds/cluster.go
shashankram: There's no default in the current impl, just based on what the configurator interface returns. Should be fine to change these to `permissiveMode, withTLS`  and remove the usage of configurator within this function for permissive mode.

pkg/envoy/cds/response.go
steeling: logging the name to maintain the original logging statement. Changed to accept just the string and not the full object.

pkg/envoy/cds/cluster.go
steeling: done!

pkg/envoy/cds/response.go
shashankram: This logging seems unnecessary, given it's deduping and clearly it's okay to have duplicates before dedup as per the current logic.

pkg/envoy/cds/cluster.go
shashankram: @steeling I am assuming the plan is to still invert negatives.

pkg/envoy/cds/cluster.go
steeling: done!

pkg/envoy/cds/cluster.go
steeling: i meant the default for the call to the unimported function, but happy to make that change!

pkg/envoy/cds/response.go
steeling: done!

pkg/envoy/cds/response_test.go
```diff
@@ -399,11 +408,16 @@ func TestNewResponseGetLocalServiceClusterError(t *testing.T) {
 	tassert.Nil(t, err)
 
 	ctrl := gomock.NewController(t)
+	mockKubeController := k8s.NewMockController(ctrl)
+	cfg := configurator.NewMockConfigurator(ctrl)
 	meshCatalog := catalog.NewMockMeshCataloger(ctrl)
```
shashankram, line 413: nit: initialize the mocks at the start of the function so it's easy to know the test is leveraging mocks vs fakes without needing to look through the business logic.

shashankram (approved):

snehachhabria (approved):

PR #3600, opened 2021-06-15
feat(pkg/envoy): Add multi cluster gateway LDS 

steeling:
Add LDS configs for the multi cluster gateway. The gateway should be setup
for SSL Passthrough, and be able to discover all upstream services.

part of #3444

**Affected area**:
| Control Plane              | [x ] |

No

1. Is this a breaking change?
No

### DIFF
--- a/pkg/envoy/lds/gateway.go
+++ b/pkg/envoy/lds/gateway.go
@@ -1,0 +1,84 @@
+package lds
+
+import (
+	"fmt"
+
+	xds_listener "github.com/envoyproxy/go-control-plane/envoy/config/listener/v3"
+	"github.com/envoyproxy/go-control-plane/pkg/cache/types"
+	"github.com/envoyproxy/go-control-plane/pkg/wellknown"
+	"google.golang.org/protobuf/types/known/wrapperspb"
+
+	"github.com/openservicemesh/osm/pkg/constants"
+	"github.com/openservicemesh/osm/pkg/envoy"
+	"github.com/openservicemesh/osm/pkg/service"
+)
+
+func (lb *listenerBuilder) buildGatewayListeners() []types.Resource {
+	if !lb.cfg.GetFeatureFlags().EnableMulticlusterMode {
+		return nil
+	}
+	return []types.Resource{
+		&xds_listener.Listener{
+			Name:         multiclusterListenerName,
+			Address:      envoy.GetAddress(constants.WildcardIPAddr, constants.EnvoyInboundListenerPort),
+			FilterChains: lb.getMultiClusterGatewayFilterChainPerUpstream(),
+			ListenerFilters: []*xds_listener.ListenerFilter{
+				{
+					// The OriginalDestination ListenerFilter is used to redirect traffic
+					// to its original destination.
+					Name: wellknown.OriginalDestination,
+				},
+			},
+		},
+	}
+}
+
+func (lb *listenerBuilder) getMultiClusterGatewayFilterChainPerUpstream() []*xds_listener.FilterChain {
+	var filterChains []*xds_listener.FilterChain
+
+	dstServices := lb.meshCatalog.ListMeshServicesForIdentity(lb.serviceIdentity)
+	if len(dstServices) == 0 {
+		log.Debug().Msgf("Proxy with identity %s does not have any allowed upstream services", lb.serviceIdentity)
+		return filterChains
+	}
+
+	// Iterate all destination services
+	for _, upstream := range dstServices {
+		if !upstream.Local() {
+			continue
+		}
+
+		log.Trace().Msgf("Building outbound filter chain for upstream service %s for proxy with identity %s", upstream, lb.serviceIdentity)
+		protocolToPortMap, err := lb.meshCatalog.GetPortToProtocolMappingForService(upstream)
+		if err != nil {
+			log.Error().Err(err).Msgf("Error retrieving port to protocol mapping for upstream service %s", upstream)
+			continue
+		}
+
+		// Create protocol specific inbound filter chains per port to handle different ports serving different protocols
+		for port := range protocolToPortMap {
+			// The gateway uses SSL passthrough, so simply uses a TCP filter.
+			filter, err := lb.getOutboundTCPFilter(upstream)
+			if err != nil {
+				log.Error().Err(err).Msgf("Error getting tcp filter for upstream service %s", upstream)
+				continue
+			}
+
+			hostnames, _ := lb.meshCatalog.GetServiceHostnames(upstream, service.RemoteCluster)
+			filterChains = append(filterChains, &xds_listener.FilterChain{
+				Name:    fmt.Sprintf("%s:%s", outboundMeshTCPFilterChainPrefix, upstream),
+				Filters: []*xds_listener.Filter{filter},
+				FilterChainMatch: &xds_listener.FilterChainMatch{
+					DestinationPort: &wrapperspb.UInt32Value{
+						Value: port,
+					},
+					ServerNames:          hostnames,
+					ApplicationProtocols: envoy.ALPNInMesh,
+					TransportProtocol:    envoy.TransportProtocolTLS,
+				},
+			})
+		}
+	}
+
+	return filterChains
+}
--- a/pkg/envoy/lds/listener.go
+++ b/pkg/envoy/lds/listener.go
@@ -20,6 +20,7 @@ import (
 const (
 	inboundListenerName           = "inbound-listener"
 	outboundListenerName          = "outbound-listener"
+	multiclusterListenerName      = "multicluster-listener"
 	prometheusListenerName        = "inbound-prometheus-listener"
 	outboundEgressFilterChainName = "outbound-egress-filter-chain"
 	egressTCPProxyStatPrefix      = "egress-tcp-proxy"
--- a/pkg/envoy/lds/response.go
+++ b/pkg/envoy/lds/response.go
@@ -18,17 +18,6 @@ import (
 // 2. Outbound listener to handle outgoing traffic
 // 3. Prometheus listener for metrics
 func NewResponse(meshCatalog catalog.MeshCataloger, proxy *envoy.Proxy, _ *xds_discovery.DiscoveryRequest, cfg configurator.Configurator, _ certificate.Manager, proxyRegistry *registry.ProxyRegistry) ([]types.Resource, error) {
-	if proxy.Kind() == envoy.KindGateway {
-		// TODO: Configure gateway
-		return nil, nil
-	}
-
-	svcList, err := proxyRegistry.ListProxyServices(proxy)
-	if err != nil {
-		log.Error().Err(err).Msgf("Error looking up MeshService for proxy %s", proxy.String())
-		return nil, err
-	}
-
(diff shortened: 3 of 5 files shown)


pkg/envoy/lds/inmesh.go
```diff
@@ -383,14 +383,12 @@ func (lb *listenerBuilder) getOutboundTCPFilter(upstream service.MeshService) (*
 func (lb *listenerBuilder) getOutboundFilterChainPerUpstream() []*xds_listener.FilterChain {
 	var filterChains []*xds_listener.FilterChain
 
-	upstreamServices := lb.meshCatalog.ListAllowedOutboundServicesForIdentity(lb.serviceIdentity)
-	if len(upstreamServices) == 0 {
+	dstServices := lb.meshCatalog.ListMeshServicesForIdentity(lb.serviceIdentity)
```
snehachhabria, line 386: This doesn't seem right, why are we altering existing behavior? If we need to build something specific for multicluster it needs to be a separate code path

pkg/envoy/lds/multicluster_gateway.go
```diff
@@ -0,0 +1,83 @@
...
+				Name: wellknown.OriginalDestination,
+			},
+		},
+	}
+}
+
+func (lb *listenerBuilder) getMultiClusterGatewayFilterChainPerUpstream() []*xds_listener.FilterChain {
+	var filterChains []*xds_listener.FilterChain
+
+	dstServices := lb.meshCatalog.ListMeshServicesForIdentity(lb.serviceIdentity)
```
snehachhabria, line 36: Will the gateway have a service identity ? as far as I am aware it won't, so you will not be able to get the mesh services from the gateway's identity

pkg/envoy/lds/multicluster_gateway.go
```diff
@@ -0,0 +1,83 @@
...
+	if len(dstServices) == 0 {
+		log.Debug().Msgf("Proxy with identity %s does not have any allowed upstream services", lb.serviceIdentity)
+		return filterChains
+	}
+
+	// Iterate all destination services
+	for _, upstream := range dstServices {
```
snehachhabria, line 43:
```suggestion
	for _, destService := range dstServices {
```

pkg/envoy/lds/response.go
```diff
@@ -44,6 +39,23 @@ func NewResponse(meshCatalog catalog.MeshCataloger, proxy *envoy.Proxy, _ *xds_d
...
+		} else {
+			ldsResources = append(ldsResources, prometheusListener)
+		}
+	}
+
+	if proxy.Kind() == envoy.KindGateway {
+		ldsResources = append(ldsResources, lb.newMultiClusterGatewayListener())
+		return ldsResources, nil
+	}
+
```
snehachhabria, line 58: Lets have this block where it was earlier and have the Gateway check after it

pkg/envoy/lds/response_test.go
```diff
@@ -108,7 +108,7 @@ func TestNewResponse(t *testing.T) {
 	assert.Len(resources, 3)
 
 	// validating outbound listener
-	listener, ok := resources[0].(*xds_listener.Listener)
```
snehachhabria, line 111: if you do not change the order above you will not have to update this test

pkg/envoy/lds/inmesh.go
shashankram: Correct, this should not program all services for the given identity, only outbound as done by `ListAllowedOutboundServicesForIdentity`.

steeling: This is just removing erroneous code. It previously called ListAllowedServices, then did nothing. Now we call ListMeshServicesForIdentity, which under the hood calls  ListAllowedServices, but combines with the Apex services from traffic splits.

shashankram: Would you mind fixing the erroneous code in a separate PR considering it is outside the scope of multicluster. This previously called `ListAllowedOutboundServicesForIdentity()`, not sure what you mean why you say `then did nothing`. Clearly we should not be programming upstream clusters for all services corresponding to the identity.

steeling: it doesn't.. take a look at the implementation of ListMeshServicesForIdentity. but ok will send anotehr PR

shashankram: No, I the understand the bug, but your fix doesn't seem not correct either. Should be calling `ListAllowedOutboundServicesForIdentity` instead of `ListMeshServicesForIdentity`

pkg/envoy/lds/multicluster_gateway.go
steeling: moved these changes to https://github.com/openservicemesh/osm/pull/3602

snehachhabria: thanks for that PR but this line still needs a change for consistency. 
```suggestion
	for _, upstream := range upstreamServices {
```

pkg/envoy/lds/multicluster_gateway.go
snehachhabria, line 36:
```suggestion
	upstreamServices := lb.meshCatalog.ListMeshServicesForIdentity(lb.serviceIdentity)
```

pkg/envoy/lds/response.go
snehachhabria: makes sense. 

I don't see the multicluster feature flag being used anywhere, could we please wrap all code in the NewResponse for multicluster under its feature flag.

pkg/envoy/lds/multicluster_gateway.go
steeling: done!

pkg/envoy/lds/multicluster_gateway.go
```diff
@@ -0,0 +1,83 @@
...
+		// Filter out to only the local and global services.
+		// TODO(steeling): local here needs to the remote name.
```
shashankram, line 45: typo in comment?

pkg/envoy/lds/response.go
steeling: done!

pkg/envoy/lds/listener.go
```diff
@@ -20,13 +20,15 @@ import (
...
 	outboundListenerName          = "outbound-listener"
+	multiclusterListenerName      = "multicluster-listener"
 	prometheusListenerName        = "inbound-prometheus-listener"
 	outboundEgressFilterChainName = "outbound-egress-filter-chain"
 	egressTCPProxyStatPrefix      = "egress-tcp-proxy"
 	singleIpv4Mask                = 32
 )
 
 func (lb *listenerBuilder) newOutboundListener() (*xds_listener.Listener, error) {
+	// TODO(steeling): I believe there only needs to be 2 filter chains, 1 HTTP, 1 tcp... not relevant to this PR.
```
shashankram, line 31: If you believe the existing code is not correct, would you mind following an issue and following up instead of dropping a TODO here.

snehachhabria (commented):
Could we update the PR title, description and the commit message to be more indicative of the changes the PR contains, right now there are 3 PR's with the same title and it makes it very hard to review .

Could we do this for all the PR's that you have open with the same title

pkg/envoy/lds/response.go
```diff
@@ -18,11 +18,6 @@ import (
 // 2. Outbound listener to handle outgoing traffic
 // 3. Prometheus listener for metrics
 func NewResponse(meshCatalog catalog.MeshCataloger, proxy *envoy.Proxy, _ *xds_discovery.DiscoveryRequest, cfg configurator.Configurator, _ certificate.Manager, proxyRegistry *registry.ProxyRegistry) ([]types.Resource, error) {
-	if proxy.Kind() == envoy.KindGateway {
```
shashankram, line 21: I am inclined to build the listener for multicluster and ingress scenarios under this condition. I don't see a reason to build outbound listeners or prometheus (atm) for gateway, so let's bring this conditional back and create `lds/gateway.go` to house common code for ingress and multicluster (ex. listener struct), and simply return the LDS resources directly within this conditional.

steeling: done!

pkg/envoy/lds/multicluster_gateway.go
```diff
@@ -0,0 +1,85 @@
+package lds
```
shashankram, line 1: Rename file to `gateway.go`, I'll be adding ingress gateway specific listeners here as we discussed offline.

pkg/envoy/lds/listener.go
steeling: oops.. that was a personal note as i was writing the other envoy doc. removed now

pkg/envoy/lds/response.go
```diff
@@ -44,6 +33,10 @@ func NewResponse(meshCatalog catalog.MeshCataloger, proxy *envoy.Proxy, _ *xds_d
 
 	lb := newListenerBuilder(meshCatalog, svcAccount.ToServiceIdentity(), cfg, statsHeaders)
 
+	if cfg.GetFeatureFlags().EnableMulticlusterMode && proxy.Kind() == envoy.KindGateway {
```
shashankram, line 36: I'd move the multicluster check inside buildGatewayListeners, considering that function needs to build the listeners for ingress.

shashankram (approved):

pkg/envoy/lds/gateway.go
```diff
@@ -0,0 +1,84 @@
...
+
+	"github.com/openservicemesh/osm/pkg/constants"
+	"github.com/openservicemesh/osm/pkg/envoy"
+	"github.com/openservicemesh/osm/pkg/service"
+)
+
+func (lb *listenerBuilder) buildGatewayListeners() []types.Resource {
+	if !lb.cfg.GetFeatureFlags().EnableMulticlusterMode {
+		return nil
+	}
```
snehachhabria, line 19: we can remove this check from here

pkg/envoy/lds/response.go
```diff
@@ -44,6 +33,10 @@ func NewResponse(meshCatalog catalog.MeshCataloger, proxy *envoy.Proxy, _ *xds_d
 
 	lb := newListenerBuilder(meshCatalog, svcAccount.ToServiceIdentity(), cfg, statsHeaders)
 
+	if proxy.Kind() == envoy.KindGateway {
```
snehachhabria, line 36: lets add the multicluster feature flag enabled check here instead please

steeling: I had it here earlier, but @shashankram asked me to move it to the current location

shashankram: @snehachhabria this check is moved inside the function that gets called, so that ingress code can be integrated without needing to move the check back to where it is after the changes mode. This looks good to me.

pkg/envoy/lds/gateway.go
shashankram: Let's keep this check here because I am going to use this flag and split the logic for ingress and multicluster appropriately.

snehachhabria: makes sense, thanks for clarifying

snehachhabria (approved):

PR #3603, opened 2021-06-15
Adding clusterId to MeshConfig for multicluster support

mtarng:
**Description**:
Adding a new property for the multicluster scenario. For now, only adding a `clusterId` field to set the cluster's identity for reference.

A component of: https://github.com/openservicemesh/osm/issues/3456, https://github.com/openservicemesh/osm/issues/3187

**Affected area**:
| New Functionality          | [X ] |
| Install                    | [ X] |


allenlsy:
We also need to update https://github.com/openservicemesh/osm/blob/main/pkg/gen/client/config/listers/config/v1alpha1/meshconfig.go

I'm also think whether clusterId can be put outside of `multicluster` section, just to make it more general. If a cluster is not part of the multicluster set up, which means it runs on it own. Then we can give it a default clusterId.

charts/osm/crds/meshconfig.yaml
```diff
@@ -187,6 +187,10 @@ spec:
                       description: Sets the service certificate validity duration, represented as a sequence of decimal numbers each with optional fraction and a unit suffix.
                       type: string
                       default: "24h"
+                clusterId:
```
snehachhabria, line 190: We want to avoid making changes to the meshconfig for multicluster work since this is an experimental feature. Would it be possible to add this else where ?

allenlsy: How about creating a `experimental` section in MeshConfig?

mtarng: Sure, so nest this under a new `experimental` property?

charts/osm/crds/meshconfig.yaml
```diff
@@ -187,6 +187,14 @@ spec:
                       description: Sets the service certificate validity duration, represented as a sequence of decimal numbers each with optional fraction and a unit suffix.
                       type: string
                       default: "24h"
+                experimental:
+                  description: Experimental configurations
```
snehachhabria, line 191: should we have another object multicluster under experimental and then the clusterId under that ? This way the entire multicluster object can be removed from the experimental section once its no longer a feature

charts/osm/crds/meshconfig.yaml
```diff
@@ -187,6 +187,14 @@ spec:
...
+                  type: object
+                  properties:
+                    clusterID:
```
snehachhabria, line 194: Maybe call this clusterName or clusterDomain

charts/osm/crds/meshconfig.yaml
```diff
@@ -187,6 +187,14 @@ spec:
...
+                      description: ID for this cluster; will be used by other clusters to reference it in a multi-cluster scenario.
+                      type: string
+                      default: "clusterID"
```
snehachhabria, line 197: will tis default actually be beneficial ?

charts/osm/crds/meshconfig.yaml
mtarng: Done

charts/osm/crds/meshconfig.yaml
mtarng: Renamed to clusterDomain along with associated properties/methods

charts/osm/crds/meshconfig.yaml
mtarng: Fair enough, removed and set to omit if empty.

snehachhabria (approved):

shashankram (approved):

PR #3608, opened 2021-06-16
Multicluster config outline

timmyreilly:
config outline, announcements and mocks

**Affected area**:

1. Is this a breaking change?


pkg/config/types.go
```diff
@@ -0,0 +1,36 @@
...
+
+	"github.com/openservicemesh/osm/pkg/kubernetes"
+	"github.com/openservicemesh/osm/pkg/logger"
+)
+
+var (
+	log = logger.New("multiclusterservice-controller")
+)
+
+// informerCollection is the type used to represent the collection of informers for the remoteservice.openservicemesh.io API group
```
snehachhabria, line 15:
```suggestion
// informerCollection is the type used to represent the collection of informers for the mulitclusterservice.openservicemesh.io API group
```

please update this in here and in all other places below

pkg/config/client.go
```diff
@@ -0,0 +1,93 @@
+package config
```
snehachhabria, line 1: could we name this package as mulitcluster or something else more relevant rather than just config

pkg/config/client.go
```diff
@@ -0,0 +1,93 @@
...
+
```
snehachhabria, line 2: Lets have the methods in this file named similar to this : https://github.com/openservicemesh/osm/blob/main/pkg/ingress/client.go

pkg/config/client.go
steeling: FYI this is related to the group, ie: config.openservicemesh.io. So this would be the "config" client for all the services in that group. Not sure if that changes your comment

pkg/config/client.go
```diff
@@ -0,0 +1,93 @@
...
+	"github.com/openservicemesh/osm/pkg/announcements"
+)
+
+const (
+	// apiGroup is the k8s API group that this package interacts with
+	apiGroup = "config.openservicemesh.io"
+)
+
+// NewConfigController returns a config.Controller interface related to functionality provided by the resources in the config.openservicemesh.io API group
+func NewConfigController(kubeConfig *rest.Config, kubeController kubernetes.Controller, stop chan struct{}) (Controller, error) {
```
steeling, line 24: "return structs, not interfaces"

PR #3610, opened 2021-06-16
Multicluster config kubernetes updates

timmyreilly:
Updating kubernetes client to work with multicluster config

**Affected area**:

1. Is this a breaking change?

### DIFF
--- a/pkg/announcements/types.go
+++ b/pkg/announcements/types.go
@@ -156,6 +156,17 @@ const (
 
 	// EgressUpdated is the type of announcement emitted when we observe an update to egress.policy.openservicemesh.io
 	EgressUpdated AnnouncementType = "egress-updated"
+
+	// ---
+
+	// MultiClusterServiceAdded is the type of announcement emitted when we observe an addition of a multiclusterservice.config.openservicemesh.io
+	MultiClusterServiceAdded AnnouncementType = "multiclusterservice-added"
+
+	// MultiClusterServiceDeleted is the type of announcement emitted when we observe an deletion of a multiclusterservice.config.openservicemesh.io
+	MultiClusterServiceDeleted AnnouncementType = "multiclusterservice-deleted"
+
+	// MultiClusterServiceUpdated is the type of announcement emitted when we observe an update of a multiclusterservice.config.openservicemesh.io
+	MultiClusterServiceUpdated AnnouncementType = "multiclusterservice-updated"
 )
 
 // Announcement is a struct for messages between various components of OSM signaling a need for a change in Envoy proxy configuration
--- a/pkg/kubernetes/client.go
+++ b/pkg/kubernetes/client.go
@@ -15,7 +15,10 @@ import (
 	"k8s.io/client-go/tools/cache"
 
 	"github.com/openservicemesh/osm/pkg/announcements"
+	"github.com/openservicemesh/osm/pkg/apis/config/v1alpha1"
 	"github.com/openservicemesh/osm/pkg/constants"
+	configinformer "github.com/openservicemesh/osm/pkg/gen/client/config/informers/externalversions"
+
 	"github.com/openservicemesh/osm/pkg/identity"
 	"github.com/openservicemesh/osm/pkg/service"
 )
@@ -32,11 +35,12 @@ func NewKubernetesController(kubeClient kubernetes.Interface, meshName string, s
 
 	// Initialize informers
 	informerInitHandlerMap := map[InformerKey]func(){
-		Namespaces:      client.initNamespaceMonitor,
-		Services:        client.initServicesMonitor,
-		ServiceAccounts: client.initServiceAccountsMonitor,
-		Pods:            client.initPodMonitor,
-		Endpoints:       client.initEndpointMonitor,
+		Namespaces:           client.initNamespaceMonitor,
+		Services:             client.initServicesMonitor,
+		ServiceAccounts:      client.initServiceAccountsMonitor,
+		Pods:                 client.initPodMonitor,
+		Endpoints:            client.initEndpointMonitor,
+		MultiClusterServices: client.initMultiClusterServicesMonitor,
 	}
 
 	// If specific informers are not selected to be initialized, initialize all informers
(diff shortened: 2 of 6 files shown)


pkg/kubernetes/types.go
```diff
@@ -103,4 +108,10 @@ type Controller interface {
 
 	// IsMetricsEnabled returns true if the pod in the mesh is correctly annotated for prometheus scrapping
 	IsMetricsEnabled(*corev1.Pod) bool
+
+	// GetMultiClusterSvc returns an v1alpha1 MulticClusterService if it exists, or nil
+	GetMultiClusterSvc(name, namespace string) (*v1alpha1.MultiClusterService, error)
+
+	// ListMultiClusterServices returns a list of v1alpha1 MulticClusterServices if any exist, or nil
+	ListMultiClusterServices(name, namespace string) ([]*v1alpha1.MultiClusterService, error)
```
timmyreilly, line 116: remove name parameter, all good with namespace?

timmyreilly: Update to "all" by removing all parameters

pkg/kubernetes/client.go
```diff
@@ -311,3 +330,28 @@ func (c Client) IsMetricsEnabled(pod *corev1.Pod) bool {
 	isScrapingEnabled, _ = strconv.ParseBool(prometheusScrapeAnnotation)
 	return isScrapingEnabled
 }
+
+// GetMultiClusterSvc returns an v1alpha1 MulticClusterService if it exists, or nil
+func (c Client) GetMultiClusterSvc(name, namespace string) (*v1alpha1.MultiClusterService, error) {
```
snehachhabria, line 335:
```suggestion
func (c Client) GetMultiClusterService(name, namespace string) (*v1alpha1.MultiClusterService, error) {
```

pkg/kubernetes/client.go
```diff
@@ -311,3 +330,28 @@ func (c Client) IsMetricsEnabled(pod *corev1.Pod) bool {
 	isScrapingEnabled, _ = strconv.ParseBool(prometheusScrapeAnnotation)
 	return isScrapingEnabled
 }
+
+// GetMultiClusterSvc returns an v1alpha1 MulticClusterService if it exists, or nil
```
snehachhabria, line 334:
```suggestion
// GetMultiClusterService returns an v1alpha1 MultiClusterService if it exists, or nil
```

pkg/kubernetes/client.go
```diff
@@ -311,3 +330,28 @@ func (c Client) IsMetricsEnabled(pod *corev1.Pod) bool {
...
+func (c Client) GetMultiClusterSvc(name, namespace string) (*v1alpha1.MultiClusterService, error) {
+	svcIf, exists, err := c.informers[MultiClusterServices].GetStore().GetByKey(name)
+	if exists && err == nil {
+		mcsvc := svcIf.(*v1alpha1.MultiClusterService)
+		return mcsvc, nil
+	}
+	return nil, nil
+}
+
+// ListMultiClusterServices returns a list of v1alpha1 MulticClusterServices if any exist, or nil
```
snehachhabria, line 344:
```suggestion
// ListMultiClusterServices returns a list of v1alpha1 MultiClusterServices if any exist, or nil
```

pkg/announcements/types.go
```diff
@@ -156,6 +156,17 @@ const (
...
+	// ---
+
+	// MultiClusterServiceAdded is the type of announcement emitted when we observe an addition of a multiclusterservice.config.openservicemesh.io
+	MultiClusterServiceAdded AnnouncementType = "multiclusterservice-added"
+
+	// MultiClusterServiceDeleted is the type of announcement emitted when we observe an deletion of a multiclusterservice.config.openservicemesh.io
+	MultiClusterServiceDeleted AnnouncementType = "multiclusterservice-deleted"
+
+	// MultiClusterServiceUpdated is the type of announcement emitted when we observe an update of a multiclusterservice.config.openservicemesh.io
+	MultiClusterServiceUpdated AnnouncementType = "multiclusterservice-updated"
```
snehachhabria, line 169: You will need to add this to this file as well : https://github.com/openservicemesh/osm/blob/113bfc328b857272dc149066d162fc950f0b584f/pkg/catalog/dispatcher.go#L43

pkg/kubernetes/client.go
```diff
@@ -98,6 +102,21 @@ func (c *Client) initServicesMonitor() {
 	c.informers[Services].AddEventHandler(GetKubernetesEventHandlers((string)(Services), providerName, c.shouldObserve, svcEventTypes))
 }
 
+func (c *Client) initMultiClusterServicesMonitor() {
```
snehachhabria, line 105: Adding the changes here doesn't seem right, if I look at both the other custom crd's that we have in osm the changes for their informers are in these files : 

1. egress crd : https://github.com/openservicemesh/osm/blob/113bfc328b857272dc149066d162fc950f0b584f/pkg/policy/client.go#L41
2. meshConfig crd : https://github.com/openservicemesh/osm/blob/113bfc328b857272dc149066d162fc950f0b584f/pkg/configurator/client.go#L29

Please could we follow these examples and make the change for the mulitcluster crd accordingly

snehachhabria (changes requested):

pkg/kubernetes/client.go
steeling: I'd argue that those ways are more incorrect for the following reasons:

1. It places more direct dependencies on the MeshCatalog to Kubernetes (these should be abstracted through providers).
2. It's kind of messy to have a separate client for each CRD.. code will grow unwieldy
3. Knowledge of CRD's will become more intertwined, meaning more types will need to reconcile MultiClusterService with Service, or other CRD's. Forcing each type to need a client for each of these CRD's (when it's all reaching out to the Kuberentes master) is a burden on the developer. For a concrete example, we would then need to modify the kube provider to also have a MultiClusterServiceController.
4. Clients are clients that talk to a "thing". In this case that thing is the K8s master, why have 9 clients for 9 of the resources backed by that same "thing".

If you'd like, we can pick up the work to remove those 2 clients and migrate the code to the Kubernetes Controller.

pkg/kubernetes/client.go
```diff
@@ -32,11 +35,12 @@ func NewKubernetesController(kubeClient kubernetes.Interface, meshName string, s
```
steeling, line 35: you'll need the kubernetes controller to accept and populate the ConfigInterface

pkg/kubernetes/client.go
steeling: https://github.com/openservicemesh/osm/pull/3635 brings in those changes for the other client types

PR #3609, opened 2021-06-16
Feature/service provider interface

whitneygriffith:
**Description**: Create Service Provider Interface 

Upcoming PRs will include: 

- move kube provider to the new directory providers and add new service provider functions
- edit to the 

Relevant Issue: #3456 

**Affected area**:
| Functional Area            |     |
| -------------------------- | --- |
| New Functionality          | [ ] |
| Documentation              | [ ] |
| Install                    | [ ] |
| CLI Tool                   | [ ] |
| Certificate Management     | [ ] |
| Control Plane              | [ ] |
| Ingress                    | [ ] |
| Egress                     | [ ] |
| Networking                 | [ ] |
| Observability              | [ ] |
| SMI Policy                 | [ ] |
| Sidecar Injection          | [ ] |
| Security                   | [ ] |
| Upgrade                    | [ ] |
| Tests                      | [ ] |
| CI System                  | [ ] |
| Demo                       | [ ] |
| Performance                | [ ] |
| Other                      | [ ] |

Please answer the following questions with yes/no.

1. Does this change contain code from or inspired by another project?
    -   Did you notify the maintainers and provide attribution?

No

1. Is this a breaking change?

No

PR #3614, opened 2021-06-16
feat(provider): move kube endpoint provider client

whitneygriffith:
**Description**: move kube provider to the new directory, providers and add new service provider functions

Related PR: #3609, #3620
 
Associated Issue: #3456 

**Affected area**:
| Functional Area            |  x  |

No 

1. Is this a breaking change?

No


snehachhabria (commented):
This PR has changes from here as well : https://github.com/openservicemesh/osm/pull/3609 which doesn't seem right if you are intending to break them up.

pkg/providers/kube/client.go
```diff
@@ -13,10 +13,17 @@ import (
...
+	"github.com/openservicemesh/osm/pkg/utils"
 )
 
-// NewProvider implements mesh.EndpointsProvider, which creates a new Kubernetes cluster/compute provider.
-func NewProvider(kubeController k8s.Controller, providerIdent string, cfg configurator.Configurator) (endpoint.Provider, error) {
+var (
+	// Ensure Client conforms to the provider interfaces
+	_ endpoint.Provider = &Client{}
+	_ service.Provider  = &Client{}
+)
```
shashankram, line 23: This seems unnecessary. The compiler should error if the client doesn't implement an interface as long as the functions that require an interface accept the interface type and not the underlying implementation type (Client in this case).

pkg/endpoint/types.go
```diff
@@ -20,9 +20,6 @@ type Provider interface {
 	// ListEndpointsForIdentity retrieves the list of IP addresses for the given service account
 	ListEndpointsForIdentity(identity.ServiceIdentity) []Endpoint
 
-	// GetServicesForServiceAccount retrieves the namespaced services for a given service account
-	GetServicesForServiceAccount(identity.K8sServiceAccount) ([]service.MeshService, error)
```
steeling, line 24: Let's keep this here on this PR, and circle back to it on the next one.

pkg/catalog/service.go
```diff
@@ -72,8 +72,8 @@ func (mc *MeshCatalog) getApexServicesForBackendService(targetService service.Me
 // getServicesForServiceAccount returns a list of services corresponding to a service account
 func (mc *MeshCatalog) getServicesForServiceAccount(sa identity.K8sServiceAccount) ([]service.MeshService, error) {
 	var services []service.MeshService
-	for _, provider := range mc.endpointsProviders {
```
steeling, line 75: Which would mean reverting this change as well

shashankram (commented):
@whitneygriffith, would you mind fixing the commit message formatting so it is close to the guidelines described at https://github.com/openservicemesh/osm/blob/main/CONTRIBUTING.md#commit-style-guideline.

image

pkg/catalog/types.go
```diff
@@ -25,6 +25,7 @@ var (
 // MeshCatalog is the struct for the service catalog
 type MeshCatalog struct {
 	endpointsProviders []endpoint.Provider
+	serviceProviders   []service.Provider
```
shashankram, line 28: What's the reason to not use a single `service.Provider` interface to also serve the endpoints for a service? I don't see a need to have 2 interfaces here.

steeling: at the end of the day these providers query a control plane. It's likely that there would be a single control plane in every case (both endpoint and service), however if there's multiple endpoint providers then there will likely be multiple service providers.

shashankram: I don't understand this comment, it seems to be making an assumption which I am not comfortable with. The idea behind having a provider was to abstract different runtimes (k8s, VMs etc.), so why not simplify and have a single provider (ServiceProvider) that provides information about the service and its endpoints?

I think you are probably referring to k8s control plane components (ex. endpoints controller), but that's not what this abstraction is intended for. The OSM control plane is always going to use an API (k8s in this case) to retrieve information regarding a service (and it's endpoints), rather than dealing directly with an internal component such as a controller provided by k8s.

steeling: > I don't understand this comment, it seems to be making an assumption which I am not comfortable with. The idea behind having a provider was to abstract different runtimes (k8s, VMs etc.), so why not simplify and have a single provider (ServiceProvider) that provides information about the service and its endpoints?

Yes, this was my initial suggestion; we even had some code that did it this way initially. I'm not sure if you remember, but you specifically requested we separate these out, as you didn't want to make that big of a change on the first pass. If you'd like to revert that decision, we can, but I'd prefer to tackle that on a later task.

```golang
type Provider interface {
   service.Provider
   endpoint.Provider
}
```

steeling: my previous comment was referring to the fact that it's a list of providers. I agree that one provider interface would have been better

shashankram: The reason I suggested introducing a separate component was to avoid refactoring the existing code for a POC. This is still okay for this POC, but eventually, it would make sense to collapse these 2 interfaces. However, this code has already moved the existing endpoints provider code around (which was something I was hoping to avoid with the initial change you made), so we could very well go the extra mile and collapse these. Handling this in a subsequent PR is fine, but please file a GH issue to track this enhancement.

whitneygriffith: Will move forward as is, other PR that implements these changes #3620 

Created issue to collapse these interfaces into one [3632](https://github.com/openservicemesh/osm/issues/3632)

shashankram (approved):

snehachhabria (approved):

PR #3620, opened 2021-06-17
feat(provider): catalog uses service and endpoint providers methods

whitneygriffith:
**Description**:  Kube Provider is aware of both Service and Endpoint Provider Interfaces and relevant functions have been updated

Related PRs: #3609 and #3614 

Associated Issues: #3456 

**Affected area**:
| New Functionality          | [ X] |
| Other                      | [X ] |

No

1. Is this a breaking change?

No

### DIFF
--- a/cmd/osm-controller/osm-controller.go
+++ b/cmd/osm-controller/osm-controller.go
@@ -44,6 +44,7 @@ import (
 	"github.com/openservicemesh/osm/pkg/metricsstore"
 	"github.com/openservicemesh/osm/pkg/policy"
 	"github.com/openservicemesh/osm/pkg/providers/kube"
+	"github.com/openservicemesh/osm/pkg/service"
 	"github.com/openservicemesh/osm/pkg/signals"
 	"github.com/openservicemesh/osm/pkg/smi"
 	"github.com/openservicemesh/osm/pkg/validator"
@@ -197,6 +198,7 @@ func main() {
 	kubeProvider := kube.NewClient(kubernetesClient, configClient, constants.KubeProviderName, cfg)
 
 	endpointsProviders := []endpoint.Provider{kubeProvider}
+	serviceProviders := []service.Provider{kubeProvider}
 
 	ingressClient, err := ingress.NewIngressClient(kubeClient, kubernetesClient, stop, cfg)
 	if err != nil {
@@ -216,7 +218,9 @@ func main() {
 		policyController,
 		stop,
 		cfg,
-		endpointsProviders...)
+		serviceProviders,
+		endpointsProviders,
+	)
 
 	var proxyMapper registry.ProxyServiceMapper
 	if cfg.GetFeatureFlags().EnableAsyncProxyServiceMapping {
--- a/pkg/catalog/inbound_traffic_policies_test.go
+++ b/pkg/catalog/inbound_traffic_policies_test.go
@@ -490,12 +490,14 @@ func TestListInboundTrafficPolicies(t *testing.T) {
 			mockKubeController := k8s.NewMockController(mockCtrl)
 			mockMeshSpec := smi.NewMockMeshSpec(mockCtrl)
 			mockEndpointProvider := endpoint.NewMockProvider(mockCtrl)
+			mockServiceProvider := service.NewMockProvider(mockCtrl)
 			mockConfigurator := configurator.NewMockConfigurator(mockCtrl)
 
 			mc := MeshCatalog{
 				kubeController:     mockKubeController,
 				meshSpec:           mockMeshSpec,
 				endpointsProviders: []endpoint.Provider{mockEndpointProvider},
+				serviceProviders:   []service.Provider{mockServiceProvider},
 				configurator:       mockConfigurator,
 			}
 
@@ -524,6 +526,19 @@ func TestListInboundTrafficPolicies(t *testing.T) {
 			}
 
 			mockConfigurator.EXPECT().IsPermissiveTrafficPolicyMode().Return(tc.permissiveMode).AnyTimes()
+
+			for _, ms := range tc.meshServices {
+				locality := service.LocalCluster
+				if ms.Namespace == tc.downstreamSA.ToK8sServiceAccount().Namespace {
+					locality = service.LocalNS
+					mockServiceProvider.EXPECT().GetHostnamesForService(ms, locality).Return(tests.ExpectedHostnames[ms.Name], nil).AnyTimes()
+				} else {
+					if ms.Name == tests.BookstoreApexServiceName {
+						mockServiceProvider.EXPECT().GetHostnamesForService(ms, locality).Return(tests.ExpectedHostnames["BookstoreApexServiceName"], nil).AnyTimes()
+					}
+				}
+			}
+
 			actual := mc.ListInboundTrafficPolicies(tc.upstreamSA, tc.upstreamServices)
 			assert.ElementsMatch(tc.expectedInboundPolicies, actual)
 		})
@@ -971,11 +986,13 @@ func TestListInboundPoliciesForTrafficSplits(t *testing.T) {
 			mockKubeController := k8s.NewMockController(mockCtrl)
 			mockMeshSpec := smi.NewMockMeshSpec(mockCtrl)
 			mockEndpointProvider := endpoint.NewMockProvider(mockCtrl)
+			mockServiceProvider := service.NewMockProvider(mockCtrl)
 
 			mc := MeshCatalog{
 				kubeController:     mockKubeController,
 				meshSpec:           mockMeshSpec,
 				endpointsProviders: []endpoint.Provider{mockEndpointProvider},
+				serviceProviders:   []service.Provider{mockServiceProvider},
 			}
 
 			for _, meshSvc := range tc.meshServices {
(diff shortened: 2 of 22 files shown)


pkg/catalog/service.go
```diff
@@ -146,25 +146,28 @@ func (mc *MeshCatalog) GetTargetPortToProtocolMappingForService(svc service.Mesh
...
+	if len(portToProtocolMap) == 0 {
+		return nil, errors.Errorf("Error fetching port:protocol mapping for service %s", svc)
 	}
 
 	return portToProtocolMap, nil
 }
 
 // listMeshServices returns all services in the mesh
+// TODO(whitneygriffith): rm or reference provider
 func (mc *MeshCatalog) listMeshServices() []service.MeshService {
```
whitneygriffith, line 171: Want to rename to CamelCase

pkg/catalog/service.go
```diff
@@ -146,19 +146,21 @@ func (mc *MeshCatalog) GetTargetPortToProtocolMappingForService(svc service.Mesh
...
 	}
 
-	for _, portSpec := range k8sSvc.Spec.Ports {
-		var appProtocol string
-		if portSpec.AppProtocol != nil {
-			appProtocol = *portSpec.AppProtocol
-		} else {
-			appProtocol = kubernetes.GetAppProtocolFromPortName(portSpec.Name)
-		}
-		portToProtocolMap[uint32(portSpec.Port)] = appProtocol
```
shashankram, line 161: Why are these checks being removed?

pkg/catalog/service.go
```diff
@@ -146,19 +146,21 @@ func (mc *MeshCatalog) GetTargetPortToProtocolMappingForService(svc service.Mesh
...
-	if k8sSvc == nil {
-		return nil, errors.Wrapf(ErrServiceNotFound, "Error retrieving k8s service %s", svc)
+	for _, provider := range mc.serviceProviders {
+		currentPortToProtocolMap, err := provider.GetTargetPortToProtocolMappingForService(svc)
+		if err != nil {
+			return nil, err
+		}
+		for key, value := range currentPortToProtocolMap {
+			if v, ok := portToProtocolMap[key]; ok && v != value {
+				return nil, errors.Errorf("Error fetching port:protocol, multiple entries found on same port for service %s", svc)
```
shashankram, line 156: Clarify log to something like:
```suggestion
				return nil, errors.Errorf("Unexpected protocol %s for port %d, expected protocol %s, service: %s, provider: %s", value, key, v, svc, provider.GetID())
```

pkg/catalog/service.go
```diff
@@ -167,22 +169,37 @@ func (mc *MeshCatalog) GetPortToProtocolMappingForService(svc service.MeshServic
...
-	if svc == nil {
-		return nil, errors.Errorf("Error fetching service %q", meshService)
+	svc := utils.K8sSvcToMeshSvc(mc.kubeController.GetService(meshService))
+
+	var hostnames []string
+	for _, provider := range mc.serviceProviders {
+		hosts, err := provider.GetHostnamesForService(svc, locality)
+
+		if err != nil {
+			return nil, errors.Errorf("Error fetching service %q", meshService)
```
shashankram, line 197:
```suggestion
			return nil, errors.Errorf("Error fetching hostnames for sercice: %s, provider: %s", meshService, provider.GetID())
```

pkg/catalog/service.go
whitneygriffith: We moved  this check to the kube provider [pkg/providers/kube/client.go#L277](https://github.com/openservicemesh/osm/blob/bf225542d9e6569d7d6c638ab779be8be272c86c/pkg/providers/kube/client.go#L277) since the catalog shouldn't necessarily  know about kubernetes. @steeling

snehachhabria (commented):
Seems like the changes in this PR are causing the e2e tests to fail and even DCO check is failing, please fix and then I can review

cmd/osm-controller/osm-controller.go
```diff
@@ -183,13 +184,11 @@ func main() {
 		}
 	}
 
-	kubeProvider, err := kube.NewClient(kubernetesClient, constants.KubeProviderName, cfg)
```
steeling, line 186: @whitneygriffith let's actually keep this cfg here.

pkg/providers/kube/client.go
```diff
@@ -17,7 +16,7 @@ import (
 )
 
 // NewClient returns a client that has all components necessary to connect to and maintain state of a Kubernetes cluster.
-func NewClient(kubeController k8s.Controller, providerIdent string, cfg configurator.Configurator) (*Client, error) {
```
steeling, line 20: and here

snehachhabria (approved):

pkg/catalog/inbound_traffic_policies_test.go
```diff
@@ -524,6 +526,37 @@ func TestListInboundTrafficPolicies(t *testing.T) {
...
+						fmt.Sprintf("%s.%s.svc.cluster", ms.Name, ms.Namespace),
+						fmt.Sprintf("%s.%s.svc.cluster.%s", ms.Name, ms.Namespace, ms.ClusterDomain),
+						fmt.Sprintf("%s.%s:8888", ms.Name, ms.Namespace),
+						fmt.Sprintf("%s.%s.svc:8888", ms.Name, ms.Namespace),
+						fmt.Sprintf("%s.%s.svc.cluster:8888", ms.Name, ms.Namespace),
+						fmt.Sprintf("%s.%s.svc.cluster.%s:8888", ms.Name, ms.Namespace, ms.ClusterDomain),
+					}, nil).AnyTimes()
+				}
+			}
+
```
shashankram, line 559: There is room for improvement here. Instead of hardcoding expectations for different test scenarios in the test body, make them a part of the `testCase` struct and set expected return types for the individual test cases in the table.

pkg/catalog/inbound_traffic_policies_test.go
```diff
@@ -990,6 +1025,36 @@ func TestListInboundPoliciesForTrafficSplits(t *testing.T) {
...
```
shashankram, line 1057: Same comment as what I suggested earlier regarding not hardcoding test expectations with conditionals in the test body.

pkg/catalog/service.go
```diff
@@ -100,16 +101,15 @@ func (mc *MeshCatalog) getServicesForServiceAccount(sa identity.K8sServiceAccoun
...
-		log.Err(err).Msgf("Error getting ServiceAccounts for Service %s", svc)
-		return nil, err
-	}
-
 	var serviceIdentities []identity.ServiceIdentity
-	for _, svcAccount := range serviceAccounts {
-		serviceIdentity := svcAccount.ToServiceIdentity()
-		serviceIdentities = append(serviceIdentities, serviceIdentity)
+	for _, provider := range mc.serviceProviders {
+		serviceIds, err := provider.ListServiceIdentitiesForService(svc)
```
shashankram, line 106: nit: s/serviceIDs/serviceIdentities

serviceIDs reads `IDs for a serivice`

pkg/catalog/service.go
```diff
@@ -100,16 +101,15 @@ func (mc *MeshCatalog) getServicesForServiceAccount(sa identity.K8sServiceAccoun
...
+		if err != nil {
+			log.Err(err).Msgf("Error getting ServiceIdentities for Service %s", svc)
+			return nil, errServiceNotFound
```
shashankram, line 109: This error is not correct, the service identity error doesn't imply the service was not found.

pkg/providers/kube/fake.go
```diff
@@ -57,19 +63,70 @@ func (f fakeClient) ListEndpointsForIdentity(serviceIdentity identity.ServiceIde
...
+		// Within the same namespace, service name is resolvable to its address
+		hostnames = append(hostnames, fmt.Sprintf("%s:%d", serviceName, port)) // service:port
+	}
+	hostnames = append(hostnames, fmt.Sprintf("%s.%s:%d", serviceName, namespace, port))                   // service.namespace:port
+	hostnames = append(hostnames, fmt.Sprintf("%s.%s.svc:%d", serviceName, namespace, port))               // service.namespace.svc:port
+	hostnames = append(hostnames, fmt.Sprintf("%s.%s.svc.cluster:%d", serviceName, namespace, port))       // service.namespace.svc.cluster:port
+	hostnames = append(hostnames, fmt.Sprintf("%s.%s.svc.cluster.local:%d", serviceName, namespace, port)) // service.namespace.svc.cluster.local:port
+	return hostnames, nil
+}
+
```
shashankram, line 120: Please avoid using fakes with the exact same logic as the actual function being tested. This is not maintainable and will introduce bugs quickly when the logic gets more complex. Instead, use mocks when managing fakes gets out of hand.

Would it be possible to just return hardcoded hostnames here or something the test can set when initializing the fake client's struct?

shashankram (changes requested):

pkg/catalog/service.go
steeling: concretely, let's return the original error here

pkg/providers/kube/fake.go
steeling: the output is dependent on the combination of 2 inputs.. so you could have a double map.. but i'm not sure that is great.

I think this more so highlights the negatives of using fakes in the first place... maybe file a ticket to remove all usage of fakes and mocks altogehter?

at its heart it should probably just be faking the calls to the kubernetes master and nothing else. The k8s client simply has logic and doesn't make any network/sys calls anyways.

steeling: anyways all that to say i don't think this is terrible to have logic twice if the decision has been made to "fake" logic.. why not get closer to to the real thing anyways?

shashankram: We very much need to use mocks and have moved away from using fakes as much as possible.  Just doesn't seem to be a good idea to duplicate actual code into fakes.

steeling: can you provide a scenario that could lead to an issue? If the original changes, then the test will either break, or the potentially the tests weren't robust enough and they don't break (either way nothing detrimental has occurred). In fact it provides more flexibility than hardcoding values

And 2 cents on mocks: only network/syscalls should be mocked. It would offer much more robust testing if there was no mock provider or meshcatalog

shashankram: > can you provide a scenario that could lead to an issue? If the original changes, then the test will either break, or the potentially the tests weren't robust enough and they don't break (either way nothing detrimental has occurred). In fact it provides more flexibility than hardcoding values
> 

It is obvious that copying actual code into fakes is a maintenance hassle. When the actual code changes, if tests don't offer 100% coverage it is likely existing tests could pass while the fakes rely on an implementation that is no longer up to date. 

> And 2 cents on mocks: only network/syscalls should be mocked. It would offer much more robust testing if there was no mock provider or meshcatalog

This is something that I disagree with. If done right, mocking allows unit testing pieces of code independently of external components that the test doesn't care about. Inputs that depend on external components (could be a different interface, or whatever the test doesn't care about) can be easily mocked allowing the function to be well tested. There are benefits to using fakes for simple scenarios but it becomes unmanageable when a lot of code gets hardcoded into fakes (which I am not a fan of). If we want to leverage fakes, we should make them configurable, just like k8s fake clientset used in tests. We have files under `tests/scenarios/` that only use the fake clientset provided by k8s, but they are also harder to write and unit test because it involves initializing and testing the entire control plane components and not just a single function.

shashankram (approved):
@whitneygriffith we can address the test enhancement in a subsequent change. Would you mind rebasing before merging if your branch hasn't been rebased, looks okay to merge otherwise.

pkg/providers/kube/fake.go
whitneygriffith: Spoke  offline and resolved it

PR #3626, opened 2021-06-18
ref(catalog): improve performance of ListAllowedEndpointsForService()

allenlsy:

**Description**:

Resolves issue #3190 

* Use map to iterate over endpoint list, in order to improve performace
* Rename `ListAllowedEndpointsForService()` to `ListEndpointsForServiceIdentity()`
* Rename `destSvcAccount` variable to `destinationServiceIdentity`

**Affected area**:
| Control Plane              | ✅ |
| Performance                | ✅ |

No

1. Is this a breaking change?

No

### DIFF
--- a/pkg/catalog/endpoint.go
+++ b/pkg/catalog/endpoint.go
@@ -38,17 +38,23 @@ func (mc *MeshCatalog) GetResolvableServiceEndpoints(svc service.MeshService) ([
 	return endpoints, nil
 }
 
-// ListAllowedEndpointsForService returns only those endpoints for a service that belong to the allowed outbound service accounts
+// ListEndpointsForServiceIdentity returns only those endpoints for a allowed outbound service accounts
 // for the given downstream identity
 // Note: ServiceIdentity must be in the format "name.namespace" [https://github.com/openservicemesh/osm/issues/3188]
-func (mc *MeshCatalog) ListAllowedEndpointsForService(downstreamIdentity identity.ServiceIdentity, upstreamSvc service.MeshService) ([]endpoint.Endpoint, error) {
+func (mc *MeshCatalog) ListEndpointsForServiceIdentity(downstreamIdentity identity.ServiceIdentity, upstreamSvc service.MeshService) ([]endpoint.Endpoint, error) {
 	outboundEndpoints, err := mc.listEndpointsForService(upstreamSvc)
 	if err != nil {
 		log.Error().Err(err).Msgf("Error looking up endpoints for upstream service %s", upstreamSvc)
 		return nil, err
 	}
+	outboundEndpointsSet := make(map[string][]endpoint.Endpoint)
+	for _, ep := range outboundEndpoints {
+		ipStr := ep.IP.String()
+		outboundEndpointsSet[ipStr] = append(outboundEndpointsSet[ipStr], ep)
+	}
+	log.Info().Msgf("outbound endpoints: %v", outboundEndpointsSet)
 
-	destSvcAccounts, err := mc.ListAllowedOutboundServiceIdentities(downstreamIdentity)
+	destSvcIdentities, err := mc.ListAllowedOutboundServiceIdentities(downstreamIdentity)
 	if err != nil {
 		log.Error().Err(err).Msgf("Error looking up outbound service accounts for downstream identity %s", downstreamIdentity)
 		return nil, err
@@ -57,16 +63,20 @@ func (mc *MeshCatalog) ListAllowedEndpointsForService(downstreamIdentity identit
 	// allowedEndpoints comprises of only those endpoints from outboundEndpoints that matches the endpoints from listEndpointsForServiceIdentity
 	// i.e. only those interseting endpoints are taken into cosideration
 	var allowedEndpoints []endpoint.Endpoint
-	for _, destSvcAccount := range destSvcAccounts {
-		podEndpoints := mc.listEndpointsForServiceIdentity(destSvcAccount)
-		for _, ep := range outboundEndpoints {
-			for _, podIP := range podEndpoints {
-				if ep.IP.Equal(podIP.IP) {
-					allowedEndpoints = append(allowedEndpoints, ep)
-				}
+	for _, destSvcIdentity := range destSvcIdentities {
+		log.Info().Msgf("ups svc endpoints: %v, %v", destSvcIdentity, mc.listEndpointsForServiceIdentity(destSvcIdentity))
+
+		for _, ep := range mc.listEndpointsForServiceIdentity(destSvcIdentity) {
+			epIPStr := ep.IP.String()
+			// check if endpoint IP is allowed
+			if _, ok := outboundEndpointsSet[epIPStr]; ok {
+				// add all allowed endpoints on the pod to result list
+				// TODO(allenlsy): only allow endpoint with matching port
+				allowedEndpoints = append(allowedEndpoints, outboundEndpointsSet[epIPStr]...)
 			}
 		}
 	}
+
 	return allowedEndpoints, nil
 }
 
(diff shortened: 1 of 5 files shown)


pkg/catalog/endpoint.go
```diff
@@ -38,17 +38,21 @@ func (mc *MeshCatalog) GetResolvableServiceEndpoints(svc service.MeshService) ([
...
 // for the given downstream identity
 // Note: ServiceIdentity must be in the format "name.namespace" [https://github.com/openservicemesh/osm/issues/3188]
-func (mc *MeshCatalog) ListAllowedEndpointsForService(downstreamIdentity identity.ServiceIdentity, upstreamSvc service.MeshService) ([]endpoint.Endpoint, error) {
+func (mc *MeshCatalog) ListEndpointsForServiceIdentity(downstreamIdentity identity.ServiceIdentity, upstreamSvc service.MeshService) ([]endpoint.Endpoint, error) {
 	outboundEndpoints, err := mc.listEndpointsForService(upstreamSvc)
 	if err != nil {
 		log.Error().Err(err).Msgf("Error looking up endpoints for upstream service %s", upstreamSvc)
 		return nil, err
 	}
+	outboundEndpointsIPSet := make(map[string]endpoint.Endpoint)
```
shashankram, line 50: Use a mapset here instead?
You could use the return value from `set.Add()` to determine if an add was successful, in which case the endpoint can be appended to the final slice.

pkg/catalog/endpoint.go
```diff
@@ -38,17 +38,21 @@ func (mc *MeshCatalog) GetResolvableServiceEndpoints(svc service.MeshService) ([
...
+	for _, ep := range outboundEndpoints {
+		outboundEndpointsIPSet[ep.IP.String()] = ep
```
shashankram, line 52: Isn't this logic flawed, where an endpoint's port is disregarded? Multiple endpoints with the same IP but different ports constitute multiple endpoints, but with this logic they all map to the same key.

shashankram (changes requested):

pkg/catalog/endpoint.go
allenlsy: The current logic is also only checking the IP address: https://github.com/openservicemesh/osm/issues/3190#L63

But I agree that port should be considered during the allowing process.

shashankram: It seems like the current logic is matching on pod IP only to filer out the endpoints corresponding to the pod. But this logic is disregarding endpoints that have the same port completely.

allenlsy: I have another question, why do we only compare IP addresses in the first place?

Assume 3 services, down-1, up-1, up-2. 

* down-1 ---denied--> up-1, which has endpoint 1.2.3.4:8000
* down-1 ---allowed-----> up-2, which has endpoint 1.2.3.4:9000

Thus, suppose we invoke this method `ListAllowedEndpointsForService(down-1, up-1)`, we expect it to return an empty list. In the current implementation [here](https://github.com/openservicemesh/osm/blob/main/pkg/catalog/endpoint.go#L64-L66), in the 3-level nested loop:

* `destSvcAccounts` => [up-2]
* `outboundEndpoints` => 1.2.3.4:8000 (endpoint of up-1)
* `podEndpoints` => [1.2.3.4:9000] (endpoints of up-2)

Because their IP matches, the `allowedEndpoints` => [1.2.3.4:8000]. This makes up-1 allowed. So in this example, even the current implementation is wrong. I think the current implementation has an assumption that different services will not have pods with the same IP, thus we only compared IP addresses.

shashankram: Your understanding is correct. Why not simply use a mapset for storing the endpoints and doing `podEndpoints.Intersect(allowedEndpoints)`?

pkg/catalog/endpoint.go
allenlsy: I didn't go with the mapset approach. The key of mapset has to be hashable, while right now endpoint.Endpoint is not.

shashankram (approved):
Could you also add tests to cover scenarios where there are duplicate IP/port combinations for the endpoints, and simulate the bug that existed previously which this change addresses.

pkg/catalog/endpoint_test.go
```diff
@@ -129,6 +129,28 @@ func TestListAllowedEndpointsForService(t *testing.T) {
 				Port: endpoint.Port(tests.ServicePort),
 			}},
 		},
+		{
+			name: `Traffic target defined for bookstore ServiceAccount.
+			This service account has bookstore-v1 bookstore-v2 services,
+			Bookstore-v2 pod has service account bookstore-v2 which is allowed in the traffic target, while bookstore-v1 is denied.
+			Hence, when asked about endpoints of upstream service bookstore-v1, it returns an empty list. This holds even if endpoint of v1 and v2 has the same IP address.`,
```
shashankram, line 136: Use a shorter test name, and add the rest of the description as a comment.

allenlsy: Will make it a bit shorter, but not as short as other one line names. This follows the style from other cases in this test.

pkg/catalog/endpoint.go
```diff
@@ -38,33 +38,40 @@ func (mc *MeshCatalog) GetResolvableServiceEndpoints(svc service.MeshService) ([
...
-	destSvcAccounts, err := mc.ListAllowedOutboundServiceIdentities(downstreamIdentity)
+	destSvcIdentities, err := mc.ListAllowedOutboundServiceIdentities(downstreamIdentity)
 	if err != nil {
 		log.Error().Err(err).Msgf("Error looking up outbound service accounts for downstream identity %s", downstreamIdentity)
 		return nil, err
 	}
 
 	// allowedEndpoints comprises of only those endpoints from outboundEndpoints that matches the endpoints from listEndpointsForServiceIdentity
 	// i.e. only those interseting endpoints are taken into cosideration
+	destPodEndpointsSet := make(map[string]endpoint.Endpoint)
```
eduser25, line 63: you are not using `destPodEndpointsSet` to really lookup anything (you end up walking it), so this map could just remain as a slice and avoid the N walk?

pkg/catalog/endpoint.go
allenlsy: Turns out we cannot use port number as key in the hash map to match endpoints. `ListEndpointsForIdentity()` returns endpoints without port, that causes matching fail in some cases. It's detected by our CI this time.

https://github.com/openservicemesh/osm/blob/0f792a579b05de5221d81e88399a4834bd0e7c55/pkg/providers/kube/client.go#L71-L98

pkg/catalog/endpoint.go
allenlsy: Right. `destPodEndpointSet` is removed.

pkg/catalog/endpoint.go
```diff
@@ -57,16 +63,20 @@ func (mc *MeshCatalog) ListAllowedEndpointsForService(downstreamIdentity identit
...
-				}
+	for _, destSvcIdentity := range destSvcIdentities {
+		log.Info().Msgf("ups svc endpoints: %v, %v", destSvcIdentity, mc.listEndpointsForServiceIdentity(destSvcIdentity))
+
+		for _, ep := range mc.listEndpointsForServiceIdentity(destSvcIdentity) {
+			epIPStr := ep.IP.String()
+			// check if endpoint IP is allowed
+			if _, ok := outboundEndpointsSet[epIPStr]; ok {
+				// add all allowed endpoints on the pod to result list
+				// TODO(allenlsy): only allow endpoint with matching port
```
shashankram, line 74: Please open a bug to track this.

shashankram (approved):

pkg/catalog/endpoint.go
allenlsy: Issue #3680

eduser25 (approved):

snehachhabria (approved):

PR #3635, opened 2021-06-21
feat(pkg/kuberentes) move CRD clients to be under the kubernetes client.

steeling:
All clients should be a part of the kubernetes controller object

-->
**Affected area**:

Refactor

No

1. Is this a breaking change?
No


cmd/osm-injector/osm-injector.go
```diff
@@ -141,8 +142,10 @@ func main() {
 	}
 	log.Info().Msgf("Initial MeshConfig %s: %v", osmMeshConfigName, meshConfig)
 
+	policyClient := policyV1alpha1Client.NewForConfigOrDie(kubeConfig)
```
shashankram, line 145: osm-injector doesn't need this. Let's just pass `nil` to be explicit with a comment?

shashankram (commented):
@steeling what's the motivation for this change? Would you mind describing this in the commit and PR description?

This goes against the existing design of carving out different functionalities into their own packages, which makes the code modular. I understand why you might have wanted to implement a monolithic client for k8s resources, but I'd like this to be explicitly stated so we can review whether this change makes sense or not.

cmd/osm-injector/osm-injector.go
snehachhabria: I propose we do not merge code refactors while we are working on a dedicated feature that is behind a feature flag. We can create an issue in OSM for improvements like these and address them in the future but lets scope the work such that it becomes to deliver on what we have planned.

Additional improvements are always welcome but they can be done in the future, with the correct tracking in place

steeling: @shashankram done, and split up the interfaces such the injector gets the minimal set.

```
// Controller implements the full set of controller methods, including those for CRD's.
type Controller interface {
	EgressController
	BasicController
}
```

@snehachhabria I'm fine with whatever, although if we can separately agree that this is a change in the right direction, then it doesn't need to be considered in concert with the multi cluster work

shashankram:
@steeling, this change doesn't seem to be consistent in its motive. Either we collapse all informers into a single package (policy, smi, ingress, core kubernetes) in `pkg/kubernetes`, or we keep them as is as separate packages. 

The benefit of having separate packages:
- Modularity
- Independent interfaces
- Limits functionality scope (ex. a client can have access to ingress/egress but not all k8s resources, which won't be possible with a single monolithic interface)

Cons:
- Duplicate boilerplate code across packages to initialize informers
- Needing to pass multiple interfaces around to the catalog

Were there other cons to keeping the packages separate?

steeling:
> @steeling, this change doesn't seem to be consistent in its motive. Either we collapse all informers into a single package (policy, smi, ingress, core kubernetes) in `pkg/kubernetes`, or we keep them as is as separate packages.
> 
> The benefit of having separate packages:
> 
> * Modularity
> * Independent interfaces
> * Limits functionality scope (ex. a client can have access to ingress/egress but not all k8s resources, which won't be possible with a single monolithic interface)
> 
> Cons:
> 
> * Duplicate boilerplate code across packages to initialize informers
> * Needing to pass multiple interfaces around to the catalog
> 
> Were there other cons to keeping the packages separate?

Wasn't aware of the other ones. I'm happy to go with whatever direction, either migrating all, or keeping as is. Let me know what you'd prefer

Issue #3164, opened 2021-04-14
Track egress traffic with WASM metrics extension


nojnhuh:
**Please describe the Improvement and/or Feature Request**

_Originally captured in #1697_

Currently, custom metrics captured by the WASM extension for SMI metrics are only recorded for in-mesh traffic. The same information like success/failure rates and latency per source/destination pod/workload/namespace would also be useful for egress traffic.

**Scope (please mark with X where applicable)**
- New Functionality          [ ]
- Install                    [ ]
- SMI Traffic Access Policy  [ ]
- SMI Traffic Specs Policy   [ ]
- SMI Traffic Split Policy   [ ]
- Permissive Traffic Policy  [ ]
- Ingress                    [ ]
- Egress                     [X]
- Envoy Control Plane        [ ]
- CLI Tool                   [ ]
- Metrics                    [X]
- Certificate Management     [ ]
- Sidecar Injection          [ ]
- Logging                    [ ]
- Debugging                  [ ]
- Tests                      [ ]
- CI System                  [ ]
- Project Release            [ ]

**Possible use cases**

* Checking success/failure rates and latency of egress requests per namespace/pod/workload.

nojnhuh:
Depending on what we need, there may be existing metrics emitted by Envoy that capture this. The main thing to verify is if we can check request success rates and durations for egress traffic separate from in-mesh traffic. Querying beyond that (like egress to particular hosts) may not be possible with Envoy's built-in metrics. If Envoy's existing metrics are enough then I think the most we should do for this issue is document those.

If we find we need to do this in WASM, I can think of a couple ways to accomplish this. One is to add something like a boolean `is_egress` label on the `osm_request_total` and `osm_request_duration_ms` metrics. Another is to make the `destination_namespace` label value only on egress traffic be something containing a `.` (since `.` isn't allowed in k8s namespace names, that would never cause a collision with an actual namespace name). I'm not married to any particular approach, just throwing out ideas.

PR #3651, opened 2021-06-23
fix(pkg/envoy/cds): configure egress Envoy cluster altStatName for prometheus

jaellio:
**Description**:

This change formats the altStatName of Envoy clusters created based on the applied egress policies so that egress traffic can be observed with Prometheus. The change only applies to egress traffic enabled using egress policies, not globally enabled egress. The AltStatName of the cluster is formatted so that the stats for the cluster are grouped with Envoy metrics we are already scraping for. Prior to this change, the egress cluster name would be interpreted incorrectly due to periods and colons in the name and the stats would be grouped under a unique metric. We were not scraping for this metric and it would be difficult for customers to interpret.

Addresses issue #3164.

This change can be demonstrated using the [egress policy demo](https://docs.openservicemesh.io/docs/tasks/traffic_management/demos/egress_policy_demo/). Metrics can be enabled for the curl namespace using `osm metrics enable --namespace curl`. The stats for egress traffic between the curl pod and httpbin.org  should be visible at stats/prometheus on the envoyAdmin.

```
# TYPE envoy_cluster_upstream_rq counter
envoy_cluster_upstream_rq{envoy_response_code="200",envoy_cluster_name="envoy-metrics-cluster"} 163
envoy_cluster_upstream_rq{envoy_response_code="200",envoy_cluster_name="httpbin_org_80"} 2
envoy_cluster_upstream_rq{envoy_response_code="200",envoy_cluster_name="osm-controller"} 2
envoy_cluster_upstream_rq{envoy_response_code="503",envoy_cluster_name="osm-controller"} 
```

Prior to this change, stats for upstream requests for the httpbin.org:80 cluster would be visible as the following:

```
# TYPE envoy_cluster_upstream_rq counter
envoy_cluster_upstream_rq{envoy_response_code="200",envoy_cluster_name="envoy-metrics-cluster"} 163
envoy_cluster_upstream_rq{envoy_response_code="200",envoy_cluster_name="osm-controller"} 2
envoy_cluster_upstream_rq{envoy_response_code="503",envoy_cluster_name="osm-controller"} 

# TYPE envoy_cluster_org_80_upstream_rq counter
envoy_cluster_upstream_rq{envoy_response_code="200",envoy_cluster_name="httpbin"} 2
```

The metric name `envoy_cluster_org_80_upstream_rq` does not intuitively represent the envoy cluster, httpbin.org:80. By default, the altStatName for an Envoy cluster is the cluster name. In this case, the cluster name includes periods and colon. Envoy interprets colons as underscores and the period as a delimiter when formatting the metrics exposed at stats/prometheus. This causes the naming issues seen above. Additionally, the metric `envoy_cluster_org_80_upstream_rq` is not scaped by prometheus.

With this change, egress traffic stats can be queried as follows:

MicrosoftTeams-image

**Note:**

- An egress prefix could be used to specify that the cluster is an egress cluster.
- There may be more characters in hostnames that could cause the naming issues seen above.

**Affected area**:
| Egress                     | [ X ] |
| Observability              | [ X ] |

### DIFF
--- a/pkg/envoy/cds/cluster.go
+++ b/pkg/envoy/cds/cluster.go
@@ -1,6 +1,7 @@
 package cds
 
 import (
+	"strings"
 	"time"
 
 	xds_cluster "github.com/envoyproxy/go-control-plane/envoy/config/cluster/v3"
@@ -24,6 +25,9 @@ const (
 	// clusterConnectTimeout is the timeout duration used by Envoy to timeout connections to the cluster
 	clusterConnectTimeout = 1 * time.Second
 )
+
+// replacer used to configure an Envoy cluster's altStatName
+var replacer = strings.NewReplacer(".", "_", ":", "_")
 
 type clusterOptions struct {
 	permissive bool
@@ -230,7 +234,7 @@ func getDNSResolvableEgressCluster(config *trafficpolicy.EgressClusterConfig) (*
 
 	return &xds_cluster.Cluster{
 		Name:           config.Name,
-		AltStatName:    config.Name,
+		AltStatName:    formatAltStatNameForPrometheus(config.Name),
 		ConnectTimeout: ptypes.DurationProto(clusterConnectTimeout),
 		ClusterDiscoveryType: &xds_cluster.Cluster_Type{
 			Type: xds_cluster.Cluster_STRICT_DNS,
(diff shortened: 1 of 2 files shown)


pkg/envoy/lds/egress.go
```diff
@@ -17,8 +17,11 @@ import (
 )
 
 const (
+	// @prometheus-metric
 	egressHTTPFilterChainPrefix = "egress-http"
-	egressTCPFilterChainPrefix  = "egress-tcp"
+
+	// @prometheus-metric
+	egressTCPFilterChainPrefix = "egress-tcp"
```
shashankram, line 24: Are these prefixes used to scrape the egress metrics?

eduser25 (approved):
lgtm for the part we discussed

pkg/envoy/lds/egress.go
jaellio: I don't believe so. I must have misunderstood your earlier message.

shashankram: Let's remove it in that case.

snehachhabria (approved):

PR #3660, opened 2021-06-24
chore(release): update charts and tags for v0.9.1-rc.1

jaellio:

- [x] I have cherry-picked any changes [labelled](https://github.com/openservicemesh/osm/labels) `needs-cherry-pick-vX.Y.Z`
- [x] I have made all of the following version and patch updates:
  1. Updated the container image tag in charts/osm/values.yaml
  2. Updated the chart **and** app version in charts/osm/Chart.yaml
  3. Updated the image tags used in the demo manifests
  4. Regenerated the Helm chart README.md

- [x] I have checked that the base branch for this PR is correct as defined by the release guide

Is this a release candidate?
Yes

### DIFF
--- a/pkg/configurator/methods_test.go
+++ b/pkg/configurator/methods_test.go
@@ -246,15 +246,15 @@ func TestCreateUpdateConfig(t *testing.T) {
 			name:                  "GetInitContainerImage",
 			initialMeshConfigData: &v1alpha1.MeshConfigSpec{},
 			checkCreate: func(assert *tassert.Assertions, cfg Configurator) {
-				assert.Equal("openservicemesh/init:v0.9.0", cfg.GetInitContainerImage())
+				assert.Equal("openservicemesh/init:v0.9.1-rc.1", cfg.GetInitContainerImage())
 			},
 			updatedMeshConfigData: &v1alpha1.MeshConfigSpec{
 				Sidecar: v1alpha1.SidecarSpec{
-					InitContainerImage: "openservicemesh/init:v0.8.2",
+					InitContainerImage: "openservicemesh/init:v0.9.0",
 				},
 			},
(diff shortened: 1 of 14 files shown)


pkg/configurator/methods_test.go
```diff
@@ -246,15 +246,15 @@ func TestCreateUpdateConfig(t *testing.T) {
 			name:                  "GetInitContainerImage",
 			initialMeshConfigData: &v1alpha1.MeshConfigSpec{},
 			checkCreate: func(assert *tassert.Assertions, cfg Configurator) {
-				assert.Equal("openservicemesh/init:v0.9.0", cfg.GetInitContainerImage())
+				assert.Equal("openservicemesh/init:v0.9.1-rc.1", cfg.GetInitContainerImage())
 			},
 			updatedMeshConfigData: &v1alpha1.MeshConfigSpec{
 				Sidecar: v1alpha1.SidecarSpec{
-					InitContainerImage: "openservicemesh/init:v0.8.2",
+					InitContainerImage: "openservicemesh/init:v0.9.0",
```
snehachhabria, line 253: Shouldn't this also be v0.9.1-rc.1 ?

nojnhuh: This value doesn't need to be updated each time since it doesn't relate to the current version at all.

I'd be in favor of (outside of this PR) changing this to something like `TEST` just to make it obvious that the exact value doesn't really matter as long as it's different from the default.

nojnhuh (approved):

pkg/configurator/methods_test.go
jaellio: This method is testing updates to the mesh config. I don't think the version number matters as long as it is different from the above version number (in this case v0.9.1-rc.1)

snehachhabria (approved):

nojnhuh (approved):

Issue #3396, opened 2021-05-24
CRD upgrades


snehachhabria:
**Please describe the Improvement and/or Feature Request**
We have two new CRD's in osm, one for policy and the other for meshconfig. We need to come up with a generic way/story to version and upgrade CRD's.

**Scope (please mark with X where applicable)**
- New Functionality          [ ]
- Install                    [ ]
- SMI Traffic Access Policy  [ ]
- SMI Traffic Specs Policy   [ ]
- SMI Traffic Split Policy   [ ]
- Permissive Traffic Policy  [ ]
- Ingress                    [ ]
- Egress                     [ ]
- Envoy Control Plane        [ ]
- CLI Tool                   [ ]
- Metrics                    [ ]
- Certificate Management     [ ]
- Sidecar Injection          [ ]
- Logging                    [ ]
- Debugging                  [ ]
- Tests                      [ ]
- CI System                  [ ]
- Demo                       [ ]
- Project Release            [ ]

**Possible use cases**

snehachhabria:
This feature is currently behind a feature flag `enableCRDConverter` and no e2e tests have been added as it is blocked on : https://github.com/openservicemesh/osm/issues/3716

snehachhabria:
By updating the nodes on CI in PR : https://github.com/openservicemesh/osm/pull/3813 , the CRD converter has now been enabled by default in OSM : https://github.com/openservicemesh/osm/pull/3838

PR #3676, opened 2021-06-26
feat(upgrade): Add helm hooks to manage CRD upgrades

snehachhabria:
**Description**:

This commit allows OSM to manage the lifecycle of CRD's by adding a pre-install and pre-upgrade helm hook to update the CRD's. The cleanup hook (post-delete helm hook)  has also been modified to ensure that all the CRD's are deleted from the cluster once OSM is uninstalled.

A custom image is built in OSM that contains all the CRD's needed by a particular version and the pod that runs as a pre-install, pre-upgrade helm hook re-applies the CRD's, thus ensuring all the CRD's are upto date. 

Part of #3396

**Affected area**:


shashankram (approved):

charts/osm/templates/cleanup-hook.yaml
```diff
@@ -9,23 +118,26 @@ metadata:
...
+  containers:
+    - name: garbage-collector
+      image: bitnami/kubectl
+      imagePullPolicy: IfNotPresent
+      command:
+        - sh
+        - -c
+        - >
+         kubectl delete --ignore-not-found meshconfig -n '{{ include "osm.namespace" . }}' osm-mesh-config;
+         kubectl delete --ignore-not-found meshconfig -n '{{ include "osm.namespace" . }}' osm-mesh-config;
```
allenlsy, line 135: Why we delete it twice? Do you mean preset-mesh-config?

snehachhabria: it was accidental I removed it.

eduser25 (approved):

ritazh:
Great PR @snehachhabria! 
> The cleanup hook (post-delete helm hook) has also been modified to ensure that all the CRD's are deleted from the cluster once OSM is uninstalled.

Wouldn't this also remove user-created custom resources?

snehachhabria:
> Great PR @snehachhabria!
> 
> > The cleanup hook (post-delete helm hook) has also been modified to ensure that all the CRD's are deleted from the cluster once OSM is uninstalled.
> 
> Wouldn't this also remove user-created custom resources?

@ritazh that's right, the reason for this is because we have implemented logic to patch a conversion webhook to each of the CRDs (primarily to support older custom resources after OSM has been updated to use the newer CRD). The conversion webhook ties the CRD very tightly to OSM and if we leave the CRD after an uninstall it becomes unusable as k8s will complain about a missing webhook service.

We have an issue to fix this : https://github.com/openservicemesh/osm/issues/3760. I am thinking that on uninstall if the customer explicitly specified not to delete the CRD we will re-patch it to remove the conversion webhook

PR #3684, opened 2021-06-28
[WIP] test(cmd/cli): add test for proxy get run cmd

SanyaKochhar:
**Description**:

**Affected area**:

### DIFF
--- a/cmd/cli/proxy_get_test.go
+++ b/cmd/cli/proxy_get_test.go
@@ -1,12 +1,19 @@
 package main
 
 import (
+	"bytes"
+	"context"
 	"fmt"
+	"github.com/openservicemesh/osm/pkg/apis/config/v1alpha1"
+	"os"
 	"testing"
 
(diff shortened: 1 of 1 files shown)


shashankram:
Instead of writing a unit test for the CLI commands that rely on network connectivity to an endpoint (port-forward to a remote pod and executing http.GET in this case), there is a lot more value in writing an e2e test and verifying these commands work as expected. The unit test will partially increase coverage, but by no means will actually test the expected behavior. To unit test this, you would need to set up an http server (could be a fake server using httptest) mimicking the Envoy admin endpoint on port 15000 that implements http handlers, and get the portforward to work with a fake.

SanyaKochhar:
@shashankram that makes sense -- I haven't been having much luck with trying to unit test the http get bit and the code coverage increases for the rest of the lines here are minute

snehachhabria:
stale PR, closing

PR #3685, opened 2021-06-28
feat(multicluster): add weight to RemoteCluster

allenlsy:

**Description**:

Add `weight` attribute to remote cluster spec. This PR prepares for implementing locality based routing. 

**Affected area**:
| New Functionality          | X |

No

1. Is this a breaking change?
No

### DIFF
--- a/pkg/apis/config/v1alpha1/multi_cluster_service.go
+++ b/pkg/apis/config/v1alpha1/multi_cluster_service.go
@@ -47,6 +47,9 @@ type ClusterSpec struct {
 
 	// Name defines the name of the remote cluster.
 	Name string `json:"name,omitempty"`
+
+	// Weight defines the load balancing weight of the remote cluster
+	Weight int `json:"weight,omitempty"`
 }
 
 // PortSpec contains information on service's port.
--- a/charts/osm/crds/config_multicluster_service.yaml
+++ b/charts/osm/crds/config_multicluster_service.yaml
@@ -81,6 +81,11 @@ spec:
                       name:
                         description: Name of the remote cluster
                         type: string
+                      weight:
+                        description: Load balancing weight of the remote cluster
+                        type: integer
+                        minimum: 0
+                        default: 0
                       certificate:
                         description: mTLS certificates (optional)
                         type: string


charts/osm/crds/multiclusterservice.yaml
```diff
@@ -58,6 +58,9 @@ spec:
                       name:
                         description: Name of the remote cluster
                         type: string
+                      weight:
+                        description: Load balancing weight of the remote cluster in percentage
+                        type: integer
```
snehachhabria, line 63: should we add a check to ensure its a positive integer and if not specified assign a default of 0 ?

allenlsy: One correction I added: the weight here is not percentage.

charts/osm/crds/multiclusterservice.yaml
```diff
@@ -58,6 +58,12 @@ spec:
                       name:
                         description: Name of the remote cluster
                         type: string
+                      weight:
+                        description: Load balancing weight of the remote cluster
+                        type: integer
+                        minimum: 1
```
snehachhabria, line 64: this doesn't seem right, the minimum should be 0

charts/osm/crds/multiclusterservice.yaml
```diff
@@ -58,6 +58,12 @@ spec:
...
+                        maximum: 10000
```
snehachhabria, line 65: Any particular reason why this is capped at 10000 ?

charts/osm/crds/multiclusterservice.yaml
allenlsy: 0 means no weight at all. If that is the case, why we add that cluster?

charts/osm/crds/multiclusterservice.yaml
allenlsy: No, I confused myself with another number.

Anyway, in Envoy, there is no limit on the load balancing weight. If we follow convention, we can do the same. https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/endpoint/v3/endpoint_components.proto#envoy-v3-api-field-config-endpoint-v3-localitylbendpoints-load-balancing-weight

charts/osm/crds/multiclusterservice.yaml
snehachhabria: yes, there can be scenarios where the cluster is mentioned but there's no traffic yet

charts/osm/crds/multiclusterservice.yaml
allenlsy: Updated minimum to 0, and no limit on maximum, as long as it's in the range of integer

ksubrmnn (approved):

pkg/apis/config/v1alpha1/multi_cluster_service.go
```diff
@@ -39,6 +39,12 @@ type ClusterSpec struct {
 
 	// Name defines the name of the remote cluster.
 	Name string `json:"name,omitempty"`
+
+	// Weight defines the load balancing weight of the remote cluster
+	Weight int `json:"weight,omitempty"`
+
+	// Certificate defines the mTLS certificate of the remote cluster
+	Certificate string `json:"certificate,omitempty"`
```
shashankram, line 47: This PR does not mention anything regarding certificates. Please drop this change or describe in the commit why this is required.

shashankram (changes requested):

snehachhabria (changes requested):
Agree with @shashankram lets please rebase based off the latest multicluster changes and determine if this is needed. Also for any changes in multicluter, going forward lets have a proposal and get it reviewed. Post the proposal approval we can work on a POC to validate the proposal before we commit changes to the repository

pkg/apis/config/v1alpha1/multi_cluster_service.go
allenlsy: Certificate field was originally in the CRD, but not in the Go struct. I think this was a mistake for leaving it out. Thus adding it back in this PR.

shashankram: Where is the certificate being used? Is including it in the CRD necessary? Like I mentioned already, the commit message does not talk about the certificate, so seeing this here is quite confusing.

allenlsy: I haven't checked if the `certificate` is used anywhere in the code base. If we find it's not needed, we can remove it in another PR.

shashankram: If you aren't sure it's needed, there's no reason to include it in this change.

snehachhabria: the crd is slate, the current crd doesn't have `certificate`. Please rebase changes and share a proposal as to why weight needs to be added for the remote cluster

snehachhabria (approved):

PR #3692, opened 2021-06-29
feat(pkg/envoy/lds): Switch the gateway listener to proxy mode

steeling:
Proxy mode should allow the gateway to pick up the original
connection string

**Description**:

**Affected area**:
| New Functionality          | [ x] |

No

1. Is this a breaking change?
No

### DIFF
--- a/pkg/envoy/lds/inmesh.go
+++ b/pkg/envoy/lds/inmesh.go
@@ -116,6 +116,18 @@ func (lb *listenerBuilder) getInboundMeshHTTPFilterChain(proxyService service.Me
 	}
 
 	filterchainName := fmt.Sprintf("%s:%s:%d", inboundMeshHTTPFilterChainPrefix, proxyService, servicePort)
+
+	serverNames := []string{proxyService.ServerName()}
+	// The sidecar needs to be aware of the names a remote cluster might refer to this service as.
+	if lb.cfg.GetFeatureFlags().EnableMulticlusterMode {
+		serverNames = append(serverNames,
+			service.MeshService{
+				Name:          proxyService.Name,
+				Namespace:     proxyService.Namespace,
+				ClusterDomain: constants.ClusterDomain(lb.cfg.GetClusterDomain())}.ServerName(),
+		)
+	}
+
 	filterChain := &xds_listener.FilterChain{
 		Name:    filterchainName,
 		Filters: filters,
@@ -129,7 +141,7 @@ func (lb *listenerBuilder) getInboundMeshHTTPFilterChain(proxyService service.Me
 
 			// The ServerName is the SNI set by the downstream in the UptreamTlsContext by GetUpstreamTLSContext()
 			// This is not a field obtained from the mTLS Certificate.
-			ServerNames: []string{proxyService.ServerName()},
+			ServerNames: serverNames,
 
 			// Only match when transport protocol is TLS
 			TransportProtocol: envoy.TransportProtocolTLS,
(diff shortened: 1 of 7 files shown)


pkg/envoy/lds/ingress.go
```diff
@@ -69,6 +69,7 @@ func (lb *listenerBuilder) newIngressHTTPFilterChain(cfg configurator.Configurat
 func (lb *listenerBuilder) getIngressFilterChains(svc service.MeshService) []*xds_listener.FilterChain {
 	var ingressFilterChains []*xds_listener.FilterChain
 
+	// TODO(steeling): shouldn't this call GetPortToProtocolMappingForService?
```
shashankram, line 72: Remove this, we discussed offline why the existing code is correct.

pkg/envoy/lds/inmesh.go
```diff
@@ -164,6 +171,12 @@ func (lb *listenerBuilder) getInboundMeshTCPFilterChain(proxyService service.Mes
 		return nil, err
 	}
 
+	serverNames, err := lb.meshCatalog.GetServiceHostnames(proxyService, service.RemoteCluster)
+	if err != nil {
+		// this error only affects multi cluster, so we don't return early here.
+		log.Error().Err(err).Msgf("Error fetching remote cluster service names for %s", proxyService)
+	}
```
shashankram, line 178: Please wrap this change with `if proxy.Kind == gateway`. We don't want to be exercising multicluster related code for inmesh app<-->app connectivity.

shashankram (changes requested):

pkg/envoy/lds/ingress.go
steeling: yup, good call

pkg/envoy/lds/inmesh.go
steeling: this actually is for the non-gateways, so that they can get the correct reference. Added a comment, but also wrapped in the ENabledMultiCluster feature flag

pkg/envoy/lds/inmesh.go
```diff
@@ -116,6 +116,15 @@ func (lb *listenerBuilder) getInboundMeshHTTPFilterChain(proxyService service.Me
 	}
 
 	filterchainName := fmt.Sprintf("%s:%s:%d", inboundMeshHTTPFilterChainPrefix, proxyService, servicePort)
+
+	serverNames := []string{proxyService.ServerName()}
+	// The sidecar needs to be aware of the names a remote cluster might refer to this service as.
+	if lb.cfg.GetFeatureFlags().EnableMulticlusterMode {
```
shashankram, line 122: Could you add a `proxyKind` field to the `listenerBuilder` struct and add an additional check of `if lb.proxyKind == envoy.KindGateway` in L122.

We don't want the additional server name processing for in-mesh clients that are not a gateway.

steeling: the gateway does TLS passthrough to the upstream service. The downstream service refers to the upstream via:

service.namespace.svc.cluster.cluster-id, so the upstream service will need the additional server name to process. The only other alternative is to not do SSL Passthrough on the gateway, which would be a large rewrite

shashankram: Okay sounds good.

shashankram (approved):

snehachhabria (commented):
please could you rebase from main, we had an issue with the CI that was fixed.

snehachhabria (approved):

PR #3696, opened 2021-06-29
e2e: Add additional check to verify injector and controller readiness

eduser25:
There's been some cases where injector was not ready after osm install
in some very specific situations and environments.

To avoid and ensuring it's not a situational timing issue or a `helm --wait`
related problem (see #3665), this commit adds an additional check to make
_sure_ osm-controller and osm-injector apps are marked as ready by
kubernetes before proceeding.

**Affected area**:

no
1. Is this a breaking change?
no

### DIFF
--- a/tests/framework/common.go
+++ b/tests/framework/common.go
@@ -283,7 +283,7 @@ nodeRegistration:
 	// After client creations, do a wait for kind cluster just in case it's not done yet coming up
 	// Ballparking pod number. kind has a large number of containers to run by default
 	if (td.InstType == KindCluster) && td.ClusterProvider != nil {
-		if err := td.WaitForPodsRunningReady("kube-system", 120*time.Second, 5); err != nil {
+		if err := td.WaitForPodsRunningReady("kube-system", 120*time.Second, 5, nil); err != nil {
 			return errors.Wrap(err, "failed to wait for kube-system pods")
 		}
 	}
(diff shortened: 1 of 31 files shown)


tests/framework/common.go
```diff
@@ -1031,6 +1036,45 @@ func (td *OsmTestData) WaitForPodsRunningReady(ns string, timeout time.Duration,
 	return fmt.Errorf("not all pods were Running & Ready in NS %s after %v", ns, timeout)
 }
 
+// waitPodsReady waits for pods fetched by app label in a namespace to be ready in N seconds of time
+func (td *OsmTestData) waitPodsReady(ns string, appLabels []string, timeout time.Duration) bool {
```
shashankram, line 1040: there's already a function `WaitForPodsRunningReady(...)` that ensures all pods in the control plane namespace are ready. Let's reuse that and integrate that check for all installs, and not just for `cert-manager`, `vault` scenarios.

shashankram (changes requested):

tests/framework/common.go
eduser25: the existing impl checks for a number of ready pod running on a namespace, it's more basic in that sense. I anything, this is better as allows checking specific applications running as opposed to just a number of pods running (which I think is needed).
what would you suggest?

shashankram: Can we instead enhance that and wait for all pods in that namespace to be running? What's the benefit to only waiting for a subset of pods in the namespace?

eduser25: that was an easy way to wait for pods without going into the gory details as of how to do it, as we generally anyway we don't have bags of mixed stuff in test namespaces.
I'm not sure all uses of the previous impl would be able to migrate to the new one as it uses app labels (do single pod tests / no deployment tests use app label?)
I'll take a look regardless. I purposefully tried to avoid changing a broadly used api for the reasons above.

shashankram: My suggestion is to collapse all checks to a single API that checks if all pods in the control plane namespace are running.

eduser25: @shashankram take a look again. I augmented the available API to optionally take selectors. lmk

shashankram (approved):

snehachhabria (approved):

SanyaKochhar (approved):

PR #3707, opened 2021-06-30
feat(pkg/apis): add a global ip address for the multiclusterservice

steeling:
The global IP address is used by the listener to determine which service is chosen, and will be assigned by a mutating webhook.

-->
**Affected area**:

No

1. Is this a breaking change?
No

### DIFF
--- a/charts/osm/crds/multiclusterservice.yaml
+++ b/charts/osm/crds/multiclusterservice.yaml
@@ -41,6 +41,9 @@ spec:
                 - serviceAccount
                 - clusters
               properties:
+                globalIP:
+                  type: string
+                  description: globalIP is the IP address for the implicit global service which references every defined cluster. It is usually assigned at random by OSM.
                 serviceAccount:
                   type: string
                 ports:
--- a/pkg/apis/config/v1alpha1/multi_cluster_service.go
+++ b/pkg/apis/config/v1alpha1/multi_cluster_service.go
@@ -24,6 +24,9 @@ type MultiClusterService struct {
 
 // MultiClusterServiceSpec is the type used to represent the multicluster service specification.
 type MultiClusterServiceSpec struct {
+	// GlobalIP defines the IP Address for the implicit global service.
+	GlobalIP string `json:"globalIP,omitempty" protobuf:"bytes,3,opt,name=globalIP"`
+
 	// ClusterSpec defines the configuration of other clusters
 	Cluster []ClusterSpec `json:"cluster,omitempty"`
 


charts/osm/crds/multiclusterservice.yaml
```diff
@@ -41,6 +41,9 @@ spec:
                 - serviceAccount
                 - clusters
               properties:
+                globalIP:
```
snehachhabria, line 44: Would you elaborate why this is needed ? The design doc doesn't talk about this, isn't the IP address of the gateway sufficient ?

steeling: The envoy listener filter is applied after dns, and uses the IP address of the service to perform the match, since that’s all that’s available on the socket since tls may not have been applied. 

Without an IP address it wouldn’t be able to use it. Correct that it was missed in the initial design doc

charts/osm/crds/multiclusterservice.yaml
```diff
@@ -41,6 +41,9 @@ spec:
...
+                  type: string
+                  description: globalIP is the IP address for the implicit global service which references every defined cluster. It is usually assigned at random by OSM.
```
shashankram, line 46: > for the implicit global service

This is not clear to me. What is this referring to? Public IP? Internal k8s service IP?

charts/osm/crds/multiclusterservice.yaml
shashankram: @steeling could you clarify the following:
1. What is this IP referring to exactly
2. Is this to be used for inbound traffic filtering on the gateway or outbound filtering on the client to access the gateway?

steeling: 1. The IP address for the svc.namespace.cluster.global target
2. inbound filtering on the downstream service (not the gateway), to pick the outbound cluster

charts/osm/crds/multiclusterservice.yaml
steeling: Internal k8s service IP.. it will need to be created by a mutating webhoook

shashankram: Why would the service IP need to be created by a mutating webhook? Would you mind adding a brief section on how this is going to be leveraged in the design doc.

steeling: yup, will do!

would it be possible to do that async and continue with this PR?

charts/osm/crds/multiclusterservice.yaml
steeling: resolved based on in person conversation

snehachhabria (approved):

snehachhabria:
not required for now

PR #3710, opened 2021-06-30
cmd/osm-*: log error codes for errors during startup

shashankram:
**Description**:
Logs error codes for errors encountered during app
startup. Additionally moves error logging to the
caller/handler of the error to avoid duplicate
logging.

Part of #2866

**Affected area**:

### DIFF
--- a/cmd/osm-injector/reconcile.go
+++ b/cmd/osm-injector/reconcile.go
@@ -1,9 +1,11 @@
 package main
 
 import (
+	"github.com/pkg/errors"
 	"k8s.io/client-go/kubernetes"
 	ctrl "sigs.k8s.io/controller-runtime"
 
+	"github.com/openservicemesh/osm/pkg/errcode"
 	"github.com/openservicemesh/osm/pkg/reconciler"
 )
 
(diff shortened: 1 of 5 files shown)


snehachhabria (approved):

cmd/osm-injector/reconcile.go
```diff
@@ -27,15 +28,15 @@ func createReconciler(kubeClient *kubernetes.Clientset) error {
 		OsmWebhook:   webhookConfigName,
 		OsmNamespace: osmNamespace,
 	}).SetupWithManager(mgr); err != nil {
-		log.Error().Err(err).Msg("Error creating controller to reconcile MutatingWebhookConfiguration")
-		return err
+		return errors.Wrap(err, "Error creating controller to reconcile MutatingWebhookConfiguration")
```
allenlsy, line 31: Does `Wrap()` log the message at this point?

allenlsy (approved):

cmd/osm-injector/reconcile.go
shashankram: Yes, that's the whole point of Wrap: https://play.golang.org/p/Aq4sdQMtzah

PR #3714, opened 2021-07-01
catalog/egress: define error codes for error logs

shashankram:
**Description**:
Defines error codes for egress policy related
errors.

**Affected area**:

### DIFF
--- a/pkg/errcode/errcode.go
+++ b/pkg/errcode/errcode.go
@@ -34,13 +34,37 @@ const (
 	ErrStartingReconcileManager
 )
 
+// Range 2000-2500 is reserved for errors related to traffic policies
+const (
+	// ErrDedupEgressTrafficMatches indicates an error related to deduplicating egress traffic matches
+	ErrDedupEgressTrafficMatches errCode = iota + 2000
+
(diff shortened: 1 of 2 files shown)


pkg/errcode/errcode.go
```diff
@@ -34,13 +34,37 @@ const (
...
 	return fmt.Sprintf("E%d", e)
 }
 
+// Note: error code description mappings must be defined in the same order
+// as they appear in the error code definitions - from lowest to highest
+// ranges in the order they appear within the range.
 //nolint: deadcode,varcheck,unused
 var errCodeMap = map[errCode]string{
+	//
+	// Range 1000-1050
```
allenlsy, line 66: Why 1050? Do we anticipate only 50 error codes? The range below has 500 codes

allenlsy (approved):

pkg/errcode/errcode.go
shashankram: In general, we shouldn't have a large number of issues related to app startup, which the 1000-1050 range is reserved for. The range starting at 2000 is reserved for policy-related errors, which can be quite a large number as you will see from the code.

Also, the ranges are fairly large as is the case when you want to account for future needs.

jaellio (approved):

PR #3774, opened 2021-07-15
envoy/lds: refactor http connection manager + wasm filter

shashankram:
**Description**:
The existing code to build the HTTP connection
manager is unnecessarily complex and complicated
to test. This change refactor the code such
that it is easier to comprehend and test.

Additionally, removes the WASM stats headers for ingress
so that internal mesh details (pod name, namespace,
deployments etc.) are not leaked as a part of the
HTTP response headers.

- Removes mesh WASM config for ingress
- WASM, external auth, tracing related code
  have been consolidated into their respective
  files.
- HTTP connection manager is now built from its
  input config as opposed to passing the Configurator
  interface around multiple times.
- Fixes incorrect test expectations when registering
  mock expectations for function calls.
- Only adds the necessary WASM filters when `statsHeaders`
  are configured on the `lisenerBuilder` type.
- Renames tests to match the function names and moves
  updated tests to their corresponding test files.

**Affected area**:


pkg/envoy/lds/http_connection_test.go
```diff
@@ -0,0 +1,135 @@
...
+		return true
+	}
+	contains := func(filters []*xds_hcm.HttpFilter, filterName string) bool {
+		return !notContains(filters, filterName)
+	}
+
+	testCases := []struct {
+		name       string
+		option     httpConnManagerOptions
+		assertFunc func(*assert.Assertions, *xds_hcm.HttpConnectionManager, error)
```
nojnhuh, line 30: I notice the `error` is never checked in any test. Since it looks like any error would ultimately come from `ptypes.MarshalAny` (which I haven't found a way to reliably make fail), I'd be in favor of removing it here and adding a `assert.NoError(err)` directly in the loop over the tests where we call `buildFromOptions()`.

pkg/envoy/lds/ingress.go
```diff
@@ -41,11 +41,29 @@ func (lb *listenerBuilder) newIngressHTTPFilterChain(cfg configurator.Configurat
...
+	if err != nil {
+		log.Error().Err(err).Msgf("Error building inbound HTTP connection manager for proxy with identity %s and service %s", lb.serviceIdentity, svc)
+		return nil
+	}
+
 	marshalledIngressConnManager, err := ptypes.MarshalAny(ingressConnManager)
 	if err != nil {
-		log.Error().Err(err).Str(errcode.Kind, errcode.ErrMarshallingXDSResource.String()).
-			Msgf("Error marshalling ingress HttpConnectionManager object for proxy %s", svc)
+		log.Error().Err(err).Msgf("Error marshalling inbound HTTP connection manager for proxy with identity %s and service %s", lb.serviceIdentity, svc)
```
nojnhuh, line 66: Should we be keeping the error code here?

pkg/envoy/lds/http_connection.go
```diff
@@ -0,0 +1,135 @@
...
+	// Additional filters
+	wasmStatsHeaders map[string]string
+	extAuthConfig    *auth.ExtAuthConfig
+
+	// Tracing options
+	enableTracing      bool
+	tracingAPIEndpoint string
+}
+
+func (options httpConnManagerOptions) buildFromOptions() (*xds_hcm.HttpConnectionManager, error) {
```
eduser25, line 47: `options` here is passed by value right, and therefore struct copy? wasmStatsHeader map and extAuthzConfig are already reference-copied if so I imagine, but not sure about connectionDirection

eduser25 (approved):

pkg/envoy/lds/inmesh_test.go
```diff
@@ -625,3 +626,28 @@ func TestGetOutboundTCPFilter(t *testing.T) {
...
+	}
+
+	mockConfigurator.EXPECT().IsTracingEnabled().Return(false).AnyTimes()
+	mockConfigurator.EXPECT().GetTracingEndpoint().Return("test-endpoint").AnyTimes()
+	mockConfigurator.EXPECT().GetInboundExternalAuthConfig().Return(auth.ExtAuthConfig{
+		Enable: false,
+	}).AnyTimes()
+	mockConfigurator.EXPECT().GetFeatureFlags().Return(v1alpha1.FeatureFlags{
+		EnableWASMStats: false,
+	}).AnyTimes()
```
nojnhuh, line 647: Similar comment above about how if the values that these return aren't important, I'd remove them (though the `.AnyTimes()` is still necessary for the last two since they're called more than once):

This still passes:
```suggestion
	mockConfigurator.EXPECT().IsTracingEnabled()
	mockConfigurator.EXPECT().GetTracingEndpoint()
	mockConfigurator.EXPECT().GetInboundExternalAuthConfig().AnyTimes()
	mockConfigurator.EXPECT().GetFeatureFlags().AnyTimes()
```

(Acknowledging that this was just moved from a previous test and not something brand new in this PR)

nojnhuh (approved):
Left a few suggestions inline, but nothing blocking this.

pkg/envoy/lds/http_connection.go
shashankram: `options` is passed by value to "indicate" the function does not mutate the input options. The pointer fields are used to represent optional fields in the struct, which is sort of a common pattern, for example with struct types that have json tags where using a pointer can differentiate between a value being unset vs being 0. `connectionDirection` is not a pointer, so it's copied.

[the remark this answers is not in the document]
shashankram: I anticipate adding more methods on this type to mutate the connection manager as we add more functionality.

snehachhabria (approved):

pkg/envoy/lds/http_connection_test.go
```diff
@@ -0,0 +1,135 @@
...
+					Address:          "test.xyz",
+					Port:             123,
+					StatPrefix:       "pref",
+					AuthzTimeout:     3 * time.Second,
+					FailureModeAllow: false,
+				},
+			},
+			assertFunc: func(a *assert.Assertions, connManager *xds_hcm.HttpConnectionManager, err error) {
+				a.True(notContains(connManager.HttpFilters, wellknown.HTTPExternalAuthorization))
+				a.Equal(wellknown.Router, connManager.HttpFilters[len(connManager.HttpFilters)-1].Name) // Router must be last
```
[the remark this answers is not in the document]
shashankram: Makes sense!

nojnhuh (approved):
Thank you @shashankram, this is a big improvement!

PR #3795, opened 2021-07-19
feat(upgrade): Add new http server to crd converter

snehachhabria:
**Description**:

Add another http server to the crd converter which servers as the health
endpoint which is used to sync the readiness of osm-controller based on this port.

Update the cleanup hook as pre-delete to ensure proper deletion order
of CRD's and services via Helm.

Part of #3396

**Affected area**:

Please answer the following questions with yes/no. `no`

### DIFF
--- a/pkg/crdconversion/crdconversion.go
+++ b/pkg/crdconversion/crdconversion.go
@@ -23,6 +23,9 @@ const (
 
 	// crdConverterServiceName is the name of the OSM crd converter webhook service
 	crdConverterServiceName = "osm-crd-converter"
+
+	// healthPort is the port on which the '/healthz` requests are served
+	healthPort = 9095
 )
 
 var crdConversionWebhookConfiguration = map[string]string{
@@ -75,15 +78,13 @@ func (crdWh *crdConversionWebhook) run(stop <-chan struct{}) {
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 
-	mux := http.NewServeMux()
-
-	mux.HandleFunc(webhookHealthPath, healthHandler)
+	webhookMux := http.NewServeMux()
 
 	// TODO (snchh): add handler and logic for conversion stratergy of each CRD in OSM
 
-	server := &http.Server{
+	webhookServer := &http.Server{
 		Addr:    fmt.Sprintf(":%d", crdWh.config.ListenPort),
-		Handler: mux,
+		Handler: webhookMux,
 	}
 
 	log.Info().Msgf("Starting conversion webhook server on port: %v", crdWh.config.ListenPort)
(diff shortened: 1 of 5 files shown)


shashankram (approved):

pkg/crdconversion/crdconversion.go
```diff
@@ -104,8 +106,28 @@ func (crdWh *crdConversionWebhook) run(stop <-chan struct{}) {
...
+
+	go func() {
+		if err := server2.ListenAndServe(); err != nil {
+			log.Error().Err(err).Msg("crd-converter health server failed to start")
+			return
+		}
+		wg.Done()
+	}()
+
+	wg.Wait()
```
shashankram, line 129: Why is the Wait() required?

You also need to allow shutting down server2 similar to server1 (L135)

eduser25 (dismissed):

pkg/crdconversion/crdconversion.go
allenlsy: Also I think, if both server fail to start, will the program wait here forever?

pkg/crdconversion/crdconversion.go
```diff
@@ -104,8 +106,28 @@ func (crdWh *crdConversionWebhook) run(stop <-chan struct{}) {
...
 		}
+		wg.Done()
 	}()
 
+	healthMux := http.NewServeMux()
+
+	healthMux.HandleFunc(webhookHealthPath, healthHandler)
+
+	server2 := &http.Server{
+		Addr:    fmt.Sprintf(":%d", 9095),
```
eduser25, line 117: maybe move the port to const section

pkg/crdconversion/crdconversion.go
shashankram: This needs to be fixed as discussed offline.

shashankram (changes requested):

pkg/crdconversion/crdconversion.go
eduser25: a bit confused by the wait too

pkg/crdconversion/crdconversion.go
snehachhabria: done

pkg/crdconversion/crdconversion.go
snehachhabria: removed the wait and shut down the server

pkg/crdconversion/crdconversion.go
```diff
@@ -75,15 +79,16 @@ func (crdWh *crdConversionWebhook) run(stop <-chan struct{}) {
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 
-	mux := http.NewServeMux()
+	wg := new(sync.WaitGroup)
```
eduser25, line 82: Not sure if you were planning to use it to wait post-shutdowns, to ensure the routines finished. Otherwise this looks unused so far.

pkg/crdconversion/crdconversion.go
```diff
@@ -75,15 +79,16 @@ func (crdWh *crdConversionWebhook) run(stop <-chan struct{}) {
...
+	wg.Add(2)
```
shashankram, line 83: This is not required.

eduser25 (approved):
leaving +1 after discussing changes offline

pkg/crdconversion/crdconversion.go
```diff
@@ -96,25 +97,47 @@ func (crdWh *crdConversionWebhook) run(stop <-chan struct{}) {
...
 
-		if err := server.ListenAndServeTLS("", ""); err != nil {
+		if err := webhookServer.ListenAndServeTLS("", ""); err != nil {
 			log.Error().Err(err).Msg("crd-converter webhook HTTP server failed to start")
 			return
 		}
 	}()
 
+	healthMux := http.NewServeMux()
+
```
shashankram, line 111: nit: remove unnecessary newline

shashankram (approved):

pkg/crdconversion/crdconversion.go
snehachhabria: I have another pr that will modify the same file, I will remove it then

allenlsy (approved):

Issue #3775, opened 2021-07-16
XDS state machine issue with SDS request being ignored


shashankram:
**Bug description**:
There seems to be a bug in osm-controller's XDS state machine where requested resources are not correctly fulfilled, potentially due to how the requested resources are tracked.

In the demo, it was observed that 2 secrets on the bookbuyer proxy were not fulfilled:
```
  {
   "@type": "type.googleapis.com/envoy.admin.v3.SecretsConfigDump",
   "dynamic_warming_secrets": [
    {
     "name": "service-cert:bookbuyer/bookbuyer",
     "version_info": "uninitialized",
     "last_updated": "2021-07-16T21:11:56.339Z",
     "secret": {
      "@type": "type.googleapis.com/envoy.extensions.transport_sockets.tls.v3.Secret",
      "name": "service-cert:bookbuyer/bookbuyer"
     }
    },
    {
     "name": "root-cert-for-mtls-outbound:bookstore/bookstore",
     "version_info": "uninitialized",
     "last_updated": "2021-07-16T21:11:56.339Z",
     "secret": {
      "@type": "type.googleapis.com/envoy.extensions.transport_sockets.tls.v3.Secret",
      "name": "root-cert-for-mtls-outbound:bookstore/bookstore"
     }
    }
   ]
  }
```

OSM controller logs: 
[osm_controller.log](https://github.com/openservicemesh/osm/files/6833594/osm_controller.log)

OSM controller logs related to bookbuyer proxy:
[osm_controller_bookbuyer.log](https://github.com/openservicemesh/osm/files/6833595/osm_controller_bookbuyer.log)

Config dump on bookbuyer:
[bookbuyer_envoy_dump.txt](https://github.com/openservicemesh/osm/files/6833605/bookbuyer_envoy_dump.txt)

This seemed to happen when switching from permissive mode to SMI. From the logs, it seems as though the controller thinks the 2 requested SDS secrets are ACK'd, when in reality this is a request from Envoy post the LDS/CDS config update after switching from permissive mode to SMI:
```
[...]

PR #3799, opened 2021-07-19
ads: Convert superset logic to equals when computing ACK

eduser25:
Unsubscriptions might be requesting less resources from a given number
of resources previously requested:

```
Original logic:
req: [N:-, A, B]
response: [N:1, A, B]
req: [N:1, A, B] (ACK)

envoy unsubscribes from B:
req: [N:1, A]
response: // OSM Never replies as nonce 1 should have already fulfilled this request, as it
          // is a superset (already had 'A') from the new resources being requested
```

This is causing issues as it seems that envoy does expect to be replied to.
Changing superset logic to equals will make OSM reply as long as the requested
resources are different from the last sent.

This fix is sitting on top of #3803, as it is needed to properly account for the subscribed and
sent resources in a proper way, for every proxy.

Related to #3775

### DIFF
--- a/pkg/envoy/ads/stream.go
+++ b/pkg/envoy/ads/stream.go
@@ -273,15 +273,9 @@ func respondToRequest(proxy *envoy.Proxy, discoveryRequest *xds_discovery.Discov
 	// Get resources last sent prior to this request
 	resourcesLastSent := proxy.GetLastResourcesSent(typeURL)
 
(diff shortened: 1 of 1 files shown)


shashankram (approved):

allenlsy:
LGTM. @snehachhabria  can you take a look also?

snehachhabria (approved):

PR #3803, opened 2021-07-20
ads: fix wrong accountability of sent resources

eduser25:
This commit introduces 2 changes:
- Knowledge of what resources a certain proxy has requested/subscribed
to, up to a current point in time and,
- the use of this new facility to avoid accounting for resources
sent that might have not been requested by envoy yet, as well as using them
to generate stub requests from ctl-driven proxy updates.

There's a potential bug in how we treated resources sent, where a
non-wildcard vertical sending more resources than the requested/subscribed ones
could make a later request be understood as an ACK.

Consider the following case:
```
req: [N:-, A]
resp: [N:1, A, B]  <- let's say vertical code generated more resources than requested
                      OSM would record that we last sent [A, B]
// Envoy will silently drop [B]
req [N:1, A, B]   <- Envoy is actually requesting B apart from A now
resp: // No response, OSM thinks this is an ACK
```

Related to #3775

No
1. Is this a breaking change?
No


eduser25:
I will add a couple unit tests for the newly added functions

pkg/envoy/ads/response.go
```diff
@@ -122,15 +125,31 @@ func (s *Server) SendDiscoveryResponse(proxy *envoy.Proxy, request *xds_discover
...
 			log.Error().Err(err).Str(errcode.Kind, errcode.ErrMarshallingXDSResource.String()).
 				Msgf("Error marshalling resource %s for proxy %s", typeURI, proxy.GetCertificateSerialNumber())
 			continue
 		}
+		// Add resource to response
 		response.Resources = append(response.Resources, proto)
-		resourcesSent.Add(cache.GetResourceName(res))
+
+		// We will preemptively filter each of the resources generated by the non-wildcard verticals and NOT count them
+		// towards `LastSentResources` if they are not subscribed resources if they were never requested by the proxy.
```
shashankram, line 140: Could you reword this statement with a double `if` for clarity

pkg/envoy/ads/stream.go
```diff
@@ -246,22 +248,24 @@ func respondToRequest(proxy *envoy.Proxy, discoveryRequest *xds_discovery.Discov
...
-	// with the new resources requested.
-	//
+	// Update subscribed resources for this TypeURI with the ones on this request.
 	// In case of LDS and CDS, "Envoy will always use wildcard mode for Listener and Cluster resources".
-	// The following logic is not needed (though correct) for LDS and CDS as request resources are also empty in ACK case.
-	//
+	// The following logic is not needed (though correct) for both LDS and CDS as both requested and last sent resources
+	// will purposefully remain as empty sets at all times, and could be avoided altogether.
+	resourcesRequested := getRequestedResourceNamesSet(discoveryRequest)
+	proxy.SetSubscribedResources(typeURL, resourcesRequested)
```
shashankram, line 260: If this can be avoided for LDS and CDS, what's the motivation to execute it?

pkg/envoy/ads/stream.go
```diff
@@ -246,22 +248,24 @@ func respondToRequest(proxy *envoy.Proxy, discoveryRequest *xds_discovery.Discov
...
+
+	// ----
+	// What's left is to check if the resources listed are the same. If they are not, we must answer
+	// with a new response.
```
shashankram, line 264:
```suggestion
	// What's left is to check if the resources listed are the same. If they are not, we must respond to the request for the requested resources
```

pkg/envoy/ads/stream.go
```diff
@@ -289,6 +293,22 @@ func getRequestedResourceNamesSet(discoveryRequest *xds_discovery.DiscoveryReque
 	return resourcesRequested
 }
 
+// Helper to turn a mapset of resource names to a string slice
```
shashankram, line 296:
```suggestion
// getResourceSliceFromMapset is a helper to convert a mapset of resource names to a string slice
```

pkg/envoy/ads/response.go
eduser25: will do

pkg/envoy/ads/stream.go
eduser25: I had removed it initially, but decided to leave it as is to keep smaller amount of changes possible and check with reviewers first. Glad we think alike here.

I had something around the lines of:
```
if IsWildCardTypeURI(typeuri) {
   // skip resource names checks, treat as ACK (which it is)
}
```

pkg/envoy/proxy.go
```diff
@@ -207,6 +210,21 @@ func (p *Proxy) SetLastResourcesSent(typeURI TypeURI, resourcesSet mapset.Set) {
...
+// If none were subscribed, empty set is returned
+func (p *Proxy) GetSubscribedResources(typeURI TypeURI) mapset.Set {
+	sentResources, ok := p.subscribedResources[typeURI]
+	if !ok {
+		return mapset.NewSet()
+	}
+	return sentResources
+}
+
+// SetSubscribedResources sets the last sent resources given a proxy for a TypeURL
```
eduser25, line 223:
```suggestion
// SetSubscribedResources sets the last subscrubed resources given a proxy for a TypeURL
```

pkg/envoy/ads/stream.go
shashankram: If skipping for wildcard type is correct, might be a good idea to do it. If this doesn't work, then our understanding of the protocol can be corrected.

Up to you on this one.

pkg/envoy/ads/response.go
eduser25: I hope I understood you right here, lmk if not

eduser25:
This commit unblocks https://github.com/openservicemesh/osm/pull/3799, which will fix #3775. We couldn't do a proper `Equal` diff of resources before without accounting properly the sent resources.

Istio's resource diff is not a superset (I took that from  go-control-plane) and is in fact an `Equal` diff of resources:
https://github.com/istio/istio/blob/da6178604559bdf2c707a57f452d16bee0de90c8/pilot/pkg/xds/ads.go#L466

pkg/envoy/types.go
```diff
@@ -17,6 +17,14 @@ var (
 // TypeURI is a string describing the Envoy xDS payload.
 type TypeURI string
 
+// IsWildcardTypeURI returns if a given TypeURI is an expected wildcard TypeURI or not.
+// XDS proto defines general client behavior as:
+// "Envoy will always use wildcard subscriptions for Listener and Cluster resources"
+// https://www.envoyproxy.io/docs/envoy/latest/api-docs/xds_protocol#client-behavior
+func IsWildcardTypeURI(t TypeURI) bool {
+	return t == TypeCDS || t == TypeLDS
```
allenlsy, line 25: Why CDS and LDS are wildcard URIs?

eduser25: https://www.envoyproxy.io/docs/envoy/latest/api-docs/xds_protocol#client-behavior

pkg/envoy/ads/response.go
```diff
@@ -49,24 +49,27 @@ func (s *Server) getTypeResources(proxy *envoy.Proxy, request *xds_discovery.Dis
...
-					continue
-				}
-			} else {
-				finalReq = &xds_discovery.DiscoveryRequest{TypeUrl: typeURI.String()}
-			}
+		if osmDrivenUpdate {
+			finalReq = &xds_discovery.DiscoveryRequest{TypeUrl: typeURI.String()}
+
+			// Fill the request resources with subscribed resources from the proxy thus far.
+			// Verticals should be hold accountable to generated the requested resources.
```
shashankram, line 66:
```suggestion
			// Verticals should be held accountable for generating the requested resources.
```

shashankram (approved):

snehachhabria (approved):

PR #3813, opened 2021-07-20
Update node count for kind cluster on CI

snehachhabria:
**Description**:

This commit increase the node count for kind cluster used in the CI.

**Affected area**:

### DIFF
--- a/.github/workflows/nightly-noinstall.yml
+++ b/.github/workflows/nightly-noinstall.yml
@@ -22,16 +22,18 @@ jobs:
           apiVersion: kind.x-k8s.io/v1alpha4
           nodes:
           - role: control-plane
+          - role: worker
             kubeadmConfigPatches:
             - |
-              kind: InitConfiguration
+              kind: JoinConfiguration
               nodeRegistration:
                 kubeletExtraArgs:
                   node-labels: "ingress-ready=true"
             extraPortMappings:
             - containerPort: 80
               hostPort: 80
-              protocol: TCP
+              protocol: TCP          
+          - role: worker
           EOF
 
           cat kind-config.yaml
--- a/tests/framework/common.go
+++ b/tests/framework/common.go
@@ -8,6 +8,7 @@ import (
 	"encoding/json"
 	"flag"
 	"fmt"
+	"io"
 	"io/ioutil"
 	"math"
 	"math/big"
@@ -222,7 +223,10 @@ func (td *OsmTestData) InitTestData(t GinkgoTInterface) error {
 			Nodes: []v1alpha4.Node{
 				{
 					Role: v1alpha4.ControlPlaneRole,
-					KubeadmConfigPatches: []string{`kind: InitConfiguration
+				},
+				{
+					Role: v1alpha4.WorkerRole,
+					KubeadmConfigPatches: []string{`kind: JoinConfiguration
 nodeRegistration:
   kubeletExtraArgs:
     node-labels: "ingress-ready=true"`},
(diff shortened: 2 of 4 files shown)


tests/framework/common_apps.go
```diff
@@ -463,7 +463,7 @@ func (td *OsmTestData) GetPrometheusPodHandle(ns string, prometheusPodName strin
 }
 
 func (td *OsmTestData) waitForOSMControlPlane(timeout time.Duration) error {
-	var errController, errInjector error
+	var errController, errInjector, errCrdConverter error
 	waitGroup := sync.WaitGroup{}
 	waitGroup.Add(2)
```
eduser25, line 468:
```suggestion
	waitGroup.Add(3)
```
😸

tests/framework/common.go
```diff
@@ -234,10 +238,15 @@ nodeRegistration:
...
 				},
+				{
+					Role: v1alpha4.WorkerRole,
+				},
 			},
 		}
 		if Td.ClusterVersion != "" {
 			clusterConfig.Nodes[0].Image = fmt.Sprintf("kindest/node:%s", td.ClusterVersion)
+			clusterConfig.Nodes[1].Image = fmt.Sprintf("kindest/node:%s", td.ClusterVersion)
+			clusterConfig.Nodes[2].Image = fmt.Sprintf("kindest/node:%s", td.ClusterVersion)
```
nojnhuh, line 249: Let's make this a loop.

tests/framework/common.go
```diff
@@ -346,14 +355,21 @@ func (td *OsmTestData) LoadImagesToKind(imageNames []string) error {
 	if err != nil {
 		return errors.Wrap(err, "failed to get image data")
 	}
+
+	imageReader, _ := ioutil.ReadAll(imageData)
```
nojnhuh, line 359: Let's handle this error just to be safe, though I don't expect we'll ever hit it.

nojnhuh (approved):

tests/framework/common.go
```diff
@@ -234,10 +238,16 @@ nodeRegistration:
...
 					},
 				},
+				{
+					Role: v1alpha4.WorkerRole,
+				},
 			},
 		}
 		if Td.ClusterVersion != "" {
-			clusterConfig.Nodes[0].Image = fmt.Sprintf("kindest/node:%s", td.ClusterVersion)
+			for i := 0; i <= 3; i++ {
```
shashankram, line 247: What's 3? Maybe define a constant?

shashankram (commented):
Could you split this change into 2 PRs - 1. update the CI infra, and 2. enable CRD convertor

tests/framework/common.go
snehachhabria: defined a constant and I removed enabling the CRD convertor in this PR

shashankram (approved):

.github/workflows/nightly-noinstall.yml
```diff
@@ -22,6 +22,7 @@ jobs:
           apiVersion: kind.x-k8s.io/v1alpha4
           nodes:
           - role: control-plane
+          - role: worker
```
nshankar13, line 25: Is there a reason that you're setting the ingress configuration and port mappings in the worker node instead of the the control-plane node?

snehachhabria: it shouldn't matter, we just need to ensure that the ingress e2e test runs on the node that this configuration is set for only. I have left the configuration on the control-plane node itself

snehachhabria: we want the nginx controller pod to be scheduled on a specific node, based on the node selector match `ingress-reday=true` . Since the control-plane node has a master label, the pod nginx pod won't be scheduled on this even if the `inress-ready=true` label selector matches, hence I moved it to a worker.

Issue #3742, opened 2021-07-07
MeshConfig values that default to "true" can't be set to "false" via Helm values


nojnhuh:
**Bug description**: Boolean fields with a default of "true" in the MeshConfig CRD aren't overridden when the corresponding Helm value is set to "false" when OSM is installed. This applies to `OpenServiceMesh.featureFlags.enableWASMStats` and `OpenServiceMesh.featureFlags.enableEgressPolicy`.

**Affected area (please mark with X where applicable)**:
- Install                    [X]
- SMI Traffic Access Policy  [ ]
- SMI Traffic Specs Policy   [ ]
- SMI Traffic Split Policy   [ ]
- Permissive Traffic Policy  [ ]
- Ingress                    [ ]
- Egress                     [ ]
- Envoy Control Plane        [ ]
- CLI Tool                   [ ]
- Metrics                    [ ]
- Certificate Management     [ ]
- Sidecar Injection          [ ]
- Logging                    [ ]
- Debugging                  [ ]
- Tests                      [ ]
- Demo                       [ ]
- CI System                  [ ]

**Expected behavior**: When the value is set to "false" on install, that value ultimately shows up in the osm-mesh-config.

**Steps to reproduce the bug (as precisely as possible)**:
```
make kind-up && demo/run-osm-demo.sh --set OpenServiceMesh.featureFlags.enableWASMStats=false
kubectl get meshconfig -n osm-system osm-mesh-config -o jsonpath='{.spec.featureFlags.en
ableWASMStats}'
```

**How was OSM installed?**: Demo

**Anything else we need to know?**:

In the preset-mesh-config the value is set correctly to `false` (`kubectl get configmap -n osm-system preset-mesh-config -o yaml`), so there must be some issue with how the osm-mesh-config is created.

Changing the default to `false` in the CRD before installing OSM gets around the issue. In that case the default in values.yaml is true then it will still be enabled by default.

**Environment**:
- OSM version (use `osm version`): `Version: dev; Commit: ead5c9e8095d55dacbe79cb67259c9ccc78bb0ff; Date: `
- Kubernetes version (use `kubectl version`):
```
[...]

PR #3831, opened 2021-07-21
fix(install): Remove boolean default values from MeshConfig CRD

allenlsy:

**Description**:

In the issue description, user tries to set `enableWASMStats` to false in the `osm install` process, whose default value is true. But after installation, WASM stats is not enabled. In `osm-mesh-config`, it shows `false`.

The root cause is at this line: https://github.com/openservicemesh/osm/blob/b370fa2689293394d39eb29388e32213f9027b41/cmd/init-osm-controller/init-osm-controller.go#L64

When `enableWASMStats` is false, it is same as the default value declared in `FeatureFlags` struct. Due to the `omitempty` tag, it is omitted. Moving on to line https://github.com/openservicemesh/osm/blob/b370fa2689293394d39eb29388e32213f9027b41/cmd/init-osm-controller/init-osm-controller.go#L117, when the MeshConfig resource is created, the default value declared in the CRD manifest is used, which is true.

> [Outdated comment]
> The solution is to remove `omitempty` in this case.

The solution is removing default values from MeshConfig CRD. Default values should be set in values.yaml

This also applies to the `FeatureFlags.EnableEgressPolicy` field.

This PR also updates the test cases for other field, to further ensure the correctness of the unmarshalling process. 

Resolves issue #3742 

**Affected area**:

No

1. Is this a breaking change?

No

--------

### Testing

Manually deployed OSM to local cluster: `demo/run-osm-demo.sh --set OpenServiceMesh.featureFlags.enableWASMStats=false`.

Observed `enableWASMStats: false` that in the MeshConfig.

### Debug process

During debugging process, the Spec used in `preset-mesh-config` ConfigMap:

```
{
    "sidecar": {
        "enablePrivilegedInitContainer": false,
        "logLevel": "debug",
        "maxDataPlaneConnections": 0,
        "envoyImage": "envoyproxy/envoy-alpine:v1.18.3",
        "initContainerImage": "localhost:5000/init:15c31eb4e27df0ba190f9d7ad064dbdee73c5ea4",
        "configResyncInterval": "0s"
    },
    "traffic": {
        "enableEgress": false,
        "useHTTPSIngress": false,
        "enablePermissiveTrafficPolicyMode": false,
        "outboundPortExclusionList": [],
        "inboundPortExclusionList": [],
        "outboundIPRangeExclusionList": []
    },
    "observability": {
        "enableDebugServer": true,
        "osmLogLevel": "trace",
        "tracing": {
            "enable": false
        }
    },
    "certificate": {
        "serviceCertValidityDuration": "24h"
    },
    "featureFlags": {
        "enableWASMStats": false,
        "enableEgressPolicy": true,
        "enableMulticlusterMode": true,
        "enableOSMGateway": false,
        "enableAsyncProxyServiceMapping": false,
        "enableValidatingWebhook": false,
        "enableIngressBackendPolicy": false
    }
}
```

After `json.Unmarshal()` of the JSON above, the unmarshalled MeshConfigSpec is:

```
{
    "sidecar": {
        "logLevel": "debug",
          "envoyImage": "envoyproxy/envoy-alpine:v1.18.3",
          "initContainerImage": "localhost:5000/init:15c31eb4e27df0ba190f9d7ad064dbdee73c5ea4",
          "configResyncInterval": "0s",
          "resources": {}
    },
    "traffic": {
        "inboundExternalAuthorization": {}
    },
    "observability": {
        "osmLogLevel": "trace",
        "enableDebugServer": true,
        "tracing": {}
    },
    "certificate": {
        "serviceCertValidityDuration": "24h"
    },
    "experimental": {
        "multicluster": {}
    },
    "featureFlags": {
        "enableEgressPolicy": true,
        "enableMulticlusterMode": true
    }
}
```

`enableWASMStats` field is omitted.


shashankram (commented):
Would removing defaults from the CRD resolve this issue?

pkg/apis/config/v1alpha1/mesh_config.go
```diff
@@ -172,10 +172,10 @@ type MeshConfigList struct {
 // FeatureFlags is a type to represent OSM's feature flags.
 type FeatureFlags struct {
```
snehachhabria, line 173: Should we remove the omit empty from all the bool values in this and not just WASM and Egress ?

allenlsy: I removed `omitempty` for all boolean fields whose default value is true in MeshConfig. We can also do it for other CRDs. But with this approach, developers meed to care about whether `omitempty` should be added, which is error prune.

allenlsy:
@shashankram In this case, I believe yes, it will solve the issue. 

Overall, I prefer remove `omitempty`  for all boolean fields. Keeping default value in CRD has its value.

shashankram (changes requested):
This change introduces a bit of inconsistency in how bool values are initialized via the CRD and API definition. Removing `omitempty` only from certain bool fields is error prone and confusing. I suggest just removing the default values from the CRD definition and initializing defaults using the `values.yaml` file instead.

allenlsy:
Discussed offline. 

We agreed on keeping `omitempty` tag for consistency. Our strategy now should be keeping default values in values.yaml rather than CRD. I also checked with other service meshes (Istio and Linkerd), and they don't set defaults in CRD.

One drawback of this approach is, we cannot (or tedious to) detect whether user configured a field.

allenlsy:
FYI, in Istio, it uses protobuf to declare resources. It then is converted to Golang struct, with default values. Roughly we can say the default value is written in Golang.

Example: https://github.com/istio/api/blob/327bcef8062a2411a9f6c7cc021dc6a9d8b3b7c2/security/v1beta1/authorization_policy.pb.go#L312

cmd/init-osm-controller/init-osm-controller_test.go
```diff
@@ -45,10 +45,10 @@ func TestCreateDefaultMeshConfig(t *testing.T) {
 	}
   },
   "certificate": {
-	"serviceCertValidityDuration": "24h"
+	"serviceCertValidityDuration": "23h"
```
michelleN, line 48: Why are we changing this value to 23h?

allenlsy: Because 24h is the default value in the CRD. Changing it to 23h in test can ensure the option is set by preset-mesh-config.

nojnhuh (approved):

charts/osm/crds/config_meshconfig.yaml
```diff
@@ -208,10 +208,10 @@ spec:
...
                       type: boolean
-                      default: true
+                      default: false
                     enableEgressPolicy:
                       type: boolean
-                      default: true
+                      default: false
                     enableMulticlusterMode:
                       type: boolean
                       default: false
```
shashankram, line 217: These defaults do not convey the correct intention. Based on our discussions, it's easier to always derive defaults from `values.yaml` and configure them via presets.

shashankram (changes requested):

shashankram (approved):

PR #3835, opened 2021-07-22
fix(e2e): upgrade test fixes

nojnhuh:
**Description**:
These changes fix current issues with the upgrade e2e test:

- Remove a check that the `kubectl` executable is present before running
  the upgrade e2e test. Because CRDs are now upgraded with hooks in the
  Helm chart instead of invoking `kubectl` in the test, this is no
  longer necessary.

- Remove the step to reinstall the SMI policy resources, since that is
  no longer necessary with the new CRD management.

- Restart the injected Pods to handle incompatible certificate common
  name formats between the old and new OSM installations.

**Affected area**:

### DIFF
--- a/tests/e2e/e2e_upgrade_test.go
+++ b/tests/e2e/e2e_upgrade_test.go
@@ -1,7 +1,7 @@
 package e2e
 
 import (
-	"os/exec"
+	"context"
 	"path/filepath"
 	"time"
 
@@ -10,6 +10,8 @@ import (
 	"helm.sh/helm/v3/pkg/action"
 	"helm.sh/helm/v3/pkg/chart/loader"
 	"helm.sh/helm/v3/pkg/cli"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	. "github.com/openservicemesh/osm/tests/framework"
 )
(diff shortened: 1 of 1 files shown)


snehachhabria (approved):

tests/e2e/e2e_upgrade_test.go
```diff
@@ -1,7 +1,6 @@
 package e2e
 
 import (
-	"os/exec"
```
snehachhabria, line 4: We can remove the `// Re-deploy policies` code in line 218 as a part of this PR too

ksubrmnn (approved):
Would it make sense to have a flag in `osm upgrade` to handle restarting these deployments automatically?

nojnhuh:
> Would it make sense to have a flag in `osm upgrade` to handle restarting these deployments automatically?

I suspect that would be possible, but might be tricky in certain cases. Like if there's a Pod in the mesh not backed by a controller, can we figure out what was originally originally created vs. what was added by a mutating webhook? In that case would recreating the Pod in the mesh from its current spec mean the osm-injector would add a second sidecar? I'm sure other non-trivial things like that would pop up.

My gut tells me it's probably not worth it. If I were a user that saw that flag, I wouldn't use it just in case it does something destructive and just restart my workloads myself since I would know best how to do that.

snehachhabria (approved):
thanks for fixing!

PR #3837, opened 2021-07-22
Make the certificate key size configurable

trstringer:
**Description**:

This PR addresses #1933. With this PR, the certificate key size is now configurable through mesh config as `certKeyBitSize`. Prior to this PR, it was hard coded.

**Affected area**:
| Security                   | [x] |

No.

1. Is this a breaking change?

No.

### DIFF
--- a/charts/osm/values.schema.json
+++ b/charts/osm/values.schema.json
@@ -133,6 +133,7 @@
                 "sidecarImage",
                 "certificateManager",
                 "serviceCertValidityDuration",
+                "certKeyBitSize",
                 "caBundleSecretName",
                 "enableDebugServer",
                 "enablePermissiveTrafficPolicy",
@@ -275,6 +276,15 @@
                     "description": "The service certificate validity duration.",
                     "examples": [
                         "24h"
+                    ]
+                },
+                "certKeyBitSize": {
+                    "$id": "#/properties/OpenServiceMesh/properties/certKeyBitSize",
+                    "type": "integer",
+                    "title": "The certKeyBitSize schema",
+                    "description": "The key size for data plane certificates.",
+                    "examples": [
+                        2048
                     ]
                 },
                 "caBundleSecretName": {
--- a/charts/osm/values.schema.json
+++ b/charts/osm/values.schema.json
@@ -131,9 +131,6 @@
             "required": [
                 "image",
                 "sidecarImage",
-                "certificateManager",
-                "serviceCertValidityDuration",
-                "certKeyBitSize",
                 "caBundleSecretName",
                 "enableDebugServer",
                 "enablePermissiveTrafficPolicy",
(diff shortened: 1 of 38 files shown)


pkg/certificate/providers/tresor/certificate_manager.go
```diff
@@ -22,7 +22,11 @@ func (cm *CertManager) issue(cn certificate.CommonName, validityPeriod time.Dura
 		return nil, errNoIssuingCA
 	}
 
-	certPrivKey, err := rsa.GenerateKey(rand.Reader, rsaBits)
+	certKeyBitSize := rsaBits
```
shashankram, line 25: Let's drop the default `rsaBits` here, and encode a default for this in `pkg/configurator` similar to `defaultServiceCertValidityDuration`.

pkg/certificate/providers/tresor/certificate_manager.go
```diff
@@ -22,7 +22,11 @@ func (cm *CertManager) issue(cn certificate.CommonName, validityPeriod time.Dura
...
+	if cm.cfg != nil {
+		certKeyBitSize = cm.cfg.GetCertKeyBitSize()
+	}
```
shashankram, line 28: If tests are passing `cfg` as `nil`, we should fix those instead. Non nil `cfg` should not be allowed when initializing a certificate provider, so we could drop this check.

pkg/configurator/methods.go
```diff
@@ -16,6 +16,15 @@ import (
 const (
 	// defaultServiceCertValidityDuration is the default validity duration for service certificates
 	defaultServiceCertValidityDuration = 24 * time.Hour
+
+	// defaultCertKeyBitSize is the default certificate key bit size
+	defaultCertKeyBitSize = 2048
+
+	// minCertKeyBitSize is the minimum certificate key bit size
+	minCertKeyBitSize = 512
```
shashankram, line 24: I'd set the min to `2048`. Decreasing this to 512 allows for a much lower key strength, which I don't see a use case for atm.

tests/framework/common.go
```diff
@@ -489,6 +491,10 @@ func (td *OsmTestData) InstallOSM(instOpts InstallOSMOpts) error {
 		)
 	}
 
+	instOpts.SetOverrides = append(instOpts.SetOverrides,
+		fmt.Sprintf("OpenServiceMesh.certKeyBitSize=%d", instOpts.CertBitSize),
+	)
+
```
shashankram, line 497: We shouldn't need this here, we should set a default key size in `pkg/configurator` if not specified via `--set`.

shashankram (dismissed):
Hey @trstringer this looks pretty good other than the few comments I left.

charts/osm/values.yaml
```diff
@@ -93,6 +93,9 @@ OpenServiceMesh:
   # -- Service certificate validity duration for certificate issued to workloads to communicate over mTLS
   serviceCertValidityDuration: 24h
 
+  # -- Certificate key bit size for data plane certificates issued to workloads to communicate over mTLS
```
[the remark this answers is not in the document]
snehachhabria: Maybe certificate manager too can go under the same object

shashankram:
Discussed this with @trstringer. There is an interesting implication of this change on the certificate issuance tool that I am working on for Ingress. Based on our offline discussion, we will make some changes to this PR to make the certificate manager dynamic properties configurable at initialization time, and allow it to be retrieved from `Configurator` if not set.

pkg/certificate/providers/tresor/certificate_manager.go
shashankram: Not required with the new proposal: https://github.com/openservicemesh/osm/pull/3837#issuecomment-885248364

charts/osm/values.yaml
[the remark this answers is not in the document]
shashankram: I think it would be a good idea to do something like:
```
certificateProvider:
  kind: <tresor|cert-manager.io|vault>
  serviceCertValiditityDuration: 24h
  rsaKeyBitSize: 4096
  ...
```

and map to the MeshConfig preset-mesh-config.yaml template.

cmd/cli/install.go
```diff
@@ -236,8 +236,8 @@ func (i *installCmd) validateOptions() error {
 	}
 
 	if setOptions, ok := s["OpenServiceMesh"].(map[string]interface{}); ok {
-		// if certificateManager is vault, ensure all relevant information (vault-host, vault-token) is available
-		if setOptions["certificateManager"] == "vault" {
+		// if the certificate manager is vault, ensure all relevant information (vault-host, vault-token) is available
```
ksubrmnn, line 239:
```suggestion
		// if the certificate provider kind is vault, ensure all relevant information (vault-host, vault-token) is available
```

charts/osm/values.schema.json
```diff
@@ -259,33 +256,46 @@
...
-                        2048
-                    ]
+                "certficateProvider": {
+                    "$id": "#/properties/OpenServiceMesh/properties/certificateProvider",
+                    "type": "object",
+                    "title": "The certificate provider schema",
+                    "description": "Certificate provider configuration parameters",
+                    "required": [
+                        "kind",
+                        "issuerKind",
```
ksubrmnn, line 266: Should this be `serviceCertValidityDuration`

charts/osm/values.schema.json
```diff
@@ -259,33 +256,46 @@
...
+                        "issuerGroup"
```
ksubrmnn, line 267: Should this be `certKeyBitSize`

charts/osm/values.schema.json
```diff
@@ -258,24 +256,46 @@
...
+                        },
+                        "certKeyBitSize": {
+                            "$id": "#/properties/OpenServiceMesh/properties/certificateProvider/properties/certKeyBitSize",
+                            "type": "integer",
+                            "title": "The certKeyBitSize schema",
+                            "description": "The key size for data plane certificates.",
+                            "examples": [
+                                2048
+                            ]
+                        }
```
shashankram, line 297: set `"additionalProperties": false` so that invalid keys are rejected for the root property.

pkg/certificate/providers/certmanager/certificate_manager.go
```diff
@@ -239,6 +245,8 @@ func NewCertManager(
 	namespace string,
 	issuerRef cmmeta.ObjectReference,
 	cfg configurator.Configurator,
+	validityPeriod time.Duration,
```
shashankram, line 248: s/validityPeriod/serviceCertValidityDuration

Same comment applies to other certificate provider init.

pkg/certificate/providers/certmanager/certificate_manager.go
```diff
@@ -82,7 +82,10 @@ func (cm *CertManager) getFromCache(cn certificate.CommonName) certificate.Certi
 func (cm *CertManager) RotateCertificate(cn certificate.CommonName) (certificate.Certificater, error) {
 	start := time.Now()
 
-	newCert, err := cm.issue(cn, cm.cfg.GetServiceCertValidityPeriod())
+	if cm.validityPeriod == 0 {
+		cm.validityPeriod = cm.cfg.GetServiceCertValidityPeriod()
+	}
```
shashankram, line 87: I'd leave a comment here why this is being done this way, otherwise it could be confusing.

pkg/certificate/providers/tresor/certificate_manager.go
```diff
@@ -159,7 +162,10 @@ func (cm *CertManager) RotateCertificate(cn certificate.CommonName) (certificate
 		return nil, errors.Errorf("Old certificate does not exist for CN=%s", cn)
 	}
 
-	newCert, err := cm.issue(cn, cm.cfg.GetServiceCertValidityPeriod())
+	if cm.validityPeriod == 0 {
+		cm.validityPeriod = cm.cfg.GetServiceCertValidityPeriod()
+	}
```
shashankram, line 167: Let's leave a comment here as to why this logic is required.

tests/framework/types.go
```diff
@@ -93,6 +93,7 @@ type InstallOSMOpts struct {
 	CertmanagerIssuerGroup string
 	CertmanagerIssuerKind  string
 	CertmanagerIssuerName  string
+	CertBitSize            int
```
michelleN, line 96: super small nit: could we call with `CertKeyBitSize` for consistency?

trstringer: Good point! Done!

shashankram (approved):
Hey @trstringer, this looks good. Remember to use the `squash & merge` option while merging this PR instead of creating a merge commit (because the commits in this PR aren't squashed).

snehachhabria (approved):

PR #3839, opened 2021-07-22
feat(demo): Make timeout configurable

ksubrmnn:

**Description**:
This makes the timeouts for osm install and the kubectl --wait commands configurable. This is needed for the demo to work well on different clusters which may require more time for operations to complete (such as OpenShift)

**Affected area**:

No
1. Is this a breaking change?
No


demo/clean-kubernetes.sh
```diff
@@ -5,14 +5,22 @@ set -aueo pipefail
...
 
 for ns in "$BOOKWAREHOUSE_NAMESPACE" "$BOOKBUYER_NAMESPACE" "$BOOKSTORE_NAMESPACE" "$BOOKTHIEF_NAMESPACE" "$K8S_NAMESPACE"; do
-    kubectl delete namespace "$ns" --ignore-not-found --wait &
+    kubectl delete namespace "$ns" --ignore-not-found --wait --timeout="$TIMEOUT" &
 done
 
 # Clean up Hashicorp Vault deployment
-kubectl delete deployment vault -n "$K8S_NAMESPACE" --ignore-not-found --wait &
-kubectl delete service vault -n "$K8S_NAMESPACE" --ignore-not-found --wait &
+kubectl delete deployment vault -n "$K8S_NAMESPACE" --ignore-not-found --wait --timeout="$TIMEOUT" &
```
shashankram, line 23: What's the default timeout, and what happens if the configures timeout exceeds?

ksubrmnn: from kubectl delete --help:

```
--timeout=0s: The length of time to wait before giving up on a delete, zero means determine a timeout from the
size of the object
```

However, I have found this is not reliable and when running the automated demo from CI on openshift, the delete command returns but the namespace is still in a "deleting state" despite the --wait flag being supplied. So the next step in the demo (to create the ns) fails.

demo/clean-kubernetes.sh
```diff
@@ -5,14 +5,22 @@ set -aueo pipefail
 # shellcheck disable=SC1091
 source .env
 
+MESH_NAME="${MESH_NAME:-osm}"
+K8S_NAMESPACE="${K8S_NAMESPACE:-osm-system}"
+BOOKBUYER_NAMESPACE="${BOOKBUYER_NAMESPACE:-bookbuyer}"
+BOOKSTORE_NAMESPACE="${BOOKSTORE_NAMESPACE:-bookstore}"
+BOOKTHIEF_NAMESPACE="${BOOKTHIEF_NAMESPACE:-bookthief}"
+BOOKWAREHOUSE_NAMESPACE="${BOOKWAREHOUSE_NAMESPACE:-bookwarehouse}"
```
shashankram, line 13: Why is this required? These should already be set when `run-osm-demo.sh` invokes this script?

ksubrmnn: I personally use this script outside of the context of the automated demo, such as the manual demo

shashankram: Got it. Instead of copying the same set of env into each script, set these in your `.env` instead? That's one of the reasons we have `.env`, so different users can customize it based on their use case.

shashankram (approved):

SanyaKochhar (approved):

PR #3855, opened 2021-07-23
feat(windows): Adds node selector to pods deployed by tests

davinci26:
Adds node selector to the spec of all pods deployed in e2e tests.
Also adds a new flag in the test cli to be able to configure the OS in preparation for
Windows testing.

### DIFF
--- a/tests/e2e/e2e_grpc_secure_origination_test.go
+++ b/tests/e2e/e2e_grpc_secure_origination_test.go
@@ -45,16 +45,18 @@ func testSecureGRPCTraffic() {
 		}
 
 		// Get simple pod definitions for the gRPC server
-		svcAccDef, podDef, svcDef := Td.SimplePodApp(
+		svcAccDef, podDef, svcDef, err := Td.SimplePodApp(
 			SimplePodAppDef{
 				Name:        destName,
 				Namespace:   destName,
 				Image:       "moul/grpcbin",
 				Ports:       []int{grpcbinSecurePort},
 				AppProtocol: "tcp",
+				OS:          Td.ClusterOS,
 			})
+		Expect(err).NotTo(HaveOccurred())
 
-		_, err := Td.CreateServiceAccount(destName, &svcAccDef)
+		_, err = Td.CreateServiceAccount(destName, &svcAccDef)
 		Expect(err).NotTo(HaveOccurred())
 		_, err = Td.CreatePod(destName, podDef)
 		Expect(err).NotTo(HaveOccurred())
--- a/tests/framework/common.go
+++ b/tests/framework/common.go
@@ -133,6 +133,7 @@ func registerFlags(td *OsmTestData) {
 	flag.StringVar((*string)(&td.CollectLogs), "collectLogs", string(CollectLogsIfErrorOnly), "Defines if/when to collect logs.")
 
 	flag.StringVar(&td.ClusterName, "kindClusterName", "osm-e2e", "Name of the Kind cluster to be created")
+
 	flag.BoolVar(&td.CleanupKindCluster, "cleanupKindCluster", true, "Cleanup kind cluster upon exit")
 	flag.BoolVar(&td.CleanupKindClusterBetweenTests, "cleanupKindClusterBetweenTests", false, "Cleanup kind cluster between tests")
 	flag.StringVar(&td.ClusterVersion, "kindClusterVersion", "", "Kind cluster version, ex. v.1.20.2")
@@ -209,6 +210,8 @@ func (td *OsmTestData) InitTestData(t GinkgoTInterface) error {
 	td.TestID = r.Uint64()
 	td.TestDirName = fmt.Sprintf("test-%d", td.TestID)
 	td.T.Log(color.HiGreenString("> ID for test: %d, Test dir (abs): %s", td.TestID, td.GetTestDirPath()))
+
+	td.ClusterOS = constants.OSLinux
 
 	// String parameter validation
 	err = td.ValidateStringParams()
--- a/tests/framework/common_apps.go
+++ b/tests/framework/common_apps.go
@@ -161,11 +161,16 @@ type SimplePodAppDef struct {
 	Args        []string
 	Ports       []int
 	AppProtocol string
+	OS          string
 }
 
 // SimplePodApp returns a set of k8s typed definitions for a pod-based k8s definition.
 // Includes Pod, Service and ServiceAccount types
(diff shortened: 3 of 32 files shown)


davinci26:
Lint does not work locally so I am going to iterate on the CI to ensure that everything looks ok

tests/framework/types.go
```diff
@@ -53,6 +53,7 @@ type OsmTestData struct {
 
 	// Kind cluster related vars
 	ClusterName                    string // Kind cluster name (used if kindCluster)
+	ClusterOS                      string // The operating system of the working nodes in the cluster. Currently only linux is supported.
```
ksubrmnn, line 56: I think it's okay to start out by scheduling these tests to work either on Linux workers or Windows workers, but we need to eventually include tests that validate Linux <-> Windows traffic

ksubrmnn: I'd add a note here that testing mixed OS traffic is not yet supported.

ksubrmnn: Also this var should be moved out of the "Kind cluster related vars" section since it's not applicable to kind

tests/framework/common_apps.go
```diff
@@ -170,6 +171,9 @@ func (td *OsmTestData) SimplePodApp(def SimplePodAppDef) (corev1.ServiceAccount,
 		Spec: corev1.PodSpec{
 			TerminationGracePeriodSeconds: new(int64), // 0
 			ServiceAccountName:            def.Name,
+			NodeSelector: map[string]string{
```
ksubrmnn, line 174: For this function and the one for SimpleDeploymentAppDef, I think you could just a local variable to be the DefaultOS if the string is empty, and then take the value if it is not. That way you don't have to modify all of the other tests to pass this in.

ksubrmnn (changes requested):

tests/framework/common_apps.go
davinci26: No strong opinions though

ksubrmnn: Images can't be cross platform across Linux and Windows though, unless something has changed. Linux nodes will always have different images than Windows nodes

davinci26: > Images can't be cross platform across Linux and Windows though,

I meant that in the sense that there is a corresponding windows image. Not that the same image will be used in both platforms.

ksubrmnn: I think modifying the tests for hybrid traffic will require a bit more planning/discussion, but I definitely see your point.

Since we're not at a stage where we are testing hybrid traffic yet (as of this PR), can we scope this PR to not modify the individual tests and keep the login in common_apps? That way the changes are more localized and changes can be made easily as you iterate in the future.

tests/framework/types.go
ksubrmnn: I agree, it'll be better to have a few dedicated tests to handle mixed OS traffic and leave the rest as is.

davinci26: fixed with the latest update

tests/framework/common_apps.go
davinci26: fixed with the latest update. I hope that you meant to add a return `err` value that tests should verify

tests/framework/common.go
```diff
@@ -133,6 +144,8 @@ func registerFlags(td *OsmTestData) {
 	flag.StringVar((*string)(&td.CollectLogs), "collectLogs", string(CollectLogsIfErrorOnly), "Defines if/when to collect logs.")
 
 	flag.StringVar(&td.ClusterName, "kindClusterName", "osm-e2e", "Name of the Kind cluster to be created")
+	flag.StringVar(&td.ClusterOS, "ClusterOS", DefaultClusterOS, "The operating system of the working nodes in the cluster. Currently only linux is supported.")
```
ksubrmnn, line 147: I'm not sure if having this field is necessary. To me this implies that the tests are configurable based on the value supplied here which isn't the case right now. At the moment we only have Linux e2es and we can't run Linux images on Windows machines. Eventually we'd need to add some abstraction where we can deploy clients/servers on Windows or Linux based on some flag the user passes in, but that seems outside the scope of just adding nodeSelectors.

This flag could be used in the future when we have Windows e2es and hybrid e2es and we want the ability to specify which ones to run, but we're not there yet.

It's still necessary to include the nodeSelectors because these tests should be able to run on hybrid clusters. In that case, I think it'll be best to make this change as minimal as possible right now by keeping the OS parameter for the pod/deployment definitions. But it should be explicitly programmed to correspond to the container image supplied, and not configurable based on how the test is run.

davinci26: removed the flag with the latest commit

davinci26:
@ksubrmnn @shashankram PTAL

tests/framework/types.go
```diff
@@ -57,6 +57,8 @@ type OsmTestData struct {
 	CleanupKindCluster             bool   // Cleanup kind cluster upon test finish
 	ClusterVersion                 string // Kind cluster version, ex. v1.20.2
 
+	ClusterOS string // The operating system of the working nodes in the cluster. Mixed OS traffic is not supported.
```
ksubrmnn, line 60: Please remove this flag from this PR and scope this PR to just adding linux node selectors to the tests.

tests/e2e/e2e_grpc_secure_origination_test.go
```diff
@@ -45,16 +45,18 @@ func testSecureGRPCTraffic() {
...
 		// Get simple pod definitions for the gRPC server
-		svcAccDef, podDef, svcDef := Td.SimplePodApp(
+		svcAccDef, podDef, svcDef, err := Td.SimplePodApp(
 			SimplePodAppDef{
 				Name:        destName,
 				Namespace:   destName,
 				Image:       "moul/grpcbin",
 				Ports:       []int{grpcbinSecurePort},
 				AppProtocol: "tcp",
+				OS:          Td.ClusterOS,
```
ksubrmnn, line 55: This should be hardcoded to be "linux" since the image is hardcoded as well. The image and the nodeSelector should be coupled

ksubrmnn (changes requested):

tests/framework/types.go
ksubrmnn: I see this is just part of the config, not exposed as a flag.

ksubrmnn (approved):

tests/e2e/e2e_grpc_secure_origination_test.go
ksubrmnn: This is being set as the linux OS constant elsewhere

michelleN (approved):

PR #3861, opened 2021-07-27
Deploy Windows Sidecar for Windows worker nodes

davinci26:
Makes OSM patch to deploy the Windows version of Envoy sidecar
when the worker node selector is Windows.


pkg/constants/constants.go
```diff
@@ -61,6 +61,11 @@ const (
 	// DefaultEnvoyImage is the default envoy proxy sidecar image if not defined in the osm MeshConfig
 	DefaultEnvoyImage = "envoyproxy/envoy-alpine:v1.18.3"
 
+	// DefaultEnvoyWindowsImage is the default envoy proxy windows sidecar image if not defined in the osm MeshConfig
+	// TODO(davinci26): This should be updated to the nanoserver based image when it becomes available
```
ksubrmnn, line 65: Can you include a reference to an OSM issue corresponding to this work?

pkg/injector/envoy_container.go
```diff
@@ -15,21 +15,46 @@ const (
...
+	}
+}
+
+func getEnvoySidecarContainerSpec(pod *corev1.Pod, cfg configurator.Configurator, originalHealthProbes healthProbes, isWindowsPod bool) corev1.Container {
 	// cluster ID will be used as an identifier to the tracing sink
 	clusterID := fmt.Sprintf("%s.%s", pod.Spec.ServiceAccountName, pod.Namespace)
 
+	var securityContext *corev1.SecurityContext
+	var containerImage string
+	if isWindowsPod {
```
ksubrmnn, line 44: Can you consolidate these into one method that takes the OS type as a param?

pkg/injector/envoy_container.go
```diff
@@ -15,21 +15,46 @@ const (
...
+			return &uid
+		}(),
+	}
+}
+
+func getWindowsSecurityContext() *corev1.SecurityContext {
+	return &corev1.SecurityContext{
+		WindowsOptions: &corev1.WindowsSecurityContextOptions{
+			RunAsUserName: func() *string {
+				userName := "EnvoyUser"
```
ksubrmnn, line 31: I thought we were using one of the well known SIDs so that we could do prefix matching?

pkg/constants/constants.go
davinci26: Will add an issue. Do you do code TODO without issues in general?

pkg/injector/envoy_container.go
davinci26: Yeah you are right, havent fully internalized that golang has function with multiple return variables

pkg/injector/envoy_container.go
davinci26: Yeah we do but there is the following caveat:

In Envoy [Dockerfile](https://github.com/envoyproxy/envoy/blob/57bfbda2a50d2f026cf64c82383fe3bebe1e0860/ci/Dockerfile-envoy-windows#L8) we create the EnvoyUser and add it to the `Network Configuration Operators` group so we can do match prefix with this group

pkg/constants/constants.go
shashankram: // TODO(#<issue>): ...

pkg/injector/envoy_container.go
shashankram: Define a constant for `EnvoyUser` in this file.

pkg/injector/patch.go
```diff
@@ -54,22 +54,29 @@ func (wh *mutatingWebhook) createPatch(pod *corev1.Pod, req *admissionv1.Admissi
...
-
-	// Build inbound port exclusion list
-	podInboundPortExclusionList, _ := wh.getPortExclusionListForPod(pod, namespace, inboundPortExclusionListAnnotation)
-	globalInboundPortExclusionList := wh.configurator.GetInboundPortExclusionList()
-	inboundPortExclusionList := mergePortExclusionLists(podInboundPortExclusionList, globalInboundPortExclusionList)
-
-	// Add the Init Container
-	initContainer := getInitContainerSpec(constants.InitContainerName, wh.configurator, wh.configurator.GetOutboundIPRangeExclusionList(), outboundPortExclusionList, inboundPortExclusionList, wh.configurator.IsPrivilegedInitContainer())
-	pod.Spec.InitContainers = append(pod.Spec.InitContainers, initContainer)
+	// On Windows we cannot use init containers to program HNS because it requires elevated privillages
```
shashankram, line 57:
```suggestion
	// On Windows we cannot use init containers to program HNS because it requires elevated privileges
```

pkg/injector/patch.go
```diff
@@ -54,22 +54,29 @@ func (wh *mutatingWebhook) createPatch(pod *corev1.Pod, req *admissionv1.Admissi
...
+	// As a result we assume that the HNS redirection policies are already programmed via a CNI plugin.
+	// Skip adding the init container and only patch the pod spec with sidecar container.
+	podOS := pod.Spec.NodeSelector["kubernetes.io/os"]
+	isWindowsPod := podOS == "windows"
```
shashankram, line 61: Let's make this check case insensitive.

pkg/injector/envoy_container.go
shashankram: Instead of passing `isWindows` bool around, let's directly pass `osType` as a string argument and have helper functions use this for appropriate config.

pkg/injector/patch_test.go
```diff
@@ -106,16 +129,23 @@ func TestCreatePatch(t *testing.T) {
 				nonInjectNamespaces: mapset.NewSet(),
 			}
 
+			if tc.os == "windows" {
+				mockConfigurator.EXPECT().GetEnvoyImage().Return("").Times(1)
+				mockConfigurator.EXPECT().GetEnvoyWindowsImage().Return("").Times(1)
+			} else {
+				mockConfigurator.EXPECT().GetEnvoyImage().Return("").Times(1)
+				mockConfigurator.EXPECT().GetEnvoyWindowsImage().Return("").Times(0)
+			}
```
shashankram, line 138: To avoid such conditionals with mocks, initialize the mock object within the loop for each test case in the table of tests and set both to `AnyTimes()`

davinci26: fixed in a8f1df1ea78e0932a6f67f3c51cdb6c6bbf8eb8f

pkg/injector/patch.go
davinci26: Thanks for catching this, fixed in a8f1df1ea78e0932a6f67f3c51cdb6c6bbf8eb8f

pkg/constants/constants.go
davinci26: fixed now, Delyan had some TODO with different formats and they appeared first on my search so I got the format wrong

pkg/injector/envoy_container.go
davinci26: fixed

pkg/injector/patch.go
davinci26: makes sense, fixed

michelleN (approved):

davinci26:
@shashankram @ksubrmnn PTAL and let me if I can help with anything

pkg/injector/envoy_container.go
```diff
@@ -15,21 +15,36 @@ const (
 	envoyProxyConfigPath     = "/etc/envoy"
 )
 
-func getEnvoySidecarContainerSpec(pod *corev1.Pod, cfg configurator.Configurator, originalHealthProbes healthProbes) corev1.Container {
+func getPlatformSpecificSpecComponents(cfg configurator.Configurator, podOS string) (*corev1.SecurityContext, string) {
```
shashankram, line 18: nit: with multiple return values (where value type != error), it's always good to add a function comment regarding the return values, or use named return values such as https://tour.golang.org/basics/7.

pkg/injector/patch.go
```diff
@@ -54,22 +55,28 @@ func (wh *mutatingWebhook) createPatch(pod *corev1.Pod, req *admissionv1.Admissi
...
-	inboundPortExclusionList := mergePortExclusionLists(podInboundPortExclusionList, globalInboundPortExclusionList)
-
-	// Add the Init Container
-	initContainer := getInitContainerSpec(constants.InitContainerName, wh.configurator, wh.configurator.GetOutboundIPRangeExclusionList(), outboundPortExclusionList, inboundPortExclusionList, wh.configurator.IsPrivilegedInitContainer())
-	pod.Spec.InitContainers = append(pod.Spec.InitContainers, initContainer)
+	// On Windows we cannot use init containers to program HNS because it requires elevated privileges
+	// As a result we assume that the HNS redirection policies are already programmed via a CNI plugin.
+	// Skip adding the init container and only patch the pod spec with sidecar container.
+	podOS := pod.Spec.NodeSelector["kubernetes.io/os"]
+	if !strings.EqualFold(podOS, "windows") {
```
shashankram, line 62: nit: define "windows" as a constant in `constants.go` if this is going to be used in other files as well.

shashankram (approved):
Looks good. Left minor comments that can be addressed in a separate PR.

PR #3863, opened 2021-07-27
Fix broken lint make rule and target the CI golangci-lint version

trstringer:
**Description**:

This rule appeared to have had two issues prior to this PR.

1. It did not run correctly/successfully:

```
$ make go-lint
go run github.com/golangci/golangci-lint/cmd/golangci-lint run --config .golangci.yml
# github.com/golangci/golangci-lint/pkg/golinters
../../go/pkg/mod/github.com/golangci/golangci-lint@v1.32.2/pkg/golinters/unused.go:16:7: undefined: unused.NewChecker
make: *** [Makefile:140: go-lint] Error 2
```

2. The version of golangci-lint that this was using (1.32) was not the same as the version for CI (1.30). These versions should be the same so that the local experience matches the CI pipeline.

This PR fixes both of these points by checking the currently installed golangci-lint bin. If it does not exist or if it is not the CI version (1.30) then it will install it. After that, it will lint with the same options that are used in CI.

I understand that the recommendation from golangci-lint is to use the GitHub Action, but we should consider creating a shell script to handle this logic, so that it can be invoked from the CI pipeline as well as our local dev machines. As it stands now, if the GitHub Action version changes, we will be out of sync again. The same situation applies if we use different args to lint.

**Affected area**:

No.

1. Is this a breaking change?

No.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -137,7 +137,15 @@ go-vet:
 
 .PHONY: go-lint
 go-lint: cmd/cli/chart.tgz pkg/envoy/lds/stats.wasm
-	go run github.com/golangci/golangci-lint/cmd/golangci-lint run --config .golangci.yml
+	@if ! golangci-lint version 2>&1 | grep 1.30; then \
+		echo Installing golangci-lint; \
+		go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.30; \
+	fi; \
+	golangci-lint run \
+		--config .golangci.yml \
+		--timeout 5m \
+		--skip-dirs pkg/gen \
+		--skip-files "zz_generated.deepcopy.go$$"
 
 .PHONY: go-fmt
 go-fmt:
--- a/Makefile
+++ b/Makefile
@@ -137,15 +137,7 @@ go-vet:
 
 .PHONY: go-lint
 go-lint: cmd/cli/chart.tgz pkg/envoy/lds/stats.wasm
-	@if ! golangci-lint version 2>&1 | grep 1.30; then \
-		echo Installing golangci-lint; \
-		go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.30; \
-	fi; \
-	golangci-lint run \
-		--config .golangci.yml \
-		--timeout 5m \
-		--skip-dirs pkg/gen \
-		--skip-files "zz_generated.deepcopy.go$$"
+	docker run --rm -v $$(pwd):/app -w /app golangci/golangci-lint:v1.41.1 golangci-lint run --config .golangci.yml
 
 .PHONY: go-fmt
 go-fmt:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -49,7 +49,7 @@ jobs:
       - name: golangci-lint
         uses: golangci/golangci-lint-action@v2
         with:
-          version: v1.30
+          version: v1.41.1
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -137,7 +137,15 @@ go-vet:
 
 .PHONY: go-lint
 go-lint: cmd/cli/chart.tgz pkg/envoy/lds/stats.wasm
-	go run github.com/golangci/golangci-lint/cmd/golangci-lint run --config .golangci.yml
+	@if ! golangci-lint version 2>&1 | grep 1.30; then \
+		echo Installing golangci-lint; \
+		go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.30; \
```
shashankram, line 142: If CI can use the latest because the version is not hardcoded, why not keep the same behavior for local checks? It would be good to avoid constantly updating this version hardcoded here.

trstringer: CI does have this [hardcoded](https://github.com/openservicemesh/osm/blob/main/.github/workflows/main.yml#L52). I don't necessarily have a problem with periodic updating of version numbers, but the fact that there are multiple sources of truth for the golangci-lint versioning is what I think isn't great and could probably use some more thinking.

trstringer: And I think what we don't want to do is have some sliding window label like `latest`, because my assumptions would be that `latest` for the GitHub Actions component and upstream are probably different. So then we are accepting a bit of version drift there between local and CI.

shashankram: I agree it is better to use the same version. I've never had to run this check locally because my IDE uses golangci-lint as the core linter, and it keeps updating it to the latest version anyway.

Makefile
```diff
@@ -137,7 +137,15 @@ go-vet:
...
+	fi; \
+	golangci-lint run \
```
shashankram, line 144: What if there are multiple versions installed already?

Makefile
```diff
@@ -137,7 +137,15 @@ go-vet:
...
+		--config .golangci.yml \
+		--timeout 5m \
+		--skip-dirs pkg/gen \
+		--skip-files "zz_generated.deepcopy.go$$"
```
shashankram, line 148: We should include this in `.golangci.yml`

Makefile
trstringer: It looks like `go install` will just overwrite the bin with the desired version. Perhaps the better approach is to use a [golang-ci container](https://golangci-lint.run/usage/install/#docker) to accomplish this. I personally don't mind the overwriting behavior, but I can understand if that's not desirable for other devs or just by principle alone.

shashankram: I worry overwriting the existing version will make IDE managed environments unhappy when such linters are installed and managed by IDEs. Is there a way to avoid overwriting and reference the version specific binary directly?

trstringer: It looks like this isn't possible with `go install`. I think the next best thing is to go the container approach to prevent this overwriting issue.

shashankram: I didn't mean for this to be a blocker. Maybe add a note in the script and provide a flag so the overwriting can be optional?

trstringer: I think the container implementation is better. We already expect docker on our local machines, easiest to use the version based off of an image. I also took the opportunity to start using golangci-lint 1.41.1.

shashankram (approved):

ksubrmnn (approved):

Issue #3281, opened 2021-05-03
Validating webhook for custom *.openservicemesh.io resources


shashankram:
**Please describe the Improvement and/or Feature Request**
Add a validating webhook to validate resources (currently Egress, MeshConfig) in the *.openservicemesh.io API group.
The webhook will be responsible for validating the resource when it is created/updated.

**Scope (please mark with X where applicable)**
- New Functionality          [X]
- Install                    [ ]
- SMI Traffic Access Policy  [ ]
- SMI Traffic Specs Policy   [ ]
- SMI Traffic Split Policy   [ ]
- Permissive Traffic Policy  [ ]
- Ingress                    [ ]
- Egress                     [ ]
- Envoy Control Plane        [ ]
- CLI Tool                   [ ]
- Metrics                    [ ]
- Certificate Management     [ ]
- Sidecar Injection          [ ]
- Logging                    [ ]
- Debugging                  [ ]
- Tests                      [ ]
- CI System                  [ ]
- Demo                       [ ]
- Project Release            [ ]

**Possible use cases**
To prevent user misconfigurations when using the custom resource.

steeling:
Shashank, can you comment on some of the validations we will be adding?

PR #3868, opened 2021-07-27
osm-validator: enable webhook to validate resources

shashankram:
**Description**:
This change enables osm-validator, a validating webhook
to validate k8s resources. It addresses existing issues
with the webhook being broken, and updates the name of
the validating webhook and associated flag for clarity
and correctness.
It also updates the min service cert validity duration
to 2 minutes which takes into account the noise added
while performing certificate rotation.

The validations are only enabled for features that have
been tested to work with the validator.

Part of #3281

**Affected area**:

### DIFF
--- a/pkg/validator/server.go
+++ b/pkg/validator/server.go
@@ -6,11 +6,9 @@ import (
 	"crypto/tls"
 	"encoding/json"
 	"fmt"
-	"io"
 	"net/http"
 
 	"github.com/pkg/errors"
-	"github.com/rs/zerolog/log"
 	admissionv1 "k8s.io/api/admission/v1"
 	"k8s.io/client-go/kubernetes"
 
@@ -21,8 +19,11 @@ import (
 var (
 	defaultValidators = map[string]Validator{}
 
-	// ValidatingWebhookPath is the path for the validating webhook.
-	ValidatingWebhookPath = "/validate"
+	// validationAPIPath is the API path for performing resource validations
+	validationAPIPath = "/validate"
+
+	// HealthAPIPath is the API path for health check
+	HealthAPIPath = "/healthz"
 )
 
 // RegisterValidator registers all validators. It is not thread safe.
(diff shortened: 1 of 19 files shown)


pkg/validator/server.go
```diff
@@ -98,27 +99,50 @@ func NewValidatingWebhook(webhookConfigName string, port int, certificater certi
...
+		log.Error().Err(err).Msgf("Responded to admission request with HTTP %v", http.StatusUnsupportedMediaType)
 		return
 	}
 
-	data, err := json.Marshal(s.handleValidation(adReq))
+	admissionRequestBody, err := webhook.GetAdmissionRequestBody(w, req)
 	if err != nil {
-		http.Error(w, err.Error(), http.StatusInternalServerError)
-		log.Error().Err(err).Msgf("Error marshaling admission response body; Responded to admission request with HTTP %v", http.StatusInternalServerError)
+		// Error was already logged and written to the ResponseWriter
```
nojnhuh, line 114: Is there any particular reason `GetAdmissionRequest()` needs to log the error and also return it? Not that there's anything wrong with this, just sticks out as a bit odd. I'd expect the error to be logged here vs. inside the function.

pkg/validator/validators.go
```diff
@@ -30,17 +36,49 @@ func init() {
...
+					break
+				}
+			}
+
+			if backend.TLS.SkipClientCertValidation && !authenticatedSourceFound {
+				return nil, errors.Errorf("HTTPS ingress with client certificate validation enabled must specify at least one 'AuthenticatedPrincipal` source")
+			}
+
+		default:
+			return nil, errors.Errorf("Expected 'Port.Protocol' to be 'http|https', got: %s", backend.Port.Protocol)
```
nojnhuh, line 74: Might be worth adding the full yaml/json path in the error so users also have the `spec.backends[i]` context and know right where to look.

And not a big deal, but IIRC validation errors usually use the yaml/json field names (`port.protocol`) vs. the Go struct field names (`Port.Protocol`).

I might also change `'http|https'` to `'http' or 'https'` just to make that clear. Based on the message as-is, I wouldn't fault a user for copy-pasting `http|https` into the YAML.

snehachhabria (approved):

pkg/validator/validators_test.go
```diff
@@ -10,13 +9,204 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 )
 
+func TestIngressBackendValidator(t *testing.T) {
+	assert := tassert.New(t)
```
nojnhuh, line 13: Let's move this inside the `t.Run()` body in the loop so the error messages have the subtest context (#3745).

pkg/validator/validators_test.go
```diff
@@ -10,13 +9,204 @@ import (
...
+									}
+								}
+							]
+						}
+					}
+					`),
+				},
+			},
+			expResp:   nil,
+			expErrStr: "Expected 'Port.Protocol' to be 'http|https', got: invalid",
```
nojnhuh, line 80: Might be overkill for now, but one way to get around hardcoding error strings in tests like this is to define a new type that implements the `error` interface or a function that returns an error so then in the code you can `return badPortProtocol{protocol: "invalid"}` or `return badPortProtocol("invalid")` and use the same type/function here in the tests. That makes tweaking the messages easy without having to worry about the tests.

nojnhuh (changes requested):
Left one comment to update the osm-controller command line flag referenced in its flag validation. Everything else is either a comment, question, or light suggestion.

pkg/validator/server.go
shashankram: `GetAdmissionRequestBody()` logs an error and writes an error response on the http server. It seems easier to write the HTTP response error and log it in the same place, for context.

pkg/validator/validators_test.go
shashankram: I am not a big fan of hardcoding error messages either, but I didn't want to refactor the existing tests too much, so the scope of this PR remains as small as possible. This can be an enhancement that can be done later on.

nojnhuh (approved):

PR #3876, opened 2021-07-28
chore(demo): Only create creds for private registry

nshankar13:

**Description**: Use PRIVATE_REGISTRY flag to skip credential creation step in demo if images are being pulled from a public registry. 

**Affected area**:

### DIFF
--- a/scripts/create-container-registry-creds.sh
+++ b/scripts/create-container-registry-creds.sh
@@ -13,7 +13,9 @@ if [[ "${CI:-}" == "true" ]]; then
     CTR_REGISTRY_PASSWORD=${DOCKER_PASS:-}
 fi
 
-[[ -z "$CTR_REGISTRY_PASSWORD" ]] && exit 0
+if [[ "$USE_PRIVATE_REGISTRY" = false ]]; then
+    exit 0
+fi
 
 REGISTRY=$(echo "$CTR_REGISTRY" | awk -F'.' '{print $1}')
 REGISTRY_URL=$(echo "$CTR_REGISTRY" | awk -F'.' '{print $1 "." $2 "." $3}')
--- a/demo/run-osm-multicluster-demo.sh
+++ b/demo/run-osm-multicluster-demo.sh
@@ -41,6 +41,7 @@ DEPLOY_WITH_SAME_SA="${DEPLOY_WITH_SAME_SA:-false}"
 ENVOY_LOG_LEVEL="${ENVOY_LOG_LEVEL:-debug}"
 DEPLOY_ON_OPENSHIFT="${DEPLOY_ON_OPENSHIFT:-false}"
 MULTICLUSTER_CONTEXTS="${MULTICLUSTER_CONTEXTS:-alpha beta}"
+USE_PRIVATE_REGISTRY="${USE_PRIVATE_REGISTRY:-false}"
 
 # For any additional installation arguments. Used heavily in CI.
 optionalInstallArgs=$*


scripts/create-container-registry-creds.sh
```diff
@@ -13,7 +14,9 @@ if [[ "${CI:-}" == "true" ]]; then
     CTR_REGISTRY_PASSWORD=${DOCKER_PASS:-}
 fi
 
-[[ -z "$CTR_REGISTRY_PASSWORD" ]] && exit 0
+if [ "$USE_PRIVATE_REGISTRY" = false ]; then
```
trstringer, line 17: Are we standardizing on POSIX compliant tests (`[ ]`), or the bash extension (`[[ ]]`)? It looks like it is mixed. The latter is more common, and although less compatible I think it's fine as long as it is consistent.

trstringer: I also think we should quote `"false"` so that it is more obvious we are dealing with strings, and viewers don't confuse it with the `false` bin (`/usr/bin/false`).

nshankar13: @trstringer Good point, I've honestly seen both used across our scripts, but for the sake of consistency I can use `[[ ]]` here.

scripts/create-container-registry-creds.sh
```diff
@@ -4,6 +4,7 @@ set -aueo pipefail
 
 # shellcheck disable=SC1091
 source .env
+USE_PRIVATE_REGISTRY="${USE_PRIVATE_REGISTRY:-false}"
```
jaellio, line 7: Why is `USE_PRIVATE_REGISTRY` redefined here? If I wanted to change the env var for the osm multicluster demo I would have to update it in  `create-container-registry-creds.sh`, `demo/run-osm-multicluster-demo.sh` and `demo/configure-app-namespaces.sh`.

jaellio (approved):
You could update the demo docs to include a note on updating this env var when using a private registry.

scripts/create-container-registry-creds.sh
nshankar13: I set it in `demo/run-osm-demo.sh` in a separate PR: https://github.com/openservicemesh/osm/pull/3872. I didn't want to change too many files in that PR so that's why I'm making these changes here. 

I just redefined it in this file here in case someone wants to run this script by itself, or in case there's a job / demo that runs the script without setting USE_PRIVATE_REGISTRY.

demo/configure-app-namespaces.sh
```diff
@@ -4,6 +4,7 @@ set -aueo pipefail
 
 # shellcheck disable=SC1091
 source .env
+USE_PRIVATE_REGISTRY="${USE_PRIVATE_REGISTRY:-false}"
```
ksubrmnn, line 7: I don't see this line being used in this script, so let's take it out

scripts/create-container-registry-creds.sh
```diff
@@ -13,7 +14,9 @@ if [[ "${CI:-}" == "true" ]]; then
     CTR_REGISTRY_PASSWORD=${DOCKER_PASS:-}
 fi
 
-[[ -z "$CTR_REGISTRY_PASSWORD" ]] && exit 0
+if [[ "$USE_PRIVATE_REGISTRY" = false ]]; then
```
ksubrmnn, line 17: Since we have this line, can you remove where we use this variable to optionally call the script in `run-osm-demo.sh`

We'll just let the check here handle it

scripts/create-container-registry-creds.sh
ksubrmnn: The expectation here is that users set their .env files for the non-automated-demo scripts. 

See discussion here: https://github.com/openservicemesh/osm/pull/3839#discussion_r675160685

But I agree it is confusing to users since I don't think it is documented that the defaults only apply to the demo scripts and not each individual script. However we don't want to keep track of maintaining the defaults in a bunch of different locations

nshankar13: Got it, thanks for linking the discussion.

.github/workflows/main.yml
```diff
@@ -298,6 +298,7 @@ jobs:
           ENABLE_EGRESS: "false"
           DEPLOY_TRAFFIC_SPLIT: "true"
           CTR_TAG: ${{ github.sha }}
+          USE_PRIVATE_REGISTRY: "false"
```
ksubrmnn, line 301:
```suggestion
```

The default is already false

nojnhuh (approved):

PR #3878, opened 2021-07-28
bug-report: implement bug report tool

shashankram:
**Description**:
Implements a tool to generate a bug report archive.
The tool collects info about app namespaces, pods,
events etc.

This is an initial implementation to enable users to generate
a bug report, without which it's not possible to debug most
issues.

Example:
```
$ osm support bug-report  --app-deployments bookbuyer/bookbuyer,bookstore/bookstore-v1 --app-pods bookthief/bookthief-7bb7f9b98c-qplq4 --app-namespaces bookbuyer,bookstore 
[+] Created staging dir /tmp/osm_bug_report_059982018 to generate bug report


[+] Collecting information about meshed namespaces
✓ Finished generating report for meshed namespaces


[+] Collecting information from individual app namespaces
✓ Collected report from Namespace "bookbuyer"
✓ Collected report from Namespace "bookstore"
✓ Finished generating report for individual app namespaces


[+] Collecting information from individual app namespaces
✓ Collected report for Pod "bookthief/bookthief-7bb7f9b98c-qplq4"
✓ Collected report for Pod "bookbuyer/bookbuyer-58689d659c-lmm8n"
✓ Collected report for Pod "bookstore/bookstore-v1-74d9cbd7c8-fw8c7"
✓ Finished generating report for individual app pods


[+] Collecting information from individual app namespaces
✓ Bug report successfully archived to /tmp/520720697_osm-bug-report.zip
```

**Affected area**:

### DIFF
--- a/cmd/cli/osm.go
+++ b/cmd/cli/osm.go
@@ -22,7 +22,7 @@ To install and configure OSM, run:
 
 var settings = cli.New()
 
-func newRootCmd(config *action.Configuration, in io.Reader, out io.Writer, args []string) *cobra.Command {
+func newRootCmd(config *action.Configuration, stdin io.Reader, stdout io.Writer, stderr io.Writer, args []string) *cobra.Command {
 	cmd := &cobra.Command{
 		Use:          "osm",
 		Short:        "Install and manage Open Service Mesh",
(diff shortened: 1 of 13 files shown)


snehachhabria (approved):

draychev (approved):
Awesome!
Thank you!

trstringer (commented):
This looks really good, and I think this is a really great thing for supportability. A couple of questions that aren't necessarily blockers, but perhaps worth discussing:

1. Maybe there should be a quick text reminder that this data may contain private information that the user might not want to publicly share?
2. Any reason for going with zip file format? I would think that if we're defaulting to something we should go with a tarball or perhaps give the user the option to select between zip/tar.gz.

michelleN (approved):
+1 on @trstringer's tarball and privacy reminder suggestions

This is so helpful!

shashankram:
> This looks really good, and I think this is a really great thing for supportability. A couple of questions that aren't necessarily blockers, but perhaps worth discussing:
> 
> 1. Maybe there should be a quick text reminder that this data may contain private information that the user might not want to publicly share?

There's a note on collecting sensitive info. Perhaps this could be expanded upon.
```
$ osm support bug-report -h

Generate a bug report.

*Note:
- Both 'osm' and 'kubectl' CLI must reside in the evironment's lookup path.
- If the environment includes sensitive information that should not be collected,
  please do not specify the associated resources.
...
```

> 2. Any reason for going with zip file format? I would think that if we're defaulting to something we should go with a tarball or perhaps give the user the option to select between zip/tar.gz.

I didn't mention this in the commit message, but the only reason I went with .zip  is because the archiver pkg I leveraged provides some extra options with .zip that is not available with .tar.gz. I think we should use .tar.gz and that will require some minor additions (manually implementing file overwrite, folder structure correctness etc.), but it's not too pressing so I thought it could be an incremental change.

PR #3887, opened 2021-07-29
ads: add e2e for permissive-smi switching

eduser25:
Adds an e2e to comprehensibly test switching from SMI to permissive.

Note: This PR is sitting on other PRs. PRs will be individually merged and PRs will be rebased.

This is part of the series of commits for #3775

No
1. Is this a breaking change?
No

### DIFF
--- a/tests/e2e/e2e_permissive_smi_switching.go
+++ b/tests/e2e/e2e_permissive_smi_switching.go
@@ -1,0 +1,207 @@
+package e2e
+
+import (
+	"context"
+	"fmt"
+	"sync"
+	"time"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	. "github.com/openservicemesh/osm/tests/framework"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var _ = OSMDescribe("Test HTTP traffic from N deployment client -> 1 deployment server, permissive to SMI switching",
+	OSMDescribeInfo{
+		Tier:   1,
+		Bucket: 2,
+	},
+	func() {
+		Context("PermissiveToSmiSwitching", func() {
+			const destApp = "server"
+			const sourceAppBaseName = "client"
+			var sourceNamespaces []string = []string{}
+
+			// Total (numberOfClientApps x replicaSetPerApp) pods
+			numberOfClientServices := 2
+			replicaSetPerService := 2
+
+			// Used across the test to wait for concurrent steps to finish
+			var wg sync.WaitGroup
+
+			for i := 0; i < numberOfClientServices; i++ {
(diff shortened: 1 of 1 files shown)


tests/e2e/e2e_permissive_smi_switching.go
```diff
@@ -0,0 +1,290 @@
...
+
+							Destination: fmt.Sprintf("%s.%s:%d", destApp, destApp, DefaultUpstreamServicePort),
+						})
+					}
+				}
+
+				var results HTTPMultipleResults
+
+				By("Test and expect traffic failure with no permissive mode and no SMI rules")
+				success := Td.WaitForRepeatedSuccess(func() bool {
```
nshankar13, line 129: For consistency, instead of switching between SMI and permissive mode and repeating the code segments that tests the HTTP requests, we could keep the requests segment as it's own function /ginkgo context and pass in a parameter for permissive mode like we do in the e2e_tcp_client_server test: https://github.com/openservicemesh/osm/blob/main/tests/e2e/e2e_tcp_client_server_test.go#L24-L29

tests/e2e/e2e_permissive_smi_switching.go
```diff
@@ -0,0 +1,290 @@
...
+					}
+					return true
+				}, 20, 150*time.Second)
+
+				Expect(success).To(BeTrue())
+			})
+		})
+	})
+
+func setPermissiveMode(b bool) error {
```
nshankar13, line 282: Instead of creating this function, we could just keep the request success block as it's own function / context as mentioned in the above comment, and just set the permissive mode upon install: https://github.com/openservicemesh/osm/blob/main/tests/e2e/e2e_tcp_client_server_test.go#L40-L42

tests/e2e/e2e_permissive_smi_switching.go
eduser25: you can't use contexts, there's cleanup between them.
I could do it between `It()`s?

eduser25: I can probably move the success/failure traffic checks on a function though, I was lazy

nshankar13: @nojnhuh is there cleanup between `It()`s?

nojnhuh: I'm almost positive there is, but I would have to double check to be sure.

eduser25:
@nshankar13 dropping the `It()` idea mainly because setup is also part of an `It` and ginkgo doesn't allow nesting `It`s.
Just since most tests are already written like this, I'd rather not go too much stray from what we have (I guess?)

nshankar13 (approved):

jaellio (approved):

PR #3889, opened 2021-07-29
chore(injector): Reduce tech debt in osm `/pkg/injector`

davinci26:
Addressing PR comments from #3861 we have the two follow ups:

1. Make literal strings "windows" and "linux" into a constant
2. Make the return variables of `getPlatformSpecificSpecComponents` named to
improve readability.

### DIFF
--- a/pkg/constants/constants.go
+++ b/pkg/constants/constants.go
@@ -225,3 +225,12 @@ const (
 	// byte in a TCP connection.
 	ProtocolTCPServerFirst = "tcp-server-first"
 )
+
+// Operating systems.
+const (
+	// OSWindows is the name for Windows operating system.
+	OSWindows string = "windows"
+
+	// OSLinux is the name for Linux operating system.
+	OSLinux string = "linux"
+)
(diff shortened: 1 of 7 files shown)


pkg/constants/constants.go
```diff
@@ -164,6 +164,12 @@ const (
 
 	// OSMMeshConfig is the name of the OSM MeshConfig
 	OSMMeshConfig = "osm-mesh-config"
+
+	// OSWindows is the name for Windows operating system.
+	OSWindows string = "windows"
+
+	// OSLinux is the name for Linux operating system.
+	OSLinux string = "linux"
 )
```
shashankram, line 173: Since we aren't using a custom string type, for grouping purposes I suggest using a new `const()` definition similar to the protocols defined below.

shashankram (approved):

draychev (approved):

pkg/constants/constants.go
davinci26: fixed now

davinci26:
@shashankram this should be ready to be merged

PR #3898, opened 2021-07-30
feat(metrics): errcode metric

jaellio:
**Description**:

Adds an errcode metric to track the number of errcodes generated by
OSM.

**Affected area**:
| Observability              | [ X ] |


pkg/catalog/egress.go
```diff
@@ -87,6 +88,7 @@ func (mc *MeshCatalog) GetEgressTrafficPolicy(serviceIdentity identity.ServiceId
 	if err != nil {
 		log.Error().Err(err).Str(errcode.Kind, errcode.ErrDedupEgressTrafficMatches.String()).
 			Msgf("Error deduplicating egress traffic matches for service identity %s", serviceIdentity)
+		metricsstore.DefaultMetricsStore.ErrcodeCounter.WithLabelValues(errcode.ErrDedupEgressTrafficMatches.String(), errcode.Egress).Inc()
```
shashankram, line 91: Adding an extra log line for each error log just to generate the metric doesn't seem idiomatic. 

I suggest 2 options to solve this:
1. Create a generic utility in `pkg/errcode`:
 ```
getErrCodeWithMetric(e errcode.ErrCode) string {
    metricsstore.DefaultMetricsStore.ErrcodeCounter.WithLabelValues(e.String())
return e.String()
}
...
func someFunc() {
    if err != nil {
        log.Error().Str(errcode.Kind, getErrCodeWithMetric(errcode.SomeError).Msg(...)
    }
}
```

2. Do the magic of capturing error code metrics in the logger's hook

File: pkg/errcode/errcode.go [hunk 2]
```diff
@@ -13,6 +13,11 @@ import (
 // ErrCode defines the type to represent error codes
 type ErrCode int
 
+// Error code groups
+const (
+	Egress string = "egress"
+)
+
```
shashankram, line 20: Seems unnecessary, just the error code should suffice to retrieve the error. The group doesn't serve much.

pkg/metricsstore/metricsstore.go
```diff
@@ -158,6 +164,21 @@ func init() {
...
 		},
 		[]string{})
+
+	/*
+	 * Errcode metrics
+	 */
+	defaultMetricsStore.ErrcodeCounter = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Namespace: metricsRootNamespace,
+			Name:      "errcode_count",
```
shashankram, line 174: Set the `Subsystem` to `error`.

shashankram (changes requested):

pkg/errcode/errcode.go
jaellio: We weren't sure if we should include `errcode_group` or not. Our original thinking was that it would help group errors when creating dashboards, and provide more specificity when error codes are used across groups (`ErrMarshallingXDSResource` is used in LDS and the Injector for example).

jaellio: To avoid additional complexity, at this moment I think it makes sense to not include `errcode_group` in the errcode metric. This label can be added later if necessary.

pkg/catalog/egress.go
jaellio: Great suggestion! I will try the first option. 

I explored using a hook, but it would have required parsing the log event's buffer to obtain the errcode.  The performance cost of this work made the option less appealing.

cmd/osm-controller/osm-controller.go
```diff
@@ -299,6 +299,7 @@ func startMetricsStore() {
 		metricsstore.DefaultMetricsStore.ProxyBroadcastEventCount,
 		metricsstore.DefaultMetricsStore.CertIssuedCount,
 		metricsstore.DefaultMetricsStore.CertIssuedTime,
+		metricsstore.DefaultMetricsStore.ErrcodeCounter,
```
shashankram, line 302:
```suggestion
		metricsstore.DefaultMetricsStore.ErrCodeCounter,
```

pkg/metricsstore/metricsstore.go
shashankram: `osm.error.count` reads better than `osm.error.errcode_count`, given the labels are enough to carry additional context.

pkg/errcode/errcode.go
```diff
@@ -373,6 +374,13 @@ func (e ErrCode) String() string {
 	return fmt.Sprintf("E%d", e)
 }
 
+// GetErrCodeWithMetric increments the ErrcodeCounter metric for the given error code
+// Returns the error code as a string
+func GetErrCodeWithMetric(e ErrCode) string {
```
shashankram, line 379: Let's add a small unit test for this (no need to verify that metrics work).

shashankram (approved):

pkg/errcode/errcode.go
jaellio: Added!

pkg/metricsstore/metricsstore.go
```diff
@@ -55,6 +55,12 @@ type MetricsStore struct {
 	// CertXdsIssuedCounter the histogram to track the time to issue a certificates
 	CertIssuedTime *prometheus.HistogramVec
 
+	/*
+	 * Errcode metrics
+	 */
+	// ErrcodeCounter is the metric counter for the number of errcodes generated by OSM
```
shalier, line 61: Nit: // ErrCodeCounter

jaellio: Good catch!

pkg/metricsstore/metricsstore_test.go
```diff
@@ -82,4 +84,30 @@ osm_proxy_connect_count %d
 `, proxiesConnected-proxiesDisconnected)
 		assert.Contains(rr.Body.String(), expectedResp)
 	})
+
+	t.Run("ErrcodeCounter", func(t *testing.T) {
```
shalier, line 88: Nit: "ErrCodeCounter"

shalier (approved):

snehachhabria (approved):

PR #3909, opened 2021-08-02
feat(metrics): count logs

shalier:
**Description**:
Add metric to count logs emitted based on the log level (trace, debug, error, etc.)

**Affected area**:
| Observability              | [ x] |

1. Is this a breaking change? no

### DIFF
--- a/pkg/logger/logger.go
+++ b/pkg/logger/logger.go
@@ -13,6 +13,7 @@ import (
 	"github.com/rs/zerolog/log"
 
 	"github.com/openservicemesh/osm/pkg/constants"
+	"github.com/openservicemesh/osm/pkg/metricsstore"
 )
 
 // AllowedLevels is the list of allowed log levels for OSM Controller.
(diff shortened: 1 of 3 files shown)


nshankar13 (approved):

allenlsy:
Curious, why is this metric useful? Suppose we have the numbers for each log level, what do they indicate?

pkg/metricsstore/metricsstore.go
```diff
@@ -185,6 +191,20 @@ func init() {
...
+	 */
+	defaultMetricsStore.LogLvlCount = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Namespace: metricsRootNamespace,
+			Subsystem: "log",
+			Name:      "level_count",
+			Help:      "Represents the number of logs per log level emitted",
+		},
+		[]string{
+			"logLevel",
```
jaellio, line 206:
```suggestion
			"log_level",
```
Nit: To be consistent with other metrics

pkg/logger/logger.go
```diff
@@ -22,10 +23,11 @@ var AllowedLevels = []string{"debug", "info", "warn", "error", "fatal", "panic",
 type CallerHook struct{}
 
 // Run adds additional context
-func (h CallerHook) Run(e *zerolog.Event, _ zerolog.Level, _ string) {
+func (h CallerHook) Run(e *zerolog.Event, lvl zerolog.Level, _ string) {
 	if _, file, line, ok := runtime.Caller(3); ok {
 		e.Str("file", fmt.Sprintf("%s:%d", path.Base(file), line))
 	}
+	metricsstore.DefaultMetricsStore.LogLvlCount.WithLabelValues(lvl.String()).Inc()
```
shashankram, line 30: I don't see why this metric is required. Counting number of logs  is not a useful metric. If the goal is to get metrics for the errors, then it needs to be done a lot differently - by inspecting the error log, and generating a metric with context, such as adding the error code information to the metric. 

If we really want to generate a generic metric for counting errors, then make the metric specific to error logs and inspect the level here for error. Counting `info/trace/debug` logs is totally unnecessary.

shashankram (changes requested):

pkg/logger/logger.go
nshankar13: @shashankram we will be adding a separate metric for a counter per Errcode

shashankram: @nshankar13 then this should be unnecessary.

johnsonshi (approved):

Issue #3779, opened 2021-07-19
Design and implement an API for an ingress backend


shashankram:
**Please describe the Improvement and/or Feature Request**
This issue pertains to designing and implementing an `IngressBackend` API that will be leveraged to program a backend that needs to be exposed via some ingress (k8s Ingress, Gateway API, Contour's HTTPProxy etc.).

The API will allow programming one or more service mesh backends to accept traffic from authorized clients, while allowing to specify the backend protocol to use and TLS configurations if any.

Proposal can be found [here](https://docs.google.com/document/d/1ex5YwF8QWYXTl8r2J2ZVy7UVO7sVLZpjuOuSio2CGTs/edit?usp=sharing).

**Scope (please mark with X where applicable)**
- New Functionality          [X]
- Ingress                    [X]

**Possible use cases**
Improve the usability of Ingress in OSM.

PR #3914, opened 2021-08-03
ingress: add support for automatic gateway cert provisioning

shashankram:
**Description**:
Adds support to automatically issue, rotate, and remove a certificate
required by an ingress gateway.

Part of #3779

**Affected area**:


pkg/ingress/gateway.go
```diff
@@ -0,0 +1,203 @@
...
+	if apierrors.IsAlreadyExists(err) {
+		_, err = c.kubeClient.CoreV1().Secrets(secret.Namespace).Update(context.Background(), sec, metav1.UpdateOptions{})
+	}
+	return err
+}
+
+// handleCertificateChange updates the gateway certificate and secret when the MeshConfig resource changes or
+// when the corresponding gateway certificate is rotated.
+func (c client) handleCertificateChange(currentCertSpec *configv1alpha1.IngressGatewayCertSpec, stop <-chan struct{}) {
+	meshConfigUpdated := events.GetPubSubInstance().Subscribe(announcements.MeshConfigUpdated)
```
eduser25, line 107: nit: spacing here looks a bit weird

eduser25: Q: do we care for add/del of meshconfig? asking as I'm not sure of current behaviour when it is deleted (i think it all takes defaults?)

pkg/ingress/gateway.go
```diff
@@ -0,0 +1,203 @@
...
+			Name:      secret.Name,
+			Namespace: secret.Namespace,
+		},
+		Type: corev1.SecretTypeTLS,
+		Data: secretData,
+	}
+
+	_, err := c.kubeClient.CoreV1().Secrets(secret.Namespace).Create(context.Background(), sec, metav1.CreateOptions{})
+	if apierrors.IsAlreadyExists(err) {
+		_, err = c.kubeClient.CoreV1().Secrets(secret.Namespace).Update(context.Background(), sec, metav1.UpdateOptions{})
```
eduser25, line 99: Not sure this applies, feel free to dismiss. 
Do multireplica osm require to sync this cert on all instances? i see each is issuing a cert, but only one will succeed in updating it. Does it matter?

eduser25 (approved):

pkg/ingress/gateway.go
shashankram: As long as the secret is created even by a single controller, it's good enough for this use case. I don't think we can synchronize such operations between multiple controllers. The only thing we could do better is wrt not logging errors when such things happen because multiple controllers tried to do the same thing. I'll follow up on this and address any misleading logs as I start testing the feature end to end.

pkg/ingress/gateway.go
shashankram: The spacing is correct and has been formatted with `go fmt -s`.

We don't care for add/del. MeshConfig will always be created prior to controller startup, and we don't expect to act on Delete operations.

pkg/ingress/gateway.go
eduser25: sgtm

snehachhabria (approved):

PR #3920, opened 2021-08-04
tests/e2e: move nginx ingress install to a helper

shashankram:
**Description**:
Moves the code to install nginx ingress to a helper
function so it can be reused by other tests. Moreover,
testing nginx install steps should not be a part of
the test body.

**Affected area**:

### DIFF
--- a/tests/e2e/e2e_http_ingress_test.go
+++ b/tests/e2e/e2e_http_ingress_test.go
@@ -3,15 +3,10 @@ package e2e
 import (
 	"context"
 	"net/http"
-	"strings"
 	"time"
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
-	"helm.sh/helm/v3/pkg/action"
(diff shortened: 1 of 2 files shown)


nojnhuh (approved):

nshankar13 (approved):

pkg/catalog/egress.go
```diff
@@ -105,6 +105,10 @@ func (mc *MeshCatalog) GetEgressTrafficPolicy(serviceIdentity identity.ServiceId
 	}, nil
 }
 
+func getErrCode(e errcode.ErrCode) string {
```
snehachhabria, line 108: should this be moved to a util ? I assume it can be used across the code

snehachhabria (approved):

pkg/catalog/egress.go
shashankram: Accidentally merged 2 branches. fixed.

PR #3921, opened 2021-08-04
feat(windows): Adds demo containers based on Windows

### LINKED ISSUE #3880: Windows based demo containers
All the containers in `dockerfiles` should have a Windows equivalent and we should expand the makefile to build them

davinci26:
Adds the makefile definitions and the checks to build Windows based
demo containers on the CI.

Fixes #3880

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -173,8 +173,11 @@ DEMO_TARGETS = bookbuyer bookthief bookstore bookwarehouse tcp-echo-server tcp-c
 DEMO_BUILD_TARGETS = $(addprefix build-, $(DEMO_TARGETS))
 .PHONY: $(DEMO_BUILD_TARGETS)
 $(DEMO_BUILD_TARGETS): NAME=$(@:build-%=%)
+ifeq ($(OS), windows)
+	EXT=.exe
+endif
 $(DEMO_BUILD_TARGETS):
-	GOOS=linux GOARCH=amd64 CGO_ENABLED=0 go build -o ./demo/bin/$(NAME)/$(NAME) ./demo/cmd/$(NAME)
+	GOOS=$(OS) GOARCH=amd64 CGO_ENABLED=0 go build -o ./demo/bin/$(NAME)/$(NAME)$(EXT) ./demo/cmd/$(NAME)
 	@if [ -f demo/$(NAME).html.template ]; then cp demo/$(NAME).html.template demo/bin/$(NAME); fi
 
 .PHONE: build-bookwatcher
@@ -189,8 +192,22 @@ DOCKER_DEMO_TARGETS = $(addprefix docker-build-, $(DEMO_TARGETS))
 .PHONY: $(DOCKER_DEMO_TARGETS)
 $(DOCKER_DEMO_TARGETS): NAME=$(@:docker-build-%=%)
 $(DOCKER_DEMO_TARGETS):
-	make build-$(NAME)
+	make OS=linux build-$(NAME)
 	docker build -t $(CTR_REGISTRY)/$(NAME):$(CTR_TAG) -f dockerfiles/Dockerfile.$(NAME) demo/bin/$(NAME)
+
+
+# docker-build-windows-bookbuyer, etc
+# This command can be used to push the images as well if the ARGS is set to --push
+# see https://docs.docker.com/engine/reference/commandline/buildx_build/#push
+# The reason for that is that on linux we can't load a Windows image so we need to build and push with one command.
+DOCKER_WINDOWS_DEMO_TARGETS = $(addprefix docker-build-windows-, $(DEMO_TARGETS))
+.PHONY: $(DOCKER_WINDOWS_DEMO_TARGETS)
+$(DOCKER_WINDOWS_DEMO_TARGETS): OS = windows
+$(DOCKER_WINDOWS_DEMO_TARGETS): NAME=$(@:docker-build-windows-%=%)
+$(DOCKER_WINDOWS_DEMO_TARGETS):
+	make OS=windows build-$(NAME)
+	@if ! docker buildx ls | grep -q "img-builder "; then  echo "Setting buildx img-builder"; docker buildx create --name img-builder --use; fi
+	docker buildx build --platform "windows/amd64" -t $(CTR_REGISTRY)/$(NAME)-windows:$(CTR_TAG) $(ARGS) -f dockerfiles/Dockerfile.$(NAME).windows demo/bin/$(NAME)
 
 docker-build-init:
 	docker build -t $(CTR_REGISTRY)/init:$(CTR_TAG) - < dockerfiles/Dockerfile.init
@@ -231,10 +248,27 @@ VERIFY_TAGS = 0
 .PHONY: $(DOCKER_PUSH_TARGETS)
 $(DOCKER_PUSH_TARGETS): NAME=$(@:docker-push-%=%)
 $(DOCKER_PUSH_TARGETS):
-	@if [ $(VERIFY_TAGS) != 1 ]; then make docker-build-$(NAME) && docker push "$(CTR_REGISTRY)/$(NAME):$(CTR_TAG)"; else bash scripts/publish-image.sh $(NAME); fi
+	@if [ $(VERIFY_TAGS) != 1 ]; then make docker-build-$(NAME) && docker push "$(CTR_REGISTRY)/$(NAME):$(CTR_TAG)"; else bash scripts/publish-image.sh $(NAME) "linux"; fi
+
+
+DOCKER_PUSH_WINDOWS_TARGETS = $(addprefix docker-push-windows-, $(DEMO_TARGETS))
+VERIFY_TAGS = 0
+.PHONY: $(DOCKER_PUSH_WINDOWS_TARGETS)
+$(DOCKER_PUSH_WINDOWS_TARGETS): NAME=$(@:docker-push-%=%)
+$(DOCKER_PUSH_WINDOWS_TARGETS):
+	@if [ $(VERIFY_TAGS) != 1 ]; then make ARGS=--output=type=registry docker-build-$(NAME); else bash scripts/publish-image.sh $(addprefix windows-, $(NAME)) "windows"; fi
+
+
+.PHONY: docker-linux-push
+docker-linux-push: $(DOCKER_PUSH_TARGETS)
+
+# notably the init container is missing here because we don't use it for windows
+OSM_CONTROL_PLANE_TARGETS = $(addprefix docker-push-, init-osm-controller osm-controller osm-injector osm-crds osm-crd-converter)
+.PHONY: docker-windows-push
+docker-windows-push: $(DOCKER_PUSH_WINDOWS_TARGETS) $(OSM_CONTROL_PLANE_TARGETS)
 
 .PHONY: docker-push
-docker-push: $(DOCKER_PUSH_TARGETS)
+docker-push: docker-linux-push docker-windows-push
 
 .PHONY: shellcheck
 shellcheck:
--- a/dockerfiles/Dockerfile.bookbuyer.windows
+++ b/dockerfiles/Dockerfile.bookbuyer.windows
@@ -1,0 +1,5 @@
+ARG BUILD_OS=mcr.microsoft.com/windows/nanoserver
+ARG BUILD_TAG=ltsc2022
+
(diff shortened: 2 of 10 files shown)


davinci26:
Lets see what the CI things of the change first

davinci26:
Updating to non-draft as the validation locally passes:

```
davinci@MININT-9CEC7NF:~/osm$ curl -X GET http://localhost:5000/v2/_catalog
{"repositories":["bookbuyer","bookbuyer-windows","bookstore","bookstore-windows","bookthief","bookthief-windows","bookwarehouse","bookwarehouse-windows","init","init-osm-controller","my-ubuntu","osm-controller","osm-crd-converter","osm-crds","osm-injector","tcp-client","tcp-client-windows","tcp-echo-server","tcp-echo-server-windows"]}
```

davinci26:
@nojnhuh do you want to take this review since you are involved with the docker images?

Makefile
```diff
@@ -180,21 +180,38 @@ DEMO_TARGETS = bookbuyer bookthief bookstore bookwarehouse tcp-echo-server tcp-c
...
 	@if [ -f demo/$(NAME).html.template ]; then cp demo/$(NAME).html.template demo/bin/$(NAME); fi
 
 .PHONY: demo-build
-demo-build: $(DEMO_BUILD_TARGETS) build-init-osm-controller build-osm-controller build-osm-injector build-osm-crds build-osm-crd-converter
+demo-build:
+	make OS=linux $(DEMO_BUILD_TARGETS) build-init-osm-controller build-osm-controller build-osm-injector build-osm-crds build-osm-crd-converter
+
+.PHONY: demo-build-windows
+demo-build-windows:
+	make OS=windows $(DEMO_BUILD_TARGETS) build-init-osm-controller build-osm-controller build-osm-injector build-osm-crds build-osm-crd-converter
```
nojnhuh, line 196: I'd be in favor of removing these altogether since I don't think they're dependencies for anything else. If anyone actually uses these then I'm fine keeping them, but anywhere we can remove a big list of images is a plus in my book.

dockerfiles/Dockerfile.bookbuyer.windows
```diff
@@ -0,0 +1,5 @@
+ARG BUILD_OS=mcr.microsoft.com/windows/nanoserver/insider
+ARG BUILD_TAG=10.0.20348.1
```
nojnhuh, line 2: Seems like it would be possible to share a single Dockerfile for windows and linux if we override the base image from the Makefile and template the executable name. Just a thought, don't want to make things too much messier than they already are.

demo/cmd/common/books.go
```diff
@@ -29,6 +31,18 @@ const (
...
 )
 
+// GetServiceURLBasedOnWorkerOS  is used to retried the service url based on the worker OS.
+func GetServiceURLBasedOnWorkerOS(namespace string, serviceName string, port int) string {
+	postfix := ""
+	if runtime.GOOS == constants.OSWindows {
+		// Due to a limitation on kubernetes on Windows we need to add the postfix
+		// otherwise DNS will not be able to resolve it.
+		// https://kubernetes.io/docs/setup/production-environment/windows/intro-windows-in-kubernetes/#dns-limitations
+		postfix = ".svc.cluster.local"
```
nojnhuh, line 41: I don't see any issue with adding this unconditionally if we can get rid of a `if windows` branch.

scripts/publish-image.sh
```diff
@@ -19,6 +20,11 @@ exists=$(echo "$version_list" | jq --arg t "${CTR_TAG}" '.tags | index($t)')
...
 then
-    make docker-build-"$IMAGE_TAG"
-    docker push "$IMAGE_REPO/$IMAGE_TAG:${CTR_TAG}" || { echo "Error pushing images to container registry $(CTR_REGISTRY)/$(NAME):$(CTR_TAG)"; exit 1; }
+    if [[ $OS == "linux" ]]
+    then
+        make docker-build-"$IMAGE_TAG"
+        docker push "$IMAGE_REPO/$IMAGE_TAG:${CTR_TAG}" || { echo "Error pushing images to container registry $(CTR_REGISTRY)/$(NAME):$(CTR_TAG)"; exit 1; }
+    else
+        make ARGS=--push "docker-build-$(NAME)"
+    fi
```
nojnhuh, line 29: Is this missing a `docker push` for the non-linux branch of this condition?

Leaving this mostly as a note to myself for later, but maybe this script should assume the images are already built, then it can basically just do, "Given an image, check if it exists. If not, push." That should be platform-agnostic.

Makefile
```diff
@@ -237,10 +254,19 @@ VERIFY_TAGS = 0
...
+VERIFY_TAGS = 0
+.PHONY: $(DOCKER_PUSH_WINDOWS_TARGETS)
+$(DOCKER_PUSH_WINDOWS_TARGETS): NAME=$(@:docker-push-%=%)
+$(DOCKER_PUSH_WINDOWS_TARGETS):
+	@if [ $(VERIFY_TAGS) != 1 ]; then make ARGS=--push docker-build-$(NAME); else bash scripts/publish-image.sh $(addprefix windows-, $(NAME)) "windows"; fi
+
 
 .PHONY: docker-push
-docker-push: $(DOCKER_PUSH_TARGETS)
+docker-push: $(DOCKER_PUSH_TARGETS) $(DOCKER_PUSH_WINDOWS_TARGETS)
```
nojnhuh, line 269: The demo script does invoke `make docker-push`, which already does more than it really needs to. Let's at least make it easy to build all the images for only windows or linux so we can use that in the demo script. Then we can have an all-in-one target we call from CI.

nojnhuh (commented):
Left a couple comments for things that might be broken (publish-image.sh, windows images looking like linux from `docker inspect`) and a few other light suggestions. Everything else looks good.

Makefile
```diff
@@ -180,21 +180,38 @@ DEMO_TARGETS = bookbuyer bookthief bookstore bookwarehouse tcp-echo-server tcp-c
...
 	docker build -t $(CTR_REGISTRY)/$(NAME):$(CTR_TAG) -f dockerfiles/Dockerfile.$(NAME) demo/bin/$(NAME)
 
+
+DOCKER_WINDOWS_DEMO_TARGETS = $(addprefix docker-build-windows-, $(DEMO_TARGETS))
+.PHONY: $(DOCKER_WINDOWS_DEMO_TARGETS)
+$(DOCKER_WINDOWS_DEMO_TARGETS): OS = windows
+$(DOCKER_WINDOWS_DEMO_TARGETS): NAME=$(@:docker-build-windows-%=%)
+$(DOCKER_WINDOWS_DEMO_TARGETS):
+	make OS=windows build-$(NAME)
+	docker buildx build -t $(CTR_REGISTRY)/$(NAME)-windows:$(CTR_TAG) $(ARGS) -f dockerfiles/Dockerfile.$(NAME) demo/bin/$(NAME)
```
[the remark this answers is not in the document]
davinci26: you probably right, I am surprised how it even built locally

demo/cmd/common/books.go
davinci26: I remember that @shashankram had some concerns when I mentioned this in a standup a month back this is why I made it conditional.

No strong opinion.

Makefile
davinci26: You mean delete the make rules `demo-build` and `demo-build-windows`

scripts/publish-image.sh
davinci26: The way it works is because we use buildx we need to push when we build because we cant load the image

demo/cmd/common/books.go
shashankram: Hey @davinci26 my concern was regarding using windows defaults for linux in e2e tests, which could introduce testing gaps in linux. But in this instance, it makes sense to remove this conditional and use the same URL for windows and linux, with a note on why an FQDN must be used.

scripts/publish-image.sh
davinci26: So added an optional args `--push` that is applied to push the image

nojnhuh: Ah, I misinterpreted the help for that flag. I thought it meant it just built the image in a way that it could be pushed to a registry without actually pushing it. Here I see it does actually push it: https://docs.docker.com/engine/reference/commandline/buildx_build/#push

Makefile
nojnhuh: Yes

scripts/publish-image.sh
davinci26: I will add a comment on the command to make sure it is clear for the next person

davinci26:
Currently blocked as the Windows insider images that we need to use cannot be built with `docker buildx`. This PR should be good to go when non-insider versions of Windows server become available

davinci26:
@nojnhuh the container image went out of public preview so we should be able to do merge it now.

```
davinci@MININT-9CEC7NF:~/osm$ make docker-build-windows-bookbuyer
make OS=windows build-bookbuyer
make[1]: Entering directory '/home/davinci/osm'
GOOS=windows GOARCH=amd64 CGO_ENABLED=0 go build -o ./demo/bin/bookbuyer/bookbuyer.exe ./demo/cmd/bookbuyer
make[1]: Leaving directory '/home/davinci/osm'
docker buildx build --platform "windows/amd64" -t osmwindevregistry.azurecr.io/bookbuyer-windows:latest  -f dockerfiles/Dockerfile.bookbuyer.windows demo/bin/bookbuyer
WARN[0000] No output specified for docker-container driver. Build result will only remain in the build cache. To push result image into registry use --push or to load image into docker use --load 
[+] Building 11.8s (7/7) FINISHED                                                                                                                                                                                   
 => [internal] booting buildkit                                                                                                                                                                                2.7s
 => => starting container buildx_buildkit_img-builder0                                                                                                                                                         2.7s
 => [internal] load build definition from Dockerfile.bookbuyer.windows                                                                                                                                         0.1s
 => => transferring dockerfile: 201B                                                                                                                                                                           0.0s
 => [internal] load .dockerignore                                                                                                                                                                              0.0s
 => => transferring context: 2B                                                                                                                                                                                0.0s
 => [internal] load metadata for mcr.microsoft.com/windows/nanoserver:ltsc2022                                                                                                                                 1.1s
 => [internal] load build context                                                                                                                                                                              0.5s
 => => transferring context: 24.44MB                                                                                                                                                                           0.5s
 => [1/2] FROM mcr.microsoft.com/windows/nanoserver:ltsc2022@sha256:3df3372895223b44475125ab3553fd39a71d89a896099d463d3216f1e22a3ab0                                                                           6.9s
 => => resolve mcr.microsoft.com/windows/nanoserver:ltsc2022@sha256:3df3372895223b44475125ab3553fd39a71d89a896099d463d3216f1e22a3ab0                                                                           0.0s
 => => sha256:ee7a9c9193e3d9bee1b6c8ddddddd5356aed50f77c78d18d377241f4aee6d676 117.98MB / 117.98MB                                                                                                             4.5s
 => => extracting sha256:ee7a9c9193e3d9bee1b6c8ddddddd5356aed50f77c78d18d377241f4aee6d676                                                                                                                      2.3s
 => [2/2] COPY bookbuyer.exe bookbuyer.html.template /               
 ```

**Comment by nojnhuh** [comment]
LGTM, but I tried it out on my Mac but I get an "operating system is not supported" error which seems to agree with the output I get for ​docker buildx ls​:

```
% docker buildx ls
NAME/NODE DRIVER/ENDPOINT STATUS PLATFORMS
desktop-linux   docker
  desktop-linux desktop-linux   running linux/amd64, linux/arm64, linux/riscv64, linux/ppc64le, linux/s390x, linux/386, linux/arm/v7, linux/arm/v6
default *       docker
  default       default         running linux/amd64, linux/arm64, linux/riscv64, linux/ppc64le, linux/s390x, linux/386, linux/arm/v7, linux/arm/v6
```

​I wonder if the GitHub actions runners would hit the same error. I don't think any of the CI jobs try to build the Windows images currently. Or do you know if we were going to have Windows CI machines build the images so this probably wouldn't matter?

nojnhuh:
At @davinci26's suggestion, running `docker buildx create --name img-builder --use` before  building any of the images fixed this issue for me.

nojnhuh (approved):

shashankram (approved):

davinci26:
@shashankram should we merge this one?

PR #3930, opened 2021-08-06
Add demo book app watcher for terminal and raw JSON from web API

trstringer:
**Description**:

When I'm testing or demo'ing OSM, I am in the terminal. The demo app requires you to use the browser or curl the markup of the different services (bookbuyer, bookthief, etc.). This is very manual and requires some visual grep'ing as well as manual refreshing.

This PR adds a `/raw` route to bookbuyer, bookthief, and bookstore to get their JSON data. It also adds a bookwatcher bin that polls the different raw endpoints and displays the output. The text color is as follows:

* Yellow - data hasn't changed since last poll iteration
* Green - data for bookbuyer and bookstore has changed
* Red - books were stolen by the bookthief since last poll iteration

This refreshes the terminal every second. Below are screenshots showing the experience.

**Affected area**:

**Testing done**:

osm_demo_bookbuyer1

This shows that since the last iteration, the bookbuyer has bought a book (or books) and the bookstore V1 has sold a book (or books).

osm_demo_bookthief1

This shows that since the last iteration, the bookthief has stolen a book (or books) and bookstore V1 has sold a book (or books).

No.

2. Is this a breaking change?

No.

### DIFF
--- a/demo/cmd/bookwatcher/bookwatcher.go
+++ b/demo/cmd/bookwatcher/bookwatcher.go
@@ -1,0 +1,186 @@
+package main
+
+import (
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"log"
+	"net/http"
+	"sync"
+	"time"
+
+	"github.com/openservicemesh/osm/demo/cmd/common"
+)
+
+func clearScreen() {
+	fmt.Print("\033[H\033[2J")
+}
+
+func printGreenln(msg string) {
+	fmt.Printf("\033[32m%s\033[0m\n", msg)
+}
+
+func printYellowln(msg string) {
+	fmt.Printf("\033[33m%s\033[0m\n", msg)
+}
+
+func printRedln(msg string) {
+	fmt.Printf("\033[31m%s\033[0m\n", msg)
+}
+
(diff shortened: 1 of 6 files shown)


demo/cmd/bookwatcher/bookwatcher.go
```diff
@@ -0,0 +1,204 @@
...
+
+func printYellowln(msg string) {
+	fmt.Printf("\033[33m%s\033[0m\n", msg)
+}
+
+func printRedln(msg string) {
+	fmt.Printf("\033[31m%s\033[0m\n", msg)
+}
+
+func getBookBuyer(bookBuyerPurchases *common.BookBuyerPurchases, wg *sync.WaitGroup, errc chan<- error) {
```
allenlsy, line 38: nit: better find a way to de-dup the code for these 4 get methods.

allenlsy (approved):

demo/cmd/bookwatcher/bookwatcher.go
trstringer: Great point, thanks! Done!

jaellio (approved):

PR #3939, opened 2021-08-10
crdUpgrades: Update osm clean up hook to patch osm crds instead of deleting them

### LINKED ISSUE #3760: CLI: Avoid deleting CRDs by default unless explicitly stated by the user
With the recent changes made to OSM for the purpose of upgrade, the CRDs installed get removed during uninstall. We need to avoid deleting CRDs unless explicitly specified because deleting the CRD will also delete all the custom resources (SMI policies, Egress policies etc.) from the cluster. This is a very destructive operation that needs to be explicitly stated by the user when required.

- Proposal 1:
Only delete CRDs if `--delete-crds`  is passed to `osm uninstall`.
When `--delete-crds` is passed, OSM CLI could patch `charts/osm/templates/cleanup-hook.yaml` to delete the CRDs, otherwise they should not be deleted.

- Proposal 2
Add an option in `MeshConfig` to allow the user to specify whether the CRD should be deleted or not, and appropriately have the cleanup hook query this value prior to deleting the CRDs.

- Proposal 3
When uninstalling (using Helm or CLI), use a hook to backup all custom resources before removing the CRDs.

- Uninstall                    [X]

**Possible use cases**
To avoid deleting all policies that the user may or may not have backed up.

snehachhabria:
**Description**:

Update the pre-delete clean up hook in osm to patch the crds, removing
the conversion webhook instead of deleting them on uninstall.

Fixes #3760

**Affected area**:

**Testing done**:

Manually verified that the CRDs remain in the cluster after an uninstall and do not contain the conversion webhook spec

### DIFF
--- a/charts/osm/templates/cleanup-hook.yaml
+++ b/charts/osm/templates/cleanup-hook.yaml
@@ -107,8 +107,8 @@ metadata:
     helm.sh/hook: pre-delete
     helm.sh/hook-delete-policy: before-hook-creation,hook-succeeded
 ---
-apiVersion: v1
-kind: Pod
+apiVersion: batch/v1
+kind: Job
 metadata:
   name: {{ .Release.Name }}-cleanup
   namespace: {{ include "osm.namespace" . }}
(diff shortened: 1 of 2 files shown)


charts/osm/templates/cleanup-hook.yaml
```diff
@@ -118,26 +118,31 @@ metadata:
...
+             kubectl patch crd/traffictargets.access.smi-spec.io -p '{"spec":{"conversion":{"strategy":"None", "webhook":null}}}' --type=merge;
+             kubectl patch crd/httproutegroups.specs.smi-spec.io -p '{"spec":{"conversion":{"strategy":"None", "webhook":null}}}' --type=merge;
+             kubectl patch crd/multiclusterservices.config.openservicemesh.io -p '{"spec":{"conversion":{"strategy":"None", "webhook":null}}}' --type=merge;
+             kubectl patch crd/egresses.policy.openservicemesh.io -p '{"spec":{"conversion":{"strategy":"None", "webhook":null}}}' --type=merge;
+             kubectl patch crd/ingressbackends.policy.openservicemesh.io -p '{"spec":{"conversion":{"strategy":"None", "webhook":null}}}' --type=merge;
+             kubectl patch crd/trafficsplits.split.smi-spec.io -p '{"spec":{"conversion":{"strategy":"None", "webhook":null}}}' --type=merge;
+             kubectl patch crd/tcproutes.specs.smi-spec.io -p '{"spec":{"conversion":{"strategy":"None", "webhook":null}}}' --type=merge;
+      nodeSelector:
+        kubernetes.io/os: linux
```
shashankram, line 147: This was previously:
```
  nodeSelector:
    kubernetes.io/arch: amd64
    kubernetes.io/os: linux
```

So why is the `arch` removed?

snehachhabria: by bad, added it back

shashankram (approved):

allenlsy (approved):

PR #3940, opened 2021-08-10
feat(metrics): osm watchdog using kube-state-metrics

jaellio:
**Description**:

Adds a kube-state-metrics instance to OSM. Adds a kube-state-metrics deployment,
service, ClusterRole, ClusterRoleBinding, ServiceAccount, and restricted PodSecurityPolicy.

The kube-state-metrics service listens to the kubernetes API server and exposes metrics
like container restarts and pod status. Adding kube-state-metrics to OSM will improve the
observability of OSM and provide the essential data required to configure alerting.

**Affected area**:

**Testing done**:


charts/osm/templates/kube-state-metrics-deployment.yaml
```diff
@@ -0,0 +1,50 @@
...
+        kubernetes.io/os: linux
+      containers:
+      - name: kube-state-metrics
+        ports:
+        - name: "metrics"
+          containerPort: {{.Values.OpenServiceMesh.kubeStateMetrics.port}}
+        args:
+        - --namespaces={{ include "osm.namespace" . }}
+        - --resources=pods
+        - --metric-allowlist=kube_pod_container_status_restarts_total,kube_pod_start_time,kube_pod_container_state_started,kube_pod_status_phase
```
nshankar13, line 39: So this will get the restart count, start time, status, etc, of all pods in the osm namespace?

allenlsy:
LGTM

charts/osm/values.yaml
```diff
@@ -65,6 +65,20 @@ OpenServiceMesh:
...
       time: 15d
 
+  #
+  # -- kube-state-metrics parameters
+  kubeStateMetrics:
+    # -- kube-state-metrics container resource parameters
+    resources:
+      limits:
+        cpu: "1"
+        memory: "2G"
```
shashankram, line 75: Why does this require such a high memory limit? Was this profiled? I'd be inclined to set both the limits and requests memory to a much lower value (64M).

charts/osm/values.yaml
```diff
@@ -65,6 +65,20 @@ OpenServiceMesh:
       # -- Prometheus data retention time
       time: 15d
 
+  #
+  # -- kube-state-metrics parameters
+  kubeStateMetrics:
+    # -- kube-state-metrics container resource parameters
+    resources:
+      limits:
+        cpu: "1"
```
shashankram, line 74: Doesn't .5 suffice?

charts/osm/values.yaml
```diff
@@ -118,6 +132,9 @@ OpenServiceMesh:
   # -- Deploy Prometheus with OSM installation
   deployPrometheus: false
 
+  # -- Deploy kube-state-metrics with OSM installation
+  deployKubeStateMetrics: false
```
shashankram, line 136: Any reason this should be optional?

jaellio: The user may want to bring their own kube-state-metrics rather than have the deployment of the service be managed by OSM.

charts/osm/values.yaml
jaellio: I was hoping to get feedback on the resource limit values. I wasn't sure what values would be appropriate for kube-state-metrics.

charts/osm/values.yaml
shashankram: Sounds good. One thing I would change is to introduce this param as a part of a new key called `metrics` in the values.yaml file.

charts/osm/values.yaml
shashankram: @eduser25, @nojnhuh can share details on how to profile this. I would start with a lower cpu and memory value and profile the process to understand the average/max values for these properties.

charts/osm/values.yaml
snehachhabria: I think its a good idea to make the deployment of this optional like promtheus, grafana. I agree with Shashank, a new key called metrics would be a good idea. 

```
metrics : 
  enable : true/false
```

charts/osm/templates/prometheus-configmap.yaml
```diff
@@ -32,6 +32,14 @@ data:
         - source_labels: [__meta_kubernetes_namespace, __meta_kubernetes_service_name, __meta_kubernetes_endpoint_port_name]
           action: keep
           regex: default;kubernetes;https
+      
+      - job_name: 'kube-state-metrics'
```
snehachhabria, line 36: Lets add this only if kube-state-metrics is enabled

PR #3936, opened 2021-08-09
feat(demo): Set USE_PRIVATE_REGISTRY to true

nshankar13:
Signed-off-by: nshankar13 <nshankar@microsoft.com>

**Description**: Change the USE_PRIVATE_REGISTRY default value to true

**Affected area**:
| Functional Area            |     |
| -------------------------- | --- |
| New Functionality          | [ ] |
| Documentation              | [ ] |
| Install                    | [ ] |
| CLI Tool                   | [ ] |
| Certificate Management     | [ ] |
| Control Plane              | [ ] |
| Ingress                    | [ ] |
| Egress                     | [ ] |
| Networking                 | [ ] |
| Observability              | [ ] |
| SMI Policy                 | [ ] |
| Sidecar Injection          | [ ] |
| Security                   | [ ] |
| Upgrade                    | [ ] |
| Tests                      | [ ] |
| CI System                  | [X] |
| Demo                       | [X] |
| Performance                | [ ] |
| Other                      | [ ] |

**Testing done**:

Please answer the following questions with yes/no.

1. Does this change contain code from or inspired by another project? No
    -   Did you notify the maintainers and provide attribution?

2. Is this a breaking change? No

PR #3946, opened 2021-08-10
ref(demo): update USE_PRIVATE_REGISTRY to false

michelleN:
The environment variable should be false by default as stated in the .env.example file. This PR changes the env default to false in the demo script to match.

-->
**Affected area**:

**Testing done**:
These changes were tested by running the demo successfully using a public docker registry where the container images were hosted.


shashankram (changes requested):
This script is predominantly used with a private registry for testing, so it makes sense to keep it as is. It's possible https://github.com/openservicemesh/osm/pull/3936 did not update the `.env.example` to reflect the change. Was there a reason for this change other than the comments in `.env.example` not matching?

SanyaKochhar:
@nshankar13

nshankar13:
@shashankram @michelleN the default value for USE_PRIVATE_REGISTRY was initially set to true, but we decided to change it to false since people would likely use private registries primarily used for testing. For the CI we override the default and set this to false since it uses kind, but in most other scenarios we probably want to keep it as true.

nshankar13:
Also the .env.example now sets it as true: https://github.com/openservicemesh/osm/blob/main/.env.example#L176

michelleN:
Thanks @nshankar13 and @shashankram. Closing this PR since I have more clarity now!

PR #3968, opened 2021-08-13
fix(helm-chart-kubeversion-eks): Fixes Helm chart to install on EKS

ironstar94:
**Description**: 
Fixes Helm chart to get it working on EKS Kubernetes clusters. Previously, the install was failing due to kubeVersion being checked too stringently. osm install was returning 

> Error: chart requires kubeVersion: >= 1.18.0 which is incompatible with Kubernetes v1.21.2-eks-0389ca3

as per [Helm's internal issue tracker](https://github.com/helm/helm/issues/3810) >= 1.18.0 would not match 1.18.0-anything, and EKS uses a naming convention that uses dashes after the version, resulting in the above error. This PR resolves the issue, allowing the installation to proceed on EKS clusters

**Affected area**:

**Testing done**:

Current release

image

vs

PR 

image

### DIFF
--- a/charts/osm/Chart.yaml
+++ b/charts/osm/Chart.yaml
@@ -21,7 +21,7 @@ version: 0.9.1
 appVersion: v0.9.1
 
 # This specifies the minimum Kubernetes version OSM is compatible with.
-kubeVersion: ">= 1.19.0-rc0"
+kubeVersion: ">= 1.19.0-0"
 
 dependencies:
 - name: contour
--- a/charts/osm/Chart.yaml
+++ b/charts/osm/Chart.yaml
@@ -21,7 +21,7 @@ version: 0.9.1
 appVersion: v0.9.1
 
 # This specifies the minimum Kubernetes version OSM is compatible with.
-kubeVersion: ">= 1.19.0"
+kubeVersion: ">= 1.19.0-rc0"
 
 dependencies:
 - name: contour


charts/osm/Chart.yaml
```diff
@@ -21,7 +21,7 @@ version: 0.9.1
 appVersion: v0.9.1
 
 # This specifies the minimum Kubernetes version OSM is compatible with.
-kubeVersion: ">= 1.19.0"
+kubeVersion: ">= 1.19.0-rc0"
```
trstringer, line 24: Reading through https://github.com/helm/helm/issues/3810 it appears as though the pre-release string can be anything. Using `rc0` is very specific and could leave other readers wondering if there is some logic that checks for common pre-release strings, which it appears there is not. My preference would be to use `-0` as it is a common logical base.

trstringer (changes requested):

charts/osm/Chart.yaml
ironstar94: Makes sense. I've updated it to `-0`

trstringer (approved):
Thank you for the contribution!

shashankram (approved):

ironstar94:
@trstringer Any chance this could be included in a backport for 0.9x release?

shashankram:
> @trstringer Any chance this could be included in a backport for 0.9x release?

@ironstar94 we will include this if we decide to cut the v0.9.2 patch release. Till now, there wasn't a need to cut a patch release on v0.9, but if this becomes a blocker for your use case we could cut backport this and cut a release.

ironstar94:
> > @trstringer Any chance this could be included in a backport for 0.9x release?
> 
> @ironstar94 we will include this if we decide to cut the v0.9.2 patch release. Till now, there wasn't a need to cut a patch release on v0.9, but if this becomes a blocker for your use case we could backport this and cut a release.

I've recompiled OSM cli locally, but I'm not the only user, and we would like to make our reproduce the environment without relying on internal hacks. Cutting a 0.9.2 patch release would be greatly appreciated.

shashankram:
> > > @trstringer Any chance this could be included in a backport for 0.9x release?
> > 
> > 
> > @ironstar94 we will include this if we decide to cut the v0.9.2 patch release. Till now, there wasn't a need to cut a patch release on v0.9, but if this becomes a blocker for your use case we could backport this and cut a release.
> 
> I've recompiled OSM cli locally, but I'm not the only user, and we would like to make our reproduce the environment without relying on internal hacks. Cutting a 0.9.2 patch release would be greatly appreciated.

@ironstar94 we will cut v0.9.2 mid next week. I'll backport this, thanks!

Issue #3953, opened 2021-08-11
Use `osm.tracingAddress` from charts/osm/templates/_helpers.tpl instead of `tracing.address` from values.yaml


shashankram:
Use `osm.tracingAddress` from charts/osm/templates/_helpers.tpl instead of `tracing.address` from values.yaml

PR #3977, opened 2021-08-13
chore: use osm.tracingAddress in preset-mesh-config

allenlsy:

**Description**:

Removed `tracing.address` field from values.yaml. Tracing address should be generated by `osm.tracingAddress` helper function.

**Affected area**:

**Testing done**:

CI test

2. Is this a breaking change?

### DIFF
--- a/charts/osm/values.yaml
+++ b/charts/osm/values.yaml
@@ -188,7 +188,7 @@ OpenServiceMesh:
     # -- Toggles Envoy's tracing functionality on/off for all sidecar proxies in the mesh
     enable: false
     # -- Address of the tracing collector service (must contain the namespace). When left empty, this is computed in helper template to "jaeger.<osm-namespace>.svc.cluster.local". Please override for BYO-tracing as documented in tracing.md
-    address: "jaeger.osm-system.svc.cluster.local"
+    address: ""
     # -- Port of the tracing collector service
     port: 9411
     # -- Tracing collector's API path where the spans will be sent to
--- a/charts/osm/README.md
+++ b/charts/osm/README.md
@@ -139,7 +139,7 @@ The following table lists the configurable parameters of the osm chart and their
 | OpenServiceMesh.prometheus.retention.time | string | `"15d"` | Prometheus data retention time |
 | OpenServiceMesh.pspEnabled | bool | `false` | Run OSM with PodSecurityPolicy configured |
 | OpenServiceMesh.sidecarImage | string | `"envoyproxy/envoy-alpine:v1.19.0"` | Envoy sidecar image |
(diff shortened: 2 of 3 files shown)


michelleN (commented):
It looks to me like we should keep the helper function and use it in the templates instead of getting rid of it. Currently, the default tracing address would be incorrect if the user installed the osm control plane in any namespace other than osm-system without manually changing the tracing.address value in the values.yaml file. The template helper function allows us a better more automated configuration experience than modifying by the address in the yaml file by hand or passing it in through a flag.

snehachhabria (changes requested):
+1 on @michelleN comment. I do not think we should be removing the helper function

shashankram:
> +1 on @michelleN comment. I do not think we should be removing the helper function

My bad, I created the issue directly from a Github comment and it didn't convey the correct fix. I have updated the bug title and description: https://github.com/openservicemesh/osm/issues/3953

charts/osm/values.yaml
```diff
@@ -187,10 +187,10 @@ OpenServiceMesh:
   tracing:
     # -- Toggles Envoy's tracing functionality on/off for all sidecar proxies in the mesh
     enable: false
-    # -- Address of the tracing collector service (must contain the namespace). When left empty, this is computed in helper template to "jaeger.<osm-namespace>.svc.cluster.local". Please override for BYO-tracing as documented in tracing.md
-    address: "jaeger.osm-system.svc.cluster.local"
     # -- Port of the tracing collector service
     port: 9411
+    # -- Address of the tracing collector service (must contain the namespace). When left empty, this is computed in helper template to "jaeger.<osm-namespace>.svc.cluster.local". Please override for BYO-tracing as documented in tracing.md
```
snehachhabria, line 192: nit: you can revert the change in this file

allenlsy: I found I cannot revert it. `address` somehow is a required field. Removing it will fail Chart checks

snehachhabria: looks like you were able to revert. I was not asking to remove the address field. The change you had just moved the lines

snehachhabria (approved):
Lets please link the issue this PR is resolving before merging

michelleN (dismissed):
The default value for OpenServiceMesh.tracing.address should be left empty: `""` so that the namespace doesn't get hardcoded in the string and the helper can inject the namespace. https://github.com/openservicemesh/osm/blob/main/charts/osm/values.yaml#L191

The documentation for the chart (charts/osm/README.md) should also be updated: https://github.com/openservicemesh/osm/blob/main/charts/osm/README.md

shashankram (approved):

PR #3985, opened 2021-08-14
feat(ci/action): add SMI conformance action

### LINKED ISSUE #3948: Integrate with SMI conformance GitHub action
Use SMI conformance GitHub action, to run SMI conformance tests. Would love some feedback from the community on how often this should be done. On every merge to main, every PR or every release?

https://github.com/layer5io/meshery-smi-conformance-action

cc/ @navendu-pottekkat who is available for a walk through and any questions

- CI System                  [X ]
- Project Release            [X ]

**Possible use cases**
Tells us and the community if OSM is conformant with SMI and which parts.

navendu-pottekkat:
Adds Meshery SMI Conformance GitHub action to workflows. Resolves #3948

Signed-off-by: Navendu Pottekkat <navendupottekkat@gmail.com>

**Description**:

Adds [SMI Conformance with Meshery GitHub action](https://github.com/marketplace/actions/service-mesh-interface-conformance-with-meshery) to the workflow to validate SMI Conformance on each release.

For this action to work and report test results, these should be configured and setup:

- [ ] Setup Meshery and download token as mentioned in this doc: [Running SMI Conformance Tests in CI/CD Pipelines](https://docs.meshery.io/guides/smi-conformance#running-smi-conformance-tests-in-cicd-pipelines).
- [ ] Create a repo secret `MESHERY_PROVIDER_TOKEN` and paste the token(see above).
- [ ] Share official OSM account details used in Meshery to auto-update the [SMI Conformance Dashboard](https://meshery.io/service-mesh-interface).

Once the token is uploaded and when this workflow runs, the results would automatically be updated on your Meshery Dashboard and the [SMI Conformance Dashboard](https://meshery.io/service-mesh-interface)(If you share the official account details).

**Affected area**:
| New Functionality          | [ X ] |
| CI System                  | [ X ] |

**Testing done**:

[Example of workflow running on my fork](https://github.com/navendu-pottekkat/osm/runs/3329332037?check_suite_focus=true).

### DIFF
--- a/.github/workflows/smi-conformance.yml
+++ b/.github/workflows/smi-conformance.yml
@@ -1,8 +1,9 @@
 name: SMI Conformance with Meshery
 on:
-  push:
-    tags:
-      - 'v*'
+  workflow_run:
+    workflows: ["Release Helm Chart"]
+    types:
+      - completed
 
 jobs:
   smi-conformance:
--- a/.github/workflows/smi-conformance.yml
+++ b/.github/workflows/smi-conformance.yml
@@ -1,0 +1,18 @@
+name: SMI Conformance with Meshery
+on:
+  push:
+    tags:
+      - 'v*'
+
+jobs:
+  smi-conformance:
+    name: SMI Conformance
+    runs-on: ubuntu-latest
+    steps:
+
+      - name: SMI conformance tests
+        uses: layer5io/mesheryctl-smi-conformance-action@master
+        with:
+          provider_token: ${{ secrets.MESHERY_PROVIDER_TOKEN }}
+          service_mesh: open_service_mesh
+          mesh_deployed: false


navendu-pottekkat:
Updated the workflow to run only after the latest Helm charts are released.

.github/workflows/smi-conformance.yml
```diff
@@ -0,0 +1,19 @@
+name: SMI Conformance with Meshery
+on:
+  workflow_run:
+    workflows: ["Release Helm Chart"]
+    types:
+      - completed
```
shashankram, line 6: This workflow ideally should be run per pull request or nightly, so we don't end up discovering non-conformance during the process of cutting a release.

.github/workflows/smi-conformance.yml
```diff
@@ -0,0 +1,19 @@
...
+jobs:
+  smi-conformance:
+    name: SMI Conformance
+    runs-on: ubuntu-latest
+    steps:
+
+      - name: SMI conformance tests
+        uses: layer5io/mesheryctl-smi-conformance-action@master
+        with:
+          provider_token: ${{ secrets.MESHERY_PROVIDER_TOKEN }}
```
shashankram, line 17: Feel free to reach out to me to have this updated.

shashankram (changes requested):

.github/workflows/smi-conformance.yml
navendu-pottekkat: We could do that, but for that, we need to publish helm charts on every nightly build. This is so because Meshery uses OSM's Helm charts to deploy it. 

So, with the current workflows, OSM's Helm charts are published during every release and this workflow runs after we make sure that the Helm charts are published and uses the latest one.

.github/workflows/smi-conformance.yml
navendu-pottekkat: Sure, since this is sensitive, we could discuss on Slack. You can find me on the OSM/CNCF Slack channel as **Navendu Pottekkat**.

navendu-pottekkat:
@shashankram Could you check the comments I made?

.github/workflows/smi-conformance.yml
shashankram: Is it possible for Meshery to use OSM's CLI? We could publish the CLI as an artifact with the PR workflow, and the CLI could reference compatible images to install OSM for testing. 
If Helm is a hard requirement, the alternative is to figure out a solution that works as a part of the regular PR workflow.
The goal here is to avoid discovery of non-conformance **after** a release is cut. We want to be able to continuously check for conformance.

navendu-pottekkat: Makes sense to run this action on every PR. We just recently moved from using OSM CLI to the Helm charts.

@DelusionalOptimist what were the caveats in using the OSM CLI? Could we do something like this?

leecalcote: @shashankram this is great feedback. @shashankram, @delqn, @michelleN, we've invested so much in the SMI Conformance program to date. Before we throw more effort into the initiative (by way of enhancements), can we first incorporate use of the tool as-is, gleen some value, feedback on the tests themselves, and other aspects of Meshery's approach to SMI conformance before undertaking additional enhancements? It'll be much easier to convince contributors to poor more into this effort having reached the initial milestone of employing it (even if its only reporting after release).

What do you think?

shashankram: @leecalcote I understand your concern. My only concern is that making this a part of the release job could cause the release Github action to fail and affect the release as a whole, which is far from ideal. I think we should do the opposite - verify for conformance per PR and once OSM is conformant and the action is considered stable enough, add this is a part of the release workflow.

leecalcote: You can always run Meshery and use other the UI or `mesheryctl` to run validation on an ad hoc basis.

shashankram: To clarify, my suggestion was to not run the conformance test on an ad hoc basis, but to run it more frequently (per PR, nightly) before integrating it with the release workflow.

PR #3993, opened 2021-08-16
[backport] ads: fix wrong accountability of sent resources (plus testfix)

eduser25:
This PR backports 2 commits:
  0cf7acf: ads: fix wrong accountability of sent resources
  352c220: test/ads: fix unit test flakyness (introduced by previous)

0cf7acf:
This commit introduces 2 changes:
- Knowledge of what resources a certain proxy has requested/subscribed
to, up to a current point in time and,
- the use of this new facility to avoid accounting for resources
sent that might have not been requested by envoy yet, as well as using them
to generate stub requests from ctl-driven proxy updates.

There's a potential bug in how we treated resources sent, where a
non-wildcard vertical sending more resources than the requested/subscribed ones
could make a later request be understood as an ACK.

Consider the following case:
```
req: [N:-, A]
resp: [N:1, A, B]  <- let's say vertical code generated more resources than requested
                      OSM would record that we last sent [A, B]
// Envoy will silently drop [B]
req [N:1, A, B]   <- Envoy is actually requesting B apart from A now
resp: // No response, OSM thinks this is an ACK
```

Related to #3775

352c220:
PR #3803 introduced a mapset (backed by golang map) to slice conversion,
which does not guarantee same output ordering for the same input,
thus was making a unit test fail from time to time.

This commit forces output of the slice to be alphabetically ordered,
ensuring determinism of the output slice for the same input mapset.

---

No
1. Is this a breaking change?
No


eduser25:
I have faliures on a test, i think it misses some changes from @shashankram . Will debug

snehachhabria (approved):

shashankram (approved):

jaellio (approved):

PR #4002, opened 2021-08-17
pkg/*: un-export k8s client implementing interface

shashankram:
**Description**:
Un-exports the struct types that are implementing
an interface because consumers should not be directly
interacting with the internal interface implementation.

**Affected area**:


snehachhabria (approved):
Good catch, thanks for fixing

eduser25 (approved):

Issue #3557, opened 2021-06-09
Build a proof of concept of the idea locality-based routing using Envoy


allenlsy:
**Please describe the Improvement and/or Feature Request**

This is a subtask of #3043 .  Refer to #3043 for the details of the POC.

**Scope (please mark with X where applicable)**

| Functional Area            |     |
| -------------------------- | --- |
| New Functionality          | [ ] |
| Documentation              | [ ] |
| Install                    | [ ] |
| CLI Tool                   | [ ] |
| Certificate Management     | [ ] |
| Control Plane              | ✅ |
| Ingress                    | [ ] |
| Egress                     | ✅ |
| Networking                 | ✅ |
| Observability              | [ ] |
| SMI Policy                 | [ ] |
| Sidecar Injection          | [ ] |
| Security                   | [ ] |
| Upgrade                    | [ ] |
| Tests                      | [ ] |
| CI System                  | [ ] |
| Demo                       | [ ] |
| Performance                | [ ] |
| Other                      | [ ] |

**Possible use cases**

allenlsy:
Here is a simple POC demo for this issue:

https://github.com/allenlsy/envoyproxy-locality-weighted-lb-demo

TLDR is, Envoy locality weight load balancing works as described in its documentation. By default when the overprovision rate is lower than 1.4, which means less than 71% of local workloads are healthy, Envoy starts to split traffic to other localities, based on weight and priority.

allenlsy:
cc @snehachhabria @delqn @shashankram

PR #4009, opened 2021-08-19
feat(multicluster): config cluster LB weight

allenlsy:

**Description**:

Multicluster mode support locality based load balancing. [Here is the POC](https://github.com/openservicemesh/osm/issues/3557). Traffic will be balanced to remote clusters when local cluster starts to lose available workloads.

[Here is the design doc for this feature](https://docs.google.com/document/d/1OxDPYkdffEJmraSMZpl83JOt3Kjxjfmoehp4T1ebpfs/edit#).

In this PR, we have some assumptions.

* Local endpoints have priority of 0, which is the highest priority

Related issue: #3043 

#### Multicluster demo changes

After adding this feature, in the original multicluster demo, traffic will be 100% routed to alpha cluster.  Locality based LB starts to work when the workload availability of the highest priority subset drops to below 71% (by Envoy default). To demonstrate this behavior, a new process of injecting faulty workloads is added to the multicluster demo, which injects 4 unresponsive bookstore pods, makes the availability to 20%. Thus, the demo can still traffic being splitted between alpha and beta clusters. Without the change, the multicluster demo will route traffic on alpha cluster only.

Here is how we check the healthiness of pods from bookbstore:

```
$ osm proxy get clusters bookbuyer-7577c96df9-tbsfw -n bookbuyer | grep 'health'

osm-controller::10.0.74.109:15128::health_flags::healthy
bookstore/bookstore-v1::10.1.9.49:14001::health_flags::healthy
bookstore/bookstore-v1::10.1.9.24:14001::health_flags::/failed_active_hc
bookstore/bookstore-v1::10.1.8.69:14001::health_flags::/failed_active_hc
bookstore/bookstore-v1::10.1.8.231:14001::health_flags::/failed_active_hc
bookstore/bookstore-v1::10.1.8.46:14001::health_flags::/failed_active_hc
bookstore/bookstore-v1::10.1.17.0:15443::health_flags::healthy
envoy-metrics-cluster::127.0.0.1:15000::health_flags::healthy
```

**Testing done**:

✅ Unit test
✅ Manual test
✅ Multicluster demo

**Affected area**:

No

2. Is this a breaking change?

No

### DIFF
--- a/charts/osm/crds/config_multicluster_service.yaml
+++ b/charts/osm/crds/config_multicluster_service.yaml
@@ -86,6 +86,10 @@ spec:
                         type: integer
                         minimum: 0
                         default: 0
+                      priority:
+                        description: Priority of the remote cluster in locality based load balancing
+                        type: integer
+                        minimum: 0
                       certificate:
                         description: mTLS certificates (optional)
                         type: string
--- a/pkg/endpoint/types.go
+++ b/pkg/endpoint/types.go
@@ -30,8 +30,13 @@ type Provider interface {
 
 // Endpoint is a tuple of IP and Port representing an instance of a service
 type Endpoint struct {
-	net.IP `json:"ip"`
-	Port   `json:"port"`
+	net.IP   `json:"ip"`
+	Port     `json:"port"`
+	Weight   `json:"weight"`
+	Priority `json:"priority,omitempty"`
+
+	// Zone is the zone the endpoint resides in.
+	Zone string `json:"name"`
 }
 
 func (ep Endpoint) String() string {
@@ -40,3 +45,9 @@ func (ep Endpoint) String() string {
 
 // Port is a numerical type representing a port on which a service is exposed
 type Port uint32
+
+// Weight is the load assignment weight to the endpoint. The assignment works on the endpoints with the same priority.
+type Weight uint32
+
+// Priority is the priority of the remote cluster in locality based load balancing
+type Priority uint32
--- a/pkg/envoy/eds/cluster_load_assignment.go
+++ b/pkg/envoy/eds/cluster_load_assignment.go
@@ -12,21 +12,24 @@ import (
 )
 
 const (
-	zone = "zone"
+	localZone             = "local"
+	localClusterPriority  = uint32(0)
+	remoteClusterPriority = uint32(1)
 )
 
 // newClusterLoadAssignment returns the cluster load assignments for the given service and its endpoints
 func newClusterLoadAssignment(svc service.MeshService, serviceEndpoints []endpoint.Endpoint) *xds_endpoint.ClusterLoadAssignment {
+	localLbEndpoints := &xds_endpoint.LocalityLbEndpoints{
+		Locality: &xds_core.Locality{
+			Zone: localZone,
+		},
+		LbEndpoints: []*xds_endpoint.LbEndpoint{},
+		Priority:    localClusterPriority,
+	}
+
 	cla := &xds_endpoint.ClusterLoadAssignment{
 		ClusterName: svc.EnvoyClusterName(),
-		Endpoints: []*xds_endpoint.LocalityLbEndpoints{
-			{
-				Locality: &xds_core.Locality{
-					Zone: zone,
-				},
-				LbEndpoints: []*xds_endpoint.LbEndpoint{},
-			},
-		},
+		Endpoints:   []*xds_endpoint.LocalityLbEndpoints{localLbEndpoints},
 	}
 
 	// If there are no service endpoints corresponding to this service, we
@@ -53,7 +56,28 @@ func newClusterLoadAssignment(svc service.MeshService, serviceEndpoints []endpoi
 				Value: uint32(lbWeightPerEndpoint),
 			},
 		}
-		cla.Endpoints[0].LbEndpoints = append(cla.Endpoints[0].LbEndpoints, lbEpt)
+
+		// Endpoint without a weight set implies it belongs to the local cluster
+		if meshEndpoint.Weight == 0 {
+			localLbEndpoints.LbEndpoints = append(localLbEndpoints.LbEndpoints, lbEpt)
+			continue
+		}
+
+		// Endpoint belongs to a remote cluster, configure its locality
+		remoteLbEndpoints := &xds_endpoint.LocalityLbEndpoints{
+			Locality: &xds_core.Locality{
+				Zone: meshEndpoint.Zone,
+			},
+			LbEndpoints: []*xds_endpoint.LbEndpoint{lbEpt},
+			Priority:    remoteClusterPriority,
+			LoadBalancingWeight: &wrappers.UInt32Value{
+				Value: uint32(meshEndpoint.Weight),
+			},
+		}
+		if meshEndpoint.Priority != 0 {
+			remoteLbEndpoints.Priority = uint32(meshEndpoint.Priority)
+		}
+		cla.Endpoints = append(cla.Endpoints, remoteLbEndpoints)
 	}
 
 	return cla
--- a/pkg/envoy/eds/cluster_load_assignment_test.go
+++ b/pkg/envoy/eds/cluster_load_assignment_test.go
@@ -17,6 +17,7 @@ import (
 )
 
 func TestNewClusterLoadAssignment(t *testing.T) {
+	remoteZoneName := "remote"
 	testCases := []struct {
 		name      string
 		svc       service.MeshService
@@ -35,7 +36,7 @@ func TestNewClusterLoadAssignment(t *testing.T) {
 				Endpoints: []*xds_endpoint.LocalityLbEndpoints{
 					{
 						Locality: &xds_core.Locality{
-							Zone: zone,
+							Zone: localZone,
 						},
 						LbEndpoints: []*xds_endpoint.LbEndpoint{
 							{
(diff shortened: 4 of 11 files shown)


snehachhabria:
> In multicluster mode, a multi cluster service always has endpoints available on local cluster.

I had a question regarding this assumption, if there is no endpoint on the local cluster would the logic change ?

pkg/envoy/eds/cluster_load_assignment.go
```diff
@@ -12,21 +12,23 @@ import (
 )
 
 const (
-	zone = "zone"
+	zone                  = "zone"
+	localClusterPriority  = uint32(0)
+	remoteClusterPriority = uint32(1)
```
snehachhabria, line 17: What if there are multiple remote clusters, are they all given equal remote cluster priority ? Shouldn't this be determined by the location of the cluster with respect to the source ?

allenlsy:
> I had a question regarding this assumption, if there is no endpoint on the local cluster would the logic change ?

@snehachhabria I checked the code again. I initially thought keeping at least one local workload is by design. 

I've updated the implementation. Local cluster can be configured with multi cluster service without any workload on local cluster. In this case, the traffic will be balanced among all remote clusters. There is a unit test for this scenario.

pkg/envoy/eds/cluster_load_assignment.go
allenlsy: In this PR:

* Yes, remote clusters are given the same priority. Any traffic outbound to the local cluster will be split based on weights. Sure we can assign priority to each remote clusters. I prefer doing in to another PR. We also need more works on how we use priority.
* `zone` field here is used as the name of the cluster. It's part of the endpoint name, and can be helpful for debugging.

snehachhabria: thanks for the clarification. I have one more concern here, we already have the SMI traffic split and we might want to think how that will co-exist with this

pkg/endpoint/types.go
```diff
@@ -32,6 +32,8 @@ type Provider interface {
 type Endpoint struct {
 	net.IP `json:"ip"`
 	Port   `json:"port"`
+	Weight `json:"weight"`
+	Name   string `json:"name"`
```
shashankram, line 36: Please add a comment for the field here. It's not clear what `Name` implies here. It seems to be used as an endpoint's zone instead, so maybe just call this `Zone`.

shashankram (changes requested):
Change itself looks good. Just 1 comment on the endpoint's name being used to refer to the cluster name that I'd like to see addressed.

pkg/endpoint/types.go
allenlsy: `Name` is the name of the endpoint, which is optional. In OSM multicluster mode, endpoints from local cluster are not assigned names, while endpoints from remote cluster have names. Each remote cluster has one endpoint, which is the one of its multicluster gateway. 

I think keeping it `Name` is better, as it's more general than `Zone`.

shashankram: I think making the struct field names specific makes it easier to understanding. I wouldn't generalize this unless there is a reason to, because it can be unnecessarily confusing. `Name` for an endpoint seems ambiguous to me, because an endpoint is just a representation of an ip:port combination, and naming this seems weird. On the other hand, an endpoint being a part of a `zone` (or availabilityZone, localityZone etc.) is much more clear and self-explanatory.

snehachhabria:
@allenlsy please could you share a design doc on the goals of the feature #3043, what your implementation plan is and how this PR ties with the final goal. Having an overall picture will make it easier to review the PR, the POC you have is good but the design doc will help understand better how this will fits OSM's multicluster design. Sample scenarios explaining the problem you are solving too will be helpful.

shashankram:
> In this PR, we have some assumptions.
> 
> * In multicluster mode, a multi cluster service always has endpoints available on local cluster.
> * Local endpoints have priority of 0, which is the highest priority
> * Remote endpoints all have priority of 1, regardless of latency or error rate.
> 

Per the design doc, it seems like remote clusters can each be assigned a priority. In that case, we cannot treat all remote clusters as the same priority. Is there a reason priority per remote cluster is not implemented?

charts/osm/crds/config_multicluster_service.yaml
```diff
@@ -86,6 +86,10 @@ spec:
                         type: integer
                         minimum: 0
                         default: 0
+                      priority:
```
snehachhabria, line 89: Even though this field is added, we aren't leveraging it in this PR, is that right?

snehachhabria (approved):

charts/osm/crds/config_multicluster_service.yaml
allenlsy: It's not fully utilized. In this PR we only have priority 0 for local and 1 for remote, as a minimal implementation. We can improve it in the future.

draychev (approved):

pkg/envoy/eds/cluster_load_assignment.go
```diff
@@ -47,7 +50,27 @@ func newClusterLoadAssignment(svc service.MeshService, serviceEndpoints []endpoi
...
+				LbEndpoints: []*xds_endpoint.LbEndpoint{&lbEpt},
+				Priority:    remoteClusterPriority,
+				LoadBalancingWeight: &wrappers.UInt32Value{
+					Value: uint32(meshEndpoint.Weight),
+				},
+			}
+			if meshEndpoint.Priority != 0 {
+				remoteLbEndpoints.Priority = uint32(meshEndpoint.Priority)
+			}
+			cla.Endpoints = append(cla.Endpoints, remoteLbEndpoints)
```
shashankram, line 72: Seems like this can be simplified by changing the if-else clause as:
```
if meshEndpoint.Weight == 0 {
    // do something for local cluster endpoint
    continue
}

// do something for remote cluster endpoint
```

shashankram (changes requested):
@allenlsy please rebase this change and add a few test cases for endpoints belonging to different localities in `cluster_load_assignment_test.go`. I recently refactored this area of the code and this change conflicts with that.

pkg/envoy/eds/cluster_load_assignment.go
```diff
@@ -53,7 +56,27 @@ func newClusterLoadAssignment(svc service.MeshService, serviceEndpoints []endpoi
 				Value: uint32(lbWeightPerEndpoint),
 			},
 		}
-		cla.Endpoints[0].LbEndpoints = append(cla.Endpoints[0].LbEndpoints, lbEpt)
+
+		// endpoint with Weight set is a remote cluster endpoint.
```
shashankram, line 60: Since this comment is associated with the code handling local cluster endpoints, I'd invert it.
```suggestion
		// endpoint without a weight set implies it belongs to the local cluster
```

pkg/envoy/eds/cluster_load_assignment.go
```diff
@@ -53,7 +56,27 @@ func newClusterLoadAssignment(svc service.MeshService, serviceEndpoints []endpoi
...
+		if meshEndpoint.Weight == 0 {
+			localLbEndpoints.LbEndpoints = append(localLbEndpoints.LbEndpoints, lbEpt)
+			continue
+		}
+
+		remoteLbEndpoints := &xds_endpoint.LocalityLbEndpoints{
```
shashankram, line 66:
```suggestion

                // Endpoint belongs to a remote cluster, configure its locality
		remoteLbEndpoints := &xds_endpoint.LocalityLbEndpoints{
```

pkg/envoy/eds/cluster_load_assignment_test.go
```diff
@@ -65,27 +69,122 @@ func TestNewClusterLoadAssignment(t *testing.T) {
...
-							Zone: zone,
+							Zone: localZone,
 						},
 						LbEndpoints: []*xds_endpoint.LbEndpoint{},
 					},
 				},
 			},
 		},
+		{
+			name: "multicluster: with remote gateway",
```
shashankram, line 87: Please rename this test. This code has no notion of gateway/cluster, just endpoints with weight, priority and zone.

pkg/envoy/eds/cluster_load_assignment_test.go
```diff
@@ -65,27 +69,122 @@ func TestNewClusterLoadAssignment(t *testing.T) {
...
+						Priority: uint32(2),
+						LoadBalancingWeight: &wrappers.UInt32Value{
+							Value: 10,
+						},
+					},
+				},
+			},
+		},
+		{
+			name: "multicluster: with remote gateway, no local endpoints",
```
shashankram, line 138: Same comment as L87.

pkg/envoy/eds/cluster_load_assignment.go
allenlsy: Thanks. Resolved.

pkg/envoy/eds/cluster_load_assignment_test.go
```diff
@@ -25,7 +27,7 @@ func TestNewClusterLoadAssignment(t *testing.T) {
 	}{
 		{
 			name: "multiple endpoints per cluster within the same locality",
-			svc:  service.MeshService{Namespace: "ns1", Name: "bookstore-1", TargetPort: 80},
+			svc:  meshSvc,
```
shashankram, line 30: Could you revert this change? The idea of not hardcoding this was to allow new test cases to use a different MeshService and verify the ClusterName corresponds to it.

allenlsy: Ok, got it.

shashankram (approved):

PR #4018, opened 2021-08-20
ref(cli): extract policy check-pod helpers into pkg/cli for reuse

SanyaKochhar:
**Description**: Extracts helpers to be reused by osm-health: https://github.com/openservicemesh/osm-health/pull/69

**Testing done**:

**Affected area**:
| CLI Tool                   | [x] |

2. Is this a breaking change?
no

### DIFF
--- a/cmd/cli/policy_check_pods.go
+++ b/cmd/cli/policy_check_pods.go
@@ -14,6 +14,7 @@ import (
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 
+	"github.com/openservicemesh/osm/pkg/cli"
 	osmConfigClient "github.com/openservicemesh/osm/pkg/gen/client/config/clientset/versioned"
 )
 
@@ -35,7 +36,6 @@ osm policy check-pods bookbuyer-client bookstore-server
 const (
 	namespaceSeparator       = "/"
 	defaultOsmMeshConfigName = "osm-mesh-config"
-	serviceAccountKind       = "ServiceAccount"
 )
 
 type trafficPolicyCheckCmd struct {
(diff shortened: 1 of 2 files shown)


pkg/cli/policy_check.go
```diff
@@ -0,0 +1,37 @@
...
+
+	// Map traffic targets to the given pods
+	if spec.Destination.Name == dstPod.Spec.ServiceAccountName && spec.Destination.Namespace == dstPod.Namespace {
+		return true
+	}
+	return false
+}
+
+// DoesTargetRefSrcPod checks whether the TrafficTarget spec refers to the source pod's service account
+func DoesTargetRefSrcPod(spec v1alpha3.TrafficTargetSpec, pod *corev1.Pod) bool {
```
jaellio, line 26: Nit:
```suggestion
func DoesTargetRefSrcPod(spec v1alpha3.TrafficTargetSpec, srcPod *corev1.Pod) bool {
```

jaellio (approved):

johnsonshi (approved):

johnsonshi:
@michelleN // @SanyaKochhar

I'm not sure if checking for the `TrafficTargets` in just the destination pod's namespace is enough. Shouldn't we find and list `TrafficTarget` policies across the entire cluster?


image


https://github.com/openservicemesh/osm/pull/4018/files#diff-123f780e0ae0fa0cd5d0dfe2a079a1aaa30f8af401fdfc624e37af40994fe654R135-R140

According to the [SMI spec for traffic target](https://github.com/servicemeshinterface/smi-spec/blob/main/apis/traffic-access/v1alpha2/traffic-access.md), there is no mention of any requirements for the `TrafficTarget` custom resource to live in a specific namespace. I'm specifically referring to this specific namespace key in the traffic target custom resource:

image

cmd/cli/policy_check_pods.go
```diff
@@ -14,6 +14,7 @@ import (
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 
+	policyCheckHelper "github.com/openservicemesh/osm/pkg/cli"
```
shashankram, line 17: No need to alias this internal pkg. Use `cli.Xyz` to reference variables/functions in the `cli` pkg.

```suggestion
	"github.com/openservicemesh/osm/pkg/cli"
```

cmd/cli/policy_check_pods.go
```diff
@@ -142,25 +142,16 @@ func (cmd *trafficPolicyCheckCmd) checkTrafficPolicy(srcPod, dstPod *corev1.Pod)
 	var foundTrafficTarget bool
 	for _, trafficTarget := range trafficTargets.Items {
 		spec := trafficTarget.Spec
-		if spec.Destination.Kind != serviceAccountKind {
-			continue
-		}
 
 		// Map traffic targets to the given pods
-		if spec.Destination.Name == dstPod.Spec.ServiceAccountName && spec.Destination.Namespace == dstPod.Namespace {
+		if policyCheckHelper.DoesTargetRefDstPod(spec, dstPod) {
```
shashankram, line 147: Could you invert this condition to reduce the indent:
```
if !condition {
    continue
}
// logic
```

michelleN:
> Source Pod and Source Svc Account lives in namespace 'src'.
> Destination Pod and destination svc account lives in namespace 'dest'.
> 'TrafficTarget' policy for these 2 pods lives in some other namespace 'other-ns'.

> Is that allowed - I'm specifically referring to the policy living in other-ns instead of either src or dest namespace.

No. A TrafficTarget should live in the same namespace as the destination service account. This should be updated in the spec as well if it isn't already clear there. @johnsonshi @SanyaKochhar

shashankram:
> > Source Pod and Source Svc Account lives in namespace 'src'.
> > Destination Pod and destination svc account lives in namespace 'dest'.
> > 'TrafficTarget' policy for these 2 pods lives in some other namespace 'other-ns'.
> 
> > Is that allowed - I'm specifically referring to the policy living in other-ns instead of either src or dest namespace.
> 
> No. A TrafficTarget should live in the same namespace as the destination service account. This should be updated in the spec as well if it isn't already clear there. @johnsonshi @SanyaKochhar

FYI, this hasn't been a hard requirement yet, and OSM allows traffic targets to be created in any monitored namespace. For example, the e2e tests create the TrafficTarget resource in the source namespace.

Let's get SMI to clearly state this is a requirement before implementing this change in OSM.

SanyaKochhar:
> Let's get SMI to clearly state this is a requirement before implementing this change in OSM.

@shashankram this PR doesn't change any of the traffic target logic. The line that Johnson is referring to is identical to the existing code in OSM: https://github.com/openservicemesh/osm/blob/main/cmd/cli/policy_check_pods.go#L137 so IMO the change in SMI should not block this PR

shashankram:
> > Let's get SMI to clearly state this is a requirement before implementing this change in OSM.
> 
> @shashankram this PR doesn't change any of the traffic target logic. The line that Johnson is referring to is identical to the existing code in OSM: https://github.com/openservicemesh/osm/blob/main/cmd/cli/policy_check_pods.go#L137 so IMO the change in SMI should not block this PR. The e2es can be updated once SMI has clearly stated this as a requirement

@SanyaKochhar Sorry for the confusion. I did not mean to suggest in any way that this PR should be blocked or is related to the issue in discussion.  I was merely suggesting to have SMI make this is a requirement before we remove this assumption in OSM.

SanyaKochhar:
Gotcha, thanks for clarifying!

PR #4022, opened 2021-08-23
feat(windows): Make Egress test pass on Windows

davinci26:
This PR modifies the `e2e_egress_test` to be able to run on Windows nodes

Verified locally in an AKS cluster:

 ```
    Ran 1 of 42 Specs in 471.548 seconds
    SUCCESS! -- 1 Passed | 0 Failed | 0 Pending | 41 Skipped
    --- PASS: TestSuite (471.56s)
```

**Description**:

**Testing done**:

**Affected area**:

2. Is this a breaking change?

### DIFF
--- a/charts/osm/values.yaml
+++ b/charts/osm/values.yaml
@@ -18,6 +18,7 @@ OpenServiceMesh:
   imagePullSecrets: []
   # -- Envoy sidecar image
   sidecarImage: envoyproxy/envoy-alpine:v1.19.0
+  sidecarWindowsImage: envoyproxy/envoy-windows:v1.19.0
 
   #
   # -- OSM controller parameters
--- a/pkg/envoy/bootstrap/config.go
+++ b/pkg/envoy/bootstrap/config.go
@@ -1,8 +1,6 @@
 package bootstrap
 
 import (
-	"time"
-
 	xds_accesslog_config "github.com/envoyproxy/go-control-plane/envoy/config/accesslog/v3"
 	xds_bootstrap "github.com/envoyproxy/go-control-plane/envoy/config/bootstrap/v3"
 	xds_cluster "github.com/envoyproxy/go-control-plane/envoy/config/cluster/v3"
@@ -13,7 +11,6 @@ import (
 	xds_upstream_http "github.com/envoyproxy/go-control-plane/envoy/extensions/upstreams/http/v3"
 	"github.com/golang/protobuf/ptypes"
 	"github.com/golang/protobuf/ptypes/any"
-	"google.golang.org/protobuf/types/known/durationpb"
 
 	"github.com/openservicemesh/osm/pkg/constants"
 	"github.com/openservicemesh/osm/pkg/envoy"
@@ -140,8 +137,7 @@ func BuildFromConfig(config Config) (*xds_bootstrap.Bootstrap, error) {
 		StaticResources: &xds_bootstrap.Bootstrap_StaticResources{
 			Clusters: []*xds_cluster.Cluster{
 				{
-					Name:           config.XDSClusterName,
-					ConnectTimeout: durationpb.New(time.Millisecond * 250),
+					Name: config.XDSClusterName,
 					ClusterDiscoveryType: &xds_cluster.Cluster_Type{
 						Type: xds_cluster.Cluster_LOGICAL_DNS,
 					},
--- a/tests/e2e/e2e_egress_test.go
+++ b/tests/e2e/e2e_egress_test.go
@@ -7,6 +7,7 @@ import (
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 
+	"github.com/openservicemesh/osm/pkg/constants"
 	. "github.com/openservicemesh/osm/tests/framework"
 )
 
@@ -32,15 +33,7 @@ var _ = OSMDescribe("HTTP and HTTPS Egress",
 				Expect(Td.AddNsToMesh(true, sourceNs)).To(Succeed())
 
 				// Get simple Pod definitions for the client
-				svcAccDef, podDef, svcDef, err := Td.SimplePodApp(SimplePodAppDef{
-					Name:      "client",
-					Namespace: sourceNs,
-					Command:   []string{"/bin/bash", "-c", "--"},
-					Args:      []string{"while true; do sleep 30; done;"},
-					Image:     "songrgg/alpine-debug",
-					Ports:     []int{80},
-					OS:        Td.ClusterOS,
-				})
+				svcAccDef, podDef, svcDef, err := Td.GetOSSpecificSleepPod(sourceNs)
 				Expect(err).NotTo(HaveOccurred())
 
 				_, err = Td.CreateServiceAccount(sourceNs, &svcAccDef)
@@ -51,8 +44,7 @@ var _ = OSMDescribe("HTTP and HTTPS Egress",
 				Expect(err).NotTo(HaveOccurred())
 
 				// Expect it to be up and running in it's receiver namespace
-				Expect(Td.WaitForPodsRunningReady(sourceNs, 60*time.Second, 1, nil)).To(Succeed())
-
+				Expect(Td.WaitForPodsRunningReady(sourceNs, 240*time.Second, 1, nil)).To(Succeed())
 				protocols := []string{
 					"http://",
 					"https://",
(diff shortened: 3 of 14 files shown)

### CODE REVIEW

File: tests/framework/constants.go [hunk 1]
```diff
@@ -66,3 +66,9 @@ const (
 	// ControlPlaneOnly will collect logs only for control plane processes
 	ControlPlaneOnly CollectLogsType = "controlPlaneOnly"
 )
+
+const (
+	// EnvoyOSMWindowsImage is Envoy Windows image used for testing.
+	// On Windows until Windows Server 2022 is publicly available we have to rely on this testing images.
+	EnvoyOSMWindowsImage = "openservicemesh/envoy-windows-nanoserver:latest"
```
michelleN, line 73: Could we pin this a specific version instead of latest to follow tagging best practices?

michelleN (changes requested):

tests/framework/constants.go
davinci26: yeah for sure

tests/e2e/e2e_egress_test.go
```diff
@@ -84,7 +75,7 @@ var _ = OSMDescribe("HTTP and HTTPS Egress",
 						}
 						Td.T.Logf("%s > REST req succeeded: %d", url, result.StatusCode)
 						return true
-					}, 5, 60*time.Second)
+					}, 5, 5*60*time.Second)
```
davinci26, line 78: I bumped this timeout because of a DNS bug that is happening on Windows. See https://github.com/envoyproxy/envoy/issues/17795

I can make it conditional based on the clusterOs if you want tighter checks for linux

pkg/constants/constants.go
```diff
@@ -206,6 +206,11 @@ const (
 	HTTPServerSmiVersionPath = "/smi/version"
 )
 
+// Windows Specific containers
+const (
+	WindowsNanoserverDockerImage = "mcr.microsoft.com/windows/nanoserver/insider:10.0.20348.1"
```
ksubrmnn, line 211: Should this be configurable?

pkg/constants/constants.go
```diff
@@ -206,6 +206,11 @@ const (
 	HTTPServerSmiVersionPath = "/smi/version"
 )
 
+// Windows Specific containers
```
shashankram, line 209:
```suggestion
// Windows Specific container image
```

pkg/envoy/bootstrap/config.go
```diff
@@ -141,7 +141,7 @@ func BuildFromConfig(config Config) (*xds_bootstrap.Bootstrap, error) {
 			Clusters: []*xds_cluster.Cluster{
 				{
 					Name:           config.XDSClusterName,
-					ConnectTimeout: durationpb.New(time.Millisecond * 250),
+					ConnectTimeout: durationpb.New(time.Second * 3),
```
shashankram, line 144: 1. Why is this being changed?
2. Why such a high timeout (order of several seconds)?

tests/e2e/e2e_egress_test.go
shashankram: Please open a GH issue to track this, and drop a comment here of the form:
`TODO(#<issue>: lower the timeout once .... is fixed)`.

Also, please make this conditional based on the OS so we don't relax the time for linux, as this could mask timing issues resulting from slow config updates and so on.

shashankram (changes requested):

pkg/envoy/bootstrap/config.go
davinci26: I saw timeouts on Envoy logs while I was testing this. Given that the default timeout is 5s then 0.25s seems a bit low.

Is this reliable on linux clusters? The error is recoverable so you need to see how this value behaves across cluster sizes/loads

pkg/constants/constants.go
davinci26: What is the scenario where someone would want to change it? Any WS2022 image should work here

pkg/envoy/bootstrap/config.go
shashankram: I agree this is something that depends on traffic patterns and the environment in general, but bumping this timeout to get tests to pass seems a bit mysterious. At the very least, we should set a well-documented constant (even prior to this change) with a comment on why this value.

pkg/constants/constants.go
shashankram: This seems to be used only in tests, so move this constant to `tests/framework`.

pkg/envoy/bootstrap/config.go
davinci26: > but bumping this timeout to get tests to pass seems a bit mysterious


The test would pass regardless, it just makes for better logging because it reduces the noise from intermittent failures when reaching the control plane.

> At the very least, we should set a well-documented constant (even prior to this change) with a comment on why this value.

I was thinking that maybe we should not set this value and let Envoy set the default. WDYT? Otherwise I can push it to a constant with a comment

pkg/constants/constants.go
davinci26: +1, there are two constants.go files and probably my idea opened the non-test one first

pkg/envoy/bootstrap/config.go
shashankram: @davinci26 if the default value works well, sure that seems reasonable.

tests/framework/constants.go
davinci26: fixed by pinning the image to the sha digest

pkg/envoy/bootstrap/config.go
davinci26: I just removed it so we let envoy decide on the default

pkg/constants/constants.go
davinci26: fixed

pkg/constants/constants.go
davinci26: moved it into the correct constants file

tests/e2e/e2e_egress_test.go
davinci26: fixed, see https://github.com/openservicemesh/osm/issues/4027

charts/osm/values.yaml
```diff
@@ -18,6 +18,7 @@ OpenServiceMesh:
   imagePullSecrets: []
   # -- Envoy sidecar image
   sidecarImage: envoyproxy/envoy-alpine:v1.19.0
+  sidecarWindowsImage: envoyproxy/envoy-windows:v1.19.0
```
shashankram, line 21: nit: add comment similar to L19 per property here for clarity

shashankram (approved):

tests/framework/common.go
```diff
@@ -387,7 +392,7 @@ func (td *OsmTestData) LoadImagesToKind(imageNames []string) error {
 	return nil
 }
 
-func setMeshConfigToDefault(instOpts InstallOSMOpts, meshConfig *v1alpha1.MeshConfig) (defaultConfig *v1alpha1.MeshConfig) {
+func setMeshConfigToDefault(instOpts InstallOSMOpts, meshConfig *v1alpha1.MeshConfig, clusterOS string) (defaultConfig *v1alpha1.MeshConfig) {
```
ksubrmnn, line 395: is clusterOS used in this function?

tests/framework/common_apps.go
```diff
@@ -472,6 +473,30 @@ func (td *OsmTestData) SimpleDeploymentApp(def SimpleDeploymentAppDef) (corev1.S
...
 
+// GetOSSpecificSleepPod returns a simple OS specific busy loop pod.
+func (td *OsmTestData) GetOSSpecificSleepPod(sourceNs string) (corev1.ServiceAccount, corev1.Pod, corev1.Service, error) {
+	if td.ClusterOS == constants.OSWindows {
+		return Td.SimplePodApp(SimplePodAppDef{
+			Name:      "client",
+			Namespace: sourceNs,
+			Command:   []string{"cmd", "/c"},
+			Args:      []string{"FOR /L %N IN () DO ping -n 30 127.0.0.1> nul"},
+			Image:     WindowsNanoserverDockerImage,
```
ksubrmnn, line 484: I love that we can use nanoserver here!

tests/framework/common.go
davinci26: you are right, this one is fixed

tests/framework/common_apps.go
davinci26: It takes some hacking but I really want to have all the images be nano based because it makes things so much easier

ksubrmnn (approved):

michelleN (approved):

charts/osm/values.yaml
davinci26: can we get this one merged and I can do this in a follow up today

Issue #1704, opened 2020-09-16
Use Go std testing library for unit tests


michelleN:
**Background:**
Historically, we leveraged [Ginkgo](https://onsi.github.io/ginkgo/), a BDD style testing framework,  and [Gomega](https://onsi.github.io/gomega/), an assertion/matcher library best paired with Gingko, for unit tests but after using it for a while we discovered:

- This is a heavier and more verbose framework than we really need
- Using the Go standard testing package is easier to work with when writing table-style tests and works well for most places in this codebase.
- Learning Gomega and Gingko also comes at a learning curve and may prove to be a barrier to entry when first contributing to this repository.

**Proposal:**
Let's move back to using the Go standard testing package for unit tests in this repository. Over time, we'll convert the Gingko tests back to using the testing package but that is not a high priority any time soon. We can and should leverage Gingko/Gomega for e2e tests like Kubernetes and other projects do. 

**Open Questions:**
Assertion/matcher libraries like Gomega and [Testify](https://github.com/stretchr/testify) help make it easier to write tests and more cleanly assert equality, replacing if/else statements. [Testify](https://github.com/stretchr/testify) is a toolkit that offers more tools outside of just their assert package. We use both Gomega and Testify in our project today. A question for the maintainers/community: _**should we continue using both assertion type libraries or move toward one or none?**_ 

When thinking about this it might be helpful to think about what is important to you. When discussing with the team, a things that bubbled up as important:
- making tests easier to read/write (simplicity is key) but not increasing the barrier to entry too much
- asserting object equality (although I think we might be able to argue that reflect.DeepEqual is enough)
- pretty output when objects are not equivalent

**_What are other factors that are important when deciding which assertion library to pick?_**

codefromthecrypt:
FYI I felt your issue helpful, so thanks for capturing rationale. I also built more rationale here in case it is useful to you. cheers https://github.com/tetratelabs/getenvoy/pull/130

snehachhabria:
OSM has been using Go std library as the default for unit testing for some time, closing this issue as all new code has been using the same. We can reopen another issue when we need to clean up old test code.

Issue #1489, opened 2020-08-10
(root) Improve test coverage (goal: 80%)


vramakrishnan:
The current go test coverage is low for many modules as shown below.
Filing this issue for tracking the test coverage improvement.

---

> As of 2021-06-30 `make go-test-coverage` reports 70% unit test coverage.

---

- [x] #3363
- [x] #3408
- [x] #3412
- [x] #3415
- [ ] #3416
- [x] #3417
- [x] #3422
- [x] #3425
- [x] #3429
- [x] #3432
- [x] #3437
- [ ] #3452
- [x] #3458
- [ ] #3559
- [ ] #3563

PR #4057, opened 2021-08-31
tests: Update tests to use Go Testing

snehachhabria:
**Description**:

Update the existing tests to use go testing and improve coverage to 80%

Part of #1704 and #1489

**Testing done**:
Unit tests updated 

**Affected area**:

1. Does this change contain code from or inspired by another project? `no

### DIFF
--- a/cmd/osm-controller/validate_test.go
+++ b/cmd/osm-controller/validate_test.go
@@ -1,160 +1,147 @@
 package main
 
 import (
-	. "github.com/onsi/ginkgo"
-	. "github.com/onsi/gomega"
+	"testing"
+
+	tassert "github.com/stretchr/testify/assert"
 
 	"github.com/openservicemesh/osm/pkg/certificate/providers"
 )
 
-var _ = Describe("Test validateCertificateManagerOptions", func() {
-	var (
-		testCaBundleSecretName = "test-secret"
-	)
+func TestValidateCertificateManagerOptions(t *testing.T) {
+	testCases := []struct {
+		name               string
+		certProvider       string
+		vaultToken         string
+		caBundleSecretName string
+		issuerName         string
+		expectError        bool
+	}{
+		{
+			name:         "Cert Provider : Tresor",
+			certProvider: providers.TresorKind.String(),
+			expectError:  false,
+		},
+		{
+			name:         "Cert Provider : Vault and token is not empty",
+			certProvider: providers.VaultKind.String(),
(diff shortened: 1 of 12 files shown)


trstringer (approved):

cmd/osm-controller/validate_test.go
```diff
@@ -1,160 +1,149 @@
...
+		},
+		{
+			name:         "Cert Provider : InvalidProvider",
+			certProvider: "InvalidProvider",
+			expectError:  true,
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
```
nojnhuh, line 68: Let's move the `assert := tassert.New(t)` from the top of the test to here inside the sub-test so if a test case fails it's easier to pinpoint which sub-test the error message refers to.

(Also applicable to `TestValidateCLIParams` below)

pkg/k8s/event_handlers_test.go
```diff
@@ -21,65 +19,57 @@ const (
...
+	a.Nil(pubsubMsg.OldObj)
+
+	// Cast New obj, expect v1.Pod
+	podObj, castOk := pubsubMsg.NewObj.(*corev1.Pod)
+	a.True(castOk)
+	a.Equal(podObj.Name, "pod-name")
+	a.Equal(podObj.Namespace, testNamespace)
+
+	podAddChannel = events.Subscribe(announcements.PodAdded)
+	defer events.Unsub(podAddChannel)
```
nojnhuh, line 60: I can't imagine this is actually a problem, but merging the two `It`s into one function here means for `podAddChannel` the test was doing

```
Subscribe()
Unsub()
Subscribe()
Unsub()
```

where now it does

```
Subscribe()
Subscribe()
Unsub()
Unsub()
```

nbd if the test reliably passes, just an observation.

nojnhuh (approved):
Left one suggestion and one comment, neither should block this.

pkg/k8s/event_handlers_test.go
shashankram: Let's not change the behavior of the existing test unless there's a bug.

snehachhabria: This test is primarily checking if correct events are being subscribed to and the assert statements are helping in verifying the same. The order of Subscribe() and Unsub() has changed but that doesn't impact behavior. We should be okay here

PR #4059, opened 2021-08-31
feat(windows): Change the deployments to deploy demo on Windows

### LINKED ISSUE #3886: Make demo run on Windows
Add an additional environment variable to make the demo deploy all the pods to Windows workers.

One way we could implement that feature is by introducing an environment variable set on `.env` file that scripts can use to switch between windows and linux containers.

Additionally, we should expand the docs to document this change.

davinci26:
With this change you can deploy the demo on cluster with Windows nodes

1. Added a `run-osm-windows-demo.sh` with the options needed to run osm on Windows
2. Configured the timeouts to accomodate for some intermittent failures on Windows
3. Added a bookbuyer service to be able to visualize the demo

Fixes #3886

**Description**:

**Testing done**:

**Affected area**:

2. Is this a breaking change?

### DIFF
--- a/demo/deploy-bookbuyer.sh
+++ b/demo/deploy-bookbuyer.sh
@@ -30,6 +30,30 @@ if [ "$DEPLOY_ON_OPENSHIFT" = true ] ; then
     fi
 fi
 
+CONTAINER_NAME="bookbuyer"
+if [ "$OS" = windows ]; then
+CONTAINER_NAME="bookbuyer-windows"
+fi
+
+echo -e "Deploy bookbuyer Service"
+kubectl apply -f - <<EOF
+apiVersion: v1
+kind: Service
+metadata:
+  name: bookbuyer
+  namespace: $BOOKBUYER_NAMESPACE
+  labels:
+    app: bookbuyer
+spec:
+  ports:
+  - port: 14001
+    name: bookbuyer-port
+
+  selector:
+    app: bookbuyer
+EOF
+
+
 echo -e "Deploy BookBuyer Deployment"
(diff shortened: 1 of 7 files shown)


demo/deploy-bookbuyer.sh
```diff
@@ -30,6 +30,30 @@ if [ "$DEPLOY_ON_OPENSHIFT" = true ] ; then
...
+spec:
+  ports:
+  - port: 14001
+    name: bookbuyer-port
+
+  selector:
+    app: bookbuyer
+EOF
+
+
```
shashankram, line 56: Per your commit message, this helps visualize the bookbuyer service. I am unsure what this means because we should ideally not be creating a service for client pods (bookbuyer is a client accessing the bookstore service), and creating this could mask bugs in the control plane.

Could you clarify why this is necessary?

davinci26: I will explain what I wanted to do and if you have any better suggestions, I am open to them.

I wanted to expose a service to the internet so we could visualize the html and display the traffic split stats. The bookbuyer seemed like a good candidate for this.

Since port forwarding is not an option for the Windows demo, I used ngnix as an ingress gateway but this can only expose services. So I made a book buyer service

shashankram: If you wanted to view the UI externally, then this seems reasonable, although be aware that the windows demo introduces a dependency on ingress, unlike the linux demo.

Could you describe in the commit message and code why the bookstore service is required because that info is missing in this PR.

Issue #2527, opened 2021-02-12
Implement traffic split in permissive mode


michelleN:
Currently, traffic split policies are not applied in permissive mode. With the completion of routes v2 work, we'll be able to apply traffic split policies in permissive mode while still ignoring any SMI traffic access policies.

cc/ @snehachhabria

PR #4060, opened 2021-08-31
test(e2e): test traffic split in permissive mode

michelleN:

**Description**: adds e2e test for #2527 - traffic split in permissive mode

**Testing done**:

**Affected area**:
| Tests                      | [X ] |

    -   Did you notify the maintainers and provide attribution? No

### DIFF
--- a/tests/e2e/e2e_permissive_trafficsplit_test.go
+++ b/tests/e2e/e2e_permissive_trafficsplit_test.go
@@ -16,30 +16,29 @@ import (
 	. "github.com/openservicemesh/osm/tests/framework"
 )
 
-var _ = OSMDescribe("Test HTTP from N Clients deployments to 1 Server deployment backed with Traffic split test",
+var _ = OSMDescribe("Test Permissive HTTP from N Clients deployments to 1 Server deployment backed with Traffic split",
 	OSMDescribeInfo{
 		Tier:   1,
-		Bucket: 10,
+		Bucket: 7,
 	},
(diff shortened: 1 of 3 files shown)


tests/e2e/e2e_permissive_trafficsplit_test.go
```diff
@@ -0,0 +1,299 @@
+package e2e
```
shashankram, line 1: I was wondering if it made sense to extend the existing e2e_trafficsplit_test.go that adds an additional variable in the top-level context for SMI vs permissive mode: https://github.com/openservicemesh/osm/blob/main/tests/e2e/e2e_trafficsplit_test.go#L24-L32. That way, we would only need to set the mode and everything in the test would remain the same, which makes it easier to maintain the test.

michelleN: yea I agree. That's a great idea.

shashankram: I extended the existing traffic split test to permissive mode in https://github.com/openservicemesh/osm/pull/4074/commits

shashankram:
Traffic split test in permissive mode has been added as a part of https://github.com/openservicemesh/osm/pull/4074/commits: https://github.com/openservicemesh/osm/blob/main/tests/e2e/e2e_trafficsplit_test.go#L35-L38. Closing this since it's not required.

Issue #4052, opened 2021-08-27
Refactor routes, clusters, and listener configurations to address implementation gaps and simplify


shashankram:
**Please describe the Improvement and/or Feature Request**
Refactor routes, clusters, and listener configurations to address implementation gaps in the control plane and simplify control plane configuration.

This issue should address the following:
- [x] https://github.com/openservicemesh/osm/issues/3777
- [x] Eliminate outbound policy merges in pkg/catalog
- [x] Remove duplicate or redundant APIs
- [x] Build LDS/CDS/RDS configurations objects using the same API in pkg/catalog as much as possible (to aid in building configs for a proxy using the same cache/system state as a part of a future effort to avoid inconsistent view of the caches)
- [x] Correctly and uniformly derive the hostnames for a MeshService (currently limited to Kubernetes)
- [x] Support filtering and routing traffic correctly when service's Port and TargetPort are different
- [x] #2527 

**Scope (please mark with X where applicable)**
- Control Plane        [X]

**Possible use cases**
Required to address implementation gaps in the control plane and simplify the existing code.

shashankram:
Resolved by https://github.com/openservicemesh/osm/pull/4074

PR #4064, opened 2021-08-31
pkg/smi: allow filter options while listing items in cache

shashankram:
**Description**:
There is a use case to only get a subset of items in the
cache for TrafficTarget and TrafficSplit resources based
on certain fields being set on the resources. This change
implements optional filters that clients can use to filter
the returned items from the cache without needing to define
additional functions. With this change, the TrafficTarget
cache can be filtered based on the destination identity,
and the TrafficSplit cache can be filtered based on the
apex and backend service.

Required for #4052

**Testing done**:
Unit tests

**Affected area**:

### DIFF
--- a/pkg/smi/fake.go
+++ b/pkg/smi/fake.go
@@ -19,6 +19,7 @@ type fakeMeshSpec struct {
 }
 
 // NewFakeMeshSpecClient creates a fake Mesh Spec used for testing.
+// TODO(DEPRECATE): This fake is not extendable enough, deprecate it and use mocks or re-implement fakes
 func NewFakeMeshSpecClient() MeshSpec {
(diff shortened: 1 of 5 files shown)


pkg/smi/fake.go
```diff
@@ -19,6 +19,7 @@ type fakeMeshSpec struct {
 }
 
 // NewFakeMeshSpecClient creates a fake Mesh Spec used for testing.
+// TODO(DEPRECATE): This fake is not extendable enough, deprecate it and use mocks or re-implement fakes
```
snehachhabria, line 22: +1

snehachhabria (approved):

allenlsy (approved):

PR #4066, opened 2021-09-01
endpoints: don't filter service endpoints in permissive mode

shashankram:
**Description**:
Currently, permissive mode bypasses EDS and uses Envoy's
OriginalDestination cluster for service discovery.
As a part of implementing support for TrafficSplit in
permissive mode, permissive mode will also leverage
EDS for endpoint discovery. The existing API leveraged
by EDS to discover the endpoints of a service are
not usable for permissive mode, so this change adds
support for the same.

This change also renames the function for clarity
and removes an implicit comment.

Required by #4052 and #2527.

**Testing done**:
Unit tests and verified permissive mode with local changes.

**Affected area**:

### DIFF
--- a/pkg/catalog/endpoint.go
+++ b/pkg/catalog/endpoint.go
@@ -38,15 +38,24 @@ func (mc *MeshCatalog) GetResolvableServiceEndpoints(svc service.MeshService) ([
 	return endpoints, nil
 }
 
-// ListEndpointsForServiceIdentity returns a list of endpoints that belongs to an upstream service accounts
-// from the given downstream identity's perspective
-// Note: ServiceIdentity must be in the format "name.namespace" [https://github.com/openservicemesh/osm/issues/3188]
-func (mc *MeshCatalog) ListEndpointsForServiceIdentity(downstreamIdentity identity.ServiceIdentity, upstreamSvc service.MeshService) ([]endpoint.Endpoint, error) {
+// ListAllowedUpstreamEndpointsForService returns the list of endpoints over which the downstream client identity
+// is allowed access the upstream service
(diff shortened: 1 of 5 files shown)


pkg/catalog/endpoint.go
```diff
@@ -38,15 +38,24 @@ func (mc *MeshCatalog) GetResolvableServiceEndpoints(svc service.MeshService) ([
 	return endpoints, nil
 }
 
-// ListEndpointsForServiceIdentity returns a list of endpoints that belongs to an upstream service accounts
-// from the given downstream identity's perspective
-// Note: ServiceIdentity must be in the format "name.namespace" [https://github.com/openservicemesh/osm/issues/3188]
-func (mc *MeshCatalog) ListEndpointsForServiceIdentity(downstreamIdentity identity.ServiceIdentity, upstreamSvc service.MeshService) ([]endpoint.Endpoint, error) {
+// ListAllowedUpstreamEndpointsForService returns the list of endpoints over which the downstream client identity
+// is allowed access the upstream service
+func (mc *MeshCatalog) ListAllowedUpstreamEndpointsForService(downstreamIdentity identity.ServiceIdentity, upstreamSvc service.MeshService) ([]endpoint.Endpoint, error) {
```
allenlsy, line 43: In this issue https://github.com/openservicemesh/osm/issues/3190, we discussed about dropping the word `allowed` from method names. Basically we don't return endpoints that are not allowed.

nojnhuh (approved):

pkg/catalog/endpoint.go
shashankram: For permissive mode, everything is allowed. For SMI, a subset of endpoints is allowed. Hope that makes sense.

allenlsy (approved):

snehachhabria (approved):

PR #4071, opened 2021-09-01
chore(ci): create release on any tag

nojnhuh:
**Description**:
This change modifies the release CI workflow to run on any tag, not just
ones starting with 'v'. To facilitate testing, nightly releases will be
created based on tags of the form `nightly-YYYY-MM-DD`.

**Testing done**: Verified by pushing test tags to my fork.

**Affected area**:


snehachhabria (approved):

nojnhuh:
> I wasn't sure if it's a good idea to create a pre-release every night if all we want are published charts and binary artifacts.

The main gap here is CLI artifacts, which are currently uploaded to the GitHub release. Where should those go if there's no release?

> do you think it makes sense to publish the charts to a branch other than `gh-pages`, something like `nightly-builds`?

The Helm repo is hosted on a GitHub Pages site, which is built from the gh-pages branch. Will GitHub let us create another GH pages site for the repo?

shashankram:
> > I wasn't sure if it's a good idea to create a pre-release every night if all we want are published charts and binary artifacts.
> 
> The main gap here is CLI artifacts, which are currently uploaded to the GitHub release. Where should those go if there's no release?
> 

The CLI binaries could be uploaded as artifacts with the nighly job, similar to how logs are uploaded when a job fails in the CI.

> > do you think it makes sense to publish the charts to a branch other than `gh-pages`, something like `nightly-builds`?
> 
> The Helm repo is hosted on a GitHub Pages site, which is built from the gh-pages branch. Will GitHub let us create another GH pages site for the repo?

I am unsure, but there's another option. We could publish a tarball of the Helm charts as an artifact of the job, and have users create a wrapper chart for testing that could specify OSM's chart as a dependency available locally (e.g. `repository: "file://../charts/osm"`. 

Would these solve the problem of testing nighly builds without needing to cut a pre-release?

trstringer (approved):

nojnhuh:
Uploading everything for nightly builds as artifacts of the workflow run would work, but it would be much more involved to retrieve than if they were uploaded to a release.

For example, the URL I see in the UI for an artifact on a job looks like https://github.com/openservicemesh/osm/suites/3665817106/artifacts/89112184, which looks like it would involve querying the GitHub API several times. The URL for a release asset looks like https://github.com/openservicemesh/osm/releases/download/v0.9.2/osm-v0.9.2-linux-amd64.tar.gz, which can be constructed trivially given a tag name and a known file name.

If we use job artifacts for nightlies and release assets for other releases, then any tooling pulling down releases would also need to know that difference. Overall I agree that publishing nightlies as releases will clutter the history, but making two completely separate paths for different release types seems like it would cause more headaches.

@draychev and @allenlsy, do you have any thoughts?

shashankram:
> Overall I agree that publishing nightlies as releases will clutter the history

This was my main concern. I am not opposed to nightly releases but that it could become a distraction and clutter the actual release history and charts uploaded to gh-pages. I don't know if there's anything additional that would need to be done so `helm install` using the repo URL doesn't pick up the nightly release as the latest released binaries.

nojnhuh:
> I don't know if there's anything additional that would need to be done so `helm install` using the repo URL doesn't pick up the nightly release as the latest released binaries.

Testing this out, it looks like the tag will need to be a valid semver, otherwise `helm package` fails. Depending on what major/minor/patch semver we use (and tag them `vMAJOR.MINOR.PATCH-nightly+YYYY-MM-DD`), we can tweak how easy they are to find. If we use something like `v1.0.0-...` (or anything greater than the latest pre-release), then nightlies will be used by default for `helm install --devel`. If we use `v0.0.0-...`, then the non-nightly pre-releases will take precedence over nightlies in that case.

michelleN (approved):

.github/workflows/release-chart.yml
```diff
@@ -2,7 +2,7 @@ name: Release Helm Chart
 on:
   push:
     tags:
-      - "v*"
+      - "*"
```
shashankram, line 5: This breaks the new release workflow where a pre-rel-* tag is pushed to create the pre-release workflow as a part of the image publishing step.

shashankram (changes requested):

PR #4073, opened 2021-09-02
feat(e2e): ensures namespace ignore label takes precedence

jaellio:
**Description**:

This test verifies that the ignore label on a namespace resource
takes precedence over the values of the monitored-by label and the
sidecar injection annotation if they have been applied to a Ns.
The ignore label prevents sidecar injection from occurring in the
namespace the label is applied to.

**Testing done**:

**Affected area**:
| Tests                      | [ X ] |

### DIFF
--- a/tests/e2e/e2e_ignore_namespace_test.go
+++ b/tests/e2e/e2e_ignore_namespace_test.go
@@ -21,9 +21,6 @@ var _ = OSMDescribe("Ignore Namespaces",
 	func() {
 		Context("Ignore Label", func() {
 			const ignoreNs = "ignore"
-			const monitorIgnoreNs = "mesh-ignore"
-			const sidecarMonitorIgnoreNs = "sidecar-monitor-ignore"
-			var ns []string = []string{ignoreNs, monitorIgnoreNs, sidecarMonitorIgnoreNs}
 
 			It("Tests the ignore label on a namespace disables sidecar injection", func() {
 				// Install OSM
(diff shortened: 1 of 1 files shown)


tests/e2e/e2e_ignore_namespace_test.go
```diff
@@ -0,0 +1,197 @@
...
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/openservicemesh/osm/pkg/constants"
+	. "github.com/openservicemesh/osm/tests/framework"
+)
+
+var _ = OSMDescribe("Ignore Namespaces",
+	OSMDescribeInfo{
+		Tier:   1,
+		Bucket: 2,
```
jaellio, line 19: Will update to `Tier: 2` after PR review is complete

nshankar13 (approved):

tests/e2e/e2e_ignore_namespace_test.go
```diff
@@ -0,0 +1,197 @@
...
+				installOpts.EnablePermissiveMode = true
+				Expect(Td.InstallOSM(installOpts)).To(Succeed())
+
+				// Create test NS in mesh with ignore label
+				for _, n := range ns {
+					Expect(Td.CreateNs(n, map[string]string{constants.IgnoreLabel: "true"})).To(Succeed())
+				}
+
+				// Add monitor-ignore to mesh with sidecar injection disabled
+				Expect(Td.AddNsToMesh(false, monitorIgnoreNs)).To(Succeed())
```
shashankram, line 40: This test case here seems unnecessary given that this test is meant to test the ignore label. Just testing ignore label with sidecar-injection enabled should suffice. This is a k8s functionality that we are testing now and not OSM, so I suggest simplifying this test to only test the ignore label and not all combinations of the `matchExpression` selector exposed by the k8s API.

nojnhuh (commented):
Could this check instead be expressed as a unit test? I don't see a test for this particular scenario, but similar tests are here: https://github.com/openservicemesh/osm/blob/2cec5d7df7d3b71d9ffff9d89e36ecc6f4f43eaa/pkg/injector/webhook_test.go#L189

shashankram:
> Could this check instead be expressed as a unit test? I don't see a test for this particular scenario, but similar tests are here:
> 
> https://github.com/openservicemesh/osm/blob/2cec5d7df7d3b71d9ffff9d89e36ecc6f4f43eaa/pkg/injector/webhook_test.go#L189

@nojnhuh this test is primarily testing whether k8s [matchExpression](https://github.com/openservicemesh/osm/blob/main/charts/osm/templates/mutatingwebhook.yaml#L23-L26) on the namespace selector configuration works as expected. I think the idea of this test was to confirm that we are correctly implementing the desired behavior (ignore label overrides sidecar-injection & monitor annotation and labels respectively). A simple e2e test as noted in https://github.com/openservicemesh/osm/pull/4073#discussion_r702011730 should suffice in my opinion. I don't think we should be testing various permutations around k8s API behavior here.

tests/e2e/e2e_ignore_namespace_test.go
shashankram: The Tier doesn't matter anymore. All tests run for PR workflows.

shashankram (approved):

PR #4094, opened 2021-09-09
feat(reconciler): Rename mutating webhook file name

snehachhabria:
**Description**:

Update filename of mutating webhook reconciler


shashankram (commented):
Could you instead rename the file to `mutating_webhook` so that the full path becomes `pkg/reconciler/mutating_webhook`? I don't think we need to suffix `reconciler` for the related code within the `reconciler` folder.

shashankram (approved):

trstringer (approved):

Issue #3715, opened 2021-07-01
container images are not pinned


KnicKnic:
Container images should include the sha256 so that if there is a supply chain attack users are less vulnerable.

Ex:   `image: envoyproxy/envoy-alpine:v1.18.3`
Should be - `image: envoyproxy/envoy-alpine:v1.18.3@sha256:6225750f76e1e995690ce4a512aaff3ba5d9d70426926ec821b867346a703c0e`

In this current instance this is made more problematic by specifying `ALWAYS` for image pull policy, meaning the second the container is updated with malicious payload it can be pulled down when a pod is created.

**Scope (please mark with X where applicable)**
- New Functionality          [ x ]

shashankram:
The scope of the issue is a bit larger given that updating the control plane image references to use the image digest requires the following additional tasks to be completed:
1. Remove image references from Go code (should only be derived from chart values or user input)
2. Tweak release process to use image digests

shashankram:
The next release of OSM (v0.11) will have the images pinned.

PR #4108, opened 2021-09-14
charts/osm: use image digest for default images

shashankram:
Default images corresponding to the latest released
image should use the image digest instead of tags
to avoid supply chain attacks, which tags are vulernable
to due to being mutatble. Image digests are immutable.

This change updates the way images are published during
a release and also the release workflow as noted in the
release guide. The change does the following:

1. Uses image digests for the control plane images instead
   of tags when the digests are specified. The digests must
   always be specified in release branches, while the code
   in the main branch can continue to use tags for better
   compatibility between the charts and images. As a result
   of this change, all images in the main branch will leverage
   the `latest-main` tag instead of release tags. This is
   necessary because the CLI/charts are not backward compatible
   with the previous release and thus unusable. The image refs
   in the main branch will thus use the `latest-main` tag.

2. Updates the release workflow to include a pre-release step
   to publish the control plane images for the release prior
   to the final release being cut. This is necessary because
   the image digests for the released images published to the
   container registry must be included in the charts and CLI
   as a part of the final release step. The release step will
   not rebuild the control plane images, but instead encode them
   into the charts/CLI and upload the release binaries.

3. Updates the release guide to reflect the new release process.

Part of #3715

**Testing done**:
- Local demo run

When tags are not specified:
```
        "initContainerImage": "openservicemesh/init@sha256:3717f5054f835ccefc751c6fe5d4b2824038dc49514e811bf867093206ec2ce1",
          image: "openservicemesh/osm-crds@"
          image: "openservicemesh/osm-bootstrap@"
          image: "openservicemesh/osm-controller@sha256:f77659d771d82c8f053bf008fd513574bde5021ea344289ffd47de863fac4461"
          image: "openservicemesh/osm-injector@sha256:c3ded5e1cd4b02474aac573a7437c8e8a91ada556e05060701af4be3982c02b2"
          image: "openservicemesh/osm-crds@"
```

When tags are specified:
```
        "initContainerImage": "openservicemesh/init:foo",
          image: "openservicemesh/osm-crds:foo"
          image: "openservicemesh/osm-bootstrap:foo"
          image: "openservicemesh/osm-controller:foo"
          image: "openservicemesh/osm-injector:foo"
          image: "openservicemesh/osm-crds:foo"
```

**Affected area**:


trstringer (approved):

nojnhuh (changes requested):

docs/release_guide.md
```diff
@@ -54,13 +54,13 @@ If there are other commits on the `main` branch to be included in the release (s
 
 Create a new commit on the patch branch to update the hardcoded version information in the following locations:
 
-* The container image tag in [charts/osm/values.yaml](/charts/osm/values.yaml)
+* The container image digests `OpenServiceMesh.image.digest` for images in [charts/osm/values.yaml](/charts/osm/values.yaml)
```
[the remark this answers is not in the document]
shashankram: @nojnhuh That's a valid question I didn't address in my commit message, but I think the release workflow should be updated to:
1. Publish the images with the appropriate release tag
2. Make the changes to retrieve the digest from the published images as a part of building the binaries for the release

Is that something that is feasible?

charts/osm/templates/cleanup-hook.yaml
```diff
@@ -128,7 +128,11 @@ spec:
       restartPolicy: Never
       containers:
         - name: garbage-collector
+          {{- if .Values.OpenServiceMesh.image.tag }}
           image: "{{ .Values.OpenServiceMesh.image.registry }}/osm-crds:{{ .Values.OpenServiceMesh.image.tag }}"
+          {{- else }}
+          image: "{{ .Values.OpenServiceMesh.image.registry }}/osm-crds@{{ .Values.OpenServiceMesh.image.digest.osmCRDs }}"
+          {{- end }}
```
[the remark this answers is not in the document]
shashankram: Yes, although templating can happen later as well.

pkg/constants/constants.go
```diff
@@ -67,7 +67,7 @@ const (
 	DefaultEnvoyWindowsImage = "envoyproxy/envoy-windows@sha256:c904fda95891ebbccb9b1f24c1a9482c8d01cbca215dd081fc8c8db36db85f85"
 
 	// DefaultInitContainerImage is the default init container image if not defined in the osm MeshConfig
-	DefaultInitContainerImage = "openservicemesh/init:v0.9.2"
+	DefaultInitContainerImage = "openservicemesh/init@sha256:3717f5054f835ccefc751c6fe5d4b2824038dc49514e811bf867093206ec2ce1"
```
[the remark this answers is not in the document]
shashankram: I'll wanted to make this a binary choice between using a tag and a digest because of the way I implemented the logic to use tags only when specified.

docs/release_guide.md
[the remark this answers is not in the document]
nojnhuh: > 1. Publish the images with the appropriate release tag

So we move image builds from the CI environment to local workstations? I don't see that being a major problem, but does open up opportunities to make mistakes.

> 2. Make the changes to retrieve the digest from the published images as a part of building the binaries for the release

Then where do you verify the retrieved digests are the same as the digests of the images you built?

> Is that something that is feasible?

Probably? I'm not totally sure. I'd like to see what it would take to get fully reproducible builds, since if we could leverage that then it wouldn't take as much tweaking to the overall process.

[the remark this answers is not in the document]
shashankram: > > 1. Publish the images with the appropriate release tag
> 
> So we move image builds from the CI environment to local workstations? I don't see that being a major problem, but does open up opportunities to make mistakes.
> 
No, we could make it a part of the CI but it could be a pre-release step, where we make a code change to `constants.go` to update the image references and publish the images with the appropriate tags. 

> > 1. Make the changes to retrieve the digest from the published images as a part of building the binaries for the release
> 
> Then where do you verify the retrieved digests are the same as the digests of the images you built?
> 
This can be retrieved from the image manifest. We wouldn't need to verify anything else other than the fact that the digest matches the image that was published. This is much like updating the tags currently, with correctness of the update verified through code reviews.

> > Is that something that is feasible?
> 
> Probably? I'm not totally sure. I'd like to see what it would take to get fully reproducible builds, since if we could leverage that then it wouldn't take as much tweaking to the overall process.

What's the use case for reproducible builds? I am not sure how using image digests could play a role here. Could you share an example?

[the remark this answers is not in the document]
nojnhuh: > No, we could make it a part of the CI but it could be a pre-release step, where we make a code change to `constants.go` to update the image references and publish the images with the appropriate tags.

There's still the nuance where the init image needs to be built before the injector image so the injector know the init image's digest. That's certainly possible to configure, just might be a bit convoluted. What would the trigger for this CI job be?

> This can be retrieved from the image manifest. We wouldn't need to verify anything else other than the fact that the digest matches the image that was published. This is much like updating the tags currently, with correctness of the update verified through code reviews.

I'm not exactly sure which digests you're suggesting to compare. I was trying to say we should make sure the digests we put in values.yaml and elsewhere come straight from the images built in CI and not after some push and pull if we're trying to protect against images that have been modified in Docker Hub.

> What's the use case for reproducible builds? I am not sure how using image digests could play a role here. Could you share an example?

If builds are reproducible, then we can build the images locally, see what the digests are, update image references accordingly like we do today with tags, then tag and push that commit. The CI would then build and publish images that result in the same digests since it's working off of the same tagged commit.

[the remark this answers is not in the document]
shashankram: @nojnhuh the digest is calculated on push to a registry. So we will not be able to calculate this beforehand. The goal is to ensure the digest is correct and bake that into the chart values **before** cutting a release.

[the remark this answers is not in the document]
nojnhuh: > @nojnhuh the digest is calculated on push to a registry. So we will not be able to calculate this beforehand.

The digest is a property of the image itself, so we only need to build it to know what the digest is. No push is necessary. `docker inspect <image> -f '{{ .ID }}'`

> The goal is to ensure the digest is correct and bake that into the chart values before cutting a release.

Yeah I think we're agreeing here. Let's make sure all the CI changes for release are captured in this PR. Commenting on code is much easier than English 😄

charts/osm/values.yaml
```diff
@@ -7,12 +7,25 @@ OpenServiceMesh:
...
-    # -- Container image registry
+    # -- Container image registry for control plane images
     registry: openservicemesh
-    # -- Container image pull policy
+    # -- Container image pull policy for control plane containers
     pullPolicy: IfNotPresent
-    # -- Container image tag
-    tag: v0.9.2
+    # -- Container image tag for control plane images
+    tag: "latest-main"
```
nojnhuh, line 15: Let's change the pull policy to `Always` by default if `latest-main` images are used by default.

docs/release_guide.md
```diff
@@ -46,6 +46,29 @@ Push the release branch to the upstream repo (NOT forked), identified here by th
 $ git push upstream release-<version> # ex: git push upstream release-v0.4
 ```
 
+## Create and push the pre-release Git tag
+
+The pre-release Git tag publishes the OSM control plane images to the Openservicemesh container registry in Dockerhub, and publishes the image digests as an artificat of the pre-release Github workflow. The image digests must be used in the next step to update the default control plane image referenced in the Helm charts.
```
nojnhuh, line 51:
```suggestion
The pre-release Git tag publishes the OSM control plane images to the Openservicemesh container registry in Dockerhub, and publishes the image digests as an artifact of the pre-release Github workflow. The image digests must be used in the next step to update the default control plane image referenced in the Helm charts.
```

docs/release_guide.md
```diff
@@ -46,6 +46,29 @@ Push the release branch to the upstream repo (NOT forked), identified here by th
...
+
+The pre-release Git tag is of the form `pre-rel-<release-version>`, e.g. `pre-rel-v0.4.0`.
+
+```console
+$ PRE_RELEASE_VERSION=<pre-release-version> # ex: PRE_RELEASE_VERSION=pre-rel-v0.4.0
+$ git tag "$PRE_RELEASE_VERSION"
+$ git push upstream "$PRE_RELEASE_VERSION"
+```
+
+Once the pre-release Git tag has been pushed, wait for the Pre-release Github workflow to complete. Upon workflow completion, retrieve the image digests for the given release. The image digests are uploaded as a part of the Pre-release workflow. Note that multiple image digest files will be present in the uploaded artificat, one per image tag, e.g. v0.4.0 and latest. Use the release version specific image digest file to extract the image digests to be used as the default for the control plane images, i.e. osm_image_digest_v0.4.0.txt instead of osm_image_digest_latest.txt if the release version if v0.4.0.
```
nojnhuh, line 61:
```suggestion
Once the pre-release Git tag has been pushed, wait for the Pre-release Github workflow to complete. Upon workflow completion, retrieve the image digests for the given release. The image digests are uploaded as a part of the Pre-release workflow. Note that multiple image digest files will be present in the uploaded artifact, one per image tag, e.g. v0.4.0 and latest. Use the release version specific image digest file to extract the image digests to be used as the default for the control plane images, i.e. osm_image_digest_v0.4.0.txt instead of osm_image_digest_latest.txt if the release version is v0.4.0.
```

docs/release_guide.md
```diff
@@ -54,28 +77,17 @@ If there are other commits on the `main` branch to be included in the release (s
 
 Create a new commit on the patch branch to update the hardcoded version information in the following locations:
 
-* The container image tag in [charts/osm/values.yaml](/charts/osm/values.yaml)
-* The chart and app version in [charts/osm/Chart.yaml](/charts/osm/Chart.yaml)
-* The default osm image tag in [osm cli mesh upgrade](/cmd/cli/mesh_upgrade.go)
+* The control plane image digests  defined by `OpenServiceMesh.image.digest` for images in [charts/osm/values.yaml](/charts/osm/values.yaml) from the image digests obtained from the Pre-release workflow. For example, if the osm-controller image digest is `sha256:eb194138abddbe271d42b290489917168a6a891a3dabb575de02c53f13879bee`, update the value of `OpenServiceMesh.image.digest.osmController` to `sha256:eb194138abddbe271d42b290489917168a6a891a3dabb575de02c53f13879bee`.
+* Replace the `latest-main` tag with the release version tag for all images throughout the repo, e.g. `v0.4.0`.
```
nojnhuh, line 81: Does this mean the places that were listed here before, like docs/example/manifests/apps/bookbuyer.yaml? I know it was really verbose and needed to be changed fairly often, but I appreciated it when it came time to make a release since it's not as easy as a global find/replace across the repo.

Even if we don't want a super verbose list, even mentioning the general locations where these tags exist would be helpful, like docs/example/manifests.

docs/release_guide.md
```diff
@@ -145,7 +157,7 @@ Make an announcement on the [OSM mailing list](https://groups.google.com/g/opens
 
 Skip this step if the release is a release candidate (RC).
 
-Open a pull request against the `main` branch making the same version updates as [above](#update-release-branch-with-patches-and-versioning-changes) so the latest release assets are referenced there.
+Open a pull request against the `main` branch to update the `version` field in `charts/osm/Chart.yaml` to the release version.
```
nojnhuh, line 160: I'm guessing we can't just keep the chart version as `latest-main` because `helm` needs it to be a valid semantic version?

docs/release_guide.md
nojnhuh: For context, Shashank and I do think this flow of build images, grab digests, update references to the digests , do the rest of the release could all be automated in a single release CI workflow in the future. Having some manual intervention is the easiest thing to do right now to unblock the rest of these changes and allows us the chance to work around any issues we may run into with this process.

nojnhuh (commented):
Left a couple nits and questions. Overall looks good.

charts/osm/values.yaml
shashankram: I didn't change this just because the demo scripts sets this to Always. We have to be careful in not setting this to Always for release versions, so it makes sense to default it to `IfNotPresent`.

docs/release_guide.md
shashankram: Added a note regarding the locations where this update is applicable.

docs/release_guide.md
shashankram: While I think it would be ideal to have a single workflow, I don't think it's possible because there are code changes necessary once the image digests are available. We might need to auto-commit a PR to achieve that or manipulate tags in a hacky way. I don't have any ideas to coalesce these 2 steps right now.

charts/osm/README.md
```diff
@@ -101,9 +101,15 @@ The following table lists the configurable parameters of the osm chart and their
 | OpenServiceMesh.fluentBit.workspaceId | string | `""` | WorkspaceId for Fluent Bit output plugin to Log Analytics |
 | OpenServiceMesh.grafana.enableRemoteRendering | bool | `false` | Enable Remote Rendering in Grafana |
 | OpenServiceMesh.grafana.port | int | `3000` | Grafana service's port |
-| OpenServiceMesh.image.pullPolicy | string | `"IfNotPresent"` | Container image pull policy |
-| OpenServiceMesh.image.registry | string | `"openservicemesh"` | Container image registry |
-| OpenServiceMesh.image.tag | string | `"v0.9.2"` | Container image tag |
+| OpenServiceMesh.image.digest | object | `{"osmBootstrap":"","osmCRDs":"","osmController":"","osmInjector":"","osmSidecarInit":""}` | Image digest (defaults to latest compatible tag) |
+| OpenServiceMesh.image.digest.osmBootstrap | string | `""` | osm-boostrap's image digest |
```
draychev, line 105: For readability - I wonder if instead of `OpenServiceMesh.image.digest.osmBootstrap` this should be `OpenServiceMesh.image.osmBootstrap.digest`  (i.e. swap the positions of `digest` and `osmBootstrap`)
This mainly because I see that the comment to this value is `osm-boostrap's image digest` (first name of image, then digest)

charts/osm/values.yaml
draychev: @nojnhuh, what's the reason behind your request to change to `Always`?  It seems to me that `IfNotPresent` is the safest bet for production environments (so Kubernetes does not do any extra unnecessary work) and `Always` perhaps for development.

charts/osm/README.md
shashankram: That works, I felt it's easier to keep `image.digest` as the top-level key and use the image names as child keys, so we have:
```
image
  digest:
     foo: ..
     bar: ...
     baz: ...
```

instead of:
```
image:
  foo:
    digest: ...
  bar:
    digest: ...
  baz:
    digest: ...
```

charts/osm/values.yaml
nojnhuh: > I didn't change this just because the demo scripts sets this to Always. We have to be careful in not setting this to Always for release versions, so it makes sense to default it to `IfNotPresent`.

I suppose we could effectively get this behavior "for free" if the `latest-main` images were instead named `latest` and take advantage of the [default behavior](https://kubernetes.io/docs/concepts/containers/images/#imagepullpolicy-defaulting) by blanking the pull policy. Not sure how many people are currently relying on `latest` images though.

> what's the reason behind your request to change to Always?

`latest-main` is a floating tag whose meaning changes frequently. If I were running a cluster and installed the chart using `latest-main` images pulled `IfNotPresent`, then installing a new version of the chart some time later without changing the tag or pull policy will still use the old images since images with the `latest-main` tag already exist on the node. The only way to get the new images to pull would be to manually delete them from the node. If the images were pulled `Always`, then each install of OSM would get the newest images automatically. I find those kinds of issues where I'm not running the versions of images that I think I am particularly frustrating.

> It seems to me that `IfNotPresent` is the safest bet for production environments (so Kubernetes does not do any extra unnecessary work) and `Always` perhaps for development.

Isn't that exactly the distinction we're making here with the tags, where the main branch is set up for development and the release branches are set up for production?

charts/osm/README.md
nojnhuh: I'm not sure it makes much of a difference now, but if we decide later to encode more things for each image, like if we make it possible to pull different images from different repos and with different tags, then it definitely makes sense to reorganize the values to `image.<name>.digest`. But reorganizing the values like that would be a breaking change to the Helm chart which we should do our best to avoid.

docs/release_guide.md
```diff
@@ -54,28 +77,17 @@ If there are other commits on the `main` branch to be included in the release (s
 
 Create a new commit on the patch branch to update the hardcoded version information in the following locations:
 
-* The container image tag in [charts/osm/values.yaml](/charts/osm/values.yaml)
-* The chart and app version in [charts/osm/Chart.yaml](/charts/osm/Chart.yaml)
-* The default osm image tag in [osm cli mesh upgrade](/cmd/cli/mesh_upgrade.go)
+* The control plane image digests  defined by `OpenServiceMesh.image.digest` for images in [charts/osm/values.yaml](/charts/osm/values.yaml) from the image digests obtained from the Pre-release workflow. For example, if the osm-controller image digest is `sha256:eb194138abddbe271d42b290489917168a6a891a3dabb575de02c53f13879bee`, update the value of `OpenServiceMesh.image.digest.osmController` to `sha256:eb194138abddbe271d42b290489917168a6a891a3dabb575de02c53f13879bee`.
+* Replace the `latest-main` tag with the release version tag for all images throughout the repo, e.g. `v0.4.0`. This includes updating the image image references in `charts/osm` and `docs/example/manifests` folders.
```
nojnhuh, line 81:
```suggestion
* Replace the `latest-main` tag with the release version tag for all images throughout the repo, e.g. `v0.4.0`. This includes updating the image references in `charts/osm` and `docs/example/manifests` folders.
```

charts/osm/values.yaml
shashankram: @nojnhuh I don't mind changing the PullPolicy to `Always` for the `main` branch, but we have to make it a part of the release checklist to ensure the PullPolicy is always set to `IfNotPresent` for released images.
I don't think this should be a major concern for this PR in any way, given that we can easily override it.

charts/osm/README.md
shashankram: For simplicity, I don't think we should be allowing mixing tags for different images, that's an unnecessary complexity I wouldn't consider at the moment.

nojnhuh (approved):

draychev (approved):
LGTM
Thank you!

PR #4117, opened 2021-09-15
feat(Windows): Add Windows nightly job

### LINKED ISSUE #3883: Testing cluster with Windows workers
In order to test OSM on Windows we need a cluster with Windows worker nodes.

Requirements:
1. Windows Server 2022
2. All the Windows containers should have the same version as the Windows worker nodes. 

Follow up on me to specify the exact Windows version needed

davinci26:
Fixes #3883 Adds a nightly job for Windows tests.

Cluster is installed via AKS-engine on the CNCF subscription.

**Description**:

**Testing done**:

**Affected area**:

2. Is this a breaking change?

### DIFF
--- a/.github/workflows/windows-nightly.yml
+++ b/.github/workflows/windows-nightly.yml
@@ -1,0 +1,36 @@
+name: Windows Nightly Job
+on: 
+  workflow_run:
+    workflows: ["Windows Images"]
+    types: [completed]
+
+env:
+  CTR_REGISTRY: openservicemesh
+  CTR_TAG: latest-main
+
+jobs:
+  test:
+    name: Windows Cluster Nightly Job
+    runs-on: ubuntu-latest
+    if: ${{ github.event.workflow_run.conclusion == 'success' }}
+    steps:
+      - name: Checkout v2
+        uses: actions/checkout@v2
+      - name: Authenticate and set context
+        uses: azure/k8s-set-context@v1
+        with:
+          method: kubeconfig
+          kubeconfig: ${{ secrets.CI_WINDOWS_KUBECONFIG }}
+        id: setcontext
+      - name: Setup Go 1.16
+        uses: actions/setup-go@v1
+        with:
+          go-version: 1.16
+      - name: Run e2es
+        run: |
+          make build-osm
+          go test ./tests/e2e -test.v -ginkgo.v -ginkgo.progress -ginkgo.focus='\[Cross-platform\]|\[windows\]' -test.timeout 180m -deployOnWindowsWorkers=true
+      - name: Cleanup resources
+        if: ${{ always() }}
+        run: |
+          kubectl delete service vault -n osm-system --ignore-not-found --wait
--- a/tests/e2e/e2e_egress_test.go
+++ b/tests/e2e/e2e_egress_test.go
@@ -15,6 +15,7 @@ var _ = OSMDescribe("HTTP and HTTPS Egress",
 	OSMDescribeInfo{
 		Tier:   1,
 		Bucket: 2,
+		OS:     OSCrossPlatform,
 	},
 	func() {
 		Context("Egress", func() {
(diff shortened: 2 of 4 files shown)


davinci26:
@shashankram let's talk offline to add the config of the cluster to the repository secrets

.github/workflows/windows-nightly.yml
```diff
@@ -0,0 +1,38 @@
...
+        with:
+          go-version: 1.16
+      - name: Run e2es
+        run: |
+          make build-osm
+          go test ./tests/e2e -test.v -ginkgo.v -ginkgo.progress -ginkgo.focus="\bCrossPlat\b" -test.timeout 180m -deployOnWindowsWorkers=true
+      - name: Cleanup resources
+        if: ${{ always() }}
+        run: |
+          kubectl delete service vault -n osm-system --ignore-not-found --wait
```
ksubrmnn, line 38: nit: newline

.github/workflows/windows-nightly.yml
```diff
@@ -0,0 +1,38 @@
...
+    steps:
+      - name: Checkout v2
+        uses: actions/checkout@v2
+      - name: Authenticate and set context
+        uses: azure/k8s-set-context@v1
+        with:
+          method: kubeconfig
+          kubeconfig: ${{ secrets.CI_WINDOWS_KUBECONFIG }}
+        id: setcontext
+      - name: Test oc
```
ksubrmnn, line 25: looks like this got left in :)

tests/e2e/e2e_egress_test.go
```diff
@@ -17,7 +17,7 @@ var _ = OSMDescribe("HTTP and HTTPS Egress",
 		Bucket: 2,
 	},
 	func() {
-		Context("Egress", func() {
+		Context("CrossPlat Egress", func() {
```
ksubrmnn, line 20: Is this renaming just temporary? Or are we going to use this naming convention going forward for all tests that should run on both Linux and Windows?

We can use this for now but we may want to reconsider when we start onboarding more tests and some are cross-plat, some just for linux, and some just for windows.

ksubrmnn:
Can you write some documentation for this pipeline in a follow up PR? 

How to invoke it, when it runs, etc. Example for openshift pipeline: https://github.com/openservicemesh/osm/blob/main/docs/development_guide/openshift/openshift_nightly.md

tests/e2e/e2e_egress_test.go
shashankram: Instead of naming it this way, would you mind extending the `OSMDescribeInfo struct to include the OS that be optionally set by tests, and use `-ginkgo.focus` to filter tests you want to run for Windows.

tests/framework/types.go
```diff
@@ -19,6 +24,9 @@ type OSMDescribeInfo struct {
 	// Bucket indicates in which test Bucket the test will run in for CI. Each
 	// Bucket is run in parallel while tests in the same Bucket run sequentially.
 	Bucket int
+
+	// OSBitMask indicates which OS the test can run on.
+	OSBitMask uint8
```
shashankram, line 29: `OS string` doesn't work?

davinci26: string is a bit weird because it can have three values, Windows linux and cross-platform, so I went with a representation that could capture all three in one variable

davinci26: string can also work but it seems that it might be more error prone. if you think its better as a string, I dont have a strong opinion

shashankram: If I were writing C code I would have done the same, but given this is in the test code and will be integrated into the CI, I think it would be better to make this human-readable and reference it in the workflows file as `linux, windows, crossPlatform` when invoking tests based on the OS.

shashankram (approved):
Thanks @davinci26

tests/framework/types.go
ksubrmnn: If you were interested, you can also create an enum in Go and give it a String() method for readability, but this is fine for now.

ksubrmnn (approved):

PR #4132, opened 2021-09-17
fix(lint): increase golangci-lint timeout

jaellio:
**Description**:

This change addresses the periodic lint failures due to timeout.
CI on main is currently failing because `golangci-lint` is timing out.

**Testing done**:

**Affected area**:
| CI System                  | [ x ] |

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -51,7 +51,7 @@ jobs:
         with:
           version: v1.41.1
           skip-pkg-cache: true
-          args: --timeout 5m --skip-dirs pkg/gen --skip-files "zz_generated.deepcopy.go$"
+          args: --timeout 10m --skip-dirs pkg/gen --skip-files "zz_generated.deepcopy.go$"
 
   codegen:
     name: Codegen
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,6 +1,6 @@
 run:
   tests: true
-  timeout: 5m
+  timeout: 10m
   skip-dirs:
     - pkg/gen
   skip-files:


.github/workflows/main.yml
```diff
@@ -51,7 +51,7 @@ jobs:
         with:
           version: v1.41.1
           skip-pkg-cache: true
-          args: --timeout 5m --skip-dirs pkg/gen --skip-files "zz_generated.deepcopy.go$"
+          args: --timeout 15m --skip-dirs pkg/gen --skip-files "zz_generated.deepcopy.go$"
```
shashankram, line 54: Does it take this long? Because it shouldn't. Let's reduce it to 10m at worst, because something else could be going wrong if it's taking that long.

snehachhabria (approved):

shalier (approved):

.github/workflows/main.yml
nojnhuh: Something is definitely fishy here. Poking around a bit it looks like it has something to do with the dot import of the `tests/framework` package that all the files in `tests/e2e` do.

After cleaning the golangci-lint cache (`golangci-lint cache clean`), I was seeing the goimports linter alone take over 15 minutes on my machine (`golangci-lint run --timeout 1h --disable-all --enable goimports --no-config -v`). After cleaning the cache again and running `rm -rf tests/e2e`, the same command took 1 minute.

I messed with one of the e2e test files and ran goimports on it directly to figure out what's making it take so long. Here's what I came up with:
```go
package e2e

import (
	"github.com/openservicemesh/osm/tests/framework"
	. "github.com/openservicemesh/osm/tests/framework"
)

var _ = OSMDescribe                // fast
var _ = OSMDescribeInfo{}.String() // fast
var _ = framework.Td.T             // fast
var _ = Td                         // fast
var _ = Td.T                       // slow
```

When the `// slow` line isn't commented out I can see goimports doing a lot more work than when it is commented out (`goimports -l -v <this file>`).

Removing all the dot imports and using `framework.` everywhere would speed up golangci-lint significantly I think, but not sure if that's worth it given the Lint CI job isn't a bottleneck.

Here's a good blog post I found that explains how goimports works that might help to debug this further: https://nikodoko.com/posts/goimports_explained/

shashankram: > Removing all the dot imports and using framework.

@nojnhuh It seems to me it would be worth making that change. I don't know why we use dot imports here because it's also something we avoid throughout the repo.

If the golangci-lint experience can be improved for local development and provide faster feedback, it seems worth it to make the change in e2e tests, given that it's one of the only places in the repo where the existing import style is not being followed.

jaellio: @nojnhuh I followed the steps you outlined (cleaning the golangci-lint cache and running `golangci-lint run --timeout 1h --disable-all --enable goimports --no-config -v` at the root of the osm repo) and I got this output:
```bash
INFO [lintersdb] Active 1 linters: [goimports]
INFO [loader] Go packages loading at mode 7 (files|name|compiled_files) took 22.3129967s
INFO [runner/filename_unadjuster] Pre-built 0 adjustments in 52.5081ms
INFO [linters context/goanalysis] analyzers took 56.6957286s with top 10 stages: goimports: 56.6957286s
INFO [runner] Issues before processing: 2, after processing: 0
INFO [runner] Processors filtering stat (out/in): skip_files: 2/2, autogenerated_exclude: 0/2, cgo: 2/2, path_prettifier: 2/2, skip_dirs: 2/2, filename_unadjuster: 2/2
INFO [runner] processing took 211.3µs with stages: autogenerated_exclude: 96.7µs, skip_dirs: 60.6µs, path_prettifier: 43.4µs, cgo: 3.1µs, filename_unadjuster: 1.1µs, nolint: 1.1µs, max_same_issues: 1µs, uniq_by_line: 700ns, identifier_marker: 600ns, max_from_linter: 400ns, source_code: 400ns, skip_files: 400ns, sort_results: 300ns, max_per_file_from_linter: 300ns, exclude-rules: 300ns, diff: 200ns, path_prefixer: 200ns, severity-rules: 200ns, path_shortener: 200ns, exclude: 100ns
INFO [runner] linters took 32.7037289s with stages: goimports: 32.7033154s
INFO File cache stats: 0 entries of total size 0B
INFO Memory: 552 samples, avg is 64.4MB, max is 96.7MB
INFO Execution took 55.0917089s
```
I have repeated this several times and the goimports linter takes around 1 minute. I am unable to reproduce the 15 minutes execution time you noted above.

Issue #4065, opened 2021-08-31
Reconciler for OSM CRDs


snehachhabria:
The goal of this GitHub issue is to create a feature flagged feature where Open Service Mesh Control plane will deploy, manage, and maintain CRDs. This feature will help customers who want OSM to maintain the consistency of the system. In the case of an accidental deletion of the CRDs or mutations, which could adversely affect the service mesh, OSM Control Plane would restore the artifacts and keep the mesh operating.

Tasks : 

- [X] Move CRD deployment from helm to osm-bootstrap
- [X] Build a reconciler client in OSM that watches for CRD updates/deleted
- [X] Add reconciliation logic for OSM to manage CRDs
- [X] Migrate existing mutating webhook reconciler to this framework
- [X] Create a reconciler for validating webhook in osm
- [X] Remove mutating and validating webhook deployment via Helm
- [X] Add documentation

snehachhabria:
Tasks : 

- [ ] Move CRD deployment from helm to osm-bootstrap
- [ ] Build a reconciler client in OSM that watches for CRD updates/deleted
- [ ] Add reconciliation logic for OSM to manage CRDs
- [ ] Migrate existing mutating webhook reconciler to this framework

michelleN:
@snehachhabria could you confirm this behavior is opt in and not the default experience?

snehachhabria:
> @snehachhabria could you confirm this behavior is opt in and not the default experience?

Yes the reconciliation process will be configurable and will be turned off by default

PR #4154, opened 2021-09-21
feat(reconciler): enable mutatingwebhook reconciler in osm

snehachhabria:
**Description**:

Enable a reconciler in osm-injector for the mutating webhook and
deprecate the previous controller runtime reconciler

Deployment of the mutating webhook is via helm or osm-injector depending
on if the reconilcer is enabled. The clean up script has also been
updated to ensure that the mutating webhook gets deleted on osm
uninstall. The clean up  hook has been updated from a pre-delete to post-delete to
ensure that the resources to be removed are deleted after the control plane is uninstalled

e2e tests for the reconciler has been extended to include testing the
mutating webhook

Part of #4065

**Testing done**: e2e tests have been updated

**Affected area**:
| Tests                      | x] |


cmd/osm-injector/osm-injector.go
```diff
@@ -177,9 +189,12 @@ func main() {
 		log.Fatal().Err(err).Msgf("Failed to start OSM metrics/probes HTTP server")
 	}
 
-	// Initialize the reconciler for the injector's MutatingWebhookConfiguration
-	if err := createReconciler(kubeClient); err != nil {
-		events.GenericEventRecorder().FatalEvent(err, events.InitializationError, "Error creating controller manager to reconcile sidecar injector webhook config")
+	if enableReconciler {
+		log.Info().Msgf("OSM reconciler enabled for mutating webhook")
```
shashankram, line 193:
```suggestion
		log.Info().Msgf("OSM reconciler enabled for sidecar injector webhook")
```

pkg/injector/webhook.go
```diff
@@ -437,7 +443,79 @@ func updateMutatingWebhookCABundle(cert certificate.Certificater, webhookName st
...
+							Key:      "control-plane",
+							Operator: metav1.LabelSelectorOpDoesNotExist,
+						},
+					},
+				},
+				Rules: []admissionregv1.RuleWithOperations{
+					{
+						Operations: []admissionregv1.OperationType{admissionregv1.Create},
+						Rule: admissionregv1.Rule{
+							APIGroups:   []string{"*"},
```
shashankram, line 498: Should this be `""` or `"*"`? If `""` works from the Go code, let's use that for consistency with the chart value.

shashankram (approved):

pkg/injector/webhook.go
snehachhabria: `""` doesn't work in go code, it will consider it as empty and omit it. which causes the creation of the webhook to fail since its a mandatory field. Hence it has to be `"*"`. Do we want to change the chart to `"*"` ?

shashankram: I think this is fine. Thanks for clarifying!

PR #4158, opened 2021-09-22
ref(pkg/k8s): add more logging

nojnhuh:
**Description**:
This change adds additional logging in the Kubernetes event handlers
passed to informers and to all pubsub operations.

**Testing done**: Manual testing with the automated demo.

**Affected area**:

### DIFF
--- a/pkg/k8s/event_handlers.go
+++ b/pkg/k8s/event_handlers.go
@@ -3,6 +3,8 @@ package k8s
 import (
 	"reflect"
 
+	"github.com/rs/zerolog"
+	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/client-go/tools/cache"
 
 	"github.com/openservicemesh/osm/pkg/announcements"
@@ -32,6 +34,7 @@ func GetKubernetesEventHandlers(shouldObserve observeFilter, eventTypes EventTyp
 			if !shouldObserve(obj) {
 				return
 			}
+			logResourceEvent(log, eventTypes.Add, obj)
 			ns := getNamespace(obj)
 			metricsstore.DefaultMetricsStore.K8sAPIEventCounter.WithLabelValues(eventTypes.Add.String(), ns).Inc()
 			events.Publish(events.PubSubMessage{
@@ -45,6 +48,7 @@ func GetKubernetesEventHandlers(shouldObserve observeFilter, eventTypes EventTyp
 			if !shouldObserve(newObj) {
 				return
 			}
+			logResourceEvent(log, eventTypes.Update, newObj)
 			ns := getNamespace(newObj)
 			metricsstore.DefaultMetricsStore.K8sAPIEventCounter.WithLabelValues(eventTypes.Update.String(), ns).Inc()
 			events.Publish(events.PubSubMessage{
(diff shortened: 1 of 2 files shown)


pkg/k8s/event_handlers.go
```diff
@@ -69,6 +100,10 @@ func GetKubernetesEventHandlers(informerName, providerName string, shouldObserve
 	}
 }
 
-func getNamespace(obj interface{}) string {
-	return reflect.ValueOf(obj).Elem().FieldByName("ObjectMeta").FieldByName("Namespace").String()
+func logWithResourceName(log zerolog.Logger, obj metav1.Object) zerolog.Logger {
```
shashankram, line 103: I'd instead create  `func getNamespacedName(obj metav1.Object) string` and let the caller log something like `Observed k8s ADD event for resource 'foo/bar'`

shashankram (changes requested):

pkg/k8s/event_handlers.go
```diff
@@ -69,6 +100,10 @@ func GetKubernetesEventHandlers(informerName, providerName string, shouldObserve
...
+	name := obj.GetName()
+	if len(obj.GetNamespace()) > 0 {
```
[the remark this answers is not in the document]
nojnhuh: Done

pkg/k8s/event_handlers.go
```diff
@@ -29,11 +36,19 @@ func GetKubernetesEventHandlers(informerName, providerName string, shouldObserve
 
 	return cache.ResourceEventHandlerFuncs{
 		AddFunc: func(obj interface{}) {
+			log := log.With().Str("event", eventTypes.Add.String()).Logger()
+			o, err := meta.Accessor(obj)
+			if err != nil {
+				log.Error().Err(err).Msg(msgErrParsingObject)
+				return
+			}
+			log = logWithResourceName(log, o)
```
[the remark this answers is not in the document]
nojnhuh: > I don't think we should log resources that are not being observed, it will create a lot of noise in larger environments with a lot of resources.

Done

> Can we move this snippet that logs the resource name observed and the event type and merge it with the `msgObservedEvent` log?

This is doing what I think you want it to. `log := log.With().Str("event", eventTypes.Add.String()).Logger()` isn't writing any logs itself, but effectively saying "create a new logger that includes an `event` field with this value" so then future `log.Msg()` calls all have that field.The new logger is local to the function so it's not modifying the package-scoped one. The resource observed message will include the event kind and name of the resource in the same line.

pkg/k8s/event_handlers.go
nojnhuh: If we're not logging anything for ignored resources, then I can fit all the new logging stuff in a single function and make one call to that in each handler. I don't see a need to create another new function just to construct this name if it's only ever called in one place.

shashankram: Right, I didn't think we'd need a function to log, just so the log itself is inlined:
`log.Trace().Str(...,...).Msgf("... %s..", getNamespacedName(obj))`

shashankram (approved):

jaellio (approved):

PR #4177, opened 2021-09-27
feat(*): ignore invalid traffic targets

michelleN:
+ ignores invalid traffic targets when building envoy config
+ provides extra guarantee that even if invalid traffic targets
exist in cluster before OSM is installed, OSM will ignore them

resovles #4116

**Description**:

**Testing done**:

**Affected area**:

2. Is this a breaking change?


pkg/catalog/traffictarget_test.go
```diff
@@ -870,156 +870,3 @@ func TestListInboundTrafficTargetsWithRoutes(t *testing.T) {
 		})
 	}
 }
-
```
draychev, line 873: Is it true that this test was defacto moved from `pkg/catalog/traffictarget_test.go` to `pkg/smi/client_test.go` ?

pkg/smi/client.go
```diff
@@ -280,6 +291,10 @@ func (c *client) ListTrafficTargets(options ...TrafficTargetListOption) []*smiAc
 			continue
 		}
 
+		if !isValidTrafficTarget(trafficTarget) {
+			continue
```
draychev, line 295: Would it make sense to `log.Trace().Msgf()` the fact that we found an invalid traffic target?

draychev: Should we have a prometheus counter for invalid TrafficTargets? That might be a really good counter to add.
I took the liberty to create the following issue for this #4187

pkg/smi/client.go
```diff
@@ -288,6 +303,40 @@ func (c *client) ListTrafficTargets(options ...TrafficTargetListOption) []*smiAc
...
+}
+
+// hasValidRules checks if the given SMI TrafficTarget object has valid rules
+func hasValidRules(rules []smiAccess.TrafficTargetRule) bool {
+	if rules == nil {
+		return false
+	}
+	if len(rules) == 0 {
+		return false
+	}
```
draychev, line 326: I was curious if this could be collapsed. (Not proposing we do - just curious.)
It looks like `len(rules)` could cover both cases: https://play.golang.org/p/LzR_HKKywFj
image

pkg/smi/client_test.go
```diff
@@ -261,6 +346,15 @@ var _ = Describe("When listing TrafficTargets", func() {
 
 		_, err := fakeClientSet.smiTrafficTargetClientSet.AccessV1alpha3().TrafficTargets(testNamespaceName).Create(context.TODO(), trafficTarget, metav1.CreateOptions{})
 		Expect(err).ToNot(HaveOccurred())
+
+		_, err = fakeClientSet.smiTrafficTargetClientSet.AccessV1alpha3().TrafficTargets("random-namespace").Create(context.TODO(), trafficTargetWithNamespaceMismatch, metav1.CreateOptions{})
+		Expect(err).ToNot(HaveOccurred())
+
+		_, err = fakeClientSet.smiTrafficTargetClientSet.AccessV1alpha3().TrafficTargets(testNamespaceName).Create(context.TODO(), trafficTargetWithInvalidRuleKind, metav1.CreateOptions{})
```
draychev, line 353: Am I correct that we expect this line to **not** create a new TrafficTarget because it is invalid. So then a few lines below when we list the traffic targets - the count will still be 1?
I wonder if a comment here would help disambiguate this.

draychev (approved):
lgtm

pkg/smi/client_test.go
```diff
@@ -261,6 +346,15 @@ var _ = Describe("When listing TrafficTargets", func() {
 
 		_, err := fakeClientSet.smiTrafficTargetClientSet.AccessV1alpha3().TrafficTargets(testNamespaceName).Create(context.TODO(), trafficTarget, metav1.CreateOptions{})
 		Expect(err).ToNot(HaveOccurred())
+
```
shashankram, line 349: @michelleN would it be possible to add new tests for these new changes instead of updating the existing ones. I plan on refactoring the existing tests to address issues in the pub-sub usage (already have locally). If it's a separate function, would be much easier for me to incorporate those changes without accidentally leaving them out as a part of the cleanup.

pkg/smi/client_test.go
michelleN: I updated these tests to make things clearer

pkg/smi/client_test.go
michelleN: sure thing. done.

pkg/catalog/egress.go
```diff
@@ -135,7 +135,7 @@ func (mc *MeshCatalog) buildHTTPRouteConfigs(egressPolicy *policyV1alpha1.Egress
 	var httpRouteMatches []trafficpolicy.HTTPRouteMatch
 	httpMatchSpecified := false
 	for _, match := range egressPolicy.Spec.Matches {
-		if match.APIGroup != nil && *match.APIGroup == smiSpecs.SchemeGroupVersion.String() && match.Kind == httpRouteGroupKind {
+		if match.APIGroup != nil && *match.APIGroup == smiSpecs.SchemeGroupVersion.String() && match.Kind == smi.HTTPRouteGroupKind {
```
snehachhabria, line 138: :thumbsup:

pkg/catalog/traffictarget.go
```diff
@@ -91,7 +77,7 @@ func (mc *MeshCatalog) getAllowedDirectionalServiceAccounts(svcIdentity identity
 	for _, trafficTarget := range allTrafficTargets {
 		spec := trafficTarget.Spec
 
-		if spec.Destination.Kind != serviceAccountKind {
+		if spec.Destination.Kind != smi.ServiceAccountKind {
```
snehachhabria, line 80: :thumbsup:

snehachhabria (approved):

eduser25 (approved):

PR #4197, opened 2021-09-30
fix(build): check if windows tag exists correctly

nojnhuh:
**Description**:
Currently, when checking if a tag on a windows image exists, the linux
image is checked. This change fixes it so the windows image is checked
instead.

**Affected area**:

### DIFF
--- a/scripts/publish-image.sh
+++ b/scripts/publish-image.sh
@@ -27,14 +27,18 @@ if [ -z "${CTR_TAG}" ]; then
 fi
 
 if [[ "$VERIFY_TAGS" == "true" ]]; then
-    tokenUri="https://auth.docker.io/token?service=registry.docker.io&scope=repository:$IMAGE_REPO/$IMAGE_NAME:pull"
+    image="$IMAGE_NAME"
+    if [[ $OS == "windows" ]]; then
+      image="$image-windows"
+    fi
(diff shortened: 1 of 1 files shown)


snehachhabria (approved):

scripts/publish-image.sh
```diff
@@ -27,14 +27,18 @@ if [ -z "${CTR_TAG}" ]; then
 fi
 
 if [[ "$VERIFY_TAGS" == "true" ]]; then
-    tokenUri="https://auth.docker.io/token?service=registry.docker.io&scope=repository:$IMAGE_REPO/$IMAGE_NAME:pull"
+    image="$IMAGE_NAME"
+    if [[ $OS == "windows" ]]; then
+      image="$image-windows"
+    fi
```
shashankram, line 33: Why can't the caller (Makefile) pass the correct IMAGE_NAME?

nojnhuh: We still use the image name when we do `make ARGS=--push "docker-build-windows-$IMAGE_NAME"`.

Basically, this is the smallest change I could think of to get this to work correctly. There are plenty of things we can do to clean all of the image building later, but I just wanted to unblock the release for now.

shashankram (approved):

shashankram:
Thanks @nojnhuh for the quick fix!

snehachhabria:
@nojnhuh could we create a PR to fix this on main too ?

nojnhuh:
Yep, #4198

PR #4200, opened 2021-09-30
chore(release): update version to v0.11.0-rc.1

jaellio:

**Changes to note:**

The logic in `_helpers.tpl` to determine the control plane and demo app image names was modified. If the `OpenServiceMesh.image.tag` is set in `charts/osm/values.yaml` it will be used by default to configure the image names. If the image tag is not set, the image digests defined by `OpenServiceMesh.image.digest` will be used. 

Prior to this change, setting the image digests in `charts/osm/values.yaml` was causing the demo to fail. By default, the image digests were used to configure the image names.  In the demo, the new images that were pushed to the local container registry did not have the same digests as the digests generated from the pre-release workflow. When attempting to pull the images from the local registry with the digests from the pre-release workflow the images didn't exist.

- [x] I have cherry-picked any changes [labelled](https://github.com/openservicemesh/osm/labels) `needs-cherry-pick-vX.Y.Z`
- [x] I have made all of the following version and patch updates:
  1. Updated the container image tag in charts/osm/values.yaml
  2. Updated the chart **and** app version in charts/osm/Chart.yaml
  3. Updated the image tags used in the demo manifests
  4. Regenerated the Helm chart README.md

- [x] I have checked that the base branch for this PR is correct as defined by the release guide

Is this a release candidate? Yes

**Description**: Update OSM version to v0.11.0-rc.1

### DIFF
--- a/charts/osm/values.yaml
+++ b/charts/osm/values.yaml
@@ -11,20 +11,20 @@ OpenServiceMesh:
     registry: openservicemesh
     # -- Container image pull policy for control plane containers
     pullPolicy: IfNotPresent
-    # -- Container image tag for control plane images
-    tag: "latest-main"
-    # -- Image digest (defaults to latest compatible tag)
+    # -- Container image tag for control plane images (must not be set on release branches)
+    tag: ""
+    # -- Image digest (ignored when image.tag is set)
     digest:
       # -- osm-controller's image digest
-      osmController: ""
+      osmController: "sha256:147e6155b6bbb69c97715e552cf2effbb26ede3380d379bb7e0de03493c85045"
       # -- osm-injector's image digest
-      osmInjector: ""
+      osmInjector: "sha256:cf301763aa21b7973b5fc963399e553dac88517d590cfb72eceb2f985fb63a86"
       # -- Sidecar init container's image digest
-      osmSidecarInit: ""
+      osmSidecarInit: "sha256:0da5863eb304717a73079054fc8b1b5c18cede6903224701a011166551e63990"
       # -- osm-crds' image digest
-      osmCRDs: ""
+      osmCRDs: "sha256:4d85f94733f2fd009c7b956276aca9ac86b1be6ecca91243486254307a5beaae"
       # -- osm-boostrap's image digest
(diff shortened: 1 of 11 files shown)


snehachhabria (approved):

nojnhuh (approved):

shashankram:
@jaellio could you rebase this PR to the HEAD of release-v0.11.

charts/osm/values.yaml
```diff
@@ -12,19 +12,19 @@ OpenServiceMesh:
     # -- Container image pull policy for control plane containers
     pullPolicy: IfNotPresent
     # -- Container image tag for control plane images
-    tag: "latest-main"
+    tag: "v0.11.0-rc.1"
```
shashankram, line 15: We want to be setting this tag only from the CI job for the PR workflow, otherwise, the default images will also use tags instead of digests.

shashankram (changes requested):
@jaellio I forgot when I suggested altering the logic in the Helm template that we only want to be setting the image tags in `values.yaml` for the PR workflow. The behavior should be that image digests get used by default for which image tags must not be encoded in `values.yaml`. So we can pass in the image tag from the CI instead of setting the default.

charts/osm/values.yaml
```diff
@@ -12,19 +12,19 @@ OpenServiceMesh:
     # -- Container image pull policy for control plane containers
     pullPolicy: IfNotPresent
     # -- Container image tag for control plane images
-    tag: "latest-main"
+    tag: ""
```
shashankram, line 15: Would be helpful to add a comment here that the tag must not be set on release branches.

.github/workflows/openshift-nightly.yml
```diff
@@ -1,5 +1,5 @@
 name: OpenShift Nightly Job
-on: 
+on:
```
shashankram, line 2: nit: unrelated to this PR.

shashankram (approved):

shashankram (approved):

snehachhabria (approved):

eduser25 (approved):

PR #4207, opened 2021-10-01
chore(hpa): removes max limit to maxReplicas for HPA

shalier:
**Description**:
Removes max limit on maxReplicas for HPA
See also PR #4201 for more details

**Testing done**:

**Affected area**:

### DIFF
--- a/charts/osm/templates/osm-injector-hpa.yaml
+++ b/charts/osm/templates/osm-injector-hpa.yaml
@@ -10,6 +10,7 @@ spec:
     kind: Deployment
     name: osm-injector
   minReplicas: {{.Values.OpenServiceMesh.injector.autoScale.minReplicas}}
+  # no max limit set for maxReplicas
   maxReplicas: {{.Values.OpenServiceMesh.injector.autoScale.maxReplicas}}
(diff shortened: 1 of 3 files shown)


charts/osm/templates/osm-injector-hpa.yaml
```diff
@@ -10,6 +10,7 @@ spec:
     kind: Deployment
     name: osm-injector
   minReplicas: {{.Values.OpenServiceMesh.injector.autoScale.minReplicas}}
+  # no max limit set for maxReplicas
```
ksubrmnn, line 13: I think this comment and the other one like this are a bit confusing. I don't think we need them, so I'd leave them out.

shashankram (changes requested):
I suggest setting an upper limit. It's not good practice to keep the replica count unbounded, as it can overwhelm the cluster when there are persistent utilization issues on the control plane. Marking the control plane nodes as critical would make this problem worse, where control plane components would multiply at an uncontrollable rate without an upper limit. 

Please profile the code and gauge what a reasonable upper limit is instead of keeping the max unbounded.

Issue #4167, opened 2021-09-24
osm-controller deadlocks, resulting in no events being processed


shashankram:
**Bug description**:
It has been observed that osm-controller hangs (most likely due to one or more deadlocks) which results in no events being processed.

I suspect this to be due to how the global pub-sub framework is leveraged in OSM, where the pub-sub could deadlock if the subscription channel is at capacity and 2 or more events are published and waiting to be read in the order they were published. This can happen because the global pub-sub instance runs a single goroutine to handle publishes for all topics, where if a message is published for a topic ,it needs to be consumed by a subscriber for that topic before other published messages can be processed by the pub-sub send routine. The problem starts to manifest when the subscriber channels are at capacity.

An example of this issue can be seen in https://github.com/openservicemesh/osm/blob/main/pkg/envoy/registry/services.go#L109-L133, where a Publish() depends on a Subscribe() to be invoked, which deadlocks when the channel reaches its capacity.
Similar issues can occur in other parts of the code.

Another problem with the existing implementation is that blocking channels result in k8s event handler callbacks to block. This should be avoided using a [workqueue](https://pkg.go.dev/k8s.io/client-go/util/workqueue). Further, when an event that results in a configuration update fails, it must be retried a few times so the system recovers in case of transient failures.

**Affected area (please mark with X where applicable)**:
- Control Plane        [X]

**Expected behavior**:
osm-controller should not hang.

eduser25:
@shashankram I would assume the following can also be closed with the new msg broker, unless msg broker were to have a similar weakness.
https://github.com/openservicemesh/osm/issues/3144

PR #4210, opened 2021-10-04
messaging: introduce message broker

shashankram:
**Description**:
Introduces message broker functionality to
handle messaging in OSM control plane. The broker
will handle proxy update events and k8s events
required by other components within the control plane.
The broker comprises of a workqueue to process events
and pub-sub instances to implement the messaging
channels. The message broker will replace the global
pub-sub functionality.

Testing done:
- 100% unit test coverage
- Verified fixes for #4167 with the use of
  message broker

Required by #4167

**Affected area**:

### DIFF
--- a/pkg/announcements/types.go
+++ b/pkg/announcements/types.go
@@ -10,6 +10,9 @@ func (at Kind) String() string {
 }
 
 const (
+	// ProxyUpdate is the event kind used to trigger an update to subscribed proxies
+	ProxyUpdate Kind = "proxy-update"
+
 	// ScheduleProxyBroadcast is used by other modules to request the dispatcher to schedule a global proxy broadcast
 	ScheduleProxyBroadcast Kind = "schedule-proxy-broadcast"
 
--- a/pkg/messaging/broker.go
+++ b/pkg/messaging/broker.go
@@ -1,0 +1,168 @@
+package messaging
+
+import (
+	"time"
+
+	"github.com/cskr/pubsub"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/util/workqueue"
+
+	"github.com/openservicemesh/osm/pkg/announcements"
+	"github.com/openservicemesh/osm/pkg/apis/config/v1alpha1"
+	"github.com/openservicemesh/osm/pkg/k8s/events"
+	"github.com/openservicemesh/osm/pkg/metricsstore"
+)
+
+// NewBroker returns a new message broker instance and starts the internal goroutine
+// to process events added to the workqueue.
+func NewBroker(stopCh <-chan struct{}) *Broker {
+	b := &Broker{
+		queue:             workqueue.NewRateLimitingQueue(workqueue.DefaultControllerRateLimiter()),
+		proxyUpdatePubSub: pubsub.New(0),
+		kubeEventPubSub:   pubsub.New(0),
+		certPubSub:        pubsub.New(0),
+	}
+
+	go b.run(stopCh)
+
+	return b
+}
+
+// GetProxyUpdatePubSub returns the PubSub instance corresponding to proxy update events
+func (b *Broker) GetProxyUpdatePubSub() *pubsub.PubSub {
+	return b.proxyUpdatePubSub
+}
+
+// GetKubeEventPubSub returns the PubSub instance corresponding to k8s events
+func (b *Broker) GetKubeEventPubSub() *pubsub.PubSub {
+	return b.kubeEventPubSub
+}
+
+// GetCertPubSub returns the PubSub instance corresponding to certificate events
+func (b *Broker) GetCertPubSub() *pubsub.PubSub {
+	return b.certPubSub
+}
+
+// run starts a goroutine to process events from the workqueue until
+// signalled to stop on the given channel.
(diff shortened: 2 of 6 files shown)


snehachhabria (approved):
looks good!

pkg/messaging/broker_test.go
```diff
@@ -0,0 +1,250 @@
...
+	)
+	defer c.Unsub(c.kubeEventPubSub, serviceChan)
+
+	meshCfgChan := c.GetKubeEventPubSub().Sub(announcements.MeshConfigUpdated.String())
+	defer c.Unsub(c.kubeEventPubSub, meshCfgChan)
+
+	certRotateChan := c.GetCertPubSub().Sub(announcements.CertificateRotated.String())
+	defer c.Unsub(c.certPubSub, certRotateChan)
+
+	numEventTriggers := 50
```
allenlsy, line 44: Are we using 50 concurrent events per type to check if there is still deadlock?

shashankram: numEventTriggers corresponds to the number of times the test generates the batch of events (pod add|update|delete, service add|update|delete), and cert updates. It's just a test variable to trigger a reasonably large number of events.

pkg/messaging/broker.go
```diff
@@ -0,0 +1,168 @@
...
+// 3. Updates metrics associated with the event
+func (b *Broker) processEvent(msg events.PubSubMessage) {
+	// Update proxies if applicable
+	if shouldUpdateProxy(msg) {
+		b.proxyUpdatePubSub.Pub(msg, announcements.ProxyUpdate.String())
+		metricsstore.DefaultMetricsStore.ProxyBroadcastEventCount.Inc()
+	}
+
+	// Publish event to other interested clients, e.g. log level changes, debug server on/off etc.
+	b.kubeEventPubSub.Pub(msg, msg.Kind.String())
```
nojnhuh, line 78: Based on the second point [here](https://pkg.go.dev/k8s.io/client-go/util/workqueue#pkg-overview):

> if an item is added multiple times before it can be processed, it will only be processed once.

Is it possible that some kubernetes events will not be propagated? I can see where that wouldn't matter in most cases, but could that cause issues in certain scenarios, like if we're counting events?

pkg/announcements/types.go
```diff
@@ -10,6 +10,9 @@ func (at Kind) String() string {
 }
 
 const (
+	// ProxyUpdate is the event kind used to trigger an update to subscribed proxies
+	ProxyUpdate Kind = "proxy-update"
```
nojnhuh, line 14: Is this functionally different from `ScheduleProxyBroadcast`?

pkg/messaging/broker.go
```diff
@@ -0,0 +1,168 @@
...
+}
+
+// Unsub unsubscribes the given channel from the PubSub instance
+func (b *Broker) Unsub(pubSub *pubsub.PubSub, ch chan interface{}) {
+	// Unsubscription should be performed from a different goroutine and
+	// existing messages on the subscribed channel must be drained as noted
+	// in https://github.com/cskr/pubsub/blob/v1.0.2/pubsub.go#L95.
+	go pubSub.Unsub(ch)
+	for range ch {
+		// Drain channel till 'Unsub' results in a close on the subscribed channel
```
nojnhuh, line 97:
```suggestion
		// Drain channel until 'Unsub' results in a close on the subscribed channel
```

pkg/messaging/workqueue.go
```diff
@@ -0,0 +1,48 @@
...
+}
+
+// GetTotalQEventCount returns the total number of events queued throughout
+// the lifetime of the workqueue.
+func (b *Broker) GetTotalQEventCount() uint64 {
+	return atomic.LoadUint64(&b.totalQEventCount)
+}
+
+// processNextItem processes the next item in the workqueue
+func (b *Broker) processNextItem() bool {
```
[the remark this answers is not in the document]
shashankram: Updated the comment.

pkg/messaging/workqueue_test.go
```diff
@@ -0,0 +1,25 @@
...
+	stop := make(chan struct{})
+	defer close(stop)
+
+	b := NewBroker(stop)
+
+	// Verify that a non PubSubMessage does not panic
+	b.queue.AddRateLimited("string")
+	a.Eventually(func() bool {
+		return b.GetTotalQEventCount() == 1
+	}, 100*time.Millisecond, 10*time.Millisecond)
```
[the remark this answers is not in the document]
shashankram: `processNexItem()` is invoked internally within the broker, and required the queue to be initialized prior to adding items. It is simplest to keep it as is.

pkg/messaging/broker.go
shashankram: I haven't seen an issue with this, and I am unsure if the link you shared applies here. We do not drop any events as such, the rate limiter could if the same object is repeatedly queued, which does not happen right now. I don't see an issue with the event counters either.

pkg/announcements/types.go
shashankram: ScheduleProxyBroadcast will be deprecated.

PR #4217, opened 2021-10-06
feat(hpa): Adds memory targetAverageUtilization

shalier:
**Description**:
Allows HPA to react to memory utilization(not only CPU)

**Testing done**:

**Affected area**:
| New Functionality          | [x ] |

1. Does this change contain code from or inspired by another project?no

2. Is this a breaking change?no

### DIFF
--- a/charts/osm/values.yaml
+++ b/charts/osm/values.yaml
@@ -61,8 +61,12 @@ osm:
       minReplicas: 1
       # -- Maximum replicas for autoscale
       maxReplicas: 5
-      # -- Average target CPU utilization (%)
-      targetAverageUtilization: 80
+      cpu:
+        # -- Average target CPU utilization (%)
+        targetAverageUtilization: 80
+      memory:
+        # -- Average target memory utilization (%)
+        targetAverageUtilization: 80
 
   #
   # -- Prometheus parameters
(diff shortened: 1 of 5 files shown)


charts/osm/values.yaml
```diff
@@ -249,8 +253,12 @@ OpenServiceMesh:
...
       # -- Maximum replicas for autoscale
       maxReplicas: 5
-      # -- Average target CPU utilization (%)
-      targetAverageUtilization: 80
+      cpu:
+        # -- Average target CPU utilization (%)
+        targetAverageUtilization: 80
+      memory:
+      # -- Average target value of memory
+        targetAverageValue: "10Mi"
```
shashankram, line 261: The memory demands for injector and controller will not be the same, so we should not be using the same value for both. It would make sense to create a unique autoScale property for each of them, nested under the `osmController` and `injector` properties for this reason.

Moreover, I would like suggest using the same units for the memory as specified in the `resource` property for the respective apps. In this case, I suggest defining the memory value as an integer that represents the bytes instead of the power-of-two notation. 
```
    resource:
      limits:
        cpu: "1.5"
        memory: "512M"
      requests:
        cpu: "0.5"
        memory: "128M"
```

Also could you document how the value `10Mi` was chosen?

shashankram (approved):

charts/osm/values.yaml
shalier: Changed it to be similar to cpu in that if the memory targetAverageUtilization is greater than 80% of the requested memory then it'll scale up

draychev (approved):
This looks great!
Thank you @shalier

PR #4285, opened 2021-10-19
ingress: deprecate k8s ingress API usage

shashankram:
**Description**:
Starting v0.10, OSM uses the IngressBackend API for
ingress configuration. This change deprecates the
usage of the k8s Ingress API in code.

Also fixes a bug where the gateway secret is not
updated when the certificate validity duration is
changed.

**Testing done**:

**Affected area**:


shashankram:
> @shashankram would you want to back port this to v0.11.1 ? since we are deprecating some fields in the meshConfig CRD it might be good to

@snehachhabria Deprecating code in a patch release does not seem right, so let's handle this in the next release being cut.

PR #4325, opened 2021-10-29
ref(tcp): support tcp-server-first inside the mesh

### LINKED ISSUE #4313: Handle tcp-server-first ports within the mesh
**Bug description**:

#3329 adds handling for ports with an `appProtocol` of `tcp-server-first` only when such services are referred to in an Egress policy. When a server-first port exists in the mesh, the TLS and HTTP inspector filters are still configured on the outbound-listener of a client to a server-first service, and Envoy's default 15s `listener_filters_timeout` is hit.

**Expected behavior**: Clients involved in server-first protocols immediately receive data from a service in the mesh when the service's port is `tcp-server-first`.

- Install OSM with permissive mode enabled
- Install a simple client and server:
```yaml
apiVersion: v1
kind: ServiceAccount
metadata:
  name: server
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: server
spec:
  replicas: 1
  selector:
    matchLabels:
      app: server
  template:
    metadata:
      labels:
        app: server
    spec:
      serviceAccountName: server
      containers:
        - name: server
          image: busybox
          command:
            - /bin/sh
            - -c
            - |
              nc -lkp 4242 -e sh -c "while yes; do :; done"
---
apiVersion: v1
kind: Service
metadata:
  name: server
spec:
  ports:
    - name: server
      port: 4242
      appProtocol: tcp-server-first
  selector:
    app: server
---
apiVersion: v1
kind: ServiceAccount
metadata:
  name: client
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: client
spec:
  replicas: 1
  selector:
    matchLabels:
      app: client
  template:
    metadata:
      labels:
        app: client
    spec:
      serviceAccountName: client
      containers:
        - name: client
          image: busybox
          command:
            - /bin/sh
            - -c
            - |
              nc server 4242
```
When client and server are outside the mesh, the client receives an endless stream of `y` printed in its logs. When both are inside the mesh, the client gets stuck in `CrashLoopBackOff` and no logs are shown.

**How was OSM installed?**:

**Bug report archive**:

**Environment**:
- Others:

nojnhuh:
**Description**:
This change adds the capability for OSM to handle ports with
`appProtocol: tcp-server-first` on services within the mesh.
Specifically, the main change is to add handling for `tcp-server-first`
ports in the mesh when we calculate the predicate function used to
disable the TLS and HTTP inspector listener filters. Previously, this
was only done for `tcp-server-first` ports specified in an Egress
policy resource.

Fixes #4313

**Testing done**: Tested manually with the example in #4313 and with the automated demo to ensure that still works as expected.

**Affected area**:

### DIFF
--- a/pkg/envoy/lds/listener.go
+++ b/pkg/envoy/lds/listener.go
@@ -58,6 +58,7 @@ func (lb *listenerBuilder) newOutboundListener() (*xds_listener.Listener, error)
 	}
 
 	if featureflags := lb.cfg.GetFeatureFlags(); featureflags.EnableEgressPolicy {
+		var trafficMatches []*trafficpolicy.TrafficMatch
 		var filterDisableMatchPredicate *xds_listener.ListenerFilterChainMatchPredicate
 		// Create filter chains for egress based on policies
 		if egressTrafficPolicy, err := lb.meshCatalog.GetEgressTrafficPolicy(lb.serviceIdentity); err != nil {
@@ -65,8 +66,10 @@ func (lb *listenerBuilder) newOutboundListener() (*xds_listener.Listener, error)
 		} else if egressTrafficPolicy != nil {
 			egressFilterChains := lb.getEgressFilterChainsForMatches(egressTrafficPolicy.TrafficMatches)
 			listener.FilterChains = append(listener.FilterChains, egressFilterChains...)
-			filterDisableMatchPredicate = getFilterMatchPredicateForTrafficMatches(egressTrafficPolicy.TrafficMatches)
+			trafficMatches = append(trafficMatches, egressTrafficPolicy.TrafficMatches...)
 		}
+		trafficMatches = append(trafficMatches, lb.meshCatalog.GetOutboundMeshTrafficPolicy(lb.serviceIdentity).TrafficMatches...)
+		filterDisableMatchPredicate = getFilterMatchPredicateForTrafficMatches(trafficMatches)
 		additionalListenerFilters := []*xds_listener.ListenerFilter{
 			// Configure match predicate for ports serving server-first protocols (ex. mySQL, postgreSQL etc.).
 			// Ports corresponding to server-first protocols, where the server initiates the first byte of a connection, will
--- a/pkg/envoy/lds/listener_test.go
+++ b/pkg/envoy/lds/listener_test.go
@@ -12,9 +12,13 @@ import (
 	. "github.com/onsi/gomega"
 	tassert "github.com/stretchr/testify/assert"
 
+	configv1alpha1 "github.com/openservicemesh/osm/pkg/apis/config/v1alpha1"
+	"github.com/openservicemesh/osm/pkg/catalog"
 	"github.com/openservicemesh/osm/pkg/configurator"
 	"github.com/openservicemesh/osm/pkg/constants"
 	"github.com/openservicemesh/osm/pkg/envoy"
+	"github.com/openservicemesh/osm/pkg/identity"
+	"github.com/openservicemesh/osm/pkg/service"
 	"github.com/openservicemesh/osm/pkg/trafficpolicy"
 )
 
@@ -165,4 +169,47 @@ func TestGetFilterMatchPredicateForTrafficMatches(t *testing.T) {
 			assert.Equal(tc.expectedMatch, actual)
 		})
 	}
+}
+
+func TestNewOutboundListener(t *testing.T) {
+	mockCtrl := gomock.NewController(t)
+	identity := identity.K8sServiceAccount{}.ToServiceIdentity()
+	meshCatalog := catalog.NewMockMeshCataloger(mockCtrl)
+	meshCatalog.EXPECT().GetEgressTrafficPolicy(gomock.Any()).Return(nil, nil).Times(1)
+	meshCatalog.EXPECT().GetOutboundMeshTrafficPolicy(identity).Return(&trafficpolicy.OutboundMeshTrafficPolicy{
+		TrafficMatches: []*trafficpolicy.TrafficMatch{
+			{
+				WeightedClusters: []service.WeightedCluster{{}},
+				DestinationIPRanges: []string{
+					"0.0.0.0/0",
+				},
+				DestinationPort:     1,
+				DestinationProtocol: constants.ProtocolTCPServerFirst,
+			},
+		},
+	}).Times(2)
+	cfg := configurator.NewMockConfigurator(mockCtrl)
+	cfg.EXPECT().IsEgressEnabled().Return(false).Times(1)
+	cfg.EXPECT().GetFeatureFlags().Return(configv1alpha1.FeatureFlags{
+		EnableEgressPolicy: true,
+	}).Times(1)
+
+	lb := newListenerBuilder(meshCatalog, identity, cfg, nil)
+
+	assert := tassert.New(t)
+	listener, err := lb.newOutboundListener()
+	assert.NoError(err)
+
+	assert.Len(listener.ListenerFilters, 3) // OriginalDst, TlsInspector, HttpInspector
+	assert.Equal(wellknown.TlsInspector, listener.ListenerFilters[1].Name)
+	assert.Equal(&xds_listener.ListenerFilterChainMatchPredicate{
+		Rule: &xds_listener.ListenerFilterChainMatchPredicate_DestinationPortRange{
+			DestinationPortRange: &xds_type.Int32Range{
+				Start: 1,
+				End:   2,
+			},
+		},
+	}, listener.ListenerFilters[1].FilterDisabled)
+	assert.Equal(wellknown.HttpInspector, listener.ListenerFilters[2].Name)
+	assert.Equal(listener.ListenerFilters[1].FilterDisabled, listener.ListenerFilters[2].FilterDisabled)
 }
--- a/tests/e2e/e2e_tcp_server_first_test.go
+++ b/tests/e2e/e2e_tcp_server_first_test.go
@@ -1,0 +1,85 @@
+package e2e
+
+import (
+	"strconv"
+	"time"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	"github.com/openservicemesh/osm/pkg/constants"
+	"github.com/openservicemesh/osm/tests/framework"
+	. "github.com/openservicemesh/osm/tests/framework"
+)
+
+var _ = OSMDescribe("TCP server-first traffic",
+	OSMDescribeInfo{
+		Tier:   1,
+		Bucket: 1,
+	},
+	func() {
+		var (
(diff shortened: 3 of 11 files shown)


nojnhuh:
Changes to docs being worked on here: https://github.com/openservicemesh/osm-docs/pull/267

pkg/envoy/lds/listener.go
```diff
@@ -35,13 +35,9 @@ func (lb *listenerBuilder) newOutboundListener() (*xds_listener.Listener, error)
 		Address:          envoy.GetAddress(constants.WildcardIPAddr, constants.EnvoyOutboundListenerPort),
 		TrafficDirection: xds_core.TrafficDirection_OUTBOUND,
 		FilterChains:     serviceFilterChains,
-		ListenerFilters: []*xds_listener.ListenerFilter{
-			{
-				// The OriginalDestination ListenerFilter is used to redirect traffic
-				// to its original destination.
-				Name: wellknown.OriginalDestination,
```
shashankram, line 42: Let's confirm whether the OriginalDestination filter is required for in-mesh connectivity on Windows. +@davinci26

shashankram (commented):
Haven't reviewed this change entirely yet, but I'd like to see an e2e test to confirm this works as expected. Follow-up is fine if you have verified this locally.

pkg/envoy/lds/listener_test.go
```diff
@@ -166,3 +170,43 @@ func TestGetFilterMatchPredicateForTrafficMatches(t *testing.T) {
 		})
 	}
 }
+
+func TestNewOutboundListener(t *testing.T) {
+	mockCtrl := gomock.NewController(t)
+	id := identity.K8sServiceAccount{}.ToServiceIdentity()
```
shashankram, line 176:
```suggestion
	identity := identity.K8sServiceAccount{}.ToServiceIdentity()
```

pkg/envoy/lds/listener_test.go
```diff
@@ -166,3 +170,43 @@ func TestGetFilterMatchPredicateForTrafficMatches(t *testing.T) {
...
+	meshCatalog.EXPECT().GetOutboundMeshTrafficPolicy(id).Return(&trafficpolicy.OutboundMeshTrafficPolicy{
+		TrafficMatches: []*trafficpolicy.TrafficMatch{
+			{
+				WeightedClusters: []service.WeightedCluster{{}},
+				DestinationIPRanges: []string{
+					"0.0.0.0/0",
+				},
+				DestinationPort:     1,
+				DestinationProtocol: constants.ProtocolTCPServerFirst,
+			},
```
shashankram, line 187: Could you also include a scenario in this test case where the protocol is not server first.

tests/e2e/e2e_tcp_server_first_test.go
```diff
@@ -0,0 +1,83 @@
...
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	"github.com/openservicemesh/osm/pkg/constants"
+	"github.com/openservicemesh/osm/tests/framework"
+	. "github.com/openservicemesh/osm/tests/framework"
+)
+
+var _ = OSMDescribe("Test TCP server-first traffic from 1 pod sever -> 1 pod client",
```
shashankram, line 15:
```suggestion
var _ = OSMDescribe("TCP server-first traffic",
```

tests/e2e/e2e_tcp_server_first_test.go
```diff
@@ -0,0 +1,83 @@
...
+		Bucket: 1,
+	},
+	func() {
+		var (
+			sourceNs = framework.RandomNameWithPrefix("client")
+			destNs   = framework.RandomNameWithPrefix("server")
+			ns       = []string{sourceNs, destNs}
+		)
+
+		It("Tests TCP traffic for client pod -> server pod", func() {
```
shashankram, line 27:
```suggestion
		It("Tests TCP server-first traffic", func() {
```

tests/e2e/e2e_tcp_server_first_test.go
```diff
@@ -0,0 +1,83 @@
...
+			})
+			Expect(err).NotTo(HaveOccurred())
+			_, err = Td.CreateServiceAccount(sourceNs, &svcAccDef)
+			Expect(err).NotTo(HaveOccurred())
+			_, err = Td.CreatePod(sourceNs, podDef)
+			Expect(err).NotTo(HaveOccurred())
+
+			Expect(Td.WaitForPodsRunningReady(sourceNs, 120*time.Second, 1, nil)).To(Succeed())
+
+			Expect(waitForLogEmission(sourceNs, podDef.Name, podDef.Name, "y")).NotTo(HaveOccurred(), "Didn't get expected response from server")
```
shashankram, line 81: Why is waiting on a log? That seems fragile to me. Instead, we should verify that netcat returns the output ASAP.

shashankram (changes requested):

tests/e2e/e2e_tcp_server_first_test.go
nojnhuh: I can ensure the timeout here is something small, like 5 seconds. I'm not sure how reliable this test would be if we didn't wait at all.

pkg/envoy/lds/listener_test.go
nojnhuh: I didn't test that here because this function calls `getFilterMatchPredicateForTrafficMatches()`, and the tests there cover that: https://github.com/openservicemesh/osm/blob/30fdefdf18b59b60084b9a485b5dc5696f47fc85/pkg/envoy/lds/listener_test.go#L131-L140

tests/e2e/e2e_tcp_server_first_test.go
```diff
@@ -0,0 +1,83 @@
...
+
+			destinationPort := 80
+
+			// Get simple pod definitions for the TCP server
+			svcAccDef, podDef, svcDef, err := Td.SimplePodApp(
+				SimplePodAppDef{
+					PodName:     framework.RandomNameWithPrefix("server"),
+					Namespace:   destNs,
+					Image:       "busybox",
+					Command:     []string{"nc", "-lkp", strconv.Itoa(destinationPort), "-e", "sh", "-c", "while yes; do :; done"},
```
davinci26, line 47: what are the flags `"-e", "sh", "-c"` for? I can't find them on https://linux.die.net/man/1/nc

A bit orthogonal but what is the motivation behind using nc as the server instead of envoy?

tests/e2e/e2e_tcp_server_first_test.go
```diff
@@ -0,0 +1,83 @@
...
+			dstSvc, err := Td.CreateService(destNs, svcDef)
+			Expect(err).NotTo(HaveOccurred())
+
+			// Expect it to be up and running in it's receiver namespace
+			Expect(Td.WaitForPodsRunningReady(destNs, 120*time.Second, 1, nil)).To(Succeed())
+
+			svcAccDef, podDef, _, err = Td.SimplePodApp(SimplePodAppDef{
+				PodName:   framework.RandomNameWithPrefix("client"),
+				Namespace: sourceNs,
+				Command:   []string{"nc", dstSvc.Name + "." + dstSvc.Namespace, strconv.Itoa(destinationPort)},
```
davinci26, line 69: Can you leverage the `TCPRequest` function here which is already cross plat?

davinci26 (commented):
i have executed the following tests on windows and they all pass:

```
davinci@MININT-9CEC7NF:~/osm$ grep -r OSCrossPlatform ./tests/e2e/
./tests/e2e/e2e_egress_test.go:         OS:     OSCrossPlatform,
./tests/e2e/e2e_controller_restart_test.go:             OS:     OSCrossPlatform,
./tests/e2e/e2e_tcp_client_server_test.go:              OS:     OSCrossPlatform,
./tests/e2e/e2e_pod_client_server_test.go:              OS:     OSCrossPlatform,
```
afaik they should cover the relevant code paths

tests/e2e/e2e_tcp_server_first_test.go
nojnhuh: Yeah, that works here. To make this test fully cross platform though I would need the windows equivalent for the `nc` command on the server side. All it should do is accept a connection and send back some response without reading any of the request. Do you know what that might look like on WIndows?

tests/e2e/e2e_tcp_server_first_test.go
shashankram: 5s is good. We just need to ensure we receive the response before the default 15 listener filter timeout.

tests/e2e/e2e_tcp_server_first_test.go
davinci26: I think swapping it out to `TCPRequest` is better regardless of the cross-platform effort as it increases the consistency among the codebase.

I kind of eluded to one approach with the above comment. If we only want a listener on a specific port Envoy can do this work easily. Where envoy is going to be used as the application container and is going to be sidecared by another envoy.

We know that Envoy is cross-platform and we already have a dependency on it. Also, the project has expertise in envoy so it can be easy to debug.

> Do you know what that might look like on WIndows?


Regarding the equivalent of `nc` on windows, the container that we use won't have that so you will need to do that with PowerShell core. 

If you want to be a good citizen, I would extract the functionality of the server into a function similar to `GetOSSpecificHTTPBinPod` and only implement this for linux and throw a not implemented exception on Windows. That way we have a building block `GetOSSpecificTCPListener` (name up to you) that we can use across all tests

pkg/envoy/lds/listener.go
```diff
@@ -57,38 +53,41 @@ func (lb *listenerBuilder) newOutboundListener() (*xds_listener.Listener, error)
...
+		{
+			// To inspect TLS metadata, such as the transport protocol and SNI
+			Name:           wellknown.TlsInspector,
+			FilterDisabled: filterDisableMatchPredicate,
+		},
+		{
+			// To inspect if the application protocol is HTTP based
+			Name:           wellknown.HttpInspector,
+			FilterDisabled: filterDisableMatchPredicate,
+		},
```
shashankram, line 90: These 2 filters are only required when EgressPolicy is enabled. Why is that behavior being changed?

nojnhuh: Didn't try to change anything extra like that. I'll change it back.

tests/e2e/e2e_tcp_server_first_test.go
nojnhuh: Talked with @shashankram offline, I'll work on making this test fully cross-platform later.

pkg/envoy/lds/listener.go
shashankram: It's fine to have them but it would result in a behavior change, where these 2 filters could time out for non-egress traffic, which was previously not the case.

shashankram (approved):

snehachhabria (approved):

PR #4326, opened 2021-11-01
update curl command for init containers

snehachhabria:
**Description**:

The args for the culr job on the init containers in osm-controller and
osm-injector has beeen updated to use curl params rather than bash

**Testing done**: Manually verified the demo and everything works as expected 

**Affected area**:

### DIFF
--- a/charts/osm/templates/osm-deployment.yaml
+++ b/charts/osm/templates/osm-deployment.yaml
@@ -38,14 +38,7 @@ spec:
       initContainers:
         - name: init-osm-controller
           image: {{ .Values.osm.curlImage }}
-          args:
-          - /bin/sh
-          - -c
-          - >
-            set -x;
-            while [ $(curl --connect-timeout 2 -sw '%{http_code}' "http://osm-bootstrap.{{ include "osm.namespace" . }}.svc.cluster.local:9095/healthz" -o /dev/null) -ne 200 ]; do
-              sleep 10;
-            done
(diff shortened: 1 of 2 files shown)


charts/osm/templates/osm-deployment.yaml
```diff
@@ -38,14 +38,9 @@ spec:
...
+          command: ["sh", "-c"]
           args:
-          - /bin/sh
-          - -c
-          - >
-            set -x;
-            while [ $(curl --connect-timeout 2 -sw '%{http_code}' "http://osm-bootstrap.{{ include "osm.namespace" . }}.svc.cluster.local:9095/healthz" -o /dev/null) -ne 200 ]; do
-              sleep 10;
-            done
+            - curl --connect-timeout 2 "http://osm-bootstrap.{{ include "osm.namespace" . }}.svc.cluster.local:9095/healthz" --retry 50 --retry-connrefused --retry-delay 5
```
nojnhuh, line 43: If the idea here is to remove the dependency on `sh` in the image, we could call `curl` directly in the `command`.

shashankram: +1

Let's avoid using `sh` as the entrypoint for the command.

snehachhabria: thanks for the suggestion @nojnhuh and @shashankram, I have updated the code. PTAL

nojnhuh (approved):

shashankram (approved):

PR #4332, opened 2021-11-02
fix(chart): set additionalProperties on root in schema

nojnhuh:
**Description**:
This change sets `"additionalProperties": false` in values.schema.json
on the root object to catch errors like `osm install --set doesntExist=true`,
which previously were allowed.

**Affected area**:

### DIFF
--- a/charts/osm/values.schema.json
+++ b/charts/osm/values.schema.json
@@ -1266,5 +1266,6 @@
                 }
             }
         }
-    }
+    },
+    "additionalProperties": false
 }


shashankram (changes requested):
This will break other projects that import our chart as a subchart:
```
commit ea70ba53ca5a35866b59a82a5ff909d04b904b0c
Date:   Tue Jun 1 09:28:21 2021 -0700

    charts/osm: allow additional root properties
    
    Currently, the schema prevents additional
    Helm chart properties other than OpenServiceMesh
    which is the root property for all values.
    
    This change enforces strict validation only
    for the OpenServiceMesh root property. Other
    root properties that might be leveraged by
    Helm charts using the OSM chart as a dependency
    will not be subjected to schema validations.
```

PR #4341, opened 2021-11-04
fix(upgrade): handle removed values

### LINKED ISSUE #4330: Upgrade should correctly handle fields removed from charts in the new version
**Bug description**:
Currently, an upgrade from version X -> Y breaks if the Helm chart in version Y removed a field that was previously referenced in X. A field used X may be removed after being marked as deprecated, and when the field is no longer required in version Y. Due to how we perform upgrades, this breaks because the removed field is no longer supported in Y, and the schema validation fails in Y.

- Upgrade [ ]

**Expected behavior**:
Upgrade should work when a field is removed from the new version's Helm chart.

nojnhuh:
**Description**:
Currently, upgrading OSM with `osm mesh upgrade` fails when the newer
chart version removes or renames values from the installed chart because
values.schema.json defines `"additionalProperties": false` on each
object and `osm mesh upgrade` uses all the values from the installed
chart.

Now, the `osm mesh upgrade` command will not use any values from the old
release. Users can specify them with the new `--set` flag.

Fixes #4330

**Testing done**: Manually ran affected tests and other upgrade scenarios based on the existing e2e test.

**Affected area**:

2. Is this a breaking change? Yes, release notes have been updated.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -248,7 +248,7 @@ jobs:
         if: ${{ github.event_name == 'pull_request' }}
         env:
           K8S_NAMESPACE: "osm-system"
-        run: go test ./tests/e2e -test.v -ginkgo.v -ginkgo.progress -installType=KindCluster -test.timeout 0 -test.failfast -ginkgo.failFast -ginkgo.focus='\[Bucket ${{ matrix.bucket }}\]' -ginkgo.skip='Upgrade'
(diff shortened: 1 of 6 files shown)


shashankram:
Looks like the e2e_upgrade_test passes with this change. Should we enable that test in the CI?
```
• [SLOW TEST:256.758 seconds]
[Tier 2][Bucket 10][] Upgrade from latest
/home/runner/work/osm/osm/tests/framework/common.go:91
  Tests upgrading the control plane
  /home/runner/work/osm/osm/tests/e2e/e2e_upgrade_test.go:27
```

nojnhuh:
Yes, those changes are included here.

shashankram (approved):

snehachhabria (approved):

PR #4342, opened 2021-11-05
feat(injector): add iptables inbound rule for inbound localhost

johnsonshi:
Adds a rule so that incoming traffic where the destination is
localhost does not go through the Envoy sidecar.

**Description**:

**Testing done**:

**Affected area**:
| Networking                 | [x] |

2. Is this a breaking change? NA

### DIFF
--- a/pkg/injector/iptables.go
+++ b/pkg/injector/iptables.go
@@ -62,6 +62,9 @@ var iptablesInboundStaticRules = []string{
 	fmt.Sprintf("iptables -t nat -A PROXY_INBOUND -p tcp --dport %d -j RETURN", readinessProbePort),
 	fmt.Sprintf("iptables -t nat -A PROXY_INBOUND -p tcp --dport %d -j RETURN", startupProbePort),
 
+	// Skip localhost traffic, doesn't need to be routed via the proxy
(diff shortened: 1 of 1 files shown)


pkg/injector/iptables.go
```diff
@@ -62,6 +62,9 @@ var iptablesInboundStaticRules = []string{
 	fmt.Sprintf("iptables -t nat -A PROXY_INBOUND -p tcp --dport %d -j RETURN", readinessProbePort),
 	fmt.Sprintf("iptables -t nat -A PROXY_INBOUND -p tcp --dport %d -j RETURN", startupProbePort),
 
+	// Skip localhost traffic, doesn't need to be routed via the proxy
+	"iptables -t nat -A PROXY_INBOUND -d 127.0.0.1/32 -j RETURN",
```
johnsonshi, line 66: @shashankram // @draychev For context, I'm trying to add an iptable rule so that a pod can communicate to itself through its own kubernetes service. Would this be the correct fix? I'm not sure if the traffic packets here would have destination `localhost` (in which case I think this rule would work) or destination `svc ip address (in which case I don't think this rule works).

johnsonshi:
Closing as @shashankram will take on this fix.

PR #4346, opened 2021-11-08
fix(init): set init container level security context

SanyaKochhar:
**Description**: Currently, pods having pod-level security contexts are causing the init container to not be able to program iptables. Setting `runAsNonRoot` and `runAsUser: 0` security context for the init container fixes this.

For example, the following security context in a bookstore demo pod causes an xtables permission denied error where the pod never finishes initializing/goes into crashLoopBackoff
```
      securityContext:
        fsGroup: 1000
        runAsGroup: 1000
        runAsUser: 1000
```

**Testing done**:

**Affected area**:

2. Is this a breaking change?

### DIFF
--- a/pkg/injector/init_container.go
+++ b/pkg/injector/init_container.go
@@ -4,6 +4,7 @@ import (
 	"strings"
 
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/utils/pointer"
 
 	"github.com/openservicemesh/osm/pkg/configurator"
 )
@@ -23,6 +24,9 @@ func getInitContainerSpec(containerName string, cfg configurator.Configurator, o
 					"NET_ADMIN",
 				},
 			},
+			RunAsNonRoot: pointer.BoolPtr(false),
+			// User ID 0 corresponds to root
+			RunAsUser: pointer.Int64Ptr(0),
 		},
 		Command: []string{"/bin/sh"},
 		Args: []string{
--- a/pkg/injector/init_container_test.go
+++ b/pkg/injector/init_container_test.go
@@ -18,6 +18,8 @@ var _ = Describe("Test functions creating Envoy bootstrap configuration", func()
 
 	privilegedFalse := false
 	privilegedTrue := true
+	runAsNonRootFalse := false
+	runAsUserID := int64(0)
 
 	mockCtrl := gomock.NewController(GinkgoT())
 	mockConfigurator := configurator.NewMockConfigurator(mockCtrl)
@@ -44,7 +46,9 @@ var _ = Describe("Test functions creating Envoy bootstrap configuration", func()
 							"NET_ADMIN",
 						},
 					},
-					Privileged: &privilegedFalse,
+					Privileged:   &privilegedFalse,
+					RunAsNonRoot: &runAsNonRootFalse,
+					RunAsUser:    &runAsUserID,
 				},
 				Stdin:     false,
 				StdinOnce: false,
@@ -76,7 +80,9 @@ var _ = Describe("Test functions creating Envoy bootstrap configuration", func()
 							"NET_ADMIN",
 						},
 					},
-					Privileged: &privilegedFalse,
+					Privileged:   &privilegedFalse,
+					RunAsNonRoot: &runAsNonRootFalse,
+					RunAsUser:    &runAsUserID,
 				},
 				Stdin:     false,
 				StdinOnce: false,
@@ -107,7 +113,9 @@ var _ = Describe("Test functions creating Envoy bootstrap configuration", func()
 							"NET_ADMIN",
 						},
 					},
-					Privileged: &privilegedTrue,
+					Privileged:   &privilegedTrue,
+					RunAsNonRoot: &runAsNonRootFalse,
+					RunAsUser:    &runAsUserID,
 				},
 				Stdin:     false,
 				StdinOnce: false,
@@ -138,7 +146,9 @@ var _ = Describe("Test functions creating Envoy bootstrap configuration", func()
 							"NET_ADMIN",
 						},
 					},
-					Privileged: &privilegedFalse,
+					Privileged:   &privilegedFalse,
+					RunAsNonRoot: &runAsNonRootFalse,
+					RunAsUser:    &runAsUserID,
 				},
 				Stdin:     false,
 				StdinOnce: false,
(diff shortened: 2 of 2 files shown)


pkg/injector/init_container.go
```diff
@@ -23,6 +25,8 @@ func getInitContainerSpec(containerName string, cfg configurator.Configurator, o
 					"NET_ADMIN",
 				},
 			},
+			RunAsNonRoot: &runInitContainerAsNonRoot,
+			RunAsUser:    &initContainerUserId,
```
shashankram, line 29: `import "k8s.io/utils/pointer"`
and
```suggestion
			RunAsNonRoot: pointer.BoolPtr(false),
			RunAsUser:    pointer.Int64Ptr(0),
```

pkg/injector/init_container.go
```diff
@@ -12,6 +12,8 @@ func getInitContainerSpec(containerName string, cfg configurator.Configurator, o
 	inboundPortExclusionList []int, enablePrivilegedInitContainer bool) corev1.Container {
 	iptablesInitCommandsList := generateIptablesCommands(outboundIPRangeExclusionList, outboundPortExclusionList, inboundPortExclusionList)
 	iptablesInitCommand := strings.Join(iptablesInitCommandsList, " && ")
+	runInitContainerAsNonRoot := false
+	initContainerUserId := int64(0)
```
shashankram, line 16: Why 0? Please add a comment.

pkg/injector/init_container.go
```diff
@@ -12,6 +12,8 @@ func getInitContainerSpec(containerName string, cfg configurator.Configurator, o
 	inboundPortExclusionList []int, enablePrivilegedInitContainer bool) corev1.Container {
 	iptablesInitCommandsList := generateIptablesCommands(outboundIPRangeExclusionList, outboundPortExclusionList, inboundPortExclusionList)
 	iptablesInitCommand := strings.Join(iptablesInitCommandsList, " && ")
+	runInitContainerAsNonRoot := false
```
shashankram, line 15: The init container should not be run as `root` by default. This should be made configurable based on whether the backing pod has a security context that prevents non-root containers from executing iptables.

I would expect the init container to not run as `root` by default, which means `SecurityContext.RunAsNonRoot = true`.

shashankram (dismissed):

pkg/injector/init_container.go
ksubrmnn: user id of 0 means root

pkg/injector/init_container.go
ksubrmnn: By default, with no security contexts, the init container runs as root. However, the pod level security context can override this. The init container must run as root to program iptables. This has been tested by setting the init container SecurityContext to runAsNonRoot and run with a non zero user id. This results in the init container failing.

The init container cannot program iptables if it is non-root. This is unaffected by pod-level security contexts.

SanyaKochhar: Have validated the above by setting `runAsNonRoot` to true and the user ID to 1000 with our default bookstore demo (no additional security contexts) - this causes the same xtables permission denied error we are trying to resolve. If we were to set it to true and keep the user ID as 0, this will be in opposition to what the [k8s API docs](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.22/#securitycontext-v1-core) say this value checks for:
> If true, the Kubelet will validate the image at runtime to ensure that it does not run as UID 0 (root) and fail to start the container if it does. If unset or false, no such validation will be performed.

pkg/injector/init_container.go
```diff
@@ -23,6 +24,9 @@ func getInitContainerSpec(containerName string, cfg configurator.Configurator, o
 					"NET_ADMIN",
 				},
 			},
+			RunAsNonRoot: pointer.BoolPtr(false),
```
allenlsy, line 27: I don't get it after reading the description. 

Isn't `RunAsNonRoot = false` and `RunAsUser = 0` the same meaning? So setting them both at the same time solves the problem?

nojnhuh (approved):

pkg/injector/init_container.go
johnsonshi: image


image

image

allenlsy: Oh man, feels hacky. But this answer solves my puzzle.

allenlsy (approved):

johnsonshi (approved):

pkg/injector/init_container.go
nojnhuh: > Setting `RunAsRoot` as `false` is redundant as it just means "do not perform such validation".

`RunAsRoot` can be set at either the pod level or the container level. Setting it here for the container is necessary in case the user specifies `runAsRoot: true` for the pod.

> Better solution would be to not touch `RunAsNonRoot` and `RunAsUser`, and instead set `Privileged` to `true`.

Privileged mode grants much higher privileges than simply running as root in an unprivileged container. Privileged doesn't seem necessary here.

This is at least my understanding. @ksubrmnn or @SanyaKochhar please correct me if I'm wrong.

SanyaKochhar: Actually just verified and you are right, using just `runAsUser = 0` is sufficient to fix the problem caused the security context originally in the description (updating now), but if that pod security context were to include `runAsNonRoot: true`, it causes the following error: 
> Error: container's runAsUser breaks non-root policy (pod: "bookstore-569cdc978b-wbd5n_bookstore(35e391ed-0d47-4066-9e5f-8b08ff974f44)", container: osm-init)

Here's why we need both to account for the different pod/container security permutations:

You're right that the nuance between the two is very slight and the naming makes it really confusing. According to k8s API docs, "If true, the Kubelet will validate the image at runtime to ensure that it does not run as UID 0 (root) and fail to start the container if it does. **If unset or false, no such validation will be performed.**"

Effectively, `runAsNonRoot: false` does not actually set the user to root, it just allows for it to be set to root. `runAsUser` additionally sets it to root user. So if the **init container** uses `runAsUser` without `runAsNonRoot: false` and the **pod** security context were to set `runAsNonRoot` to true, it would not allow `runAsUser: 0` at the **container** security context level.

SanyaKochhar:
@johnsonshi see https://github.com/openservicemesh/osm/pull/4346#discussion_r747043732 about why both values need to be set. As for Privileged - privileged is not actually the same as running as root. Not sure why but on testing with `privileged: true`, the xtables issue still arises. 

>Setting RunAsRoot as false is redundant as it just means "do not perform such validation".
>Better solution would be to not touch RunAsNonRoot and RunAsUser, and instead set Privileged to true.

snehachhabria (approved):

pkg/injector/init_container.go
SanyaKochhar: oops @nojnhuh missed your comment and didn't see that you had already addressed this, thanks!!

ksubrmnn: Exactly @nojnhuh

Issue #4371, opened 2021-11-18
Upgrading OSM does not upgrade default values in the MeshConfig


trstringer:
**Bug description**:

When you upgrade OSM, if there are changes that need to be in the MeshConfig (e.g. sidecar image) then these do not take effect after the upgrade.

**Affected area (please mark with X where applicable)**:
- Install                    [x]

**Expected behavior**:

Certain configurations in MeshConfig should be upgradeable as well, as long as they were not altered by the user.

**Steps to reproduce the bug (as precisely as possible)**:

Install OSM version v0.9.2:

```bash
$ osm version
Version: v0.9.2; Commit: 7df9badece2f251515a2bc5e5ac79438b0c58812; Date: 2021-08-19-21:02

$ osm install
OSM installed successfully in namespace [osm-system] with mesh name [osm]
```

Notice the settings for the sidecar image (this is expected for v0.9.2):

```bash
$ kubectl get meshconfig -n osm-system osm-mesh-config -o jsonpath='{.spec.sidecar.envoyImage}'
envoyproxy/envoy-alpine:v1.18.3
```

Now upgrade OSM (both the CLI and the cluster components) to v1.0.0-rc.2:

```bash
$ osm version
Version: v1.0.0-rc.2; Commit: bee1cac865a33360f1669d139969636c5ee6fd2f; Date: 2021-11-15-17:21
~
 $ osm mesh upgrade
OSM successfully upgraded mesh [osm] in namespace [osm-system]
~
 $ kubectl logs -n osm-system osm-controller-9db65477b-wz7rd | head -n 1
{"level":"info","component":"osm-controller/main","time":"2021-11-18T15:47:37Z","file":"osm-controller.go:124","message":"Starting osm-controller v1.0.0-rc.2; 7351ad3173ddd7595447c49371f5d056ce1b4cc9; 2021-11-12-16:10"}
```

And see that the MeshConfig has not changed (not desired) for the sidecar image:

```bash
$ kubectl get meshconfig -n osm-system osm-mesh-config -o jsonpath='{.spec.sidecar.envoyImage}'
envoyproxy/envoy-alpine:v1.18.3
```

Just to prove that an installation (not an upgrade) of v1.0.0-rc.2 does not have this same setting, start with a new cluster and install OSM v1.0.0-rc.2:

```bash
$ osm version
Version: v1.0.0-rc.2; Commit: bee1cac865a33360f1669d139969636c5ee6fd2f; Date: 2021-11-15-17:21
[...]

PR #4386, opened 2021-12-03
Add MeshConfig upgrade when upgrading a mesh

trstringer:
**Description**:

This fixes issue #4371. Prior to this fix if there was a change to the envoy sidecar image then this would not get reflected or upgraded in the mesh config. Upon further testing we cannot just patch that in because if there are missing required keys (as indicated by the CRD) then the patch itself will fail (see the *Testing done* section below on the missing required key error when attempting to patch).

This PR adds the functionality for a complete upgrade of the mesh config by doing two things:

1. Compare CRD required fields with the current MeshConfig. If there are missing required keys then patch them in.
2. Compare the envoy sidecar image. If the default has been upgraded between the old and the new version _and_ the user has not modified the default in their mesh config, then upgrade this value.

**Testing done**:

Integration tests and manual testing.

Prior to this change:

```
 $ osm version
Version: v0.9.2; Commit: 7df9badece2f251515a2bc5e5ac79438b0c58812; Date: 2021-08-19-21:02

 $ osm install
OSM installed successfully in namespace [osm-system] with mesh name [osm]

 $ git log --oneline -1
8dccabbf (HEAD -> main, upstream/main) injector: rename iptables chains for clarity (#4379)

 $ go run ./cmd/cli mesh upgrade
OSM successfully upgraded mesh [osm] in namespace [osm-system]
```

Even though the mesh was upgraded, the envoy sidecar image remains the old version in the mesh config:

```
 $ kubectl get configmap -n osm-system preset-mesh-config -o jsonpath='{.data.preset-mesh-config\.json}' | jq -r '.sidecar.envoyImage'
envoyproxy/envoy-alpine@sha256:6502a637c6c5fba4d03d0672d878d12da4bcc7a0d0fb3f1d506982dde0039abd

 $ kubectl get meshconfig -n osm-system osm-mesh-config -o jsonpath='{.spec.sidecar.envoyImage}'
envoyproxy/envoy-alpine:v1.18.3
```

If you try to patch that it errors because of a missing key:

```
 $ kubectl patch meshconfig -n osm-system osm-mesh-config --type=json --patch='[{"op":"replace","path":"/spec/sidecar/envoyImage","value":"envoyproxy/envoy-alpine@sha256:6502a637c6c5fba4d03d0672d878d12da4bcc7a0d0fb3f1d506982dde0039abd"}]'
The MeshConfig "osm-mesh-config" is invalid: spec.certificate.certKeyBitSize: Required value
```

With this PR:

```
 $ kubectl get meshconfig -n osm-system osm-mesh-config -o jsonpath='{.spec.sidecar.envoyImage}'
envoyproxy/envoy-alpine@sha256:6502a637c6c5fba4d03d0672d878d12da4bcc7a0d0fb3f1d506982dde0039abd

 $ kubectl get meshconfig -n osm-system osm-mesh-config -o jsonpath='{.spec.certificate.certKeyBitSize}'
2048
```

**Affected area**:

No.

2. Is this a breaking change?

No.

### DIFF
--- a/charts/osm/templates/osm-injector-deployment.yaml
+++ b/charts/osm/templates/osm-injector-deployment.yaml
@@ -94,11 +94,11 @@ spec:
               valueFrom:
                 fieldRef:
                   fieldPath: metadata.name
-            - name: OSM_ENVOY_IMAGE
+            - name: OSM_DEFAULT_ENVOY_IMAGE
               value: "{{ .Values.osm.sidecarImage }}"
-            - name: OSM_ENVOY_WINDOWS_IMAGE
+            - name: OSM_DEFAULT_ENVOY_WINDOWS_IMAGE
               value: "{{ .Values.osm.sidecarWindowsImage }}"
-            - name: OSM_INIT_CONTAINER_IMAGE
+            - name: OSM_DEFAULT_INIT_CONTAINER_IMAGE
               value: '{{ include "osmSidecarInit.image" . }}'
     {{- if .Values.osm.imagePullSecrets }}
       imagePullSecrets:
--- a/charts/osm/templates/osm-injector-deployment.yaml
+++ b/charts/osm/templates/osm-injector-deployment.yaml
@@ -94,6 +94,12 @@ spec:
               valueFrom:
                 fieldRef:
                   fieldPath: metadata.name
+            - name: OSM_ENVOY_IMAGE
+              value: "{{ .Values.osm.sidecarImage }}"
+            - name: OSM_ENVOY_WINDOWS_IMAGE
+              value: "{{ .Values.osm.sidecarWindowsImage }}"
+            - name: OSM_INIT_CONTAINER_IMAGE
+              value: '{{ include "osmSidecarInit.image" . }}'
     {{- if .Values.osm.imagePullSecrets }}
       imagePullSecrets:
 {{ toYaml .Values.osm.imagePullSecrets | indent 8 }}
--- a/pkg/configurator/methods.go
+++ b/pkg/configurator/methods.go
@@ -127,7 +127,7 @@ func (c *client) GetEnvoyLogLevel() string {
 func (c *client) GetEnvoyImage() string {
 	image := c.getMeshConfig().Spec.Sidecar.EnvoyImage
 	if image == "" {
-		image = os.Getenv("OSM_ENVOY_IMAGE")
+		image = os.Getenv("OSM_DEFAULT_ENVOY_IMAGE")
 	}
 	return image
 }
@@ -136,7 +136,7 @@ func (c *client) GetEnvoyImage() string {
 func (c *client) GetEnvoyWindowsImage() string {
 	image := c.getMeshConfig().Spec.Sidecar.EnvoyWindowsImage
 	if image == "" {
-		image = os.Getenv("OSM_ENVOY_WINDOWS_IMAGE")
+		image = os.Getenv("OSM_DEFAULT_ENVOY_WINDOWS_IMAGE")
 	}
 	return image
 }
@@ -145,7 +145,7 @@ func (c *client) GetEnvoyWindowsImage() string {
 func (c *client) GetInitContainerImage() string {
 	image := c.getMeshConfig().Spec.Sidecar.InitContainerImage
 	if image == "" {
-		image = os.Getenv("OSM_INIT_CONTAINER_IMAGE")
+		image = os.Getenv("OSM_DEFAULT_INIT_CONTAINER_IMAGE")
 	}
 	return image
 }
--- a/pkg/configurator/methods.go
+++ b/pkg/configurator/methods.go
@@ -3,6 +3,7 @@ package configurator
 import (
 	"encoding/json"
 	"fmt"
+	"os"
 	"time"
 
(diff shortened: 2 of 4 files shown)


nojnhuh:
I know this is still WIP so I haven't taken a close look yet, but a few initial thoughts:

It seems like the simplest solution here would be to default the image name/tag in the injector and leave it empty in the MeshConfig. Then the upgraded injector would implicitly start using its updated default. This was the behavior until #4131. Of the reasons @shashankram describes there, I think the biggest problem with using a default was that the default would be used if there was some error reading the value from the MeshConfig. I'd argue we should fail the injection altogether in that case. Otherwise, I don't see any issues with using a default if no value is specified in the MeshConfig.

Going forward, adding new required fields to the CRD should result in an appropriate version bump for the API (v1alpha1 -> v1alpha2) and we should either have a conversion webhook to handle both versions or force the user to update their MeshConfigs to the new API themselves. That would handle any new required fields, so working around that here doesn't seem worth it to me unless we need this behavior for upgrading v0.9 to v0.11.

shashankram:
> I know this is still WIP so I haven't taken a close look yet, but a few initial thoughts:
> 
> It seems like the simplest solution here would be to default the image name/tag in the injector and leave it empty in the MeshConfig. Then the upgraded injector would implicitly start using its updated default. This was the behavior until #4131. Of the reasons @shashankram describes there, I think the biggest problem with using a default was that the default would be used if there was some error reading the value from the MeshConfig. I'd argue we should fail the injection altogether in that case. Otherwise, I don't see any issues with using a default if no value is specified in the MeshConfig.
> 
> Going forward, adding new required fields to the CRD should result in an appropriate version bump for the API (v1alpha1 -> v1alpha2) and we should either have a conversion webhook to handle both versions or force the user to update their MeshConfigs to the new API themselves. That would handle any new required fields, so working around that here doesn't seem worth it to me unless we need this behavior for upgrading v0.9 to v0.11.

The reason defaults are not to be encoded in the binary is because the image digests are not known. We should avoid going back to encoding defaults in the Go code if possible, and always infer it through the MeshConfig or other means.

nojnhuh:
> The reason defaults are not to be encoded in the binary is because the image digests are not known. We should avoid going back to encoding defaults in the Go code if possible, and always infer it through the MeshConfig or other means.

The default doesn't necessarily have to be a constant in Go. We could pass it on the command line or as an environment variable when the injector starts so it can still be defined once in values.yaml like it is now.

trstringer:
This was a complete rewrite of the original PR, so the first message in this thread is no longer completely relevant.

From the feedback, I refactored this to use environment variables if there is no setting in the MeshConfig for the following images: envoy image, envoy windows image, and the init container image.

The default was also removed from the mesh config CRD, as that would auto-populate the MeshConfig resource with the defaults.

This does *not* solve the problem with missing required parameters (in the original case, `certKeyBitSize`) but as @nojnhuh mentioned this is more of a process issue with incrementing the resource version and dealing with the changes through a conversion webhook.

charts/osm/templates/osm-injector-deployment.yaml
```diff
@@ -94,6 +94,12 @@ spec:
               valueFrom:
                 fieldRef:
                   fieldPath: metadata.name
+            - name: OSM_ENVOY_IMAGE
+              value: "{{ .Values.osm.sidecarImage }}"
+            - name: OSM_ENVOY_WINDOWS_IMAGE
+              value: "{{ .Values.osm.sidecarWindowsImage }}"
+            - name: OSM_INIT_CONTAINER_IMAGE
+              value: '{{ include "osmSidecarInit.image" . }}'
```
nojnhuh, line 102: I would include `DEFAULT` somewhere in these names to make it clearer that these are fallbacks and not authoritative.

pkg/configurator/methods.go
```diff
@@ -124,17 +125,29 @@ func (c *client) GetEnvoyLogLevel() string {
 
 // GetEnvoyImage returns the envoy image
 func (c *client) GetEnvoyImage() string {
-	return c.getMeshConfig().Spec.Sidecar.EnvoyImage
+	image := c.getMeshConfig().Spec.Sidecar.EnvoyImage
+	if image == "" {
+		image = os.Getenv("OSM_ENVOY_IMAGE")
```
nojnhuh, line 130: I remember @shashankram mentioning when we had a default before, the default would be used if there was some error in `getMeshConfig()` since it returns an empty MeshConfig then. For users that override the image to pull from their own secure registries, using the default in that case probably isn't what users would expect. Since users can control what the default is here though, maybe that's less of a concern.

trstringer: I think that would probably be a documentation exercise, right? If the user specifies `osm.sidecarImage` and `osm.sidecarWindowsImage` when they installed OSM then it would effectively set the defaults to whatever they want. That way if we get no data back from the MeshConfig then it falls back to their defaults.

nojnhuh: Yeah, that sounds reasonable. Whether or not we should fail to inject if looking up the image names in the MeshConfig fails seems like a separate discussion.

nojnhuh (approved):

shashankram (approved):

PR #4395, opened 2021-12-11
feat(cli): add remote version to osm version output

jaellio:
**Description**:

Adds remote version to `osm version` CLI command output. By default,
the CLI and Mesh version will be included in the output. A remote
flag (by default set to true) can be set to false to only output
the CLI version. If there is any failure getting the mesh version,
the CLI version is printed along with the failure message.

A different mesh can be specified by overriding the default osm namespace
 via the OSM_CONFIG or the global `--osm-namespace` flag.

Part of #4289

**Previous `osm version` output:**

```shell
Version: v0.11.1; Commit: c01aefae509d59735d7908a32a359327ff3f2322; Date: 2021-10-20-20:33
```

**New `osm version` output:**

```shell
$ osm version
CLI Version: version.Info{Version:"v0.11.1", GitCommit:"c01aefae509d59735d7908a32a359327ff3f2322", BuildDate:"2021-10-20-20:33"}
Mesh [osm] Version: version.Info{Version:"v0.11.1", GitCommit:"c01aefae509d59735d7908a32a359327ff3f2322", BuildDate:"2021-10-20-20:33"}

$ osm version --remote=false
CLI Version: version.Info{Version:"v0.11.1", GitCommit:"c01aefae509d59735d7908a32a359327ff3f2322", BuildDate:"2021-10-20-20:33"}
```
Example output when there is an error getting the mesh version:
```shell
$ osm version
CLI Version: version.Info{Version:"v0.11.1", GitCommit:"c01aefae509d59735d7908a32a359327ff3f2322", BuildDate:"2021-10-20-20:33"}
Error: Failed to get mesh version: No mesh found in namespace [osm-system]
```

**Testing done**:
- Wrote unit tests for getting the controller pods and mesh version, and for verifying the output format
- Manually verified the `osm version` outputs with OSM installed on a kind and AKS cluster

**Affected area**:
| CLI Tool                   | [ x ] |

2. Is this a breaking change?

### DIFF
--- a/cmd/cli/version.go
+++ b/cmd/cli/version.go
@@ -1,32 +1,147 @@
 package main
 
 import (
+	"context"
+	"encoding/json"
 	"fmt"
 	"io"
+	"strconv"
 
+	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
+	"k8s.io/client-go/kubernetes"
 
+	"github.com/openservicemesh/osm/pkg/constants"
+	httpserverconstants "github.com/openservicemesh/osm/pkg/httpserver/constants"
 	"github.com/openservicemesh/osm/pkg/version"
 )
 
 const versionHelp = `
-This command prints out all the version information used by OSM
+This command prints the OSM CLI and remote mesh version information
 `
 
-// PrintCliVersion prints the version
-func PrintCliVersion(out io.Writer) {
-	_, _ = fmt.Fprintf(out, "Version: %s; Commit: %s; Date: %s\n", version.Version, version.GitCommit, version.BuildDate)
+type versionCmd struct {
+	out           io.Writer
+	namespace     string
+	clientOnly    bool
+	clientset     kubernetes.Interface
+	remoteVersion remoteVersionGetter
+}
+
+type remoteVersionGetter interface {
+	proxyGetMeshVersion(pod string, namespace string, clientset kubernetes.Interface) (*version.Info, error)
+}
+
+type remoteVersion struct{}
+
+type remoteVersionInfo struct {
+	meshName string
+	version  *version.Info
+}
+
+type versionInfo struct {
+	cliVersionInfo    *version.Info
+	remoteVersionInfo *remoteVersionInfo
 }
 
 func newVersionCmd(out io.Writer) *cobra.Command {
+	versionCmd := &versionCmd{
+		out: out,
+	}
 	cmd := &cobra.Command{
 		Use:   "version",
-		Short: "osm cli version",
+		Short: "osm cli and mesh version",
 		Long:  versionHelp,
 		Args:  cobra.NoArgs,
-		Run: func(cmd *cobra.Command, args []string) {
-			PrintCliVersion(out)
+		RunE: func(cmd *cobra.Command, args []string) error {
+			var versionInfo versionInfo
+			var err error
+
+			cliVersionInfo := version.GetInfo()
+			versionInfo.cliVersionInfo = &cliVersionInfo
+
+			if !versionCmd.clientOnly {
+				err = versionCmd.setKubeClientset()
+				if err == nil {
+					versionCmd.namespace = settings.Namespace()
+					versionCmd.remoteVersion = &remoteVersion{}
+					versionInfo.remoteVersionInfo, err = versionCmd.getMeshVersion()
+				}
+			}
+
+			versionCmd.outputVersionInfo(versionInfo)
+			if err != nil {
+				return errors.Wrap(err, "Failed to get mesh version")
+			}
+			return nil
 		},
 	}
+
+	f := cmd.Flags()
+	f.BoolVar(&versionCmd.clientOnly, "client-only", false, "only show the OSM CLI version")
+
 	return cmd
+}
+
+func (v *versionCmd) setKubeClientset() error {
+	config, err := settings.RESTClientGetter().ToRESTConfig()
+	if err != nil {
+		return errors.Wrap(err, "Error fetching kubeconfig")
+	}
+
+	clientset, err := kubernetes.NewForConfig(config)
+	if err != nil {
+		return errors.Wrap(err, "Could not access Kubernetes cluster, check kubeconfig")
+	}
+	v.clientset = clientset
+	return nil
+}
+
+func (v *versionCmd) getMeshVersion() (*remoteVersionInfo, error) {
+	var version *version.Info
+	var versionInfo *remoteVersionInfo
+
+	controllerPods, err := getControllerPods(v.clientset, v.namespace)
+	if err != nil {
+		return nil, err
+	}
+	if len(controllerPods.Items) == 0 {
+		return nil, errors.Errorf("No mesh found in namespace [%s]", v.namespace)
+	}
+
+	controllerPod := controllerPods.Items[0]
+	version, err = v.remoteVersion.proxyGetMeshVersion(controllerPod.Name, v.namespace, v.clientset)
+	if err != nil {
+		return nil, err
+	}
+	versionInfo = &remoteVersionInfo{
+		meshName: controllerPod.Labels[constants.OSMAppInstanceLabelKey],
+		version:  version,
+	}
+	return versionInfo, nil
+}
+
+func (r *remoteVersion) proxyGetMeshVersion(pod string, namespace string, clientset kubernetes.Interface) (*version.Info, error) {
+	resp, err := clientset.CoreV1().Pods(namespace).ProxyGet("", pod, strconv.Itoa(constants.OSMHTTPServerPort), httpserverconstants.VersionPath, nil).DoRaw(context.TODO())
+	if err != nil {
+		return nil, errors.Wrapf(err, "Error retrieving mesh version from pod [%s] in namespace [%s]", pod, namespace)
+	}
+	if len(resp) == 0 {
+		return nil, errors.Errorf("Empty response received from pod [%s] in namespace [%s]", pod, namespace)
+	}
+
+	versionInfo := &version.Info{}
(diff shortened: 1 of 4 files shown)


cmd/cli/version.go
```diff
@@ -1,32 +1,146 @@
...
+		return nil, errors.Errorf("Error unmarshalling retrieved mesh version from pod %s in namespace %s: %s", pod, namespace, err)
+	}
+
+	return versionInfo, nil
+}
+
+func (v *versionCmd) outputVersionInfo(versionInfo versionInfo) {
+	fmt.Fprintf(v.out, "CLI Version: %s\n", fmt.Sprintf("%#v", *versionInfo.cliVersionInfo))
+	if v.remote && versionInfo.remoteVersionInfo != nil {
+		fmt.Fprintf(v.out, "Mesh [%s] Version: %s\n", versionInfo.remoteVersionInfo.meshName, fmt.Sprintf("%#v", *versionInfo.remoteVersionInfo.version))
```
jaellio, line 144: "Mesh" could be replaced with  "Server", "Remote", or "Control Plane". Would appreciate any feedback!

cmd/cli/version_test.go
```diff
@@ -0,0 +1,188 @@
...
+type fakeRemoteVersion struct {
+	err         error
+	versionInfo *version.Info
+}
+
+func (f *fakeRemoteVersion) proxyGetMeshVersion(pod string, namespace string, clientset kubernetes.Interface) (*version.Info, error) {
+	if f.err != nil {
+		return nil, f.err
+	}
+	return f.versionInfo, nil
```
nojnhuh, line 29: I suppose this could be simplified a bit since the current test cases only define one of these anyway.
```suggestion
	return f.versionInfo, f.err
```

cmd/cli/version_test.go
```diff
@@ -0,0 +1,188 @@
...
+						Name:      "controllerPod",
+						Namespace: "test",
+						Labels: map[string]string{
+							constants.AppLabel:               constants.OSMControllerName,
+							constants.OSMAppInstanceLabelKey: "osm",
+						},
+					},
+				},
+			},
+			err: errors.Errorf("Error retrieving mesh version from pod %s in namespace %s)", "controllerPod", settings.Namespace()),
```
nojnhuh, line 66: I find inlining string constants in format strings easier to read than making them format parameters.
```suggestion
			err: errors.Errorf("Error retrieving mesh version from pod controllerPod in namespace %s", settings.Namespace()),
```

cmd/cli/version_test.go
```diff
@@ -0,0 +1,188 @@
...
+				},
+				remote: true,
+			}
+
+			remoteVersionInfo, err := cmd.getMeshVersion()
+			if err != nil && test.err != nil {
+				assert.Equal(test.err.Error(), err.Error())
+			} else {
+				assert.Equal(test.err, err)
+			}
```
nojnhuh, line 126: Because `test.err` is used as both the error returned by the fake `proxyGetMeshVersion()` and an expected return value from `getMeshVersion()`, the actual contents defined in the test case don't really matter since they'll always match, so checking them for equality doesn't make much sense here. Off the top of my head I'm not coming up with a decent way to refactor this though.

cmd/cli/version.go
```diff
@@ -1,32 +1,146 @@
...
+	if err != nil {
+		return nil, errors.Errorf("Error unmarshalling retrieved mesh version from pod %s in namespace %s: %s", pod, namespace, err)
+	}
+
+	return versionInfo, nil
+}
+
+func (v *versionCmd) outputVersionInfo(versionInfo versionInfo) {
+	fmt.Fprintf(v.out, "CLI Version: %s\n", fmt.Sprintf("%#v", *versionInfo.cliVersionInfo))
+	if v.remote && versionInfo.remoteVersionInfo != nil {
```
nojnhuh, line 143: Is there a reason you need to check `v.remote` again here even though `versionInfo.remoteVersionInfo` is only ever set after the same check in the `cmd`'s `.Run`?

cmd/cli/version_test.go
```diff
@@ -0,0 +1,188 @@
...
+			assert := tassert.New(t)
+
+			buf := bytes.NewBuffer(nil)
+			cmd := versionCmd{
+				out:    buf,
+				remote: test.remote,
+			}
+			cmd.outputVersionInfo(test.versionInfo)
+
+			assert.Contains(buf.String(), test.expected)
```
nojnhuh, line 185: `Equal` also passes here, so I'd use that just to avoid any mistakes where e.g. we expect only the client version but both the client and remote versions get written.
```suggestion
			assert.Equal(buf.String(), test.expected)
```

cmd/cli/version.go
jaellio: Nope, good point!

cmd/cli/version.go
```diff
@@ -1,32 +1,146 @@
...
+	f := cmd.Flags()
+	f.BoolVarP(&versionCmd.remote, "remote", "r", true, "If false, shows only the OSM CLI version")
+
 	return cmd
 }
+
+func (v *versionCmd) setKubeClientset() error {
+	config, err := settings.RESTClientGetter().ToRESTConfig()
+	if err != nil {
+		return errors.Errorf("Error fetching kubeconfig: %s", err)
```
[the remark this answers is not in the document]
jaellio: Good to know, thank you!

cmd/cli/version.go
```diff
@@ -1,32 +1,146 @@
...
+}
+
+func (r *remoteVersion) proxyGetMeshVersion(pod string, namespace string, clientset kubernetes.Interface) (*version.Info, error) {
+	resp, err := clientset.CoreV1().Pods(namespace).ProxyGet("", pod, strconv.Itoa(constants.OSMHTTPServerPort), httpserverconstants.VersionPath, nil).DoRaw(context.TODO())
+	if err != nil {
+		return nil, errors.Errorf("Error retrieving mesh version from pod %s in namespace %s: %s", pod, namespace, err)
+	}
+	if len(resp) == 0 {
+		return nil, errors.Errorf("Error retrieving mesh version from pod %s in namespace %s", pod, namespace)
+	}
```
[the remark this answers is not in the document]
jaellio: Good point, thanks.

cmd/cli/version.go
```diff
@@ -1,32 +1,146 @@
...
+
+			versionCmd.outputVersionInfo(versionInfo)
+			if err != nil {
+				fmt.Fprintf(versionCmd.out, "Failed to get mesh version: %v", err)
+			}
 		},
 	}
+
+	f := cmd.Flags()
+	f.BoolVarP(&versionCmd.remote, "remote", "r", true, "If false, shows only the OSM CLI version")
```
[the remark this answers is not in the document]
jaellio: Updated flag to `client-only` which defaults to false. Adjusted my logic accordingly.

cmd/cli/version.go
```diff
@@ -1,32 +1,146 @@
...
+	versionCmd := &versionCmd{
+		out: out,
+	}
 	cmd := &cobra.Command{
 		Use:   "version",
-		Short: "osm cli version",
+		Short: "osm cli and mesh version",
 		Long:  versionHelp,
 		Args:  cobra.NoArgs,
 		Run: func(cmd *cobra.Command, args []string) {
```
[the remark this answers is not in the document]
jaellio: Updated `Run` to `RunE`. I had originally kept `Run` here since I wanted to format the error to indicate that the failure occurred when getting the mesh version. There is no reason I can't wrap the error and then return it as an error.

After changes - an example output when getting the mesh version fails:
```shell
CLI Version: version.Info{Version:"dev", GitCommit:"5dcc0d0f087f99205cc85a220568432b234b048c", BuildDate:""}
Error: Failed to get mesh version: No mesh found in namespace [kube-system]
```

Previous output:
```shell
CLI Version: version.Info{Version:"dev", GitCommit:"5dcc0d0f087f99205cc85a220568432b234b048c", BuildDate:""}
Failed to get mesh version: No mesh found in namespace [kube-system]
```

cmd/cli/version_test.go
jaellio: I was checking the error contents in case `getControllerPods` failed. I think this would be most helpful when writing the unit tests and the controller pods were created incorrectly. If in the test we expected one error, but a different error was returned due to incorrectly setting up the test we wouldn't know and the expected scenarios wouldn't be tested.

nojnhuh: In general I would probably define the mocked error returned by `proxyGetMeshVersion()` in the test case separately from the error expected to be returned by `getMeshVersion()`. For the cases currently defined, these would always be the same, so sharing one field for both works.

If you define a test case where `getControllerPods()` is expected to fail, you can set `test.err` to that error you expect, and since `proxyGetMeshVersion()` wouldn't get called then, the test would still pass. But if `getControllerPods()` doesn't fail where you expect it to in that case, then `proxyGetMeshVersion()` will still return that `test.err` and the test will pass, even though it's not failing where you expect it to.

cmd/cli/version.go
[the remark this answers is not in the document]
nojnhuh: The "Error:" is slightly redundant, but returning an error here will also ultimately result in a non-zero exit code returned to the shell. Getting that to happen with plain `Run` wouldn't be as easy.

nojnhuh (approved):

cmd/cli/version_test.go
jaellio: Added a `proxyGetMeshVersionErr` field

draychev (approved):
This is great!  Thank you @jaellio!

PR #4411, opened 2022-01-03
injector: allow redirection of app traffic to itself

### LINKED ISSUE #4340: POD cannot communicate to itself via k8s-svc
I have k8s deployment with 4 PODS and 1 corresponding k8s-SVC 

when I exec into one of these 4 PODS and start curling on k8s-SVC I have 3 correct answers (from 3 other PODS) but 4th session goes to this POD that I'm on that 

and I got the message: 
```
upstream connect error or disconnect/reset before headers. reset reason: connection failure, transport failure reason: TLS error: 268435703:SSL routines:OPENSSL_internal:WRONG_VERSION_NUMBER
```
when I scale deployment to 2 PODS I have 1 correct answer and 1 with error 

and analogously when I scale to 6 PODS I have 5 correct answers but when it hit that POD that I'm on that I got once per 6 error message 
```
upstream connect error or disconnect/reset before headers. reset reason: connection failure, transport failure reason: TLS error: 268435703:SSL routines:OPENSSL_internal:WRONG_VERSION_NUMBER
```

seems like POD cannot communicate to itself via k8s-svc 

tested on AKS with OSM 0.9.2 and 0.11

shashankram:
**Description**:
This change allows an app to invoke itself by enabling
the necessary iptables rules. It allows an app to direct
traffic to itself when:
1. app directs traffic to its pod IP
2. app directs traffic to its k8s service IP
   which is then resolved to the local pod IP
   by Envoy.

Resolves #4340

**Testing done**:
Tested app sending traffic to itself via k8s service
and pod IP.

**Affected area**:

### DIFF
--- a/pkg/injector/init_container_test.go
+++ b/pkg/injector/init_container_test.go
@@ -52,6 +52,8 @@ var _ = Describe("Test functions creating Envoy bootstrap configuration", func()
 -A OSM_PROXY_OUT_REDIRECT -p tcp -j REDIRECT --to-port 15001
 -A OSM_PROXY_OUT_REDIRECT -p tcp --dport 15000 -j ACCEPT
 -A OUTPUT -p tcp -j OSM_PROXY_OUTBOUND
+-A OSM_PROXY_OUTBOUND -o lo ! -d 127.0.0.1/32 -m owner --uid-owner 1500 -j OSM_PROXY_IN_REDIRECT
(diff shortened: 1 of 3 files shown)


trstringer (approved):

snehachhabria (approved):

draychev (approved):
Great comments! Thank you!

PR #4437, opened 2022-01-13
bug(*): Change memory targetAverageUtilization to targetAverageValue

shalier:
**Description**:
Could not use targetAverageUtilization for memory, as metrics-server showed memory in bytes instead of a percentage.
image
Change memory from targetAverageUtilization to targetAverageValue, memory values were set based on memory usage of the bookstore demo for osm-controller and osm-injector, so that only 1 replica was needed.
image

**Testing done**:
Ran bookstore demo to determine memory limits.

**Affected area**:

### DIFF
--- a/charts/osm/values.yaml
+++ b/charts/osm/values.yaml
@@ -65,8 +65,8 @@ osm:
         # -- Average target CPU utilization (%)
         targetAverageUtilization: 80
       memory:
-        # -- Average target memory utilization (%)
-        targetAverageUtilization: 80
+        # -- Average target memory value
+        targetAverageValue: "50M"
 
   #
   # -- Prometheus parameters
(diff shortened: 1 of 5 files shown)


charts/osm/values.yaml
```diff
@@ -65,8 +65,8 @@ osm:
         # -- Average target CPU utilization (%)
         targetAverageUtilization: 80
       memory:
-        # -- Average target memory utilization (%)
-        targetAverageUtilization: 80
+        # -- Average target memory value
+        targetAverageValue: "50M"
```
shashankram, line 69: > memory values were set based on memory usage of the bookstore demo for osm-controller and osm-injector

@shalier the bookstore demo should not be used as a benchmark env to set default values. We need to use a more realistic testbed, such as one with 100 pods, so we can have defaults that work in most environments. In other words, using defaults based on the demo env is not something I agree with.

shashankram (changes requested):
@shalier we should be able to use a percentage for memory consumption. What's the problem with doing so? I could not understand this from the commit description.

shalier:
Closing this issue, as targetAverageUtilization should be possible with memory, will look into this issue, created #4451

Issue #4439, opened 2022-01-13
"osm uninstall mesh" prompts user to uninstall non-existent mesh if there are no meshes installed


johnsonshi:
**Bug description**:

When osm is not installed in the cluster but “osm uninstall” cli command is run, the cli prompts the user to uninstall a non-existent mesh.

**Affected area (please mark with X where applicable)**:
- Install                    [X]
- SMI Traffic Access Policy  [ ]
- SMI Traffic Specs Policy   [ ]
- SMI Traffic Split Policy   [ ]
- Permissive Traffic Policy  [ ]
- Ingress                    [ ]
- Egress                     [ ]
- Envoy Control Plane        [ ]
- CLI Tool                   [X]
- Metrics                    [ ]
- Certificate Management     [ ]
- Sidecar Injection          [ ]
- Logging                    [ ]
- Debugging                  [ ]
- Tests                      [ ]
- Demo                       [ ]
- CI System                  [ ]

**Expected behavior**: If no mesh is installed, do not prompt user to uninstall non-existent mesh.

**Steps to reproduce the bug (as precisely as possible)**:

![image](https://user-images.githubusercontent.com/13926417/149273281-6a4b3bd7-a47e-4a0d-8af1-a5e93f4aac3c.png)

**How was OSM installed?**:

**Anything else we need to know?**:

**Bug report archive**:

**Environment**:
- OSM version (use `osm version`): v1.0.0-rc.3
- Kubernetes version (use `kubectl version`):
- Size of cluster (number of worker nodes in the cluster):
- Others:

shashankram:
This is pretty cosmetic given that the CLI correctly detects that a mesh does not exist:
```
$ osm uninstall mesh -f
$ osm uninstall mesh 

List of meshes present in the cluster:
No osm mesh control planes found

Uninstall OSM [mesh name: osm] in namespace [osm-system] ? [y/n]: y
Error: No OSM control plane with mesh name [osm] found in namespace [osm-system]
```

The enhancement would be to avoid the prompt if no mesh exists.

PR #4463, opened 2022-01-20
Command won`t prompt if mesh is non-existent

mudit-01:
Fixes https://github.com/openservicemesh/osm/issues/4439

**Description**:

**Testing done**: Yes

**Affected area**:
| CLI Tool                   | [X ] |

2. Is this a breaking change?

### DIFF
--- a/cmd/cli/uninstall_mesh.go
+++ b/cmd/cli/uninstall_mesh.go
@@ -96,7 +96,7 @@ func newUninstallMeshCmd(config *action.Configuration, in io.Reader, out io.Writ
 func (d *uninstallMeshCmd) run() error {
 	meshInfoList, _ := getMeshInfoList(d.config, d.clientSet)
 	if len(meshInfoList) == 0 {
-		return errors.Errorf("No osm mesh control planes found")
+		return errors.Errorf("No osm mesh found")
 	}
 
 	ctx, cancel := context.WithCancel(context.Background())
--- a/cmd/cli/uninstall_mesh.go
+++ b/cmd/cli/uninstall_mesh.go
@@ -96,6 +96,9 @@ func newUninstallMeshCmd(config *action.Configuration, in io.Reader, out io.Writ
 func (d *uninstallMeshCmd) run() error {
 	meshInfoList, _ := getMeshInfoList(d.config, d.clientSet)
 	if len(meshInfoList) == 0 {
+		if d.force {
+			return nil
+		}
 		return errors.Errorf("No osm mesh found")
 	}
 
--- a/cmd/cli/uninstall_mesh.go
+++ b/cmd/cli/uninstall_mesh.go
@@ -94,6 +94,11 @@ func newUninstallMeshCmd(config *action.Configuration, in io.Reader, out io.Writ
 }
 
 func (d *uninstallMeshCmd) run() error {
+	meshInfoList, _ := getMeshInfoList(d.config, d.clientSet)
+	if len(meshInfoList) == 0 {
+		return errors.Errorf("No osm mesh control planes found")
+	}
+
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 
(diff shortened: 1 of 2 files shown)


mudit-01:
Integration Test with Tresor is failing it might be the case that `osm-system namespace` is missing hence throwing an error for it. Is there any workaround to it?

cmd/cli/uninstall_mesh.go
```diff
@@ -94,6 +94,11 @@ func newUninstallMeshCmd(config *action.Configuration, in io.Reader, out io.Writ
 }
 
 func (d *uninstallMeshCmd) run() error {
+	meshInfoList, _ := getMeshInfoList(d.config, d.clientSet)
+	if len(meshInfoList) == 0 {
+		return errors.Errorf("No osm mesh control planes found")
```
shashankram, line 99:
```suggestion
		return errors.Errorf("No osm mesh found")
```

cmd/cli/uninstall_mesh_test.go
```diff
@@ -117,18 +117,8 @@ var _ = Describe("Running the uninstall command", func() {
...
-			It("should prompt for confirmation", func() {
-				Expect(out.String()).To(ContainSubstring("Uninstall OSM [mesh name: testing] in namespace [" + settings.Namespace() + "] ? [y/n]: "))
-			})
-			It("should error", func() {
-				Expect(err).To(MatchError("No OSM control plane with mesh name [testing] found in namespace [" + settings.Namespace() + "]"))
-			})
-			It("should not give a message confirming the successful uninstall", func() {
-				Expect(out.String()).ToNot(ContainSubstring("OSM [mesh name: testing] in namespace [" + settings.Namespace() + "] uninstalled\n"))
-			})
-
```
shashankram, line 131: Please do not remove existing tests. You need to extend the existing test for the scenario you are addressing.

shashankram (changes requested):

cmd/cli/uninstall_mesh_test.go
mudit-01: hi @shashankram but these tests will become irrelevant, as when we don't have an osm mesh its prompt for the confirmation of its uninstallation so these 3 tests will not be required

shashankram: I see that the test already covers the 2 scenarios - 1. mesh exists, 2. mesh doesn't exist. My bad I missed this.

shashankram (approved):
Thanks @mudit-01!

snehachhabria (approved):

jaellio:
> Integration Test with Tresor is failing it might be the case that `osm-system namespace` is missing, hence throwing an error for it. Is there any workaround to it?

Hi @mudit-01! Thanks so much for this change!! I think I found a small fix to work around the failing integration tests.

The test is failing in a clean-up script (clean-kubernetes.sh) we call when running the automated demo (run-osm-demo.sh). The clean-up script calls `bin/osm uninstall mesh -f --mesh-name "$MESH_NAME" --osm-namespace "$K8S_NAMESPACE"` which will now error when there is no mesh installed. A `force` or `f` flag on the `osm uninstall mesh` command hides any errors or output. The slight modification below to your change will return nil instead of an error when no mesh is installed and `force` is true.

**uninstall_mesh.go**
```go
func (d *uninstallMeshCmd) run() error {
    meshInfoList, _ := getMeshInfoList(d.config, d.clientSet)
    if len(meshInfoList) == 0 {
        if d.force {
                return nil
        }
        return errors.Errorf("No osm mesh found")
    }
```

mudit-01:
Thanks, @jaellio for helping in fixing the issue :)

mudit-01:
Is this PR ok to merge @jaellio @shashankram @snehachhabria ?

jaellio (approved):

mudit-01:
Thanks @jaellio

PR #4471, opened 2022-01-24
tests/framework: wait for deletion and dump namespace

shashankram:
**Description**:
Waits for resources created by Helm to delete prior
to performing deletion of the entire namespace.
Without this, some dangling resources can be left
behind in the cluster such and the namespace could
be stuck in the terminating state.
Also dumps the namespaces that fail to terminate. This
is necessary to root cause namespace deletion issues.

**Testing done**:
Used this change to root cause a bug in uninstall that left
namespaces in a dangling state.

**Affected area**:

### DIFF
--- a/tests/framework/common.go
+++ b/tests/framework/common.go
@@ -1240,7 +1240,7 @@ func (td *OsmTestData) Cleanup(ct CleanupType) {
 				},
 			)
 			if err != nil {
-				td.T.Logf("Cleanup of namespaces timed out: %s", err)
+				td.T.Logf("Error polling namespaces for deletion: %s", err)
 
 				// Dump the namespaces that failed to terminate
 				stdout, stderr, _ := td.RunLocal("kubectl", "get", "namespace", "-l", osmTest, "-o", "yaml")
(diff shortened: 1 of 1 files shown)


snehachhabria (approved):

tests/framework/common.go
```diff
@@ -1239,7 +1240,13 @@ func (td *OsmTestData) Cleanup(ct CleanupType) {
 				},
 			)
 			if err != nil {
-				td.T.Logf("Poll err: %v", err)
+				td.T.Logf("Cleanup of namespaces timed out: %s", err)
```
nojnhuh, line 1243: This error could happen for reasons other than just timeouts. If it does time out, the error will say that, so I'd make this message more generic just to avoid confusion if the error isn't really a timeout.

tests/framework/common.go
```diff
@@ -1239,7 +1240,13 @@ func (td *OsmTestData) Cleanup(ct CleanupType) {
...
+
+				// Dump the namespaces that failed to terminate
+				stdout, stderr, _ := td.RunLocal("kubectl", "get", "namespace", "-l", osmTest, "-o", "yaml")
+				td.T.Logf("Namespace info:")
+				td.T.Logf("stdout:\n%s", stdout)
+				td.T.Logf("stderr:\n%s", stderr)
```
nojnhuh, line 1249: For something this simple, could you use the k8s Go client instead to do this? There are only a couple places we invoke `kubectl` directly and I'd rather not add more if we can help it.

shashankram: I'd rather not, in the interest of simplifying how the object is dumped nicely by kubectl.

tests/framework/common.go
shashankram: Updated.

jaellio (approved):

Issue #4445, opened 2022-01-13
"osm namespace add ns" and "osm metrics enable ns" should abort if namespace has ignore label


johnsonshi:
**Bug description**: Language in [osm docs for `osm ignore`](https://release-v0-11.docs.openservicemesh.io/docs/guides/app_onboarding/namespaces/#ignore-a-namespace) suggests that as long as a namespace has the `ignore` label, it should NEVER be a part of a mesh. As a consequence, if a pod has been marked with ignore, adding the namespace or enabling metrics should abort.

**Affected area (please mark with X where applicable)**:
- Install                    [ ]
- SMI Traffic Access Policy  [ ]
- SMI Traffic Specs Policy   [ ]
- SMI Traffic Split Policy   [ ]
- Permissive Traffic Policy  [ ]
- Ingress                    [ ]
- Egress                     [ ]
- Envoy Control Plane        [ ]
- CLI Tool                   [X]
- Metrics                    [ ]
- Certificate Management     [ ]
- Sidecar Injection          [ ]
- Logging                    [ ]
- Debugging                  [ ]
- Tests                      [ ]
- Demo                       [ ]
- CI System                  [ ]

**Expected behavior**:

**Steps to reproduce the bug (as precisely as possible)**: Below shows 

```
bin/osm install
kubectl create ns abc
bin/osm namespace ignore abc
bin/osm namespace add abc # should NOT succeed
bin/osm metrics enable --namespace abc # should NOT succeed

kubectl get namespace abc -o yaml
```

**How was OSM installed?**:

**Anything else we need to know?**:

**Bug report archive**:

**Environment**:
- OSM version (use `osm version`): v1.0.0-rc.3
- Kubernetes version (use `kubectl version`):
- Size of cluster (number of worker nodes in the cluster):
- Others:

mudit-01:
I would like to work on it pls assign it to me.

johnsonshi:
Sure @mudit-01 that would be greatly appreciated! <3

allenlsy:
What does `"if a pod has been marked with ignore, "` in the description? Is it possible that a namespace is not ignored, but only the pod it the namespace has the ignore label?

PR #4473, opened 2022-01-25
If ignore label exist don`t add namespace

mudit-01:
PR partially fixes https://github.com/openservicemesh/osm/issues/4445

**Description**:

**Testing done**: Yes

**Affected area**:

2. Is this a breaking change?

### DIFF
--- a/cmd/cli/namespace_add.go
+++ b/cmd/cli/namespace_add.go
@@ -15,6 +15,8 @@ import (
 	"github.com/openservicemesh/osm/pkg/constants"
 )
 
+const trueValue = "true"
+
 const namespaceAddDescription = `
 This command adds a namespace or set of namespaces to the mesh so that the osm
 control plane with the given mesh name can observe resources within that namespace
@@ -122,7 +124,7 @@ func (a *namespaceAddCmd) run() error {
 		}
 
 		// if ignore label exits don`t add namespace
-		if namespace.ObjectMeta.Labels[constants.IgnoreLabel] == "true" {
+		if namespace.ObjectMeta.Labels[constants.IgnoreLabel] == trueValue {
 			return errors.Errorf("Cannot add ignored namespace")
 		}
 
--- a/cmd/cli/namespace_add.go
+++ b/cmd/cli/namespace_add.go
@@ -124,7 +124,7 @@ func (a *namespaceAddCmd) run() error {
 		}
 
 		// if ignore label exits don`t add namespace
-		if namespace.ObjectMeta.Labels[constants.IgnoreLabel] == trueValue {
+		if val, ok := namespace.ObjectMeta.Labels[constants.IgnoreLabel]; ok && val == trueValue {
 			return errors.Errorf("Cannot add ignored namespace")
 		}
 
--- a/cmd/cli/namespace_add.go
+++ b/cmd/cli/namespace_add.go
@@ -121,6 +121,11 @@ func (a *namespaceAddCmd) run() error {
 			continue
 		}
 
+		// if ignore label exits don`t add namespace
+		if namespace.ObjectMeta.Labels[constants.IgnoreLabel] == "true" {
+			return errors.Errorf("Cannot add ignored namespace")
+		}
+
 		var patch string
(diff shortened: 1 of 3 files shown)


mudit-01:
The e2e_ignore_namespace_test.go test is failing as it is trying to add an ignored namespace to the mesh that is forbidden by this PR. 
My query is can we remove this test? After this PR adding an ignored namespace to a mesh will not be possible, hence disabling sidecar injection will not be required.
What are your views?

allenlsy:
@mudit-01 I think what the original e2e test wants to verify is still valid, which is sidecar injection should not work for ignored namespace. But the way it sets the environment is not right. It:

* creates a namespace with ignore label on it.
* adds it to the mesh

Rather than doing so, it should:

* create a namespace
* use `osm ignore` to ignore the namespace

Then proceed with adding a pod to the namespace and check the existence of sidecar.

Can you update the e2e test case in this PR also?

shashankram:
> @mudit-01 I think what the original e2e test wants to verify is still valid, which is sidecar injection should not work for ignored namespace. But the way it sets the environment is not right. It:
> 
> * creates a namespace with ignore label on it.
> * adds it to the mesh
> 
> Rather than doing so, it should:
> 
> * create a namespace
> * use `osm ignore` to ignore the namespace
> 
> Then proceed with adding a pod to the namespace and check the existence of sidecar.
> 
> Can you update the e2e test case in this PR also?

@allenlsy there is nothing wrong with how the current e2e test sets the label directly on the namespace. The logic was meant to test the ignore label functionality in relation to sidecar injection, and not the CLI behavior. We do not need to test CLI behavior in e2e as a unit test should suffice to test CLI.

mudit-01:
@shashankram Should I update the e2e test to `Expect(Td.AddNsToMesh(true, ignoreNs)).NotTo(Succeed())` as once the ignore lable is found it should not add the namespace to the mesh?

johnsonshi:
Summarizing what @jaellio and I discussed:

* `osm namespace ignore` currently patches a `openservicemesh.io/ignore: true` label to a namespace when run.
* Mutating Webhook and Validating Webhook use the *existence* of the `openservicemesh.io/ignore` label (not the boolean value) as the top-level admission criteria. If the `ignore` label does not exist, the webhooks further check criteria by looking at the value of `openservicemesh.io/enable-sidecar-injection`.
* From running `osm namespace ignore --help`, the intent of the command is to (1) allow a namespace to be added to osm but (2) prevent further NEW sidecars to be injected.
```
$ osm namespace ignore --help
This command will prevent a namespace or a set of namespaces from
participating in the mesh. Automatic sidecar injection on pods
belonging to the given namespace or set of namespaces will be prevented.
The command will not remove previously injected sidecars on pods belonging
to the given namespaces.
```
* The `enable-sidecar-injection` label and `ignore` labels currently seem to do duplicate work. One should supersede the other, and one should be deprecated imho.

shashankram:
> Proposal:
> 
> * `osm namespace ignore [namespace]` should patch a `openservicemesh.io/enable-sidecar-injection: false` label to a namespace instead of an `openservicemesh.io/ignore: true` label.

The `ignore` label is working the way it is intended to. It is meant to ignore resources (mutating webhooks, ingresses, etc.). We should not be tweaking the sidecar injection label in relation to the ignore label. The ignore label should supersede the sidecar injection annotation.

mudit-01:
@allenlsy @shashankram @johnsonshi any suggestions on this or this is ok to merge?

shashankram (approved):

jaellio (approved):

cmd/cli/namespace_add.go
```diff
@@ -121,6 +123,11 @@ func (a *namespaceAddCmd) run() error {
 			continue
 		}
 
+		// if ignore label exits don`t add namespace
+		if namespace.ObjectMeta.Labels[constants.IgnoreLabel] == trueValue {
```
shashankram, line 127:
```suggestion
		if val, ok := namespace.ObjectMeta.Labels[constants.IgnoreLabel]; ok && val == trueValue {
```

PR #4479, opened 2022-01-25
configurator: remove ip/port exclusion APIs

shashankram:
**Description**:
Simplifies the Configurator interface by leveraging
the `GetMeshConfig()` API instead of exposing iptables
related exclusion fields as separate APIs.

Note that this change only removes the APIs related
to IP/Port exclusions, to begin with, but the 
general concept applies to other fields as well.

**Affected area**:


snehachhabria (approved):
Good change!

trstringer (approved):

PR #4484, opened 2022-01-26
injector: simplify port exclusion code

shashankram:
**Description**:
Simplifies the code and tests related to port
exclusions and consolidates exclusions related
code in a separate file for better maintenance.

- Removes unnecessary wrapper function
- Removes unnecessary variable assignment to nil
- Returns nil for error type when there is no error
- Decouples functions from the `mutatingWebhook` type
- Cleans test to remove Ginkgo usage
- Cleans test to remove mocks and other unnecessary objects

**Testing done**:
Ensured code coverage for updated changes is 100%

**Affected area**:


snehachhabria (approved):
thanks for simplifying!

nojnhuh (approved):

Issue #4396, opened 2021-12-14
Option to configure set of outbound IP range inclusion for traffic interception at pod level


SoujanyaMangipudi:
The ask is to provide an option to configure include OutboundIPRanges for egress interception at pod level. 

- New Functionality          [X]
- Install                    [ ]
- SMI Traffic Access Policy  [ ]
- SMI Traffic Specs Policy   [ ]
- SMI Traffic Split Policy   [ ]
- Permissive Traffic Policy  [ ]
- Ingress                    [ ]
- Egress                     [X]
- Envoy Control Plane        [ ]
- CLI Tool                   [ ]
- Metrics                    [ ]
- Certificate Management     [ ]
- Sidecar Injection          [ ]
- Logging                    [ ]
- Debugging                  [ ]
- Tests                      [ ]
- CI System                  [ ]
- Demo                       [ ]
- Project Release            [ ]

The ask is to provide an option to configure include OutboundIPRanges for egress interception at pod level. We currently use similar feature with istio service mesh. This gives us the flexibility of configuring the service IP ranges for egress interception without having to explicitly exclude every infra related IP range (like Storage, Active Directory etc.). Having this option in OSM will make our migration simpler.

shashankram:
This is done. The feature should be available in the next release (v1.1.0). 
Preview documentation can be found at: https://main--osm-docs.netlify.app/docs/guides/traffic_management/iptables_redirection/#outbound-ip-range-inclusions

PR #4488, opened 2022-01-27
injector: support pod specific IP range exclusions

shashankram:
**Description**:
Adds support for specifying pod specific outbound
IP range exclusions similar to port exclusions.
The pod specific and global IP range exclusions
will both apply to a pod being injected.

Also makes the following changes:
- Fails patching on invalid annotations
- Moves OS specific init container code to
  its own function

Related to #4396

**Testing done**:
Unit tests

**Affected area**:

3. Has documentation corresponding to this change been updated in the [osm-docs](https://github.com/openservicemesh/osm-docs) repo (if applicable)? no, will be done later

### DIFF
--- a/pkg/injector/exclusions.go
+++ b/pkg/injector/exclusions.go
@@ -1,12 +1,24 @@
 package injector
 
 import (
+	"net"
 	"strconv"
 	"strings"
 
 	mapset "github.com/deckarep/golang-set"
 	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
+)
+
+const (
+	// outboundPortExclusionListAnnotation is the annotation used for outbound port exclusions
+	outboundPortExclusionListAnnotation = "openservicemesh.io/outbound-port-exclusion-list"
+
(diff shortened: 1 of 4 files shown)


snehachhabria (approved):

SanyaKochhar (approved):

trstringer (approved):
This looks good. Great work, @shashankram! Let's also file an issue in osm-docs to document this behavior and an example how it works and how to implement it. Also, this PR should fully address #4396, right?

shashankram:
> This looks good. Great work, @shashankram! Let's also file an issue in osm-docs to document this behavior and an example how it works and how to implement it. Also, this PR should fully address #4396, right?

Hey @trstringer, I was planning to update osm-docs once I complete implementing #4396. This change does not address #4396 as the issue pertains to IP range `inclusions` whereas this change enhances the `exclusions` behavior. The reason I implemented this change is so that both inclusions and exclusions can be configured similarly. I will start working on #4396 next.

Issue #3632, opened 2021-06-21
Unified Provider Interface for Endpoint and Service Providers


whitneygriffith:
**Please describe the Improvement and/or Feature Request**

A separate service provider was created in #3614 to enable multicluster functionality. 

The mesh catalog is now aware of `endpoint.Provider` and `service.Provider` interfaces and we will like to collapse these two interfaces into one `service.Provider`. 

**Scope (please mark with X where applicable)**
- New Functionality          [X ]
- Install                    [ ]
- SMI Traffic Access Policy  [ ]
- SMI Traffic Specs Policy   [ ]
- SMI Traffic Split Policy   [ ]
- Permissive Traffic Policy  [ ]
- Ingress                    [ ]
- Egress                     [ ]
- Envoy Control Plane        [ ]
- CLI Tool                   [ ]
- Metrics                    [ ]
- Certificate Management     [ ]
- Sidecar Injection          [ ]
- Logging                    [ ]
- Debugging                  [ ]
- Tests                      [ ]
- CI System                  [ ]
- Demo                       [ ]
- Project Release            [ ]

**Possible use cases**

[Multi Cluster Proposal Addendum
](https://docs.google.com/document/d/1sA03IwqC6tuJpCleGMakQpHzwFJ5ulfxzCq7iSXSqZc/edit#heading=h.ae8gbsiygb1m)

steeling:
One other recommendation.. there's some duplication between the provider methods and the meshcatalog. You could do the following:

1. Create a single Provider interface and have a new field on the meshcatalog:

```
type MeshCatalog struct {
   ...
   provider Provider
   ...
}
```

as opposed to an array of providers. If you ever really needed multiple providers you could have 

type AggregatedProvider struct {
   providers []Provider
}

This would remove a good amount of code.

2. Following this train of thought, why not make the provider an embedded interface?
type MeshCatalog struct {
   ...
   Provider
   ...
}

This will allow some of the xDS files to directly call these methods, removing some more boiler plate.

PR #4495, opened 2022-01-31
Feature/single provider

steeling:
Simplify the provider interface into a single interface. This abstracts the provider API as a whole, which will likely be tied to a single system (like k8s). It is unclear if there would ever be a scenario where an endpoint and service provider would not be coupled, and even if there were, it would be trivial to separate them out and join them into a single struct using DI.

Removing multiple providers on the service mesh further simplifies the code, while still allowing a future MultipleProvider struct if ever desired.

Finally embedding the provider interface allows us to directly access its methods and remove even more boiler plate.

Functionality and tests not changed

**Affected area**:
| Other                      | [ X] |

No

2. Is this a breaking change?

No

Yes, PR https://github.com/openservicemesh/osm-docs/pull/320


steeling:
I can't link issues (or reopen them), but https://github.com/openservicemesh/osm/issues/3632

shashankram (changes requested):
@steeling let's not make this change at this moment. We want to minimize widespread code refactor as we aim to prioritize more important fixes and features. When we decide to refactor again, we need to propose this via a design doc before the implementation.

PR #4499, opened 2022-02-02
feature: extend demo to work on arm64 

schristoff:
**Description**: Demo currently uses nodeSelector for scheduling - but this runs into issues when expanding usage to work with arm64. 

**Testing done**: `make kind-demo` builds properly and was able to run through the demo

**Affected area**:

   no

2. Is this a breaking change?
no

n/a


Makefile
```diff
@@ -23,7 +23,7 @@ BUILD_DATE_VAR := github.com/openservicemesh/osm/pkg/version.BuildDate
 BUILD_VERSION_VAR := github.com/openservicemesh/osm/pkg/version.Version
 BUILD_GITCOMMIT_VAR := github.com/openservicemesh/osm/pkg/version.GitCommit
 DOCKER_GO_VERSION = 1.17
-DOCKER_BUILDX_PLATFORM ?= linux/amd64
+DOCKER_BUILDX_PLATFORM ?= linux/arm64
```
schristoff, line 26: @nojnhuh  opinion on what we should do about this?

shashankram: Please keep the default as before `linux/amd64`.

shashankram (changes requested):
The `nodeSelector` is still required on linux apps so they don't get scheduled onto Windows nodes.

demo/deploy-mysql.sh
```diff
@@ -46,10 +46,8 @@ spec:
         app: mysql
     spec:
       serviceAccountName: mysql
-      nodeSelector:
-        kubernetes.io/os: linux
       containers:
-      - image: mysql:5.6
+      - image: mysql/mysql-server
```
nojnhuh, line 50: For the sake of not changing more than we need to, is there a 5.6 or similar tag we can use? I doubt using latest will cause any problems, just want to keep the scope of these changes narrow.

demo/deploy-mysql.sh
```diff
@@ -46,10 +46,8 @@ spec:
         app: mysql
     spec:
       serviceAccountName: mysql
-      nodeSelector:
-        kubernetes.io/os: linux
```
nojnhuh, line 50: In this case, keeping the nodeSelector for linux is probably a good idea since it doesn't look like this image works on Windows.

demo/deploy-vault.sh
```diff
@@ -27,9 +27,6 @@ spec:
       labels:
         app: vault
     spec:
-      nodeSelector:
-        kubernetes.io/arch: amd64
-        kubernetes.io/os: linux
```
nojnhuh, line 32: I'd keep the nodeSelector for os but skip arch for the same reason as mysql re: no windows support.

demo/multicluster-fault-injection.sh
```diff
@@ -43,9 +43,6 @@ spec:
         openservicemesh.io/sidecar-injection: "false"
     spec:
       serviceAccountName: "$BOOKSTORE_SVC"
-      nodeSelector:
-        kubernetes.io/arch: amd64
-        kubernetes.io/os: linux
```
nojnhuh, line 48: Same here to keep os but remove arch.

nojnhuh (changes requested):

demo/deploy-mysql.sh
schristoff: Yeah we can do something like that.

demo/deploy-mysql.sh
schristoff: Ah, good call, I did a sweeping pass. Totally agree.

Makefile
schristoff: Hi @shashankram  - thanks for your suggestion! If I keep it as the default of `linux/amd64` it negates the reasoning of this PR - do you think the suggestion from @nojnhuh would be a reasonable middle ground?
Thanks!

nojnhuh: FWIW, if we decide to stick with linux/amd64 hardocded as the default, you could set this as an environment variable to be whatever default you want and I don't think you'd need to do anything extra for that to be picked up.

shashankram: This can be overriden as an argument to `make`, but I see @nojnhuh's suggestion and agree with it.

schristoff: Thanks everyone!

demo/multicluster-fault-injection.sh
schristoff: Why is it needed here? Just curious.

nojnhuh: curl doesn't have Windows images but supports several architectures for linux.

schristoff: Ah, I only had to look a measly two lines down. Thank you so much :) That makes sense.

charts/osm/templates/cleanup-hook.yaml
```diff
@@ -85,9 +85,6 @@ spec:
              kubectl delete --ignore-not-found secret -n '{{ include "osm.namespace" . }}' {{ .Values.osm.caBundleSecretName }} mutating-webhook-cert-secret validating-webhook-cert-secret crd-converter-cert-secret;
              kubectl delete mutatingwebhookconfiguration -l app.kubernetes.io/name=openservicemesh.io,app.kubernetes.io/instance={{ .Values.osm.meshName }},app.kubernetes.io/version={{ .Chart.AppVersion }},app=osm-injector --ignore-not-found;
              kubectl delete validatingwebhookconfiguration -l app.kubernetes.io/name=openservicemesh.io,app.kubernetes.io/instance={{ .Values.osm.meshName }},app.kubernetes.io/version={{ .Chart.AppVersion }},app=osm-controller --ignore-not-found;
-      nodeSelector:
-        kubernetes.io/arch: amd64
-        kubernetes.io/os: linux
```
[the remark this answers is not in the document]
SanyaKochhar: Heads up: https://github.com/openservicemesh/osm/pull/4527 replaces nodeselectors with node affinities

.env.example
```diff
@@ -4,9 +4,9 @@
 # env variables required for running the demo
 
 # mandatory: URL of the container registry to use.
-# Default (local Kind registry): localhost:5000
+# Default (local Kind registry): 127.0.0.1:5000
 # Azure Container Registry (ACR) example: osmci.azurecr.io/osm
-export CTR_REGISTRY=localhost:5000
+export CTR_REGISTRY=127.0.0.1:5000
```
schristoff, line 9: Sean and I were running into odd kind registry networking issues when communicating to localhost vs 127.0.0.1
I switched this over, and I do not see any issues hopefully :)

nojnhuh: Did you figure out why localhost doesn't work?

charts/osm/templates/jaeger-deployment.yaml
```diff
@@ -22,9 +22,20 @@ spec:
...
+              - matchExpressions:
+                - key: kubernetes.io/os
+                  operator: In
+                  values:
+                  - linux
+                - key: kubernetes.io/arch
+                  operator: In
+                  values: 
+                  - amd64
+                  - arm64
```
nojnhuh, line 38: Since it looks like the jaegertracing/all-in-one image supports more archs, we might as well add them here:
```suggestion
                  values: 
                  - amd64
                  - arm64
                  - ppc64le
                  - s390x
```
https://hub.docker.com/r/jaegertracing/all-in-one/tags

charts/osm/templates/prometheus-deployment.yaml
```diff
@@ -22,9 +22,20 @@ spec:
...
+              - matchExpressions:
+                - key: kubernetes.io/os
+                  operator: In
+                  values:
+                  - linux
+                - key: kubernetes.io/arch
+                  operator: In
+                  values: 
+                  - amd64
+                  - arm64
```
nojnhuh, line 38: The Prometheus image also supports more architectures:
```suggestion
                  values: 
                  - amd64
                  - arm64
                  - arm
                  - s390x
```
https://hub.docker.com/layers/prom/prometheus/v2.18.1/images/sha256-575eb2a6d4c2fb7b15a55a2788294925ab148da77b21dc1c92d0557693fdf749?context=explore

demo/deploy-mysql.sh
```diff
@@ -49,7 +49,7 @@ spec:
       nodeSelector:
         kubernetes.io/os: linux
       containers:
-      - image: mysql:5.6
+      - image: mysql/mysql-server:8.0.28
```
nojnhuh, line 52: @allenlsy Are there any concerns with bumping the MySQL version as long as the CI passes?

demo/reset-counters.sh
```diff
@@ -7,10 +7,10 @@ source .env
 
 # This script assumes that port forwarding has already been established.
 # See ./scripts/port-forward-all.sh to enable fort forwarding for Bookstore demo.
-curl -I -X GET http://localhost:8080/reset &
-curl -I -X GET http://localhost:8081/reset &
-curl -I -X GET http://localhost:8082/reset &
-curl -I -X GET http://localhost:8083/reset &
+curl -I -X GET http://127.0.0.1:8080/reset &
```
nojnhuh, line 10: Did this also not work for you as localhost?

.env.example
schristoff: No clue, still debugging. It works fine with any other application, though.

charts/osm/templates/jaeger-deployment.yaml
schristoff: Do you just want to remove arch and leave linux?

.env.example
shashankram: Since this issue is very specific to your env and this is an example file, you could override it without needing to change the default here.

nojnhuh: It's entirely possible 127.0.0.1 is an objectively better default, but we won't know that for sure until we figure out why localhost isn't working.

shashankram: Regardless, this change seems unrelated to this PR. I'd make this and other similar changes in a separate PR so they can be reviewed in isolation.

shashankram: > running into odd kind registry networking issues when communicating to localhost vs 127.0.0.1
I switched this over

@schristoff This led me to think this is an issue specific to your environment (Kind cluster). Are you suggesting localhost doesn't work on arm64? If yes, I'd add this to the commit and PR description, and also a comment in the example file to not use `localhost` as we'd want to keep the demo as generic as possible.

schristoff: I think I've now narrowed down the issue (with the help of @nojnhuh ) - your original suggestion is correct @shashankram  - I will open up another PR with the appropriate changes. :)

demo/reset-counters.sh
schristoff: I just cmd+f changed them all. I'll leave them in the next PR

PR #4302, opened 2021-10-21
chore(cleanup): remove cleanup script

SanyaKochhar:
Signed-off-by: Sanya Kochhar <kochhars@microsoft.com>

**Description**: Deletes the cleanup script per https://github.com/openservicemesh/osm-docs/pull/254 and updates the release guide to ensure docs are updated when uninstall hook is updated

TODO: update link once 254 is merged

**Testing done**:

**Affected area**:
| Functional Area            |     |
| -------------------------- | --- |
| New Functionality          | [ ] |
| CI System                  | [ ] |
| CLI Tool                   | [ ] |
| Certificate Management     | [ ] |
| Control Plane              | [ ] |
| Demo                       | [ ] |
| Documentation              | [ ] |
| Egress                     | [ ] |
| Ingress                    | [ ] |
| Install                    | [ ] |
| Networking                 | [ ] |
| Observability              | [ ] |
| Performance                | [ ] |
| SMI Policy                 | [ ] |
| Security                   | [ ] |
| Sidecar Injection          | [ ] |
| Tests                      | [ ] |
| Upgrade                    | [ ] |
| Other                      | [ ] |

Please answer the following questions with yes/no.

1. Does this change contain code from or inspired by another project?
    -   Did you notify the maintainers and provide attribution?

2. Is this a breaking change?

PR #4506, opened 2022-02-03
chore(cleanup): remove cleanup script

SanyaKochhar:

**Description**: 
Deletes the cleanup script as this is no longer used and is documented as user-facing steps here: https://github.com/openservicemesh/osm-docs/pull/254 

**Testing done**:

**Affected area**:
| Documentation              | [x] |

Reopen of #4302

no
2. Is this a breaking change?
no


snehachhabria (approved):

docs/release_guide.md
```diff
@@ -152,6 +152,8 @@ On the docs site's main branch, edit the file [https://github.com/openservicemes
 1. Copy the table and replace the existing table in the file [https://github.com/openservicemesh/osm-docs/blob/main/content/docs/guides/troubleshooting/control_plane_error_codes.md](https://github.com/openservicemesh/osm-docs/blob/main/content/docs/guides/troubleshooting/control_plane_error_codes.md).
 1. If there were updates to the table, make a PR in the OSM docs repository.
 
+### 5. Update uninstall documentation
+If any changes were made to the [OSM cleanup hook](https://github.com/openservicemesh/osm/blob/main/charts/osm/templates/cleanup-hook.yaml), make a PR in the OSM docs repository updating the list of resources deleted during uninstall in the [uninstall guide](https://github.com/openservicemesh/osm-docs/blob/main/content/docs/guides/uninstall.md#uninstall-osm-control-plane-and-remove-user-provided-resources) and the commands in the [uninstall troubleshooting](https://github.com/openservicemesh/osm-docs/blob/main/content/docs/guides/troubleshooting/uninstall.md) manual deletion instructions.
```
shashankram, line 156: I feel as though this seems out of place in the release guide. This should be a part of the PR lifecycle so that it is detected much earlier during code review, vs adding to process overhead during the release workflow.

I'd rather have a workflow as follows:
1. Someone updates the cleanup hook
2. Reviewer detects doc changes required
3. Opens an issue in osm-docs to track this and makes it a release blocker if applicable

shashankram (dismissed):
@SanyaKochhar Could you drop the update to the release guide and only have this PR remove the cleanup script? I'd like to avoid unnecessary overhead in the release workflow.

docs/release_guide.md
nojnhuh, line 156: I would expect the docs changes to land at more or less the same time as the changes to the hook vs. catching up only when we cut a release. Then this info might be best communicated in a comment in the hook.

nojnhuh: I didn't see this before posting my comment, but mine says basically the same thing.

nojnhuh (approved):

PR #4507, opened 2022-02-03
fix(cli): uninstall mesh assume yes flag

johnsonshi:
When `osm uninstall mesh -f` is run, the `force`
flag is intended for the user to skip the uninstall
confirmation prompt only.

Even if this code is specified, the code in uninstall
mesh does not supress diagnostic messages or errors
when it encounters an error during mesh uninstallation.

This proposed fix instead changes the force flag to
'--assume-yes' flag to make this behavior clear, only
skipping the confirmation prompt but not supressing
diagnostic messages and errors.

**Testing done**: Local testing + unit testing

**Affected area**:

3. Has documentation corresponding to this change been updated in the [osm-docs](https://github.com/openservicemesh/osm-docs) repo (if applicable)? NA

### DIFF
--- a/cmd/cli/uninstall_mesh.go
+++ b/cmd/cli/uninstall_mesh.go
@@ -93,7 +93,7 @@ func newUninstallMeshCmd(config *action.Configuration, in io.Reader, out io.Writ
 
 	f := cmd.Flags()
 	f.StringVar(&uninstall.meshName, "mesh-name", defaultMeshName, "Name of the service mesh")
-	f.BoolVarP(&uninstall.force, "force", "f", false, "Attempt to uninstall the osm control plane instance without prompting for confirmation.  If the control plane with specified mesh name does not exist, do not display a diagnostic message or modify the exit status to reflect an error.")
+	f.BoolVarP(&uninstall.force, "force", "f", false, "Attempt to uninstall the osm control plane instance without prompting for confirmation.")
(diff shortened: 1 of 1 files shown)


cmd/cli/uninstall_mesh.go
```diff
@@ -93,7 +93,7 @@ func newUninstallMeshCmd(config *action.Configuration, in io.Reader, out io.Writ
 
 	f := cmd.Flags()
 	f.StringVar(&uninstall.meshName, "mesh-name", defaultMeshName, "Name of the service mesh")
-	f.BoolVarP(&uninstall.force, "force", "f", false, "Attempt to uninstall the osm control plane instance without prompting for confirmation.  If the control plane with specified mesh name does not exist, do not display a diagnostic message or modify the exit status to reflect an error.")
+	f.BoolVarP(&uninstall.assumeYes, "assume-yes", "y", false, "Assumes yes to prompts and attempt to uninstall the osm control plane instance without prompting for confirmation.")
```
snehachhabria, line 96: Lets only remove ` If the control plane with specified mesh name does not exist, do not display a diagnostic message or modify the exit status to reflect an error` and not change the name of the flag please

snehachhabria (changes requested):

cmd/cli/uninstall_mesh.go
nojnhuh: I agree that renaming the flag here doesn't seem necessary given that would be a breaking change that needs to be documented in the release notes doc.

snehachhabria (approved):

nojnhuh (approved):

PR #4509, opened 2022-02-04
feat(cli): add support controller-debug-server cmd

johnsonshi:
Adds "osm support controller-debug-server" cmd,
which performs port-redirection and opens the browser
to facilitate opening the osm-controller debug server.

This is useful for both project maintainers and osm
users that want to debug a running osm instance without
needing to clone the repo to run
"scripts/port-forward-osm-debug.sh" or to remember
kubectl port-forward commands.

**Testing done**: Built and ran the command -> browser with redirect to debug server opened automatically.

**Affected area**:

3. Has documentation corresponding to this change been updated in the [osm-docs] NA(https://github.com/openservicemesh/osm-docs) repo (if applicable)?

### DIFF
--- a/cmd/cli/support.go
+++ b/cmd/cli/support.go
@@ -20,6 +20,7 @@ func newSupportCmd(config *action.Configuration, stdout io.Writer, stderr io.Wri
 		Args:  cobra.NoArgs,
 	}
 	cmd.AddCommand(newSupportErrInfoCmd(stdout))
+	cmd.AddCommand(newOsmControllerDebugServerCmd(stdout))
 	cmd.AddCommand(newSupportBugReportCmd(config, stdout, stderr))
 
 	return cmd
--- a/cmd/cli/support_debug_server.go
+++ b/cmd/cli/support_debug_server.go
@@ -1,0 +1,129 @@
+package main
+
+import (
+	"fmt"
+	"io"
+	"os"
+	"os/signal"
+
+	"github.com/pkg/browser"
+
+	"k8s.io/client-go/rest"
+
+	"github.com/openservicemesh/osm/pkg/constants"
+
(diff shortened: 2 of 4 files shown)


snehachhabria (changes requested):
@johnsonshi I don't see any issue in our queue for this. Before a PR is created it is best if an issue is created, some members of the team review the need of the issue and add it to the roadmap. Only once an issue is prioritized and made a part of the roadmap should it be worked on, and we should not be creating PRs that add new functionality to OSM (including a new cli command) without following a process.

Further if one send a out a PR it is mandatory to add unit test and not just manually test

shashankram:
@johnsonshi I think it would make more sense to implement this under an `osm debug` command vs `osm support`. The support command is an umbrella for supportability tooling such as generating a bug report, listing error codes etc., which is subtly different than something used for live debugging required by developers as in the case of this command. It would be clear `osm debug` is targeted at developers, while `osm support` is for admins/users.

nojnhuh:
I'm not sure if there's enough added value here to warrant a new CLI command when `kubectl port-forward -n osm-system svc/osm-controller 9092` does everything this does except open a browser window.

Overall I'd rather not set a precedent of adding functionality like this that can be replicated with `kubectl` or other tools users and developers likely already use and are comfortable with.

snehachhabria:
> I'm not sure if there's enough added value here to warrant a new CLI command when `kubectl port-forward -n osm-system svc/osm-controller 9092` does everything this does except open a browser window.
> 
> Overall I'd rather not set a precedent of adding functionality like this that can be replicated with `kubectl` or other tools users and developers likely already use and are comfortable with.

I completely agree with @nojnhuh

johnsonshi:
Closed as I agree with @nojnhuh's comment.

PR #4521, opened 2022-02-10
feat(errors): handle errors exactly once.

steeling:
Errors should be handled exactly once. Logging an error and
returning an error is an example of handling errors twice.
This also contributes to disjoint log lines that are hard to tie together.

Also lowercase error messages to comply with lint: https://staticcheck.io/docs/checks#ST1005

-->
**Affected area**:
| Control Plane              | [ x] |

No

2. Is this a breaking change?

No

No

### DIFF
--- a/ci/cmd/maestro/kubernetes_tools.go
+++ b/ci/cmd/maestro/kubernetes_tools.go
@@ -112,8 +112,7 @@ func GetPodName(kubeClient kubernetes.Interface, namespace, selector string) (st
 	}
 
 	if len(podList.Items) == 0 {
-		log.Error().Msgf("Zero pods found for selector %s in namespace %s", selector, namespace)
-		return "", errNoPodsFound
+		return "", fmt.Errorf("zero pods found for selector %s in namespace %s", selector, namespace)
 	}
 
 	sort.SliceStable(podList.Items, func(i, j int) bool {
@@ -235,7 +234,7 @@ func WaitForPodToBeReady(kubeClient kubernetes.Interface, totalWait time.Duratio
 
 		podName, err := GetPodName(kubeClient, namespace, selector)
 		if err != nil {
-			log.Error().Err(err).Msgf("Error getting Pod w/ selector %q", selector)
+			log.Error().Err(err)
 			time.Sleep(WaitForPod)
 			// Pod might not be up yet, try again
 			continue
--- a/ci/cmd/maestro/types.go
+++ b/ci/cmd/maestro/types.go
@@ -1,7 +1,6 @@
 package maestro
 
 import (
-	"errors"
 	"time"
 
 	"github.com/openservicemesh/osm/pkg/logger"
@@ -46,6 +45,5 @@ var (
 	// FailureLogsFromTimeSince is the interval we go back in time to get pod logs
 	FailureLogsFromTimeSince = 10 * time.Minute
 
-	log            = logger.NewPretty("ci/maestro")
-	errNoPodsFound = errors.New("no pods found")
+	log = logger.NewPretty("ci/maestro")
 )
--- a/cmd/cli/install.go
+++ b/cmd/cli/install.go
@@ -90,7 +90,7 @@ func newInstallCmd(config *helm.Configuration, out io.Writer) *cobra.Command {
 		RunE: func(_ *cobra.Command, args []string) error {
 			kubeconfig, err := settings.RESTClientGetter().ToRESTConfig()
 			if err != nil {
-				return errors.Errorf("Error fetching kubeconfig: %s", err)
+				return errors.Errorf("error fetching kubeconfig: %s", err)
 			}
 
 			clientset, err := kubernetes.NewForConfig(kubeconfig)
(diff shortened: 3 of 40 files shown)


steeling:
@shashankram draft PR to improve disjoint error messages

steeling:
> > @shashankram draft PR to improve disjoint error messages
> 
> > Errors should be handled exactly once. Logging an error and
> > returning an error is an example of handling errors twice.
> 
> I agree with the change to fix the starting case for error object messages. Also please split the change that adjusts the casing from the changes that remove log lines. It seems to be this change pertains to adjusting the error message starting case.
> 
> Regarding removing log lines, have you verified that the log lines you have removed are correctly being logged with the same message down up the call stack? I don't think that's true, because we aren't returning wrapped errors using `errors.Wrap()` to preserve error messages across the call stack. Please do not remove the error logs unless the error are wrapped and error stack is preserved up the call stack.
> 
> Consider the below code, where logging the error becomes important because `errors.Wrap()` isn't being used.
> 
> ```
> func foo() error {
>     log.Error().Msgf("foo error")
>     return err1
> }
> 
> func bar() error {
>     if err := foo(); err != nil {
>         return errors.Errorf("bar error")
>     }
> }
> ```

Cool I'll split them into 2 separate PR's.

Not sure I'm understanding your comment on wrapping. FWIW I'm using fmt.Errorf with the [error wrapping directive](https://go.dev/blog/go1.13-errors#wrapping-errors-with-w).

It is true that some of the content of the error message will differ from the previous incarnation, but I suspect collocating these messages is a larger benefit

shashankram:
> > > @shashankram draft PR to improve disjoint error messages
> > 
> > 
> > > Errors should be handled exactly once. Logging an error and
> > > returning an error is an example of handling errors twice.
> > 
> > 
> > I agree with the change to fix the starting case for error object messages. Also please split the change that adjusts the casing from the changes that remove log lines. It seems to be this change pertains to adjusting the error message starting case.
> > Regarding removing log lines, have you verified that the log lines you have removed are correctly being logged with the same message down up the call stack? I don't think that's true, because we aren't returning wrapped errors using `errors.Wrap()` to preserve error messages across the call stack. Please do not remove the error logs unless the error are wrapped and error stack is preserved up the call stack.
> > Consider the below code, where logging the error becomes important because `errors.Wrap()` isn't being used.
> > ```
> > func foo() error {
> >     log.Error().Msgf("foo error")
> >     return err1
> > }
> > 
> > func bar() error {
> >     if err := foo(); err != nil {
> >         return errors.Errorf("bar error")
> >     }
> > }
> > ```
> 
> Cool I'll split them into 2 separate PR's.
> 
> Not sure I'm understanding your comment on wrapping. FWIW I'm using fmt.Errorf with the [error wrapping directive](https://go.dev/blog/go1.13-errors#wrapping-errors-with-w).
> 
> It is true that some of the content of the error message will differ from the previous incarnation, but I suspect collocating these messages is a larger benefit

We have been using `github.com/pkg/errors` for all error objects. If you decide to change this to use `fmt.Error`, please open an issue and mark it for discussion before proceeding with the change.

shashankram:
> > > > > @shashankram draft PR to improve disjoint error messages
> > > > 
> > > > 
> > > > > Errors should be handled exactly once. Logging an error and
> > > > > returning an error is an example of handling errors twice.
> > > > 
> > > > 
> > > > I agree with the change to fix the starting case for error object messages. Also please split the change that adjusts the casing from the changes that remove log lines. It seems to be this change pertains to adjusting the error message starting case.
> > > > Regarding removing log lines, have you verified that the log lines you have removed are correctly being logged with the same message down up the call stack? I don't think that's true, because we aren't returning wrapped errors using `errors.Wrap()` to preserve error messages across the call stack. Please do not remove the error logs unless the error are wrapped and error stack is preserved up the call stack.
> > > > Consider the below code, where logging the error becomes important because `errors.Wrap()` isn't being used.
> > > > ```
> > > > func foo() error {
> > > >     log.Error().Msgf("foo error")
> > > >     return err1
> > > > }
> > > > 
> > > > func bar() error {
> > > >     if err := foo(); err != nil {
> > > >         return errors.Errorf("bar error")
> > > >     }
> > > > }
> > > > ```
> > > 
> > > 
> > > Cool I'll split them into 2 separate PR's.
> > > Not sure I'm understanding your comment on wrapping. FWIW I'm using fmt.Errorf with the [error wrapping directive](https://go.dev/blog/go1.13-errors#wrapping-errors-with-w).
> > > It is true that some of the content of the error message will differ from the previous incarnation, but I suspect collocating these messages is a larger benefit
> > 
> > 
> > We have been using `github.com/pkg/errors` for all error objects. If you decide to change this to use `fmt.Error`, please open an issue and mark it for discussion before proceeding with the change.
> 
> Does this warrant a larger discussion? That package is in maintenance mode, and fmt.Errorf is part of the std lib. Seems like a no-brainer to use that? I get the desire to have "one way of doing things", but having 2 for a short period doesn't seem like it would create a large burden on the developer.

@steeling This doesn't need a large discussion per se, but it would make sense to open an issue to share more context. Since the existing errors pkg we use is in maintenance mode, I agree using `fmt.Error` makes sense. However, when we make large changes spanning the codebase, it's always ideal for people to discuss it over an issue if they have any concerns. Moreover, your commit does not describe the motivation to use `fmt.Errorf` vs `errors.Errof`, at the very least you should add more details to your commit (and PR) description.

Hope that makes sense.

Issue #2147, opened 2020-12-05
EnforceSingleMesh label is only enforced when using OSM CLI


ksubrmnn:
**Bug description**:
The enforceSingleMesh label is only enforced when using the osm cli as that is where the logic for checking that label exists. A new mechanism should be developed to ensure enforceSingleMesh is honored when deployed without the CLI (such as with helm directly)

**Affected area (please mark with X where applicable)**:
- Install                    [X]
- SMI Traffic Access Policy  [ ]
- SMI Traffic Specs Policy   [ ]
- SMI Traffic Split Policy   [ ]
- Permissive Traffic Policy  [ ]
- Ingress                    [ ]
- Egress                     [ ]
- Envoy Control Plane        [ ]
- CLI Tool                   [ ]
- Metrics                    [ ]
- Certificate Management     [ ]
- Sidecar Injection          [ ]
- Logging                    [ ]
- Debugging                  [ ]
- Tests                      [ ]
- CI System                  [ ]

**Expected behavior**:
enforceSingleMesh will always block new osm installations.
**Steps to reproduce the bug (as precisely as possible)**:
Install  osm with osm cli with --enforce-single-mesh flag

Install osm with helm chart directly

**How was OSM installed?**:

**Anything else we need to know?**:

**Environment**:
- OSM version (use `osm version`):
- Kubernetes version (use `kubectl version`):
- Size of cluster (number of worker nodes in the cluster):
- Others:

ksubrmnn:
This is also true for other checks such as namespace/mesh name/ etc

PR #4522, opened 2022-02-10
ref(install): move enforceSingleMesh validation to chart

nojnhuh:
**Description**:
This change moves the logic enforcing `osm.enforceSingleMesh` from the
`osm install` command to a Helm pre-install hook so using the chart with
Helm directly will also perform the same validation.

Part of #2147

**Testing done**:
- Unit tests updated
- Did some manual smoke testing which will be verified by CI

**Affected area**:


cmd/osm-preinstall/osm-preinstall.go
```diff
@@ -0,0 +1,100 @@
...
+
+func singleMeshOK(clientset kubernetes.Interface, enforceSingleMesh bool) func() error {
+	return func() error {
+		deps, err := clientset.AppsV1().Deployments("").List(context.Background(), metav1.ListOptions{
+			LabelSelector: labels.SelectorFromSet(map[string]string{
+				constants.AppLabel: constants.OSMControllerName,
+			}).String(),
+		})
+		if err != nil {
+			return errors.Wrap(err, "listing OSM deployments enforcing single mesh")
```
nshankar13, line 77: Do we need the "enforcing single mesh" part here in this error message? Looks like this error is just listing all the OSM deployments in general.

nojnhuh: I tend to do this just to help distinguish different errors since otherwise there could be several "listing deployments" errors and it might be important exactly which one you're trying to find. Since the log includes the line number though making the message unique probably isn't important.

nojnhuh: And I think while I was working on this I was listing only deployments with the enforceSingleMesh label but didn't change the error message. I'll change this.

cmd/osm-preinstall/osm-preinstall.go
```diff
@@ -0,0 +1,100 @@
...
+		var existingSingleMeshes []string
+		for _, dep := range deps.Items {
+			mesh := fmt.Sprintf("namespace: %s, name: %s", dep.Namespace, dep.Labels["meshName"])
+			existingMeshes = append(existingSingleMeshes, mesh)
+			if dep.Labels["enforceSingleMesh"] == "true" {
+				existingSingleMeshes = append(existingSingleMeshes, mesh)
+			}
+		}
+
+		if len(existingSingleMeshes) > 0 {
```
nshankar13, line 90: Do we necessarily need an array for existingSingleMeshes here? Assuming that there's one mesh that enforces single mesh on the cluster, then theoretically there shouldn't be any others - can we just give an error in the previous for loop itself if we come across a mesh that enforces it? 

(Unless we're also accounting for scenarios where enforceSingleMesh didn't work as expected)

cmd/osm-preinstall/osm-preinstall.go
nshankar13: Maybe something like "error listing OSM deployments prior to single mesh enforcement" or something could also work

cmd/osm-preinstall/osm-preinstall.go
nojnhuh: In theory there would only ever be at most one, but I think if there are multiple that would be useful information to expose. If we only show the first one, then we might be needlessly hiding that info.

cmd/osm-preinstall/osm-preinstall.go
```diff
@@ -0,0 +1,100 @@
...
+		})
+		if err != nil {
+			return errors.Wrap(err, "listing OSM deployments enforcing single mesh")
+		}
+
+		var existingMeshes []string
+		var existingSingleMeshes []string
+		for _, dep := range deps.Items {
+			mesh := fmt.Sprintf("namespace: %s, name: %s", dep.Namespace, dep.Labels["meshName"])
+			existingMeshes = append(existingSingleMeshes, mesh)
```
jaellio, line 84: Is `existingSingleMeshes` meant to be `existingMeshes` here?
```suggestion
			existingMeshes = append(existingMeshes, mesh)
```

nojnhuh: Yep, good catch.

jaellio (approved):

cmd/osm-preinstall/osm-preinstall.go
```diff
@@ -0,0 +1,100 @@
...
+
+func main() {
+	log.Info().Msgf("Starting osm-preinstall %s; %s; %s", version.Version, version.GitCommit, version.BuildDate)
+
+	var verbosity string
+	var enforceSingleMesh bool
+
+	flags := pflag.NewFlagSet("osm-preinstall", pflag.ExitOnError)
+	flags.StringVarP(&verbosity, "verbosity", "v", "info", "Set log verbosity level")
+	flags.BoolVar(&enforceSingleMesh, "enforce-single-mesh", false, "Enforce only deploying one mesh in the cluster")
```
snehachhabria, line 31:
```suggestion
	flags.BoolVar(&enforceSingleMesh, "enforce-single-mesh", true, "Enforce only deploying one mesh in the cluster")
```

By default we enforce single mesh in OSM

snehachhabria (approved):
thanks for this fix, just remember to squash and merge this PR :)

Issue #4500, opened 2022-02-02
Add support for circuit breaking


shashankram:
**Please describe the Improvement and/or Feature Request**
Add support for circuit breaking.
Ref: https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/circuit_breaking

The tasks include:
- [X] [Design doc](https://docs.google.com/document/d/1vh22nXuUM2k9_y9Tnbeyrxu-FWcG7yFA-wonylQisws/edit?usp=sharing)
- [x] Implementation
- [x] Testing
- [x] Documentation: how-to guide
- [x] Documentation: demo

**Scope (please mark with X where applicable)**
- New Functionality          [X]

shashankram:
Initial support for circuit breaking is complete. 

A new issue will track further enhancements which will be a part of the next milestone.

PR #4535, opened 2022-02-14
apis: add UpstreamTraffic API types

shashankram:
**Description**:
Adds the Go types for  UpstreamTraffic API.

*Note the defaults are not to be encoded in
the JSON tags and will be implemented within
the controller subsequently.

Part of #4500

**Affected area**:


pkg/apis/policy/v1alpha1/upstreamtraffic.go
```diff
@@ -0,0 +1,103 @@
...
+	// +optional
+	HTTP *HTTPConnectionSettings `json:"http,omitempty"`
+}
+
+// TCPConnectionSettings defines the TCP connection settings for an
+// upstream host.
+type TCPConnectionSettings struct {
+	// MaxConnections specifies the maximum number of TCP connections
+	// allowed to the upstream host.
+	// Defaults to 1024 if not specified.
```
nojnhuh, line 58: I wonder if it's worth distinguishing whether it's OSM or Envoy who's actually applying this default. I could interpret this as "if I don't specify this when the resource is created and I read it later, it will have the default" (where OSM is expected to apply the default), but if we simply pass the value to Envoy and let it apply the default then that wouldn't be the case.

pkg/apis/policy/v1alpha1/upstreamtraffic.go
```diff
@@ -0,0 +1,103 @@
...
+	// +optional
+	MaxConnections uint32 `json:"maxConnections,omitempty"`
```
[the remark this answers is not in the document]
shashankram: The fields with finite defaults do not accept values that mean unlimited. Most of these have a ceil of 2^32 - 1.

The behavior of defaults is completely left to OSM regardless of Envoy defaults. 0 is not equivalent to not specified unless stated so. A validating webhook will prevent such ambiguity from occurring.

snehachhabria (approved):

pkg/apis/policy/v1alpha1/upstreamtraffic.go
shashankram: This is an OSM API, not Envoy's. So the defaults will be applied by OSM controller regardless of what Envoy uses as defaults. This is primarily because some Envoy defaults may not be suitable defaults. At present, I have set these to match Envoy defaults, but these could change later, e.g. to 2^32 - 1 based on the most common use case. At the moment it's hard to reason how sensible these defaults are. Given the fields are optional in Envoy with defaults values if unspecified, I prefer to explicitly specify defaults from within OSM for such fields.

pkg/apis/policy/v1alpha1/upstreamtraffic.go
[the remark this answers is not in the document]
nojnhuh: Is zero still a valid value? If so, doesn't this field need to be a pointer then so we can differentiate between specified (use the given value) and unspecified (use the default)?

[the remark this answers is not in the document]
shashankram: Zero is not a valid value unless specified. A validating webhook will prevent invalid values. The controller will only override the default if this is non-zero.

jaellio (approved):

pkg/apis/policy/v1alpha1/upstreamtraffic.go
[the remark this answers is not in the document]
nojnhuh: > A validating webhook will prevent invalid values.

If all the webhook sees is a `uint32` that equals 0, how does it know whether or not that was specified? That's why I think this needs to be a pointer where `nil` means unspecified.

pkg/apis/policy/v1alpha1/upstreamtraffic.go
```diff
@@ -0,0 +1,103 @@
...
+	// requests allowed to the upstream host.
+	// Defaults to 1024 if not specified.
+	// +optional
+	MaxPendingRequests uint32 `json:"maxPendingRequests,omitempty"`
+
+	// MaxRetries specifies the maximum number of parallel retries
+	// allowed to the upstream host.
+	// Defaults to 3 if not specified.
+	// +optional
+	MaxRetries uint32 `json:"maxRetries,omitempty"`
```
steeling, line 93: To Jon's point, wouldn't 0 be a valid value for MaxRetries? If that's the case then the statement "Defaults to 3 if not specified" is incorrect, since we can't distinguish between 0 and not specified

pkg/apis/policy/v1alpha1/upstreamtraffic.go
```diff
@@ -0,0 +1,103 @@
...
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// UpstreamTraffic defines the policy applicable to traffic destined
+// to an upstream host.
+// +genclient
+// +genclient:noStatus
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+type UpstreamTraffic struct {
```
steeling, line 12: COmmented on the design, but commenting here for posterity. Traffic would indicate the actual bits flowing through the system, whereas this is Setting for upstream traffic. Therefore I think a more appropriate name would be UpstreamTrafficSetting

pkg/apis/policy/v1alpha1/upstreamtraffic.go
shashankram: At this point, there is no implementation backing the API to render defaults. The controller would have logic like
`if (maxRetries > 0) // configure max retries`. I don't yet know whether this can be set to zero, needs testing, so at this point made the assumption that a non-zero value will not be allowed for max_retries similar to the other uint32 values.

pkg/apis/policy/v1alpha1/upstreamtraffic.go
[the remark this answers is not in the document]
shashankram: The CRD schema will enforce a min value of 1.

pkg/apis/policy/v1alpha1/upstreamtraffic.go
shashankram: I chose to keep this name concise similar to other API Kinds, e.g [Ingress](https://kubernetes.io/docs/concepts/services-networking/ingress/), [Egress](https://release-v1-0.docs.openservicemesh.io/docs/api_reference/policy/v1alpha1/#policy.openservicemesh.io/v1alpha1.Egress) etc. I can see how renaming this to `UpstreamTrafficSetting` would make this less ambiguous, so I'll rename it.

pkg/apis/policy/v1alpha1/upstreamtrafficsetting.go
```diff
@@ -0,0 +1,103 @@
+package v1alpha1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// UpstreamTrafficSetting defines the settings applicable to traffic destined
+// to an upstream host.
```
draychev, line 8: Examples being Circuit Breaking

draychev (approved):
This is good stuff! Glad to see this being built out!
Thanks @shashankram !

Issue #4533, opened 2022-02-14
Create certificate.Provider interface that manages shared cert provider functionality


steeling:
Currently, each cert provider implements a suite of shared functionality like cache management, and certificate rotation. The current Cert Provider interface is:

```go
// Manager is the interface declaring the methods for the Certificate Manager.
type Manager interface {
	// IssueCertificate issues a new certificate.
	IssueCertificate(CommonName, time.Duration) (Certificater, error)

	// GetCertificate returns a certificate given its Common Name (CN)
	GetCertificate(CommonName) (Certificater, error)

	// RotateCertificate rotates an existing certificate.
	RotateCertificate(CommonName) (Certificater, error)

	// GetRootCertificate returns the root certificate in PEM format and its expiration.
	GetRootCertificate() (Certificater, error)

	// ListCertificates lists all certificates issued
	ListCertificates() ([]Certificater, error)

	// ReleaseCertificate informs the underlying certificate issuer that the given cert will no longer be needed.
	// This method could be called when a given payload is terminated. Calling this should remove certs from cache and free memory if possible.
	ReleaseCertificate(CommonName)
}
```

Currently, the only method that results in API calls is `IssueCertificate`. The rest of the methods deal strictly with the cache. Because of this, we can greatly reduce the amount of code a provider is responsible for, by implementing a new interface, certificate.Provider:

```go
type Provider interface {
	// IssueCertificate issues a new certificate.
	IssueCertificate(CommonName, time.Duration) (Certificater, error)

        // Not currently needed, but we may add it later if we want to get a cert from outside the cache.
	// GetCertificate returns a certificate given its Common Name (CN)
	// GetCertificate(CommonName) (Certificater, error)

        // Not currently needed, but may be added later.
	// ListCertificates lists all certificates issued
	// ListCertificates() ([]Certificater, error)

        // Not currently needed, but we may add it later.
	// DeleteCertificate informs the underlying certificate issuer that the given cert will no longer be needed.
	// This method could be called when a given payload is terminated. Calling this should remove certs from cache and free memory if possible.
	// DeleteCertificate(CommonName)
}
```

Now that we have a Provider that *only* deals in API calls, we will change the certificate.Manager interface to a struct, which handles higher level functions like cert rotation, caching, etc. This will reduce the amount of duplicated code, align code that is not properly duplicated, and greatly lower the bar for adding a new cert provider.

steeling:
[...]

Issue #4532, opened 2022-02-14
Remove certificate.Certificater


steeling:
This is part of a larger set of refactoring opportunities for Certifacte Management in OSM

Each cert provider currently implements near-identical code for certificate.Certificater. The certificate.Certificater interface makes no API calls, and just returns data stored from the struct. Because of this, there is no need to have an interface at all. We can simply replace all references with the raw struct.

From here we can reduce to one implementation of certificate.Certificate. There's also some misuse between returning pointers and returning structs. We will switch to return only pointers as part of this effort, and migrate all methods to be pointer receivers.

PR #4536, opened 2022-02-14
feat(certificate): Remove Certificater interface in favor of struct

steeling:
Currently there is a Certificater interface that is used to abstract a Certificate.
There are 3 identical implementations of Certificate, and some misuse between treating
a certificater as either a pointer or a struct. This PR simplifies this by consolidating
to a single implementation of Certificate, and removing the need for an interface all
together.

**Description**:

**Testing done**:

**Affected area**:

2. Is this a breaking change?


steeling:
Linked issue: #4532

nojnhuh (approved):
LGTM, I agree that this makes more sense as a struct than an interface.

Looks like there are a few more remnants of `Certificater` probably worth updating:
```
$ grep -rF Certificater .
./docs/certificate_management.md:	IssueCertificate(CommonName, time.Duration) (Certificater, error)
./docs/certificate_management.md:	GetCertificate(CommonName) (Certificater, error)
./docs/certificate_management.md:	RotateCertificate(CommonName) (Certificater, error)
./docs/certificate_management.md:	GetRootCertificate() (Certificater, error)
./docs/certificate_management.md:	ListCertificates() ([]Certificater, error)
./DESIGN.md:	IssueCertificate(CommonName, time.Duration) (Certificater, error)
./DESIGN.md:	GetCertificate(CommonName) (Certificater, error)
./DESIGN.md:	RotateCertificate(CommonName) (Certificater, error)
./DESIGN.md:	GetRootCertificate() (Certificater, error)
./DESIGN.md:	ListCertificates() ([]Certificater, error)
./pkg/certificate/providers/certmanager/certificate_manager_test.go:func TestCertificaterFromCertificateRequest(t *testing.T) {
./pkg/certificate/providers/vault/certificate_manager_test.go:		It("implements Certificater correctly", func() {
./pkg/ingress/gateway.go:				log.Error().Msgf("Received unexpected message %T on cert rotation channel, expected Certificater", cert)
```

The inline Go in the docs might be worth getting rid of entirely, but it may be better to do that all at once vs. just the `certificate.Manager` interface here.

docs/certificate_management.md
```diff
@@ -128,7 +128,7 @@ When using cert-manager as the certificate manager for Open Service Mesh, it wil
 To get the certificate issued from cert-manager, OSM controller will create a certificate request from `crypto/x509`:
 
 ```go
-x509.CertificateRequest{
+x509.*CertificateReques{
```
nojnhuh, line 131: If I'm being pedantic (and fixing a typo):
```suggestion
&x509.CertificateRequest{
```

(Also applies to the next two changed lines in this file)

steeling: good catch, fixed!

draychev (commented):
Glad to see certificate issuance being rethought! Much needed TLC for these modules!

I would love to understand how you, @steeling, think about Interface vs Struct.

Some background on the `Certificater` - to fulfill the O of OSM and keep it open to all certificate issuers out there, we wanted an abstraction that would keep this flexible.  The `Certificater` interface provides that flexibility. This interface allows us to have a different implementation for the various issuers. For example:
- in something that relies on AKV for certs - `GetIssuingCA()` could be a wrapper around an RPC call (same for Hashi Vault)
- in the case of our own Tresor module - `GetIssuingCA()` could be a read on a k8s secret

I am not opposed to switching this to a `struct`, especially if tests, integrations, and demo works. I do wonder if we are going to lose some flexibility going forward. OR perhaps the interface and my thinking around it is an unnecessary future-proofing.

steeling:
> Glad to see certificate issuance being rethought! Much needed TLC for these modules!
> 
> I would love to understand how you, @steeling, think about Interface vs Struct.
> 
> Some background on the `Certificater` - to fulfill the O of OSM and keep it open to all certificate issuers out there, we wanted an abstraction that would keep this flexible. The `Certificater` interface provides that flexibility. This interface allows us to have a different implementation for the various issuers. For example:
> 
> * in something that relies on AKV for certs - `GetIssuingCA()` could be a wrapper around an RPC call (same for Hashi Vault)
> * in the case of our own Tresor module - `GetIssuingCA()` could be a read on a k8s secret
> 
> I am not opposed to switching this to a `struct`, especially if tests, integrations, and demo works. I do wonder if we are going to lose some flexibility going forward. OR perhaps the interface and my thinking around it is an unnecessary future-proofing.

Hey Delyan, thanks for the thought-out reply! A couple thoughts on my end:

1. I actually have another bug to tackle refactoring the CertManager interface, as there is currently too much duplication with respect to shared code. Work on that will start after this is merged in. https://github.com/openservicemesh/osm/issues/4533

3. A good rule of thumb is that API calls belong to interfaces. Strict data types should be used as structs. There's also the old Go homage "accept interfaces, return structs", which provides users more flexibility with their API's.

As Rob Pike has said, "the bigger the interface, the weaker the abstraction". So Interfaces should do one thing, and that's it. Currently all the Certificate Providers do multiple things: They call out to their respective API's to issue/delete certs. They deal with cached state, they start a certificate rotor, deal with message handling, etc.
If the Certificater needs to reach out to something for GetIssuingCA, then the Certificater now does multiple things. Instead, GetIssuingCA should move to CertManager (or the soon-to-be CertProvider).

Final thought, I don't think there's any scenario where this reduces flexibility, particularly after we move the GetIssuingCA call. Since a user simply puts raw bytes into the certificate, and it is the CertManager (user-provided interface) that is responsible for populating it, they can populate it as they see fit. Until that GetIssuingCA call gets moved, we can require the user to make that API call behind the scene (as the vault implementation currently does), and then stuff the result into the Cert itself.

steeling:
@draychev  I'd also tend to air on the side of YAGNI when it comes to interfaces. And since this change moves in the direction of simplification, I'd say it makes it easier to pivot and change directions in the future than the current implementation

draychev (approved):
@steeling I love the principle of `accept interfaces, return structs` - we should add it to our developer guide!

Thanks for taking the time to walk me through your thinking of why this change is necessary!

PR #4537, opened 2022-02-15
injector: Simplify test

draychev:
This PR is a tiny simplification of a unit test.

1. Uses a UUID (different one for each run) for the namespace of the k8s primitives used in the test
2. Uses a UUID for the service account
3. Leverages `pointer.BoolPtr` package instead of `isTrue := true`

### DIFF
--- a/pkg/injector/envoy_config_test.go
+++ b/pkg/injector/envoy_config_test.go
@@ -16,6 +16,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes/fake"
+	"k8s.io/utils/pointer"
 
 	"github.com/openservicemesh/osm/pkg/certificate/providers/tresor"
 	"github.com/openservicemesh/osm/pkg/configurator"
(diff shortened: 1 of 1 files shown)


pkg/injector/envoy_config_test.go
```diff
@@ -286,7 +288,7 @@ var _ = Describe("Test functions creating Envoy bootstrap configuration", func()
 				Args: []string{
 					"--log-level", "debug",
 					"--config-path", "/etc/envoy/bootstrap.yaml",
-					"--service-cluster", "svcacc.namespace",
+					"--service-cluster", fmt.Sprintf("%s.%s", serviceAccountName, namespace),
```
nojnhuh, line 291: In general I find using UUIDs in tests like this make them kind of a pain to debug since you can't distinguish one from another, so you end up with error messages like:
```
expected 642fc408-1b83-4cc3-abd7-ae4abb233f8e.de2c19c7-824b-402e-8af7-203ce0a8dd6b, got de2c19c7-824b-402e-8af7-203ce0a8dd6b.642fc408-1b83-4cc3-abd7-ae4abb233f8e
```

I think if anything, we do something like we do in e2es where there's a static component and a (short) random component to the name. For a test this small though I'm not sure if there's really any benefit to random names vs. static names beyond verifying that the same name isn't hardcoded in the function being tested, which (I suspect) would break the bookstore demo and would almost certainly be caught in code review.

draychev: I embrace the use of UUIDs for such values as a great general principle of writing tests so I apply it across the board.

I believe that randomizing these values would help catch areas where we accidentally (or not) hard-coded values.

To address the particular concern, you @nojnhuh have, we could perhaps have a hard-coded prefix like `namespace-<UUID>` so then in error messages it is clear what this it (should you even know what that string is).

Generally speaking - having the line at which a test broke and having that line be written well (where the variable names carry proper semantics) - it should be easy to understand that `namespace` expectation does not match etc.

PR #4538, opened 2022-02-15
envoy/ads: Simplify unit test

draychev:
This PR is a tiny simplification of a unit test. 

This uses `assert.True(x)` instead of `assert.Equal(x, true)`

### DIFF
--- a/pkg/envoy/ads/health_test.go
+++ b/pkg/envoy/ads/health_test.go
@@ -19,5 +19,5 @@ func TestReadiness(t *testing.T) {
 func TestServerGetID(t *testing.T) {
 	assert := tassert.New(t)
 	actual := (&Server{}).GetID()
-	assert.Equal(actual, ServerType)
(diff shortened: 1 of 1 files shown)


pkg/envoy/ads/health_test.go
```diff
@@ -8,27 +8,16 @@ import (
...
+	assert.True((&Server{ready: true}).Readiness())
 }
 
 func TestServerGetID(t *testing.T) {
 	assert := tassert.New(t)
-
-	expected := ServerType
 	actual := (&Server{}).GetID()
-	assert.Equal(actual, expected)
+	assert.Equal(actual, ServerType)
```
nojnhuh, line 22: I know this is how it was before, but while we're updating this let's flip these to match what testify wants for expected/actual:
```suggestion
	assert.Equal(ServerType, actual)
```

nojnhuh (approved):

snehachhabria (approved):

PR #4540, opened 2022-02-15
ref(install): move namespace controller check to chart

nojnhuh:
**Description**:
This change moves the validation that no meshes exist in the namespace
where OSM is to be installed from the CLI to the chart so using Helm
directly will have the same validation performed.

Part of #2147

**Testing done**:
- Added unit tests

**Affected area**:

### DIFF
--- a/cmd/osm-preinstall/osm-preinstall.go
+++ b/cmd/osm-preinstall/osm-preinstall.go
@@ -25,10 +25,12 @@ func main() {
 
 	var verbosity string
 	var enforceSingleMesh bool
+	var namespace string
 
 	flags := pflag.NewFlagSet("osm-preinstall", pflag.ExitOnError)
 	flags.StringVarP(&verbosity, "verbosity", "v", "info", "Set log verbosity level")
 	flags.BoolVar(&enforceSingleMesh, "enforce-single-mesh", true, "Enforce only deploying one mesh in the cluster")
+	flags.StringVar(&namespace, "namespace", "", "The namespace where the new mesh is to be installed")
 
 	err := flags.Parse(os.Args)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


shashankram (approved):

jaellio (approved):

snehachhabria (approved):

cmd/osm-preinstall/osm-preinstall.go
```diff
@@ -98,3 +101,24 @@ func singleMeshOK(clientset kubernetes.Interface, enforceSingleMesh bool) func()
 		return nil
 	}
 }
+
+func namespaceHasNoMesh(clientset kubernetes.Interface, namespace string) func() error {
+	return func() error {
+		deps, err := clientset.AppsV1().Deployments(namespace).List(context.TODO(), metav1.ListOptions{
```
draychev, line 107: Sometimes deps could mean dependencies. I would have spelled it out as `deployments` to avoid ambiguity.

draychev (approved):

cmd/osm-preinstall/osm-preinstall.go
```diff
@@ -98,3 +101,24 @@ func singleMeshOK(clientset kubernetes.Interface, enforceSingleMesh bool) func()
...
+			LabelSelector: labels.SelectorFromSet(map[string]string{
+				constants.AppLabel: constants.OSMControllerName,
+			}).String(),
```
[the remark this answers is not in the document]
nojnhuh: Personally, I actually find this easier to read as-is where I can more or less read the whole thing top-to-bottom/left-to-right without having to keep track of too many variables in my head. For that reason I'd rather keep this the way it is, but I'm happy to change it if others agree pulling these out into their own variables makes things more readable.

cmd/osm-preinstall/osm-preinstall.go
nojnhuh: I think at least in this case the scope is narrow enough that there isn't too much ambiguity.

PR #4550, opened 2022-02-17
api/UpstreamTrafficSetting: add informer client

shashankram:
**Description**:
- Adds the informer client
- Adds plumbing for related events
- Adds the API to the group scheme
- Fixes missing cache sync for Retry client

Part of #4500

**Affected area**:

### DIFF
--- a/pkg/policy/client.go
+++ b/pkg/policy/client.go
@@ -30,15 +30,17 @@ func newClient(kubeController k8s.Controller, policyClient policyClientset.Inter
 	informerFactory := policyInformers.NewSharedInformerFactory(policyClient, k8s.DefaultKubeEventResyncInterval)
 
 	informerCollection := informerCollection{
-		egress:         informerFactory.Policy().V1alpha1().Egresses().Informer(),
-		ingressBackend: informerFactory.Policy().V1alpha1().IngressBackends().Informer(),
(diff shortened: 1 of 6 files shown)


pkg/policy/client.go
```diff
@@ -30,15 +30,17 @@ func newClient(kubeController k8s.Controller, policyClient policyClientset.Inter
 	informerFactory := policyInformers.NewSharedInformerFactory(policyClient, k8s.DefaultKubeEventResyncInterval)
```
steeling, line 30: What're your thoughts on putting all of these types of clients into pkg/k8s. There's a few benefits of this:

* Single location for all k8s API references
* More intuitive reasoning about a single client/server relation vs client per type
* reduced boilerplate when adding new types

We can still use interfaces however we want, but I see no reason to separate the specific API calls, and maintain multiple implementations of these clients

shashankram: I agree in general it would make sense to put all k8s informers within `pkg/k8s`. That's a bigger refactor though which should not be a part of this change. The existing design is to house policy.openservicemesh.io APIs within `pkg/policy`. 

Could you file an issue as an enhancement proposal for the same?

nojnhuh (approved):

jaellio (approved):

PR #4554, opened 2022-02-22
(feature): update kind-registry port forwarding

schristoff:
**Description**: In recent versions of Docker for Mac there is a translation issue when talking to the registry at `localhost` - this issue does not appear to occur when using `127.0.0.1` - [here is a relevant issue on this. ](https://github.com/docker/for-mac/issues/6055#issuecomment-985359465)

**Testing done**: While running Docker 4.5.0 I was unable to communicate to the registry at localhost, only via 127.0.0.1 - I uninstalled 4.5.0, installed 4.0.0 and no longer had the localhost issue. I added these changes and ran `make kind-demo` successfully.

**Affected area**:

no
2. Is this a breaking change?
no
4. n/a


.env.example
```diff
@@ -4,9 +4,9 @@
 # env variables required for running the demo
 
 # mandatory: URL of the container registry to use.
-# Default (local Kind registry): localhost:5000
+# Default (local Kind registry): 127.0.0.1:5000
 # Azure Container Registry (ACR) example: osmci.azurecr.io/osm
-export CTR_REGISTRY=localhost:5000
+export CTR_REGISTRY=127.0.0.1:5000
```
schristoff, line 9: We could also do `0.0.0.0` @nojnhuh  - your call. :)

scripts/kind-with-registry.sh
```diff
@@ -48,12 +48,12 @@ nodes:
 - role: worker
 containerdConfigPatches:
 - |-
-  [plugins."io.containerd.grpc.v1.cri".registry.mirrors."localhost:${reg_port}"]
+  [plugins."io.containerd.grpc.v1.cri".registry.mirrors."127.0.0.1:${reg_port}"]
     endpoint = ["http://${reg_host}:${reg_port}"]
```
nojnhuh, line 52: Let's keep localhost and add another entry here for 127.0.0.1 so both work.

.env.example
nojnhuh: I'm fine to keep it as 127.0.0.1

demo/get-multicluster-envoy-configs.sh
```diff
@@ -29,7 +29,7 @@ for CONTEXT in $MULTICLUSTER_CONTEXTS; do
 
     ( kubectl port-forward -n "${NS}" "${OSM_GATEWAY_POD}" "${PORT}" > /dev/null & echo $! >&3 ) 3> "${TMP_PID_FILE}"
     sleep 1
-    curl -s "http://localhost:${PORT}/config_dump?include_eds"  >  "envoy_config_${CONTEXT}_${NS}_${OSM_GATEWAY_POD}.json"
+    curl -s "http://127.0.0.1:${PORT}/config_dump?include_eds"  >  "envoy_config_${CONTEXT}_${NS}_${OSM_GATEWAY_POD}.json"
```
nojnhuh, line 32: This isn't related to the container registry. Does this not work for you either?

nojnhuh (commented):
This is a reasonable change, but I'm running Docker desktop 4.5.0 and not seeing this issue, so let's not conclude that's the issue just yet.

nojnhuh (approved):
The idea to try this came from kind's script that this is based on where this same change was recently made to prevent the registry from being accessible outside localhost: https://kind.sigs.k8s.io/docs/user/local-registry/

This also fixes an issue I had on my Mac where I would need to `docker network disconnect kind kind-registry` to be able to push to the registry and then `docker network connect kind kind-registry` for the cluster to be able to pull from the registry.

I didn't figure out exactly why this fixes the issue, but I noticed prior to this change:
```
$ docker port kind-registry
5000/tcp -> 0.0.0.0:5000
5000/tcp -> :::5000
```
and after this change:
```
$ docker port kind-registry
5000/tcp -> 127.0.0.1:5000
```
Changing the added 127.0.0.1 to 0.0.0.0 also fixed my particular issue.

I suspect it has something to do with `docker push/pull` using ipv6 and that not propagating through to the registry properly since `curl` never had a problem accessing it via `localhost`.

~~@schristoff Can you update the title/description to reflect that this doesn't seem like an issue with any particular version or distribution of Docker? At least I don't think we know enough to be able to make that claim just yet.~~
Looking at it again, I think what's there is good.

jaellio (approved):

Issue #4444, opened 2022-01-13
"osm uninstall mesh" and "osm namespace ignore [ns]" should remove osm-related labels and annotations from monitored namespace + pods


johnsonshi:
**Bug description**: During `osm uninstall mesh`, all monitored namespaces (and monitored pods in these ns) should have their osm-related annotations and labels REMOVED. Another scenario is `osm namespace ignore [ns]` - [language from osm docs](https://release-v0-11.docs.openservicemesh.io/docs/guides/app_onboarding/namespaces/#ignore-a-namespace) indicate that once a mesh is set to be `ignored`, it should no longer participate in the mesh so osm-related labels/annotations have to be removed.

Labels that should be removed include osm's sidecar-injection, osm's monitored-by, openserviceh.io/metrics.

Additional candidates for removal include osm's prometheus scraping labels and osm's pod ip exclusion labels.

Labels that should NOT be removed include osm-ignore label, as the ignore label is designed to mark a namespace as being permanently excluded from osm-related operations (as long as the ignore label is in place and not explicitly removed by the user).

**Affected area (please mark with X where applicable)**:
- Install                    [ ]
- SMI Traffic Access Policy  [ ]
- SMI Traffic Specs Policy   [ ]
- SMI Traffic Split Policy   [ ]
- Permissive Traffic Policy  [ ]
- Ingress                    [ ]
- Egress                     [ ]
- Envoy Control Plane        [ ]
- CLI Tool                   [X]
- Metrics                    [ ]
- Certificate Management     [ ]
- Sidecar Injection          [ ]
- Logging                    [ ]
- Debugging                  [ ]
- Tests                      [ ]
- Demo                       [ ]
- CI System                  [ ]

**Expected behavior**: osm-related labels and annotations should be removed from monitored namespaces. After running `osm uninstall mesh`, `osm namespace list` should not list namespaces like in the scenario below.

Additionally, `osm namespace list` should show a list of "orphaned" ignored namespaces. These are namespaces where the `monitored-by` label has been removed but the namespace still has the `ignored` label left in place.

**Steps to reproduce the bug (as precisely as possible)**:

![image](https://user-images.githubusercontent.com/13926417/149275380-438bc217-bf15-444c-8493-1587825fabb8.png)

**How was OSM installed?**:

**Anything else we need to know?**:

**Bug report archive**:

**Environment**:
- OSM version (use `osm version`): v1.0.0-rc.3
- Kubernetes version (use `kubectl version`):
[...]

PR #4555, opened 2022-02-23
cli: NS removed from mesh when uninstalling mesh

mudit-01:
"osm uninstall mesh" removes osm-related labels and annotations from the monitored namespace, partially fixes https://github.com/openservicemesh/osm/issues/4444

**Description**:

**Testing done**: yes(Manual) below are the screenshots

**Affected area**:
| New Func | [ ] |
| CLI Tool                   | [x ] |

2. Is this a breaking change?


mudit-01:
@snehachhabria @shashankram @jaellio @nojnhuh please suggest, if there are any changes for the PR.

cmd/cli/uninstall_mesh.go
```diff
@@ -141,6 +141,24 @@ func (d *uninstallMeshCmd) run() error {
 		}
 
 		if err == nil {
+			namespaces, _ := selectNamespacesMonitoredByMesh(d.meshName, d.clientSet)
+			if len(namespaces.Items) != 0 {
```
draychev, line 145: @mudit-01 I don't think this check is necessary - we could remove one level of indentation. `range` is expected to handle nils (valid empty list) like this: https://go.dev/play/p/J7JQ8LKVYi6

cmd/cli/uninstall_mesh.go
```diff
@@ -141,6 +141,24 @@ func (d *uninstallMeshCmd) run() error {
...
+						namespace: ns.Name,
+						meshName:  d.meshName,
+						clientSet: d.clientSet,
+					}
+
+					err := removeCmd.run()
+
+					// Unable to remove namespace in mesh
+					if err != nil {
+						return err
```
draychev, line 158: I wonder if it would be helpful to log the reason for the inability to remove namespaces with `log.Error().Err(err).Msgf("Error removing namespaces: %+v", namespaces.Items)`
OR we could assume that the caller would do that for us based on the `err` they get.

draychev (commented):
@mudit-01 Thank you so much for helping with this project.
I left a small comment questioning the need to check whether namespaces is empty.

cmd/cli/uninstall_mesh.go
mudit-01: Thanks, @draychev for the review 👍🏽

cmd/cli/uninstall_mesh.go
mudit-01: I believe caller one is good.
what do other people suggest?

mudit-01: Can we mark this as resolved @draychev?

mudit-01:
Thanks @shashankram I will work on resolving asap

mudit-01:
Hi @shashankram need help in resolving the lint error(cyclomatic complexity). How can I remove that currently it's 32 just 2 more than the threshold?

cmd/cli/uninstall_mesh.go
```diff
@@ -179,6 +179,22 @@ func (d *uninstallMeshCmd) run() error {
 			}
```
steeling, line 179: to fix the lint, let's try to reduce some of the nesting. How about at the end of this err != nil check, add a continue, and then we can remove the `if err == nil` check.

mudit-01: Hi @steeling thanks for the review I have removed the check

cmd/cli/uninstall_mesh.go
```diff
@@ -179,6 +179,18 @@ func (d *uninstallMeshCmd) run() error {
...
+				for _, ns := range namespaces.Items {
+					removeCmd := &namespaceRemoveCmd{
+						out:       d.out,
+						namespace: ns.Name,
+						meshName:  m.name,
+						clientSet: d.clientSet,
+					}
+
+					_ = removeCmd.run()
+
```
jaellio, line 192: To resolve this lint error we can remove the newline on  line192.

mudit-01: Thanks, @jaellio for the review.

mudit-01:
Thanks, @shalier there seems to be a problem with a test after the commit. How can we rerun it later?

cmd/cli/uninstall_mesh.go
```diff
@@ -171,6 +171,17 @@ func (d *uninstallMeshCmd) run() error {
...
+				namespaces, _ := selectNamespacesMonitoredByMesh(m.name, d.clientSet)
+				for _, ns := range namespaces.Items {
+					removeCmd := &namespaceRemoveCmd{
+						out:       d.out,
+						namespace: ns.Name,
+						meshName:  m.name,
+						clientSet: d.clientSet,
+					}
+
+					_ = removeCmd.run()
```
jaellio, line 183: @mudit-01 I see in a previous commit you did an error check on this command. Is there a reason this was removed? I think you were on the right track adding the error! I think it would be good to add a print statement to indicate that the namespace cleanup failed. Thank you!
```suggestion
					if err := removeCmd.run(); err != nil {
					    fmt.Fprintf("Could not cleanup OSM mesh [%s] related labels and annotations for namespace  [%s]....", ...)
					}
```

cmd/cli/uninstall_mesh.go
```diff
@@ -171,6 +171,17 @@ func (d *uninstallMeshCmd) run() error {
 			}
 
 			if err == nil {
+				namespaces, _ := selectNamespacesMonitoredByMesh(m.name, d.clientSet)
```
jaellio, line 174: @mudit-01 Could you add a unit test for the namespace cleanup on mesh uninstall? Thanks!

jaellio (changes requested):
Please see my above comments! Thank you!

cmd/cli/uninstall_mesh.go
mudit-01: Sure @jaellio I will do that thanks for the review

allenlsy (approved):

mudit-01:
Thanks! @allenlsy for the review :)

cmd/cli/uninstall_mesh.go
allenlsy: @mudit-01 we are pending a updated unit test here. 

https://github.com/openservicemesh/osm/blob/d2c37a6279dae347e65144b175fed94d661cb711/cmd/cli/uninstall_mesh_test.go#L631

You can check if the OSM related labels/annotations are removed from existing namespaces after uninstallation.

mudit-01: @jaellio @allenlsy added test for checking labels/annotations.

cmd/cli/uninstall_mesh_test.go
```diff
@@ -268,6 +286,16 @@ func TestUninstallCmd(t *testing.T) {
 
 				if test.userPromptsYes {
 					if test.meshExists {
+						if len(test.nsInMesh) != 0 {
```
allenlsy, line 289: Optional:

Using too many if...else is not preferred. Since removing these labels and annotation applies to the whole uninstall command, we can make the test case more general.

i.e.

```
tests := []struct {
	...
	workloadNamespaces        []string
	...
}

// when setting up test env
for workloadNs := range test.workloadNamespaces {
	// create workload namespace
	nsSpec := createNamespaceSpec(workloadNs, test.meshName, true, false, true)
	_, err = fakeClientSet.CoreV1().Namespaces().Create(context.TODO(), nsSpec, metav1.CreateOptions{})
	assert.Nil(err)
}

// verifying in test case (line 289)
for workloadNs := range test.workloadNamespaces {
	// check if labels and annotations are removed for workloadNs, same as your implementation from L290-L298
}
```

mudit-01: @allenlsy I have implemented the requested change, PTAL

jaellio (approved):

keithmattix (approved):

mudit-01:
Thanks!! @jaellio @keithmattix for the review

PR #4556, opened 2022-02-23
envoy: update image and go-control-plane version

shashankram:
**Description**:
- Updates Envoy image to v1.21.1, which is the latest
  release with security patches applied

- Updates go-control-plane module to v0.10.1. This is
  required to use latest changes to the XDS API, e.g.
  to set HTTP MaxRequestsPerConnection as a part of the
  HttpProtocolOptions.
  Resolves snapshot cache (experimental) compatibility
  with new SDK.

- Removes Envoy cmd line argument for bootstrap version
  which is now the only supported version in Envoy. The
  cmd line flag has been removed in Envoy.

- Accounts for the change in envoy where custom metrics from
   WASM extensions are no longer prefixed with `envoy_`. The metrics as
   they exist in Prometheus are unchanged.

- Update drain interval to 30s from 600s and set drain strategy to
   immediate so that connections are drained sooner when the
   listener is removed.

**Testing done**:
CI

**Affected area**:

### DIFF
--- a/charts/osm/values.yaml
+++ b/charts/osm/values.yaml
@@ -31,10 +31,10 @@ osm:
 
   # -- `osm-controller` image pull secret
   imagePullSecrets: []
-  # -- Envoy sidecar image for Linux workloads (v1.19.1)
-  sidecarImage: envoyproxy/envoy-alpine@sha256:6502a637c6c5fba4d03d0672d878d12da4bcc7a0d0fb3f1d506982dde0039abd
-  # -- Envoy sidecar image for Windows workloads (v1.19.1)
-  sidecarWindowsImage: envoyproxy/envoy-windows@sha256:c904fda95891ebbccb9b1f24c1a9482c8d01cbca215dd081fc8c8db36db85f85
+  # -- Envoy sidecar image for Linux workloads (v1.21.1)
+  sidecarImage: envoyproxy/envoy-alpine:v1.21.1@sha256:c2c0b7c136f9daf09c36ef550635eaf5db26465529a1354f725f11390e530879
+  # -- Envoy sidecar image for Windows workloads (v1.21.1)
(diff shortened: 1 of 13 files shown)


charts/osm/values.yaml
```diff
@@ -31,10 +31,10 @@ osm:
 
   # -- `osm-controller` image pull secret
   imagePullSecrets: []
-  # -- Envoy sidecar image for Linux workloads (v1.19.1)
-  sidecarImage: envoyproxy/envoy-alpine@sha256:6502a637c6c5fba4d03d0672d878d12da4bcc7a0d0fb3f1d506982dde0039abd
-  # -- Envoy sidecar image for Windows workloads (v1.19.1)
-  sidecarWindowsImage: envoyproxy/envoy-windows@sha256:c904fda95891ebbccb9b1f24c1a9482c8d01cbca215dd081fc8c8db36db85f85
+  # -- Envoy sidecar image for Linux workloads (v1.21.1)
```
nojnhuh, line 34: We might as well remove the version from these comments if it's included in the value itself.

nojnhuh (approved):
LGTM, please also update the docs: https://github.com/openservicemesh/osm-docs/blob/97baa0d481928a5607c2dcb3dfb453e93cbf9acd/config.toml#L87

jaellio (approved):

charts/osm/values.yaml
shashankram: > LGTM, please also update the docs: https://github.com/openservicemesh/osm-docs/blob/97baa0d481928a5607c2dcb3dfb453e93cbf9acd/config.toml#L87

> LGTM, please also update the docs: https://github.com/openservicemesh/osm-docs/blob/97baa0d481928a5607c2dcb3dfb453e93cbf9acd/config.toml#L87

Yes, I have a change that I will push to osm-docs once this merges.

PR #4557, opened 2022-02-24
Create release process documentation

trstringer:
**Description**:

It is important that as an open source software project we have clarity
and transparency on how releases are handled. This includes the
frequency as well as how the community can request for changes. This
also goes into depth about the lifecycle of a request for change and how
it makes it into a release.

**Affected area**:
| Documentation              | [X] |

3. Has documentation corresponding to this change been updated in the [osm-docs](https://github.com/openservicemesh/osm-docs) repo (if applicable)? Not applicable.

### DIFF
--- a/docs/release_process.md
+++ b/docs/release_process.md
@@ -1,0 +1,35 @@
+# Release process
+
+## Frequency
+
+Open Service Mesh will target a release every quarter. These releases will be targeted in the first half of February, May, August, and November. These target release months are subject to change if a release is pushed. In other words, if a release misses February and is pushed out in March then the next release will target June instead of May.
+
+## Process for requesting a change
+
+If you would like to see a new feature added to Open Service Mesh, please [open an issue](https://github.com/openservicemesh/osm/issues). This issue will be discussed and, if approved, prioritized in the backlog.
+
+## Creating a new release
+
+When there is a new release being planned and developed, a [milestone](https://github.com/openservicemesh/osm/milestones) will be created. This new milestone will have issues added to it that will be prioritized from the [vFuture milestone](https://github.com/openservicemesh/osm/milestone/32). Once the release milestone is complete, a new release of Open Service Mesh will be created and this milestone will be closed. At that moment, the next release milestone will be created and items from vFuture will be added.
+
+## Larger projects
+
+For larger projects where development will likely span multiple releases we should have a separate feature milestone created. This feature milestone should have a projected end date. Once a release milestone is opened up that has an end date that includes the feature milestone end date, this feature should be considered to be included in the next release.
+
+## Backlog handling
+
+The progression of approved work items will go through this lifecycle:
+
+- Added to the [vFuture milestone](https://github.com/openservicemesh/osm/milestone/32) when approved
+- Added to the [next release milestone](https://github.com/openservicemesh/osm/milestones) when set for the next release
+- Added to the [Todo](https://github.com/orgs/openservicemesh/projects/1) column when prioritized for "next work"
+- Moved to the **In progress** column when being actively worked on
+- Moved to the **Done** column when completed
+
+At the end of this lifecycle it will be included in the next release of OSM.
+
+## Release Candidates
+
+Release candidates (RCs) should be created before each significant release so final testing can be performed. RCs are tagged as `vX.Y.Z-rc.W`. After the following steps have been performed to publish the RC, perform any final testing with the published release artifacts for about one week.
+
+If issues are found, submit patches to the RC's release branch and create a new RC with the tag `vX.Y.Z-rc.W+1`. Apply those same patches to the `main` branch. Repeat until the release is suitably stable.
--- a/docs/release_guide.md
+++ b/docs/release_guide.md
@@ -10,12 +10,6 @@ This guide describes the process to create a GitHub Release for this project.
 
 **Note**: These steps assume that all OSM components are being released together, including the CLI, container images, and Helm chart, all with the same version.
 
-## Release Candidates
-
-Release candidates (RCs) should be created before each significant release so final testing can be performed. RCs are tagged as `vX.Y.Z-rc.W`. After the following steps have been performed to publish the RC, perform any final testing with the published release artifacts for about one week.
-
-If issues are found, submit patches to the RC's release branch and create a new RC with the tag `vX.Y.Z-rc.W+1`. Apply those same patches to the `main` branch. Repeat until the release is suitably stable.
-
 Once an RC has been found to be stable, cut a release tagged `vX.Y.Z` using the following steps.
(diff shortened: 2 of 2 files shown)


docs/release_process.md
```diff
@@ -0,0 +1,25 @@
...
+
+## Process for requesting a change
+
+If you would like to see a new feature added to Open Service Mesh, please [open an issue](https://github.com/openservicemesh/osm/issues). This issue will be discussed and, if approved, prioritized in the backlog.
+
+## Creating a new release
+
+When there is a new release being planned and developed, a [milestone](https://github.com/openservicemesh/osm/milestones) will be created. This new milestone will have issues added to it that will be prioritized from the [vFuture milestone](https://github.com/openservicemesh/osm/milestone/32). Once the release milestone is complete, a new release of Open Service Mesh will be created and this milestone will be closed. At that moment, the next release milestone will be created and items from vFuture will be added.
+
+## Backlog handling
```
steeling, line 15: How are longer projects handled, ie: cert rotation? These are not yet designed, not yet estimated projects. Even when they are estimated larger projects often tend to slip past deadlines. Do we assign these to a release after they are complete? When they are near completion?

One thing I've seen work is to use an additional label, "release-blocker",  and then we can be more aggressive about adding items to releases.

docs/release_process.md
```diff
@@ -0,0 +1,25 @@
...
+
+Open Service Mesh will target a release every quarter. These releases will be targeted in the first half of February, May, August, and November. These target release months are subject to change if a release is pushed. In other words, if a release misses February and is pushed out in March then the next release will target June instead of May.
+
+## Process for requesting a change
+
+If you would like to see a new feature added to Open Service Mesh, please [open an issue](https://github.com/openservicemesh/osm/issues). This issue will be discussed and, if approved, prioritized in the backlog.
+
+## Creating a new release
+
+When there is a new release being planned and developed, a [milestone](https://github.com/openservicemesh/osm/milestones) will be created. This new milestone will have issues added to it that will be prioritized from the [vFuture milestone](https://github.com/openservicemesh/osm/milestone/32). Once the release milestone is complete, a new release of Open Service Mesh will be created and this milestone will be closed. At that moment, the next release milestone will be created and items from vFuture will be added.
```
steeling, line 13: I think it makes sense to have at least 2-4 future release milestones in existence at once, as opposed to one at a time. It provides more clarity for users, and allows for better planning.

steeling, line 13: I'd be interested in discussing this prioritization process. Do we have any documentation around that?

docs/release_process.md
trstringer: That's a good question and something I should include in this documentation. For longer projects they should likely have their own milestone that isn't tied to a release. They should be worked in parallel to the release(s) during that time but considered for "next release" when the feature milestone completion date lines up with a release milestone completion date.

steeling: Otherwise folks will be having tasks in the  *in progress* column that are not attributed to a release.

docs/release_process.md
trstringer: What would be the differences between the multiple future release milestones? What's the benefit in splitting up approved work that isn't prioritized yet?

trstringer: The prioritization process is a combination of roadmap influence as well as roundtable discussion with the maintainers and the community.

steeling: it allows us to prioritize and plan past a 3 month horizon, while giving users some clarity on what's to come. 

I think it makes the most sense when we add larger work items (from comment above), to a release, but maybe with a "release-candidate" label, meaning its not a blocker.

steeling: if any of those conversations/roadmap are public, it may be worth including links, dates, etc in this doc

docs/release_process.md
```diff
@@ -0,0 +1,25 @@
+# Release process
+
+## Frequency
+
+Open Service Mesh will target a release every quarter. These releases will be targeted in the first half of February, May, August, and November. These target release months are subject to change if a release is pushed. In other words, if a release misses February and is pushed out in March then the next release will target June instead of May.
```
nojnhuh, line 5: The release guide has a section regarding release candidates that I think would fit better in this doc: https://github.com/openservicemesh/osm/blob/edd9ea76246f8f9f673de2db30be1eadef232a18/docs/release_guide.md?plain=1#L13-L19

Then we'd have a clear line between **when** to cut a release (this doc) vs. **how** (the release guide).

trstringer: Just be clear, @nojnhuh, you're recommending that we move that `Release Candidates` section from `release_guide.md` to `release_process.md` (the new file in this PR)?

nojnhuh: Yeah. We could of course change the language to make it more consistent with the rest of this doc, but I think that information would fit better here than in the release guide.

trstringer: Agreed. Done!

docs/release_process.md
trstringer: I think that extra granularity doesn't add enough value to compensate the additional overhead of maintaining that. But perhaps in the future if we feel like that is a gap then we should modify accordingly. I think it is too much on the first pass of this without really having a clear need of that. But I'll defer to other maintainers of the project. // @shashankram @snehachhabria @nojnhuh @jaellio

shashankram: +1 @trstringer.

shashankram (approved):
@trstringer thanks for drafting this!

docs/release_process.md
nojnhuh: I don't think we need to always have a set number of upcoming version milestones. We can always have at least one and if we think it would be worth it to have more we can create them as needed. I suspect committing to always having 3 or 4 would be a pain, but giving ourselves a chance to do some extra planning further out wouldn't hurt.

docs/release_process.md
```diff
@@ -0,0 +1,31 @@
...
+
+- Added to the [vFuture milestone](https://github.com/openservicemesh/osm/milestone/32) when approved
+- Added to the [next release milestone](https://github.com/openservicemesh/osm/milestones) when set for the next release
+- Added to the [Todo](https://github.com/orgs/openservicemesh/projects/1) column when prioritized for "next work"
+- Moved to the **In progress** column when being actively worked on
+- Moved to the **Done** column when completed
+
+At the end of this lifecycle it will be included in the next release of OSM.
+
+## Release Candidates
```
nojnhuh, line 27: @trstringer Could you also remove this from the release guide? I don't think we need it in both places.

trstringer: Good catch, that was an oversight. Removed from the release guide.

snehachhabria (approved):

PR #4558, opened 2022-02-24
fix(healthprobes): add support for TCPSocket probes

### LINKED ISSUE #4434: tcpSocket Kubernetes probes can't fail
**Bug description**: When specifying an unopened port for a `tcpSocket` probe, the probe passes when it should fail.

- Sidecar Injection          [X]

**Expected behavior**: A `tcpSocket` probe that fails when the Pod is outside the mesh should also fail when the Pod is inside the mesh.

First, add the following probe to docs/example/manifests/samples/httpbin/httpbin.yaml:
```
readinessProbe:
  tcpSocket:
    port: 14002 # not the port the server is running on
```

Then,
```
osm install
osm namespace add default
kubectl apply -f docs/example/manifests/samples/httpbin
```

At this point the pod becomes ready when it shouldn't. To verify that the same probe fails as expected outside the mesh:
```
osm namespace remove default
kubectl rollout restart deployment httpbin
kubectl describe po <new pod> # should show an event that the readiness probe is failing
```

**How was OSM installed?**: `osm install`

I dug around to see if I could figure out why this is happening. The first thing I noticed is that the sidecar injector modifies the port for `tcpSocket` probes like it does for HTTP probes and programs the following listener:
```json
{
  "listener": {
    "@type": "type.googleapis.com/envoy.config.listener.v3.Listener",
    "name": "readiness_listener",
    "address": {
      "socket_address": {
        "address": "0.0.0.0",
        "port_value": 15902
      }
    },
    "filter_chains": [
      {
        "filters": [
          {
            "name": "envoy.filters.network.tcp_proxy",
            "typed_config": {
              "@type": "type.googleapis.com/envoy.extensions.filters.network.tcp_proxy.v3.TcpProxy",
              "stat_prefix": "health_probes",
              "cluster": "readiness_cluster",
              "access_log": [...]
            }
          }
        ]
      }
    ]
  }
}
```

And the cluster it references looks like:
```json
{
  "cluster": {
    "@type": "type.googleapis.com/envoy.config.cluster.v3.Cluster",
    "name": "readiness_cluster",
    "type": "STATIC",
    "load_assignment": {
      "cluster_name": "readiness_cluster",
      "endpoints": [
        {
          "lb_endpoints": [
            {
              "endpoint": {
                "address": {
                  "socket_address": {
                    "address": "127.0.0.1",
                    "port_value": 14002
                  }
                }
              }
            }
          ]
        }
      ]
    }
  }
}
```

Could this listener be causing all probes to port 15902 to succeed whether or not the port it proxies (14002) is actually open?

I tried modifying the injector to keep the same port (and skip adding the cluster and listener above), but I see the same behavior:
```diff
diff --git a/pkg/injector/health_probes.go b/pkg/injector/health_probes.go
index 04af3a6d..8c4df074 100644
--- a/pkg/injector/health_probes.go
+++ b/pkg/injector/health_probes.go
@@ -79,8 +79,6 @@ func rewriteProbe(probe *corev1.Probe, probeType, path string, port int32, conta
 			probe.HTTPGet.Path = path
 			newPath = probe.HTTPGet.Path
 		}
-	} else if probe.TCPSocket != nil {
-		definedPort = &probe.TCPSocket.Port
 	} else {
 		return nil
 	}
```

For reference, here's how the kubelet does `tcpSocket` probes: https://github.com/kubernetes/kubernetes/blob/v1.22.1/pkg/probe/tcp/tcp.go#L50

**Bug report archive**:

**Environment**:
```
CLI Version: version.Info{Version:"v1.0.0-rc.3", GitCommit:"edb1d4531f350909d3eca77edee61a9f02056bd8", BuildDate:"2022-01-10-19:16"}
Mesh [osm] Version: version.Info{Version:"v1.0.0-rc.3", GitCommit:"abc559bb2282319bb2005231e3d5556560a50fa2", BuildDate:"2022-01-10-17:46"}
```
```
Client Version: version.Info{Major:"1", Minor:"23", GitVersion:"v1.23.1", GitCommit:"86ec240af8cbd1b60bcc4c03c20da9b98005b92e", GitTreeState:"clean", BuildDate:"2021-12-16T11:33:37Z", GoVersion:"go1.17.5", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"20", GitVersion:"v1.20.9", GitCommit:"a5e4de7e277a707bd28d448bd75de58b4f1cdc22", GitTreeState:"clean", BuildDate:"2021-11-16T01:09:55Z", GoVersion:"go1.15.14", Compiler:"gc", Platform:"linux/amd64"}
```
- Size of cluster (number of worker nodes in the cluster): 6
- Others:

jaellio:
**Description**:

Currently, OSM rewrites the port value on TCPSocket probes
to the osm reserved port for the respective probe type. An envoy
listener and cluster are configured at the rewritten port.
When performing the health probe, the kubelet will attempt to
open a socket at the specified port. In this case, the specified
port is open on the envoy container. Therefore, the kubelet will
consider the container healthy and the probe will succeed even
though it never checked if a connection could be established at
the original port. False-positive TCP connections are a known
limitation of Envoy.

This solution transforms TCPSocket probes on injection into
HTTPGet probes. Rewrites the probe's port to a new osm reserved
port (`15904`) for the osm-healthcheck HTTP server and updates 
the path to /osm-healthcheck. A header `Original-Tcp-Port` is also
added to the new HTTPGet probe and set to the port specified on
the original TCPSocket probe.

osm-healthcheck will handle HTTP health probe requests from
kubelet, get the original TCP port from the request's headers
and attempt to open a socket at the specified port. The response
status code corresponds to the success or failure of establishing
a connection.

Resolves #4434

**Original TCPSocket probe**
```yaml
livenessProbe:
      tcpSocket:
        port: 8080
      initialDelaySeconds: 15
      periodSeconds: 20
```

**Rewritten probe**
```yaml
 readinessProbe:
      httpGet:
        httpHeaders:
        - name: Original-Tcp-Port
          value: "8080"
        path: /osm-healthcheck
        port: 15904
        scheme: HTTP
      initialDelaySeconds: 15
      periodSeconds: 20
```

**Testing done**:
- added a test to the health probe e2e tests to check that the TCPSocket health probes
  fail as expected when misconfigured (port specified is not open on the container)
- updated unit tests for the osm-healthcheck handler 
- manually verified the osm-healthcheck container is only added when a TCPSocket probe
  is defined
- manually verified the original TCPSocket probes fail and succeed as expected
- manually verified TCPSockt -> HTTPGet handles named ports

**Affected area**:
| Sidecar Injection          | [x] |

    I explored multiple solutions which are documented in the linked issue. Until there is support for TCP connections in WASM, an osm healthcheck service is the best approach. Istio took a similar approach to support TCPSocket probes.
    -   Did you notify the maintainers and provide attribution? No

3. Is this a breaking change? No

4. Has documentation corresponding to this change been updated in the [osm-docs](https://github.com/openservicemesh/osm-docs) repo (if applicable)? No, but it needs to be.


dockerfiles/Dockerfile.osm-healthcheck
```diff
@@ -0,0 +1,14 @@
+ARG GO_VERSION
```
steeling, line 1: I think we should discuss the option of including the healthcheck within the sidecar container. This lets this not require its own container resources. This could be expensive for a large cluster.

* A small bonus is that users don't have to see yet another container
* Having a single sidecar for intercepting traffic makes logical sense. It's what I would expect as a user.
* It provides a platform for us to package more things in the sidecar (static configs, wasm filters, startup scripts, etc). Although I don't think this particular point should be a used in a debate on whether or not to use a single container.
* I see no concern of performance impact here. The code is tiny, will have a tiny footprint, and any erratic probe that hits this too hard would have had a similar performance impact to envoy itself.
* Other sidecars take this approach

steeling (changes requested):

snehachhabria (approved):

shashankram (approved):

tests/e2e/e2e_health_probe_test.go
```diff
@@ -253,5 +262,27 @@ server {
...
+
+			_, err := Td.Client.CoreV1().Secrets(ns).Create(context.TODO(), tls, metav1.CreateOptions{})
+			Expect(err).NotTo(HaveOccurred())
+			_, err = Td.Client.CoreV1().ConfigMaps(ns).Create(context.TODO(), conf, metav1.CreateOptions{})
+			Expect(err).NotTo(HaveOccurred())
+			_, err = Td.Client.CoreV1().Pods(ns).Create(context.TODO(), pod, metav1.CreateOptions{})
+			Expect(err).NotTo(HaveOccurred())
+
+			// Expect it to not be ready
+			Expect(Td.WaitForPodsRunningReady(ns, 60*time.Second, 1, nil)).To(MatchError("not all pods were Running & Ready in NS healthprobe after 1m0s"))
```
nojnhuh, line 284: Nit: define the timeout once and use it to parameterize the expected error message. Or omit the duration from the message if the expected error doesn't have to match the full error.

pkg/injector/healthcheck_container.go
```diff
@@ -0,0 +1,23 @@
+package injector
+
+import (
+	"os"
+
+	corev1 "k8s.io/api/core/v1"
+)
+
+func getHealthcheckContainerSpec(containerName string) corev1.Container {
```
nojnhuh, line 9: Nit: It looks like this is only called once where `constants.HealthcheckContainerName` is passed, so maybe it makes more sense to inline that value in this function.

dockerfiles/Dockerfile.osm-healthcheck
steeling: FWIW at injection time we can supply an arg to not run this. I'm imagining the new entrypoint is a small bash script that starts this up (conditionally) along with envoy and any provided args.

With appropriate testing I see the risk as relatively minimal. I'd be in favor of trying to determine what a desired end goal is, and once we have that decision, then circle back to risk/effort to get there. There may be larger caveats in terms of upgrading sidecars (currently users can override their sidecar version in the helm chart). And then we can weight up which path to take to move this forward. Or at least create an issue to circle back to this. Thoughts?

cc @jaellio, @shashankram, @snehachhabria, @trstringer

shashankram: I am in favor of keeping the healthcheck service a separate container. It keeps things simple, modular, and allows setting individual container resource limits for the data plane (envoy) and control plane (health check service). Data plane vs control plane separation is important. This approach has worked well so far so I suggest not tweaking this unless there is a strong motivation to do so, which I do not see with this change.

steeling: FWIW this is moving functionality currently in Envoy (forwarding traffic -- even healthchecks) to a separate container, so I would call this part of the data plane as well. On the resource consumption, many managed K8s services have min resources. If I have say a large zookeeper cluster this could become expensive

shashankram: We can disagree on the terminology of "control plane traffic", but to me, data traffic refers to application traffic while control plane traffic refers to anything required to ensure the healthy functioning of the application itself (e.g. probes). I also disagree that this is moving functionality in Envoy to a separate container because this fix is specifically targeting a scenario Envoy cannot address for TCP ports. Given this service is lightweight, I see no issues keeping this as is. If resource consumption is a concern, add resource limits to the health check container. 

@jaellio Feel free to merge this fix after addressing the pending comments. Revamping the existing architecture to augment this with the Envoy container is a much broader discussion that we should not be blocked on.

cmd/osm-healthcheck/osm-healthcheck.go
```diff
@@ -0,0 +1,84 @@
...
+	"github.com/openservicemesh/osm/pkg/constants"
+	"github.com/openservicemesh/osm/pkg/logger"
+	"github.com/openservicemesh/osm/pkg/signals"
+)
+
+var log = logger.New("osm-healthcheck/main")
+
+func main() {
+	stop := signals.RegisterExitHandlers()
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
```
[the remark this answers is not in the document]
jaellio: The context is only being passed to the server shutdown command. The timeout specifies the maximum time to wait for all connections to return idle. If shutdown exceeds the timeout, then it will return the context's error.

pkg/injector/healthcheck_container.go
```diff
@@ -0,0 +1,23 @@
...
+	return corev1.Container{
+		Name:            containerName,
+		Image:           os.Getenv("OSM_DEFAULT_HEALTHCHECK_CONTAINER_IMAGE"),
+		ImagePullPolicy: corev1.PullIfNotPresent,
```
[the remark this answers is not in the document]
jaellio: Renamed the `init-container-pull-policy` flag to `osm-container-pull-policy` to be used by the init and healthcheck container.

cmd/osm-healthcheck/osm-healthcheck.go
```diff
@@ -0,0 +1,84 @@
...
+	"time"
+
+	"github.com/openservicemesh/osm/pkg/constants"
+	"github.com/openservicemesh/osm/pkg/logger"
+	"github.com/openservicemesh/osm/pkg/signals"
+)
+
+var log = logger.New("osm-healthcheck/main")
+
+func main() {
```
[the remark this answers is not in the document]
jaellio: The osm-healthcheck log level is set to the same log level as the osm-injector.

cmd/osm-healthcheck/osm-healthcheck_test.go
```diff
@@ -0,0 +1,39 @@
...
+
+import (
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	tassert "github.com/stretchr/testify/assert"
+)
+
+func TestGetHealthcheckHander(t *testing.T) {
```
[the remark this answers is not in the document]
jaellio: I've updated the tests to include a success case and a failed connection case.

nojnhuh (approved):

dockerfiles/Dockerfile.osm-healthcheck
steeling: Agreed on not blocking this issue, but I think this requires further discussion. Kicked off #4563. I mentioned in that bug but I think it's worth pointing out again that some platforms have minimum resource limits that are non-trivial, so the cost here can still be high.

PR #4571, opened 2022-03-03
Upgrade docker distribution to v2.8.0

trstringer:
**Description**:

Resolves [CWE-843](https://cwe.mitre.org/data/definitions/843.html).

**Affected area**:
| Security                   | [X] |

3. Has documentation corresponding to this change been updated in the [osm-docs](https://github.com/openservicemesh/osm-docs) repo (if applicable)? N/A.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -293,8 +293,3 @@ require (
 	sigs.k8s.io/structured-merge-diff/v4 v4.1.2 // indirect
 	sigs.k8s.io/yaml v1.2.0 // indirect
 )
-
-replace (
-	github.com/docker/distribution => github.com/docker/distribution v0.0.0-20191216044856-a8371794149d
-	github.com/docker/docker => github.com/moby/moby v17.12.0-ce-rc1.0.20200618181300-9dc6525e6118+incompatible
-)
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -102,7 +102,7 @@ require (
 	github.com/davecgh/go-spew v1.1.1 // indirect
 	github.com/denis-tingajkin/go-header v0.3.1 // indirect
 	github.com/docker/cli v20.10.7+incompatible // indirect
-	github.com/docker/distribution v2.7.1+incompatible // indirect
+	github.com/docker/distribution v2.8.0+incompatible // indirect
```
nojnhuh, line 105: I imagine because of the `replace` directive this is effectively a no-op. It looks like the entire `replace` block can be removed now though since Helm doesn't need it anymore. Then this change would actually bump the version we build with.

More context: https://github.com/openservicemesh/osm/pull/1306#pullrequestreview-460228804

trstringer: We should be able to remove that `replace` directive then right?

nojnhuh: Yes

trstringer: Sorry just reread your comment. Ok, sounds good. I'll remove the `replace` block and retest.

nojnhuh (approved):

shashankram (approved):

PR #4580, opened 2022-03-09
feat(certificates) begin to abstract the cert manager patterns

steeling:
This change pulls out the shared logic between the cert providers into
a new manager struct. This is done in pieces, to avoid a single large PR
with O(1000)'s of LOC changed. The goal is to reduce all of the existing
"providers" to clients that only deal with certificates (no caching, msg broker,
logging, or timing).

Once all of these are gone, we will remove the certificate.Manager interface
and promote the certificate.manager struct to the exported certificate.Manager
(struct).

Eventually the rotor functionality will also be on the certificate.Manager. It looks like this 
was done as an artifact of the current split. I had to move rotor.ShouldRotate to certificate
within this PR as opposed to later to avoid circular dependencies.

No

2. Is this a breaking change?

No

Part of #4533


pkg/certificate/certificate.go
```diff
@@ -35,3 +47,15 @@ func (c *Certificate) GetPrivateKey() pem.PrivateKey {
...
 }
+
+// ShouldRotate determines whether a certificate should be rotated.
+func (c *Certificate) ShouldRotate() bool {
+	// The certificate is going to expire at a timestamp T
+	// We want to renew earlier. How much earlier is defined in renewBeforeCertExpires.
+	// We add a few seconds noise to the early renew period so that certificates that may have been
+	// created at the same time are not renewed at the exact same time.
+
+	intNoise := rand.Intn(maxNoiseSeconds-minNoiseSeconds) + minNoiseSeconds /* #nosec G404 */
```
shashankram, line 58: For style consistency
```suggestion
	intNoise := rand.Intn(maxNoiseSeconds-minNoiseSeconds) + minNoiseSeconds // #nosec G404 
```

pkg/certificate/manager.go
```diff
@@ -0,0 +1,146 @@
...
+	"github.com/pkg/errors"
+	"github.com/rs/zerolog/log"
+
+	"github.com/openservicemesh/osm/pkg/announcements"
+	"github.com/openservicemesh/osm/pkg/k8s/events"
+	"github.com/openservicemesh/osm/pkg/messaging"
+)
+
+var errNoIssuingCA = errors.New("no issuing CA")
+var errCertNotFound = errors.New("certificate not found")
```
shashankram, line 15:
```suggestion
var (
	errNoIssuingCA = errors.New("no issuing CA")
	errCertNotFound = errors.New("certificate not found")
)
```

pkg/certificate/manager.go
```diff
@@ -0,0 +1,146 @@
...
+
+	m := &manager{
+		// The root certificate signing all newly issued certificates
+		ca:                          ca,
+		client:                      client,
+		serviceCertValidityDuration: serviceCertValidityDuration,
+		msgBroker:                   msgBroker,
+	}
+
+	// TODO(steeling) start the cert rotation here.
```
shashankram, line 35: If this is a part of a refactoring issue, could you please reference the issue number, e.g. `TODO(#xyz): ...`

pkg/certificate/manager.go
```diff
@@ -0,0 +1,146 @@
...
+
+	return m, nil
+}
+
+func (m *manager) deleteFromCache(cn CommonName) {
+	m.cache.Delete(cn)
+}
```
shashankram, line 42: Why is this wrapper necessary? Because this is un-exported, *manager implementation can use `m.cache.Delete()` directly.

pkg/certificate/manager.go
```diff
@@ -0,0 +1,146 @@
...
+
+func (m *manager) getFromCache(cn CommonName) *Certificate {
+	if certInterface, exists := m.cache.Load(cn); exists {
```
shashankram, line 45: Avoid unnecessary nesting by changing this to:
```
certInterface, exists := m.cache.Load(cn)
if !exists {
    return nil
}
...
return cert
```

pkg/certificate/manager.go
```diff
@@ -0,0 +1,146 @@
...
+	}
+
+	cert, err := m.client.IssueCertificate(cn, validityPeriod)
+	if err != nil {
+		return cert, err
+	}
+
+	m.cache.Store(cn, cert)
+
+	log.Trace().Msgf("It took %+v to issue certificate with SerialNumber=%s", time.Since(start), cert.GetSerialNumber())
```
shashankram, line 72:
```suggestion
	log.Trace().Msgf("It took %s to issue certificate with SerialNumber=%s", time.Since(start), cert.GetSerialNumber())
```

pkg/certificate/providers/certmanager/certificate_manager.go
```diff
@@ -149,19 +41,12 @@ func (cm *CertManager) certificateFromCertificateRequest(cr *cmapi.CertificateRe
...
+// IssueCertificate will request a new signed certificate from the configured cert-manager issuer.
+func (cm *CertManager) IssueCertificate(cn certificate.CommonName, validityPeriod time.Duration) (*certificate.Certificate, error) {
 	duration := &metav1.Duration{
 		Duration: validityPeriod,
 	}
 
-	// Key bit size should remain static during the lifetime of the CertManager. In the event that this
-	// is a zero value, we make the call to config to get the setting and then cache it for future
-	// certificate operations.
-	if cm.keySize == 0 {
```
shashankram, line 162: Where is this validation being performed now?

pkg/certificate/providers/config.go
```diff
@@ -338,14 +342,19 @@ func (c *Config) getCertManagerOSMCertificateManager(options CertManagerOptions)
...
+		return nil, nil, errors.Errorf("Error instantiating Jetstack cert-manager client: %+v", err)
 	}
 
-	return certmanagerCertManager, certmanagerCertManager, nil
+	certManager, err := certificate.NewManager(rootCert, cmClient, c.cfg.GetServiceCertValidityPeriod(), c.msgBroker)
+	if err != nil {
+		return nil, nil, errors.Errorf("error instantiating osm certificate.Manager for Jetstack cert-manager : %+v", err)
+	}
+
+	// TODO(steeling): push this into the certificate.manager object.
```
shashankram, line 356:
```suggestion
	// TODO(#<issue>): push this into the certificate.manager object.
```

shashankram (commented):
Overall the refactor makes sense to me. Since this is refactoring sensitive parts of the code pertaining to certificates, could you kindly describe in your commit && PR description the testing that was done? Since we use 24h expiration for short-lived certs by default and we do not have an e2e tests to verify cert rotation, so manually testing cert rotation should suffice. You can do this by changing the `serviceCertValidityDuration` install option to small duration such as `2m`.

pkg/certificate/providers/certmanager/certificate_manager.go
```diff
@@ -242,45 +127,29 @@ func (cm *CertManager) issue(cn certificate.CommonName, validityPeriod time.Dura
...
 
-	cm.cacheLock.Lock()
-	defer cm.cacheLock.Unlock()
-	cm.cache[cert.GetCommonName()] = cert
-
 	return cert, nil
 }
 
-// NewCertManager will construct a new certificate.Certificate implemented
+// New will construct a new certificate.Certificate implemented
```
nojnhuh, line 133: To match the function signature:
```suggestion
// New will construct a new CertManager implemented
```

nojnhuh (commented):
These changes absolutely make sense, looking forward to seeing the rest.

pkg/certificate/manager.go
steeling: FYI all of these function were just copy-pasted, I didn't write anything new in this file. Happy to make the changes though

shashankram: No problem, feel free to update it later if you agree.

pkg/certificate/manager.go
```diff
@@ -0,0 +1,146 @@
...
+}
+
+func (m *manager) getFromCache(cn CommonName) *Certificate {
+	if certInterface, exists := m.cache.Load(cn); exists {
+		cert := certInterface.(*Certificate)
+		log.Trace().Msgf("Certificate found in cache SerialNumber=%s", cert.GetSerialNumber())
+		if cert.ShouldRotate() {
+			log.Trace().Msgf("Certificate found in cache but has expired SerialNumber=%s", cert.GetSerialNumber())
+			return nil
+		}
```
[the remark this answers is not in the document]
jaellio: `getFromCache` is called in `IssueCertificate`. I thought `ShouldRotate()` was used instead of the cert's expiration directly to avoid returning a certificate that would need to be rotated soon after being issued.

pkg/certificate/certificate.go
steeling: done!

pkg/certificate/manager.go
steeling: done!

pkg/certificate/providers/certmanager/certificate_manager.go
steeling: It was essentially an erroneous validation in the first place. Since it was always initially set from `cm.cfg.GetCertKeyBitSize()`, and if that's zero it would reset to cm.cfg.GetCertKeyBitSize(), which still may be zero. Eventually this would result in an error regardless of this check.

Ideally we add this to either a) validating webhook, b) give a hard error on instantiation, or c) both a,b.

for now i re-added back to the instantiation of the client

pkg/certificate/manager.go
```diff
@@ -0,0 +1,146 @@
...
+		return cert
+	}
+	return nil
+}
+
+// IssueCertificate implements Manager and returns a newly issued
```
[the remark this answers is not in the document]
steeling: yup, fixed!

pkg/certificate/manager.go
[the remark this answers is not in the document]
steeling: I agree with Jon here, I think returning a soon-to-expired cert is better than no cert.

Particularly, it allows us to bump up the ShouldRotate time to say 1/3 of total lifetime of cert, without affecting this.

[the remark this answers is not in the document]
steeling: Actually I spoke to soon. It may make sense to trigger a re-issue on certificates if it is expired, even on a GetCertificate call. Will dig through this a bit more, but since this PR isn't changing existing functionality I propose we don't block. Thoughts?

Will keep this open for further discussion

pkg/certificate/manager.go
```diff
@@ -0,0 +1,146 @@
...
+		return nil, err
+	}
+
+	m.cache.Store(cn, newCert)
+
+	m.msgBroker.GetCertPubSub().Pub(events.PubSubMessage{
+		Kind:   announcements.CertificateRotated,
+		NewObj: newCert,
+		OldObj: oldCert,
+	}, announcements.CertificateRotated.String())
```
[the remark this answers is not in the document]
steeling: Since we're going down to a single place where we do this, I'm think we can be less concerned about introducing bugs than the current state where there was one instance per cert provider.

I think you highlight a good refactoring opportunity for the msgbroker though.

It may make more sense to provide more targeted methods, ie: `PublishCertRotation`, or provide generic methods that type switch based on the incoming type. Haven't looked at this code too much though

pkg/certificate/manager.go
```diff
@@ -0,0 +1,146 @@
...
+		certs = append(certs, certInterface.(*Certificate))
+		return true // continue the iteration
+	})
+	return certs, nil
+}
+
+// GetRootCertificate returns the root
+func (m *manager) GetRootCertificate() (*Certificate, error) {
+	return m.ca, nil
+}
```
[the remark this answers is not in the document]
steeling: added a todo

pkg/certificate/providers/certmanager/certificate_manager.go
steeling: Changed the comment a bit. In general a struct implementing an interface doesn't need to know what interface it is implementing. This fit's the go paradigm: "define interfaces where they are consumed not where they are implemented"

pkg/certificate/providers/certmanager/certificate_manager.go
shashankram: This is not an erroneous validation. The reason it was done this way was to allow instantiating the certificate provider instances outside the osm-controller (e.g. from the CLI) with a specified key size. There was a use case to be able to issue a certificate on-demand using OSM CLI for ingress at some point which would make use of the CLI to issue a cert.

steeling:
> Overall the refactor makes sense to me. Since this is refactoring sensitive parts of the code pertaining to certificates, could you kindly describe in your commit && PR description the testing that was done? Since we use 24h expiration for short-lived certs by default and we do not have an e2e tests to verify cert rotation, so manually testing cert rotation should suffice. You can do this by changing the `serviceCertValidityDuration` install option to small duration such as `2m`.

Good call! Done both by adding an automated test, and verifying manually through the demo with cert-manager deployed, and the settings as you suggested. Certs are successfully being rotated and present in envoy configs

shashankram (approved):

pkg/certificate/providers/certmanager/certificate_manager_test.go
```diff
@@ -285,3 +119,25 @@ func TestCertificateFromCertificateRequest(t *testing.T) {
...
+	fakeClient := cmfakeclient.NewSimpleClientset()
+	_, err := New(
+		&certificate.Certificate{},
+		fakeClient,
+		"osm-system",
+		cmmeta.ObjectReference{Name: "osm-ca"},
+		0,
+	)
+
+	assert.Error(err, "expected error from key size of zero, got")
```
nojnhuh, line 134: Nit only because this message is a little awkward when it does fail:
```
--- FAIL: TestNew (0.00s)
    certificate_manager_test.go:134:
                Error Trace:    certificate_manager_test.go:134
                Error:          An error is expected but got nil.
                Test:           TestNew
                Messages:       expected error from key size of zero, got
FAIL
FAIL    github.com/openservicemesh/osm/pkg/certificate/providers/certmanager    0.517s
FAIL
```

```suggestion
	assert.Error(err, "expected error from key size of zero")
```

nojnhuh (approved):

pkg/certificate/providers/certmanager/certificate_manager_test.go
steeling: good call, fixed

PR #4418, opened 2022-01-08
feat(sidecar): add tls protocol version and cipher suites config to

allenlsy:
meshconfig

Signed-off-by: Allen Leigh <allenlsy@gmail.com>

**Description**:

Add TLS protocol related sidecar configuration to MeshConfig.

* TLS minimum protocol version
* TLS maximum protocol version
* Cipher suites
* ECDH curves

These configuration details can be found on [Envoy documentation](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/transport_sockets/tls/v3/common.proto)

Resovles: #4401 

**Testing done**:

* Unit test

**Affected area**:
| Functional Area            |     |
| -------------------------- | --- |
| New Functionality          | [ ] |
| CI System                  | [ ] |
| CLI Tool                   | [ ] |
| Certificate Management     | [ ] |
| Control Plane              | [ ] |
| Demo                       | [ ] |
| Documentation              | [ ] |
| Egress                     | [ ] |
| Ingress                    | [ ] |
| Install                    | [ ] |
| Networking                 | [ ] |
| Observability              | [ ] |
| Performance                | [ ] |
| SMI Policy                 | [ ] |
| Security                   | [X] |
| Sidecar Injection          | [X] |
| Tests                      | [ ] |
| Upgrade                    | [ ] |
| Other                      | [ ] |

Please answer the following questions with yes/no.

1. Does this change contain code from or inspired by another project?
    -   Did you notify the maintainers and provide attribution?

No

2. Is this a breaking change?

No

PR #4582, opened 2022-03-10
feat(ads): xDS response add TLS config

allenlsy:

**Description**:

This is a follow up PR of https://github.com/openservicemesh/osm/pull/4418. 

The original PR only applies TLS config to Envoy bootstrap config file, which means only the static resources are updated.

This PR applies the config to the dynamic resources in xDS response from control plane. SidecarSpec from MeshConfig is passed to the config generation sections.

**Testing done**:

* Unit test
* Manual test with kind cluster. Cipher suite is manually edited in mesh config. Confirmed that both static and dynamic resources are updated accordingly.

Screen Shot 2022-03-10 at 11 38 51 AM
Screen Shot 2022-03-10 at 11 38 45 AM

**Affected area**:
| Security                   | [X] |

No

2. Is this a breaking change?

No

No. It needs to be done later.

### DIFF
--- a/pkg/envoy/cds/cluster.go
+++ b/pkg/envoy/cds/cluster.go
@@ -17,6 +17,7 @@ import (
 	"google.golang.org/protobuf/types/known/durationpb"
 	"google.golang.org/protobuf/types/known/wrapperspb"
 
+	configv1alpha2 "github.com/openservicemesh/osm/pkg/apis/config/v1alpha2"
 	policyv1alpha1 "github.com/openservicemesh/osm/pkg/apis/policy/v1alpha1"
 
 	"github.com/openservicemesh/osm/pkg/catalog"
@@ -33,11 +34,11 @@ var replacer = strings.NewReplacer(".", "_", ":", "_")
 
 // getUpstreamServiceCluster returns an Envoy Cluster corresponding to the given upstream service
 // Note: ServiceIdentity must be in the format "name.namespace" [https://github.com/openservicemesh/osm/issues/3188]
-func getUpstreamServiceCluster(downstreamIdentity identity.ServiceIdentity, config trafficpolicy.MeshClusterConfig) *xds_cluster.Cluster {
+func getUpstreamServiceCluster(downstreamIdentity identity.ServiceIdentity, config trafficpolicy.MeshClusterConfig, sidecarSpec configv1alpha2.SidecarSpec) *xds_cluster.Cluster {
 	httpProtocolOptions := getDefaultHTTPProtocolOptions()
 
 	marshalledUpstreamTLSContext, err := anypb.New(
-		envoy.GetUpstreamTLSContext(downstreamIdentity, config.Service))
+		envoy.GetUpstreamTLSContext(downstreamIdentity, config.Service, sidecarSpec))
 	if err != nil {
 		log.Error().Err(err).Msgf("Error marshalling UpstreamTLSContext for upstream cluster %s", config.Name)
 		return nil
@@ -331,11 +332,11 @@ func formatAltStatNameForPrometheus(clusterName string) string {
 	return replacer.Replace(clusterName)
 }
 
-func upstreamClustersFromClusterConfigs(downstreamIdentity identity.ServiceIdentity, configs []*trafficpolicy.MeshClusterConfig) []*xds_cluster.Cluster {
+func upstreamClustersFromClusterConfigs(downstreamIdentity identity.ServiceIdentity, configs []*trafficpolicy.MeshClusterConfig, sidecarSpec configv1alpha2.SidecarSpec) []*xds_cluster.Cluster {
 	var clusters []*xds_cluster.Cluster
 
 	for _, c := range configs {
-		clusters = append(clusters, getUpstreamServiceCluster(downstreamIdentity, *c))
+		clusters = append(clusters, getUpstreamServiceCluster(downstreamIdentity, *c, sidecarSpec))
 	}
 	return clusters
 }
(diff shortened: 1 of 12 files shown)


pkg/envoy/xdsutil.go
```diff
@@ -76,11 +77,24 @@ func GetAddress(address string, port uint32) *xds_core.Address {
 }
 
 // GetTLSParams creates Envoy TlsParameters struct.
-func GetTLSParams() *xds_auth.TlsParameters {
-	return &xds_auth.TlsParameters{
-		TlsMinimumProtocolVersion: xds_auth.TlsParameters_TLSv1_2,
-		TlsMaximumProtocolVersion: xds_auth.TlsParameters_TLSv1_3,
+func GetTLSParams(sidecarSpec v1alpha2.SidecarSpec) *xds_auth.TlsParameters {
+	minVersionInt := xds_auth.TlsParameters_TlsProtocol_value[sidecarSpec.TLSMinProtocolVersion]
```
nojnhuh, line 81: I know the validating webhook already checks that the strings defined for min/max TLS versions is valid, but for something like this that's relatively far removed from that I think it's worth checking these are valid again anyway and returning an error if they aren't. If something invalid does come through this function somehow this will evaluate to 0 (`TLS_AUTO`), but I'm not sure how important it is to protect against that.

pkg/envoy/lds/inmesh.go
```diff
@@ -37,14 +38,14 @@ func (lb *listenerBuilder) getInboundMeshFilterChains(proxyService service.MeshS
 	switch strings.ToLower(proxyService.Protocol) {
 	case constants.ProtocolHTTP, constants.ProtocolGRPC:
 		// Filter chain for HTTP port
-		filterChainForPort, err := lb.getInboundMeshHTTPFilterChain(proxyService, uint32(proxyService.TargetPort))
+		filterChainForPort, err := lb.getInboundMeshHTTPFilterChain(proxyService, uint32(proxyService.TargetPort), lb.cfg.GetMeshConfig().Spec.Sidecar)
```
shashankram, line 41: Please store the `SidecarSpec` as a part of the `listenerBuilder` struct so you don't have to pass this around in every function call.

shashankram (changes requested):

pkg/envoy/lds/inmesh.go
allenlsy: Actually since `listenerBuilder` contains the configurator, `SidecarSpec` is stored in `listenerBuilding` already. We only can minimize the argument passing.

shashankram: For now, you can do either of the following:
1. Cache SidecarSpec in listenerBuilder
2. Cache MeshConfig in listenerBuilder
3. Use `lb.cfg.Xyz()` in lds/*

We want to avoid unnecessary access to the informer caches, as these incur a perf issue due to RW locks. So it's best to cache these in listenerBuilder for this purpose instead of option 3 above. The existing usage of 3 can be refactored later on.

allenlsy: Ok I see the intentions behind now.

pkg/envoy/cds/cluster.go
```diff
@@ -17,8 +17,7 @@ import (
 	"google.golang.org/protobuf/types/known/durationpb"
 	"google.golang.org/protobuf/types/known/wrapperspb"
 
-	policyv1alpha1 "github.com/openservicemesh/osm/pkg/apis/policy/v1alpha1"
-
+	"github.com/openservicemesh/osm/pkg/apis/config/v1alpha2"
```
shashankram, line 20: Please keep the import alias for custom APIs as before for consistency.

pkg/envoy/cds/response_test.go
```diff
@@ -27,6 +27,7 @@ import (
 
 	configv1alpha2 "github.com/openservicemesh/osm/pkg/apis/config/v1alpha2"
 
+	"github.com/openservicemesh/osm/pkg/apis/config/v1alpha2"
```
shashankram, line 30: Leave a line after custom API imports.
```
Go internal imports

3rd party imports

OSM custom API imports

OSM local imports
```

pkg/envoy/xdsutil.go
```diff
@@ -16,6 +16,7 @@ import (
 	"google.golang.org/protobuf/types/known/wrapperspb"
 	v1 "k8s.io/api/core/v1"
 
+	"github.com/openservicemesh/osm/pkg/apis/config/v1alpha2"
```
shashankram, line 19: Please follow the style convention for custom API imports.

shashankram (changes requested):

cmd/cli/policy_check_pods_test.go
```diff
@@ -14,6 +14,7 @@ import (
 	"k8s.io/client-go/kubernetes/fake"
 
 	configv1alpha2 "github.com/openservicemesh/osm/pkg/apis/config/v1alpha2"
+
```
shashankram, line 17: This can be a separate change (cleaning up imports)

pkg/envoy/lds/ingress_test.go
```diff
@@ -10,6 +10,7 @@ import (
 	tassert "github.com/stretchr/testify/assert"
 	"google.golang.org/protobuf/types/known/wrapperspb"
 
+	configv1alpha2 "github.com/openservicemesh/osm/pkg/apis/config/v1alpha2"
```
shashankram, line 13: leave line after this import

Issue #4568, opened 2022-03-03
Update control plane metrics


nojnhuh:
**Please describe the Improvement and/or Feature Request**

Based on [this doc](https://docs.google.com/document/d/1NB42lJ0Uvrevuqsv3ab8h-q_4bvX0FTWWrN_EPNf4NQ/edit?usp=sharing), I propose the following changes to the metrics generated by OSM's control plane components:

#### Keep

Metric|Type|Description
-|-|-
k8s_api_event_count|Counter|Represents the number of events received from the Kubernetes API Server
resource_namespace_count|Gauge|Represents the number of monitored namespaces in the service mesh
proxy_connect_count|Gauge|Represents the number of proxies connected to OSM controller
proxy_config_update_time|Histogram|Histogram to track time spent for proxy configuration
cert_issued_count|Counter|Represents the total number of XDS certificates issued to proxies
err_code_count|Counter|Respesents the number of errcodes generated by OSM
proxy_broadcast_event_count|Counter|Represents the number of ProxyBroadcast events published by the OSM controller

#### Add

Metric|Type|Description
-|-|-
proxy_xds_request|Counter|Total number of XDS requests made by each proxy
proxy_max_connections_rejected|Counter|Number of proxies rejected due to the configured connection limit
events_queued|Gauge|Number of events seen but yet to be processed
feature_flag_enabled|Gauge|Indicates whether each feature flag is enabled (1) or disabled (0)
http_response_count|Counter|Counter of HTTP responses sent handled
http_response_duration_ms|Histogram|Time in milliseconds taken to respond to HTTP requests
reconciliation_count|Counter|Total number of resource reconciliations invoked

#### Change

Metric|Type|Description|Change
-|-|-|-
proxy_response_send_success_count|Counter|Represents the number of responses successfully sent to proxies|Add labels for proxy and XDS type URI
proxy_response_send_error_count|Counter|Represents the number of responses that errored when being set to proxies|Add labels for proxy and XDS type URI
injector_sidecar_count|Counter|Counts the number of injector webhooks dealt with over time|Make this generic for all webhooks, add success indicator and resource kind
injector_rq_time|Histogram|Histogram for time taken to perform sidecar injection|Make this generic for all webhooks, add success indicator and resource kind

#### Remove

Metric|Type|Description
-|-|-
proxy_reconnect_count|Counter|Represents the number of reconnects from known proxies to OSM controller
cert_issued_time|Histogram|Histogram to track time spent to issue xds certificate

[...]

PR #4584, opened 2022-03-10
feat(metrics): add proxy XDS request metric

nojnhuh:
**Description**:
This change adds a new `osm_proxy_xds_request_count` metric to count XDS
requests made by proxies, labeled with the proxy's certificate CN and
the XDS type.

Example:

    osm_proxy_xds_request_count{common_name="369c9347-5b8f-42ff-ab1b-427ced15ee68.sidecar.bookbuyer.bookbuyer.cluster.local",type="type.googleapis.com/envoy.extensions.transport_sockets.tls.v3.Secret"} 4

Part of #4568

**Testing done**:
- This part of the codebase doesn't have unit tests, but I did some manual smoke testing.

**Affected area**:

3. Has documentation corresponding to this change been updated in the [osm-docs](https://github.com/openservicemesh/osm-docs) repo (if applicable)? Yes, https://github.com/openservicemesh/osm-docs/pull/337

### DIFF
--- a/pkg/envoy/ads/stream.go
+++ b/pkg/envoy/ads/stream.go
@@ -116,6 +116,8 @@ func (s *Server) StreamAggregatedResources(server xds_discovery.AggregatedDiscov
 
 			<-s.workqueues.AddJob(newJob(typesRequest, &discoveryRequest))
 
+			metricsstore.DefaultMetricsStore.ProxyXDSRequestCount.WithLabelValues(certCommonName.String(), discoveryRequest.TypeUrl).Inc()
+
 		case <-proxyUpdateChan:
 			log.Info().Str("proxy", proxy.String()).Msg("Broadcast update received")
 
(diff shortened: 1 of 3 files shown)


pkg/envoy/ads/stream.go
```diff
@@ -116,6 +116,8 @@ func (s *Server) StreamAggregatedResources(server xds_discovery.AggregatedDiscov
 
 			<-s.workqueues.AddJob(newJob(typesRequest, &discoveryRequest))
 
+			metricsstore.DefaultMetricsStore.ProxyXDSRequestCount.WithLabelValues(certCommonName.String(), discoveryRequest.TypeUrl).Inc()
```
shashankram, line 119: Ideally, this counter should be incremented for every request, not just the requests that are responded to. So it would make sense to move it above L110. Keeping it here only counts the number of requests that are going to be responded to, which should ideally be similar to the number of responses sent.

This metric would be valuable to count all requests from Envoy, even those that are not responded to, as it would help determine whether the controller is ignoring requests when this counter keeps incrementing but there are no responses from the contoller.

shashankram (changes requested):

shashankram (approved):

snehachhabria (approved):
Remember to squash before completing the PR :)

Issue #4524, opened 2022-02-10
Improve error handling


steeling:
There's a few things we can do to improve error messages and error handling:

1. lowercase error messages to comply with lint: https://staticcheck.io/docs/checks#ST1005
2. remove usage of github.com/pkg/errors (in maintenance mode), in favor of stdlib fmt.Errorf & errors.Unwrap, errors.As, errors.Is
3. Errors should be handled exactly once. There are many places where we log then immediately return an error. We should do one or the other. Most often we want to return a wrapped error.

#3 will result in slightly different error messages, that may affect any JSON parsers, but will colocate errors in a single log line so that they inherently associated vs spread throughout the log files.

Any lack of ability to parse error messages with JSON could be solved externally with full text search or regex search (recommended). Alternatively we could create our own errors that fit in neatly with zerolog, to log wrapped errors in a JSON friendly format. I believe the latter is not worth the effort at this time.

https://github.com/openservicemesh/osm/pull/4521 as an example draft PR

allenlsy:
This issue is not completely resolved. PR #4904 only fixed item 2 in the description. Most important item is ensuring error handled exactly once. We haven't worked on it. I suggest re-open this issue.

PR #4596, opened 2022-03-16
feat(certificates) rework vault certificate provider 

schristoff:
Builds on the work done in #4580 - this removes all of the caching, messaging broker and more to allow the Vault Provider to only focus on certificates. 
(accidentally put this in the wrong block so was hidden, oops!) 

**Affected area**:
| Certificate Management     | [X ] |

no

2. Is this a breaking change?
no


pkg/certificate/providers/vault/types.go
```diff
@@ -2,36 +2,16 @@
...
-	"time"
-
 	"github.com/hashicorp/vault/api"
-
-	"github.com/openservicemesh/osm/pkg/certificate"
-	"github.com/openservicemesh/osm/pkg/configurator"
-	"github.com/openservicemesh/osm/pkg/messaging"
 )
 
 // CertManager implements certificate.Manager and contains a Hashi Vault client instance.
```
jaellio, line 8: `CertManager` now implements the `client` interface.

pkg/certificate/providers/vault/certificate_manager.go
```diff
@@ -30,23 +25,13 @@ const (
...
-// NewCertManager implements certificate.Manager and wraps a Hashi Vault with methods to allow easy certificate issuance.
-func NewCertManager(
-	vaultAddr,
-	token string,
-	role string,
-	cfg configurator.Configurator,
-	serviceCertValidityDuration time.Duration,
-	msgBroker *messaging.Broker) (*CertManager, error) {
+// New will construct a new certificate client using Vault's cert-manager
+func New(vaultAddr, token string, role string) (*CertManager, error) {
```
steeling, line 32: nit: if you're specifying the param type once for 2 variables, may as well do once for all 3

pkg/certificate/providers/vault/certificate_manager_test.go
```diff
@@ -1,88 +1,18 @@
...
-				vaultAddr,
-				vaultToken,
-				vaultRole,
-				mockConfigurator,
-				mockConfigurator.GetServiceCertValidityPeriod(),
-				nil,
-			)
-			Expect(err).To(HaveOccurred())
-			vaultError := err.(*url.Error)
-			expected := `unsupported protocol scheme "foo"`
```
steeling, line 82: this test is. valid, we can keep it

steeling (approved):

pkg/certificate/providers/vault/certificate_manager.go
schristoff: I actually love this nit, good call out.

pkg/certificate/providers/vault/certificate_manager_test.go
shashankram: +1, please add this back

pkg/certificate/providers/vault/types.go
```diff
@@ -2,36 +2,16 @@
...
 	role vaultRole
-
-	cfg configurator.Configurator
-
-	serviceCertValidityDuration time.Duration
-
-	msgBroker *messaging.Broker
 }
 
 type vaultRole string
```
steeling, line 17: the benefit of aliasing types is typically to reduce the allowed set of values. ie:

```
type vaultRole string // unexported, so needs to be specified

var (
   defaultRole vaultRole = "default-role"
)
```

Since we're not doing that, let's just go ahead and get rid of these and call them strings.

pkg/certificate/providers/vault/tools.go
```diff
@@ -11,12 +11,8 @@ func getDurationInMinutes(validityPeriod time.Duration) string {
 	return fmt.Sprintf("%dh", validityPeriod/time.Hour)
 }
 
-func getIssueURL(role vaultRole) vaultPath {
-	return vaultPath(fmt.Sprintf("pki/issue/%+v", role))
-}
-
-func getRoleConfigURL(role vaultRole) vaultPath {
-	return vaultPath(fmt.Sprintf("pki/roles/%s", role))
+func getroleURL(role string) string {
```
steeling, line 14: Did you intend to make change?  The functionality should stay the same (as well. as the func names), with simply changing the type from. vaultRole to string

pkg/certificate/providers/vault/certificate_manager.go
```diff
@@ -60,42 +45,12 @@ func NewCertManager(
...
-	cm.ReleaseCertificate(cert.GetCommonName())
-
-	return issuingCA, cert.GetSerialNumber(), err
-}
-
-func (cm *CertManager) issue(cn certificate.CommonName, validityPeriod time.Duration) (*certificate.Certificate, error) {
-	secret, err := cm.client.Logical().Write(getIssueURL(cm.role).String(), getIssuanceData(cn, validityPeriod))
+// IssueCertificate will request a new signed certificate from the configured Vault issuer.
+func (cm *CertManager) IssueCertificate(cn certificate.CommonName, validityPeriod time.Duration) (*certificate.Certificate, error) {
+	secret, err := cm.client.Logical().Write(getroleURL(cm.role), getIssuanceData(cn, validityPeriod))
```
steeling, line 53: this is returning a new URL which would be a bug.

pkg/certificate/providers/vault/tools.go
```diff
@@ -11,12 +11,8 @@ func getDurationInMinutes(validityPeriod time.Duration) string {
 	return fmt.Sprintf("%dh", validityPeriod/time.Hour)
 }
 
-func getIssueURL(role vaultRole) vaultPath {
-	return vaultPath(fmt.Sprintf("pki/issue/%+v", role))
-}
-
-func getRoleConfigURL(role vaultRole) vaultPath {
```
steeling, line 18: note: this is valid to remove since it is never used. We can delete the corresponding test as well

pkg/certificate/providers/vault/tools_test.go
```diff
@@ -25,16 +25,16 @@ var _ = Describe("Test tools", func() {
...
 		It("creates the URL for issuing a new certificate", func() {
-			actual := getIssueURL(role)
-			expected := vaultPath(fmt.Sprintf("pki/issue/%s", role))
+			actual := getroleURL(role)
+			expected := fmt.Sprintf("pki/issue/%s", role)
 			Expect(actual).To(Equal(expected))
 		})
 	})
 
 	Context("Test role config URL", func() {
```
steeling, line 34: You can delete this whole test

pkg/certificate/providers/vault/tools.go
schristoff: 🤦‍♀️ went too hard on the cleanup and thought "getIssueURL" and "getRoleConfigURL" were the same thing. my b!

pkg/certificate/providers/vault/certificate_manager.go
```diff
@@ -30,72 +25,36 @@ const (
...
-		msgBroker:                   msgBroker,
+		role: role,
 	}
 	config := api.DefaultConfig()
 	config.Address = vaultAddr
 
 	var err error
 	if c.client, err = api.NewClient(config); err != nil {
-		return nil, errors.Errorf("Error creating Vault CertManager without TLS at %s", vaultAddr)
+		return nil, errors.Errorf("Error creating Vault CertManager without TLS at %s", c.client.Address())
```
steeling, line 45: this could result in nil pointer dereference if an err is returned and client is nil, this should be vaultAddr

pkg/certificate/providers/vault/certificate_manager_test.go
```diff
@@ -114,99 +48,115 @@ var _ = Describe("Test client helpers", func() {
...
+		{description: "no err", wantErr: false},
+		{description: "all nil", wantErr: true},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.description, func(t *testing.T) {
+			if tc.wantErr {
+				_, err := New("", "", "")
+				if err == nil {
+					t.Fatalf("expected error, got nil")
```
steeling, line 70: Fatal prevents future tests from running, this should be an error

pkg/certificate/providers/vault/certificate_manager_test.go
```diff
@@ -114,99 +48,115 @@ var _ = Describe("Test client helpers", func() {
...
+				}
+				_, err = New(addr, token, vaultRole)
```
steeling, line 72: You're actually running this inside the wantErr as well it looks like

pkg/certificate/providers/vault/certificate_manager_test.go
```diff
@@ -114,99 +48,115 @@ var _ = Describe("Test client helpers", func() {
...
+		description string
+		wantErr     bool
+	}{
+		{description: "no err", wantErr: false},
+		{description: "all nil", wantErr: true},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.description, func(t *testing.T) {
+			if tc.wantErr {
```
steeling, line 67: doesn't look like. your testing the success case, so table driven tests aren't needed here.

Typically in a TDT, you would add the inputs to the test case struct.

pkg/certificate/providers/vault/certificate_manager_test.go
```diff
@@ -114,99 +48,115 @@ var _ = Describe("Test client helpers", func() {
...
-			certBytes := uuid.New().String()
-			issue := func(certificate.CommonName, time.Duration) (*certificate.Certificate, error) {
-				cert := &certificate.Certificate{
-					IssuingCA:    pem.RootCertificate(certBytes),
-					SerialNumber: certSerialNum,
+	testCases := []struct {
+		description string
+		wantErr     bool
+	}{
+		{description: "no err", wantErr: false},
```
steeling, line 61: Instead of err and no err, let's provide the description of why it fails, ie:

"errors when an invalid address is provided"

pkg/certificate/providers/vault/certificate_manager_test.go
```diff
@@ -114,99 +48,115 @@ var _ = Describe("Test client helpers", func() {
...
+}
 
-		It("gets certs from cache", func() {
-			// This cert does not exist - returns nil
-			Expect(cm.getFromCache("nothing")).To(BeNil())
-
-			// This cert has expired -- returns nil
-			Expect(cm.getFromCache(expiredCertCN)).To(BeNil())
+func Test_IssueCertificate(t *testing.T) {
+	var common_name certificate.CommonName = "localhost"
```
steeling, line 82: No snake case in go

pkg/certificate/providers/vault/certificate_manager_test.go
```diff
@@ -114,99 +48,115 @@ var _ = Describe("Test client helpers", func() {
...
-			Expect(actual).To(Equal(rootCert))
-		})
+func mockVault(t *testing.T) (*vault.Core, string, string) {
+	coreConfig := &vault.CoreConfig{
+		LogicalBackends: map[string]logical.Factory{
+			"pki": pki.Factory,
+		},
+	}
+	core, _, rootToken := vault.TestCoreUnsealedWithConfig(t, coreConfig)
+	_, addr := vhttp.TestServer(t, core)
```
steeling, line 125: We should eventually call Close on the. returned listener. 

can we rework this to use golang's test setup?

pkg/certificate/providers/vault/certificate_manager_test.go
```diff
@@ -114,99 +48,115 @@ var _ = Describe("Test client helpers", func() {
...
+	}
+}
 
-		It("gets certs from cache", func() {
-			// This cert does not exist - returns nil
-			Expect(cm.getFromCache("nothing")).To(BeNil())
-
-			// This cert has expired -- returns nil
-			Expect(cm.getFromCache(expiredCertCN)).To(BeNil())
+func Test_IssueCertificate(t *testing.T) {
```
steeling, line 81: No snake case

pkg/certificate/providers/vault/certificate_manager_test.go
```diff
@@ -114,99 +48,115 @@ var _ = Describe("Test client helpers", func() {
...
+	if err != nil {
+		t.Fatalf("did not expect error, got %v", err)
+	}
+	mockVaultwithPKI(t, cm)
+
+	testCases := []struct {
+		description string
+		wantErr     bool
+	}{
+		{description: "no err", wantErr: false},
```
steeling, line 98: same notes as above

pkg/certificate/providers/vault/certificate_manager_test.go
```diff
@@ -114,99 +48,115 @@ var _ = Describe("Test client helpers", func() {
...
-			Expect(actual).To(Equal(validCert))
-		})
+}
 
-		It("can fetch root cert", func() {
-			actual, err := cm.GetRootCertificate()
-			Expect(err).ToNot(HaveOccurred())
-			Expect(actual).To(Equal(rootCert))
-		})
+func mockVault(t *testing.T) (*vault.Core, string, string) {
```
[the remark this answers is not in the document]
schristoff: That's Vault running - why do you find them not useful to what we're testing?

pkg/certificate/providers/vault/certificate_manager.go
```diff
@@ -30,72 +25,36 @@ const (
...
-	vaultAddr,
-	token string,
-	role string,
-	cfg configurator.Configurator,
-	serviceCertValidityDuration time.Duration,
-	msgBroker *messaging.Broker) (*CertManager, error) {
+// New will construct a new certificate client using Vault's cert-manager
+func New(vaultAddr, token, role string) (*CertManager, error) {
+	if role == "" {
+		return nil, errors.Errorf("Error creating Vault CertManager, expected role got %s", role)
```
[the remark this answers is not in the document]
schristoff: good catch

pkg/certificate/providers/vault/certificate_manager_test.go
```diff
@@ -114,99 +48,115 @@ var _ = Describe("Test client helpers", func() {
...
 
-		It("implements Certificate correctly", func() {
-			actual, err := cm.GetCertificate(validCertCN)
-			Expect(err).ToNot(HaveOccurred())
-			Expect(actual.CommonName).To(Equal(validCertCN))
-			Expect([]byte(actual.GetCertificateChain())).To(Equal([]byte(validCert.GetCertificateChain())))
-			Expect(actual.GetExpiration()).To(Equal(validCert.GetExpiration()))
-			Expect([]byte(actual.IssuingCA)).To(Equal([]byte(issuingCA)))
-		})
+func mockVaultwithPKI(t *testing.T, cm *CertManager) {
```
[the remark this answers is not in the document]
schristoff: This does do similar things to `mockVault()`
The insight behind my decision is for testing `New()` it doesn't matter if the PKI secret's engine for Vault is turned on or not - so I moved enabling Vault's PKI side into a separate function so it doesn't need to be used for `New()`. 

Basically `mockVault()` right now is an agent running that isn't configured at all, just hanging out.
`mockVaultwithPKI` takes what we made in `mockVault()` and configures that agent.

pkg/certificate/providers/vault/certificate_manager_test.go
schristoff: `doesn't look like. your testing the success case` - this was an accident 

Yeah,  I was flip flopping around on how to test this in a normal way. I could just add input and actually test what happens if we send in stuff, which _could_ be more useful than "nil" and "not nil". If you don't think so though I can take out the tables.

pkg/certificate/providers/vault/certificate_manager.go
```diff
@@ -30,72 +25,36 @@ const (
...
 
-// NewCertManager implements certificate.Manager and wraps a Hashi Vault with methods to allow easy certificate issuance.
-func NewCertManager(
-	vaultAddr,
-	token string,
-	role string,
-	cfg configurator.Configurator,
-	serviceCertValidityDuration time.Duration,
-	msgBroker *messaging.Broker) (*CertManager, error) {
+// New will construct a new certificate client using Vault's cert-manager
```
jaellio, line 31: nit:
```suggestion
// New constructs a new certificate client using Vault's cert-manager
```

pkg/certificate/providers/vault/certificate_manager.go
```diff
@@ -30,72 +25,36 @@ const (
...
-
-	// We are not going to need this certificate - remove it
-	cm.ReleaseCertificate(cert.GetCommonName())
-
-	return issuingCA, cert.GetSerialNumber(), err
-}
-
-func (cm *CertManager) issue(cn certificate.CommonName, validityPeriod time.Duration) (*certificate.Certificate, error) {
-	secret, err := cm.client.Logical().Write(getIssueURL(cm.role).String(), getIssuanceData(cn, validityPeriod))
+// IssueCertificate will request a new signed certificate from the configured Vault issuer.
```
jaellio, line 55: nit:
```suggestion
// IssueCertificate requests a new signed certificate from the configured Vault issuer.
```

pkg/certificate/providers/vault/certificate_manager_test.go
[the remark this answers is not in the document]
steeling: FWIW I agree with Jon, I think we can simplify to have the single thing running. mostly because YAGNI :)

In general modularity is good, so I see where you're coming from as well!

pkg/certificate/providers/vault/certificate_manager_test.go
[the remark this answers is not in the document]
nojnhuh: Any log that's not an error seems to indicate that Vault is running as expected, so that seems like pure noise to me. I imagine if there's an easy way to turn them off it should be simple to turn them on again when debugging these tests locally if there's any doubt Vault isn't working right where those extra logs would be helpful.

Again, not a big deal though if there's no obvious way to adjust the logs.

pkg/certificate/providers/vault/certificate_manager_test.go
```diff
@@ -51,26 +52,70 @@ var _ = Describe("Test client helpers", func() {
...
+			role:        vaultRole,
+			token:       "",
+			wantErr:     true,
+		},
 	}
 
 	for _, tc := range testCases {
 		t.Run(tc.description, func(t *testing.T) {
+			_, err := New(tc.vaultaddr, tc.token, tc.role)
 			if tc.wantErr {
```
steeling, line 113: nit:

```
if (err != nil) != tc.wantErr {
  t.Errorf(...)
}
```

pkg/certificate/providers/vault/certificate_manager_test.go
```diff
@@ -115,15 +173,15 @@ func TestIssueCertificate(t *testing.T) {
...
 	coreConfig := &vault.CoreConfig{
 		LogicalBackends: map[string]logical.Factory{
 			"pki": pki.Factory,
 		},
 	}
 	core, _, rootToken := vault.TestCoreUnsealedWithConfig(t, coreConfig)
-	_, addr := vhttp.TestServer(t, core)
-	return core, rootToken, addr
+	ln, addr := vhttp.TestServer(t, core)
+	return core, rootToken, addr, ln
```
steeling, line 184: since you're passing in t, you can do:


```
mockVault(...) (string, string) {
   ...
   t.Cleanup(func(){
       ln.Close()
       core.Shutdown()
   })
   return rootToken, addr
}
```

steeling (changes requested):

pkg/certificate/providers/vault/certificate_manager_test.go
```diff
@@ -114,99 +49,172 @@ var _ = Describe("Test client helpers", func() {
...
+			vaultaddr:   addr,
+			role:        vaultRole,
+			token:       "",
+			wantErr:     true,
+		},
+		{
+			description: "bad addr",
+			vaultaddr:   "127.1.2.3",
+			role:        vaultRole,
+			token:       "",
```
steeling, line 105: This test will fail due to no token

pkg/certificate/providers/vault/certificate_manager_test.go
```diff
@@ -169,8 +169,14 @@ func mockVault(t *testing.T) (string, string) {
 	ln, addr := vhttp.TestServer(t, core)
 
 	t.Cleanup(func() {
-		ln.Close()
-		core.Shutdown()
+		err := ln.Close()
+		if err != nil {
+			t.Fatalf("could not close net listener")
```
steeling, line 174: Probably t.error so we best effort shutdown core vs bail early

steeling (approved):

pkg/certificate/providers/vault/certificate_manager.go
```diff
@@ -30,180 +25,65 @@ const (
...
-	cm.msgBroker.GetCertPubSub().Pub(events.PubSubMessage{
-		Kind:   announcements.CertificateRotated,
-		NewObj: newCert,
-		OldObj: oldCert.(*certificate.Certificate),
-	}, announcements.CertificateRotated.String())
-
-	log.Debug().Msgf("Rotated certificate (old SerialNumber=%s) with new SerialNumber=%s took %+v", oldCert.(*certificate.Certificate).GetSerialNumber(), newCert.SerialNumber, time.Since(start))
-
-	return newCert, nil
+	//TODO(2068): implement a delete cert
```
shashankram, line 79: https://github.com/openservicemesh/osm/issues/2068 is closed. Does the TODO still apply?

shashankram (approved):

pkg/certificate/providers/vault/certificate_manager_test.go
```diff
@@ -52,6 +53,7 @@ var _ = Describe("Test client helpers", func() {
 })
 
 func TestNew(t *testing.T) {
+	tassert := assert.New(t)
```
nojnhuh, line 56: Let's move this into the `t.Run` so the error message shows exactly which test case failed. As-is the error message looks like this:
```
--- FAIL: TestNew (0.01s)
    certificate_manager_test.go:107:
                Error Trace:    certificate_manager_test.go:107
                Error:          An error is expected but got nil.
                Test:           TestNew
                Messages:       expected error, got nil
```

But with that change it would be:
```
--- FAIL: TestNew (0.01s)
    --- FAIL: TestNew/valid_inputs (0.00s)
        certificate_manager_test.go:107:
                Error Trace:    certificate_manager_test.go:107
                Error:          An error is expected but got nil.
                Test:           TestNew/valid_inputs
                Messages:       expected error, got nil
```

pkg/certificate/providers/config.go
```diff
@@ -296,19 +296,28 @@ func (c *Config) getHashiVaultOSMCertificateManager(options VaultOptions) (certi
...
-		c.msgBroker,
 	)
 	if err != nil {
 		return nil, nil, errors.Errorf("Error instantiating Hashicorp Vault as a Certificate Manager: %+v", err)
 	}
 
-	return vaultCertManager, vaultCertManager, nil
+	vaultCert, err := vaultClient.GetRootCertificate()
+	if err != nil {
+		return nil, nil, errors.Errorf("Error getting Vault Root Certificate, got: %v", err)
```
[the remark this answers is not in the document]
steeling: https://github.com/openservicemesh/osm/issues/4524 :)

PR #4599, opened 2022-03-18
fix(logs): add missing Msg calls to logs

nojnhuh:
**Description**:
Log calls that don't include `.Msg()` or `.Msgf()` never get logged.
This change adds these where I could find they were missing.

Demo: https://go.dev/play/p/zaZ58Teo7y3

**Testing done**:
- See Go playground link above

**Affected area**:


trstringer (approved):
Good catch, @nojnhuh. I wonder if we should spend some time in getting a regex put together and grep through the codebase. This is definitely significant and we should try to catch all occurrences of this.

nojnhuh:
> Good catch, @nojnhuh. I wonder if we should spend some time in getting a regex put together and grep through the codebase. This is definitely significant and we should try to catch all occurrences of this.

This is what I put together to find these (using ripgrep):
```
rg '^\s*log\.' | rg -v '\.$' | rg -vF Msg
```

Which is roughly "lines starting with `log.` that don't end in `.` (since Msg was broken out into the next line in the instances I checked) that don't contain `Msg`."

That does throw a few false-positives.

jaellio (approved):

PR #4600, opened 2022-03-20
test(*): add retry policy e2e

shalier:
**Description**:
Adds E2E testing for retry policy

**Testing done**:

**Affected area**:
| Tests                      | [x ] |

2. Is this a breaking change?


nojnhuh (approved):

tests/e2e/e2e_retry_policy_test.go
```diff
@@ -0,0 +1,239 @@
...
+					req := HTTPRequestDef{
+						SourceNs:        client,
+						SourcePod:       clientPod.Name,
+						SourceContainer: podDef.GetName(),
+						Destination:     fmt.Sprintf("%s.%s.svc.cluster.local:9091", serverSvc.Name, server),
+					}
+
+					By("A request that will be retried NumRetries times then succeed")
+					// wait for server
+					time.Sleep(3 * time.Second)
```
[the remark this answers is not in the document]
shalier: https://github.com/openservicemesh/osm/issues/4739

demo/cmd/retry/retry.go
```diff
@@ -0,0 +1,40 @@
+package main
```
shashankram, line 1: Why do we need a custom app to test retries?

Shouldn't this be possible with `httpbin` and the `curl` client where we can simulate failures by curling `/status/<status-code>`?

shashankram (changes requested):

tests/e2e/e2e_retry_policy_test.go
```diff
@@ -125,8 +125,11 @@ var _ = OSMDescribe("Test Retry Policy",
 					By("A request that will be retried NumRetries times then succeed")
 					time.Sleep(3 * time.Second)
 					result := Td.HTTPRequest(req)
-					stdout, _, err := Td.RunLocal(filepath.FromSlash("../../bin/osm"), "proxy", "get", "stats", clientPod.Name, "--namespace", client)
+					stdout, stderr, err := Td.RunLocal(filepath.FromSlash("../../bin/osm"), "proxy", "get", "stats", clientPod.Name, "--namespace", client)
 					Expect(err).ToNot((HaveOccurred()))
+					if stderr != nil {
```
nojnhuh, line 130: I'm pretty sure `stderr` is never nil even if it's empty, so I'd change this to `if err != nil`.

nojnhuh: And I think we'll want to check this before `Expect(err).ToNot((HaveOccurred()))` since I think the test will return as soon as that fails.

tests/e2e/e2e_retry_policy_test.go
```diff
@@ -0,0 +1,284 @@
...
+					Expect(err).ToNot((HaveOccurred()))
+
+					req := HTTPRequestDef{
+						SourceNs:        client,
+						SourcePod:       clientPod.Name,
+						SourceContainer: podDef.GetName(),
+						Destination:     fmt.Sprintf("%s.%s.svc.cluster.local:80/status/503", serverSvc.Name, server),
+					}
+
+					By("A request that will be retried NumRetries times then succeed")
```
jaellio, line 125: I think the request eventually fails so should this line be the following?
```suggestion
					By("A request that will be retried NumRetries times and then fail")
```

tests/e2e/e2e_retry_policy_test.go
```diff
@@ -0,0 +1,284 @@
...
+					time.Sleep(3 * time.Second)
```
jaellio, line 126: For my own understanding, why are we sleeping 3 seconds here?

tests/e2e/e2e_retry_policy_test.go
```diff
@@ -0,0 +1,284 @@
...
+					Expect(err).ToNot((HaveOccurred()))
+
+					req := HTTPRequestDef{
+						SourceNs:        client,
+						SourcePod:       clientPod.Name,
+						SourceContainer: podDef.GetName(),
+						Destination:     fmt.Sprintf("%s.%s.svc.cluster.local:80/status/503", serverSvc.Name, server),
+					}
+
+					By("A request that will be retried NumRetries times then succeed")
```
jaellio, line 238:
```suggestion
					By("A request that will be retried 0 times and then fail")
```

tests/e2e/e2e_retry_policy_test.go
shashankram, line 126: Let's remove this sleep and poll instead, so the test is resilient to timing delays.

shalier: The sleep was to wait for the retry policy to be applied, there's an issue to add a Status field for Retry in the future so that this won't be needed. https://github.com/openservicemesh/osm/pull/4600#discussion_r869584107

tests/e2e/e2e_retry_policy_test.go
```diff
@@ -0,0 +1,291 @@
...
+
+var meshNs = []string{client, server}
+
+var retryStats = map[string]string{"upstream_rq_retry": "", "upstream_rq_retry_limit_exceeded": "", "upstream_rq_retry_backoff_exponential": ""}
+var thresholdUintVal uint32 = 5
+
+var _ = OSMDescribe("Test Retry Policy",
+	OSMDescribeInfo{
+		Tier:   2,
+		Bucket: 8,
```
shashankram, line 32: Nit: move to bucket 4 which typically completes sooner than other buckets. This helps in keeping the bucket times more balanced.

shashankram (approved):

jaellio (approved):

PR #4601, opened 2022-03-21
ref(crdconvert): respond with errors from all requested resources

nojnhuh:
**Description**:
Currently, CRD conversion webhooks only respond with the error returned
by the first failed resource converted, but there may be several. This
change aggregates errors from all resources sent in a request.

In addition, for #4568 I'm planning to add logic to record metrics labeled with the requested resources' `apiVersion` and `kind` from objects parsed out of the conversion request. Those changes (WIP) are shown here: https://github.com/openservicemesh/osm/commit/e5587d6c8a084559701a9c0ff537ebae851e3e19#diff-c4fc6ee6327a461684ba57d2d0cc26326c4cb0a980119d85cf93c730c1a7ec65. Without this change, resources in the request beyond the first one that fails will not have metrics recorded showing they were also failed to be converted.

**Testing done**:
- Added/updated unit tests

**Affected area**:

### DIFF
--- a/pkg/crdconversion/framework.go
+++ b/pkg/crdconversion/framework.go
@@ -47,6 +47,7 @@ func conversionResponseFailureWithMessagef(msg string, params ...interface{}) *v
 // failures will be reported as Reason in the conversion response.
 func doConversion(convertRequest *v1beta1.ConversionRequest, convert convertFunc) *v1beta1.ConversionResponse {
 	var convertedObjects []runtime.RawExtension
+	// aggregate errors from all objects in the request vs. only returning the first
 	var errs []string
 	for _, obj := range convertRequest.Objects {
 		cr := unstructured.Unstructured{}
--- a/pkg/crdconversion/framework.go
+++ b/pkg/crdconversion/framework.go
@@ -47,7 +47,9 @@ func conversionResponseFailureWithMessagef(msg string, params ...interface{}) *v
 // failures will be reported as Reason in the conversion response.
 func doConversion(convertRequest *v1beta1.ConversionRequest, convert convertFunc) *v1beta1.ConversionResponse {
 	var convertedObjects []runtime.RawExtension
-	// aggregate errors from all objects in the request vs. only returning the first
+	// aggregate errors from all objects in the request vs. only returning the
+	// first so errors from all objects are sent in the request and metrics are
+	// recorded as accurately as possible.
 	var errs []string
 	for _, obj := range convertRequest.Objects {
 		cr := unstructured.Unstructured{}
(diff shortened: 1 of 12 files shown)


shashankram:
> Currently, CRD conversion webhooks only respond with the error returned
> by the first failed resource converted, but there may be several. This
> change aggregates errors from all resources sent in a request.

Could you comment on why the aggregation of errors is necessary here? I am guessing it may be important in some circumstances to know about all the errors that might occur to be able to diagnose an issue pertaining to the conversion. I am curious to know why this is necessary in this particular case.

nojnhuh:
> Could you comment on why the aggregation of errors is necessary here? I am guessing it may be important in some circumstances to know about all the errors that might occur to be able to diagnose an issue pertaining to the conversion. I am curious to know why this is necessary in this particular case.

It's not strictly necessary for the conversion to work, but I see it as generally useful to return as much information as we have available for users to be able to fix errors faster than one at a time.

My primary motivation for this change is to accurately count metrics which I tried to describe above. Currently, if a request comes in to convert two resources and converting the first resource fails, the response would indicate converting both failed but the metrics would only count one failed.

shashankram:
> > Could you comment on why the aggregation of errors is necessary here? I am guessing it may be important in some circumstances to know about all the errors that might occur to be able to diagnose an issue pertaining to the conversion. I am curious to know why this is necessary in this particular case.
> 
> It's not strictly necessary for the conversion to work, but I see it as generally useful to return as much information as we have available for users to be able to fix errors faster than one at a time.
> 
> My primary motivation for this change is to accurately count metrics which I tried to describe above. Currently, if a request comes in to convert two resources and converting the first resource fails, the response would indicate converting both failed but the metrics would only count one failed.

Please also document this in the code, so someone doesn't revert this back to return the first error as a form of simplification for the existing code.

pkg/crdconversion/framework.go
```diff
@@ -43,43 +43,44 @@ func conversionResponseFailureWithMessagef(msg string, params ...interface{}) *v
...
-	return metav1.Status{
-		Status: metav1.StatusSuccess,
-	}
-}
-
 // doConversion converts the requested object given the conversion function and returns a conversion response.
 // failures will be reported as Reason in the conversion response.
 func doConversion(convertRequest *v1beta1.ConversionRequest, convert convertFunc) *v1beta1.ConversionResponse {
 	var convertedObjects []runtime.RawExtension
+	// aggregate errors from all objects in the request vs. only returning the first
```
shashankram, line 50: nit: This comment is missing why aggregating errors helps in this case.

shashankram (approved):
Left a minor comment that can be ignored if you think it so

jaellio (approved):

nojnhuh:
GitHub Actions seems to be acting up at the moment per https://githubstatus.com, so the regular CI tests aren't running for the newest commit here. Since the previous commit passed aside from the known image scan issue for alpine that has a fix upstream in progress and I've only changed a code comment since then, I'm going to merge this.

PR #4607, opened 2022-03-23
k8s/version: drop v1.19 support and bump CI version

### LINKED ISSUE #4610: Drop support for k8s v1.19 and update the CI to use latest k8s version and APIs

Drop support for k8s v1.19 which has been deprecated upstream.
Further, update the CI to use the latest k8s version by default (v1.23) and update the tests to use the `networking.k8s.io/v1` API for ingress.

- CI System                  [X]

shashankram:
**Description**:
Drops support for v1.19 of k8s which is no longer
supported upstream.

Also updates the `kind` pkg version to use the
latest k8s version v1.23 by default. Since v1.22
deprecated the networking.k8s.io/v1beta1 API,
updates the e2e to use the stable v1 API .

Resolves #4610 

**Testing done**:
CI: `> k8s server version: v1.23.4`

**Affected area**:

3. Has documentation corresponding to this change been updated in the [osm-docs](https://github.com/openservicemesh/osm-docs) repo (if applicable)? `no, will be done once this change is merged`

### DIFF
--- a/tests/e2e/e2e_k8s_version_test.go
+++ b/tests/e2e/e2e_k8s_version_test.go
@@ -18,11 +18,11 @@ var _ = OSMDescribe("Test HTTP traffic for different k8s versions",
 		Bucket: 6,
 	},
 	func() {
-		Context("Version v1.22.0", func() {
-			testK8sVersion("v1.22.0")
+		Context("Version v1.22.8", func() {
+			testK8sVersion("v1.22.8")
 		})
-		Context("Version v1.19.7", func() {
-			testK8sVersion("v1.19.7")
+		Context("Version v1.21.11", func() {
+			testK8sVersion("v1.21.11")
 		})
 	})
 
--- a/README.md
+++ b/README.md
@@ -82,7 +82,7 @@ Read more about [OSM's high level goals, design, and architecture](DESIGN.md).
 ## Install
 
 ### Prerequisites
-- Kubernetes cluster running Kubernetes v1.19.0 or greater
+- Kubernetes cluster running Kubernetes v1.20.0 or greater
 - kubectl current context is configured for the target cluster install
   - ```kubectl config current-context```
 
--- a/charts/osm/Chart.yaml
+++ b/charts/osm/Chart.yaml
@@ -21,7 +21,7 @@ version: 1.0.0
 appVersion: latest-main
 
 # This specifies the minimum Kubernetes version OSM is compatible with.
-kubeVersion: ">= 1.19.0-0"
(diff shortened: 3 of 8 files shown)


trstringer (approved):

tests/e2e/e2e_k8s_version_test.go
```diff
@@ -18,11 +18,11 @@ var _ = OSMDescribe("Test HTTP traffic for different k8s versions",
...
 	func() {
-		Context("Version v1.22.0", func() {
-			testK8sVersion("v1.22.0")
+		Context("Version v1.22.8", func() {
+			testK8sVersion("v1.22.8")
 		})
-		Context("Version v1.19.7", func() {
-			testK8sVersion("v1.19.7")
+		Context("Version v1.21.11", func() {
+			testK8sVersion("v1.21.11")
```
nojnhuh, line 25: IIRC the intent here was for this to be the oldest version we support. Then this should be the latest v1.20 release.

nojnhuh (changes requested):
Left one super small comment. Other than that, calling this out in the release notes wouldn't hurt, even though people probably shouldn't be using v1.19 anyway if it's not supported upstream.

tests/e2e/e2e_k8s_version_test.go
shashankram: @nojnhuh there's no value testing v1.20.x. It has been dropped as well. The only reason I didn't deprecate support for v1.20.x from the Helm chart enforcement is that we avoid dropping installation support for multiple k8s versions in the same release. From a product perspective, there is absolutely no reason to test v1.20 in this test given we do not have product changes that use specific APIs from that version. This test was added as a best-effort validation to validate at least 2 prior versions.

nojnhuh (approved):

shashankram:
> Left one super small comment. Other than that, calling this out in the release notes wouldn't hurt, even though people probably shouldn't be using v1.19 anyway if it's not supported upstream.

I plan on updating the release notes in a subsequent PR, as I need to add a few other things such as circuit breaking there.

nojnhuh:
> > Left one super small comment. Other than that, calling this out in the release notes wouldn't hurt, even though people probably shouldn't be using v1.19 anyway if it's not supported upstream.
> 
> I plan on updating the release notes in a subsequent PR, as I need to add a few other things such as circuit breaking there.

I think it makes the most sense to keep release note updates with the relevant changes they describe where we can. Please at least note specific PRs/commits in the commit description for release note updates if they were made separately.

PR #4612, opened 2022-03-24
Decouple Conversion Patch and CRD Reconciler

### LINKED ISSUE #4608: CRD Reconciler depends on conversion webhook patch
**Bug description**:

While trying to troubleshoot #4575, I noticed that deleted CRDs weren't being reconciled and added back as expected. Upon further inspection, I realized that CRDs are only reconciled [if a specific reconciler label](https://github.com/openservicemesh/osm/blob/6bda5c3a34bf20ed33b3479ae554665d0f882f67/pkg/reconciler/client.go#L63) is present on the manifest. That label is only added to the CRD during the [conversion webhook patch](https://github.com/openservicemesh/osm/blob/6bda5c3a34bf20ed33b3479ae554665d0f882f67/pkg/crdconversion/crdconversion.go#L198) (if reconciliation is enabled). After #4569 was merged, only the `MeshConfig` CRD can be reconciled since it still receives the conversion patch. 

- Install                    [x]

**Expected behavior**:

When I enable reconciliation, CRDs should be managed and reconciled as expected.

```shell
make kind-up
demo/run-osm-demo.sh --set osm.enableReconciler=true
kubectl delete crd ingressbackends.policy.openservicemesh.io # The CRD is not re-created after this step
```

**How was OSM installed?**:
Auto demo

**Bug report archive**:

**Environment**:
- OSM version (use `osm version`): commit 7e0674c3
```
Client Version: version.Info{Major:"1", Minor:"21", GitVersion:"v1.21.2", GitCommit:"092fbfbf53427de67cac1e9fa54aaa09a28371d7", GitTreeState:"clean", BuildDate:"2021-06-16T12:59:11Z", GoVersion:"go1.16.5", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"21", GitVersion:"v1.21.1", GitCommit:"5e58841cce77d4bc13713ad2b91fa0d961e69192", GitTreeState:"clean", BuildDate:"2021-05-21T23:01:33Z", GoVersion:"go1.16.4", Compiler:"gc", Platform:"linux/amd64"}
```
- Size of cluster (number of worker nodes in the cluster): 3 kind nodes
- Others:

keithmattix:
Instead of using the CRD conversion webhook patching process
to add the osm reconciler label, use kubectl and helm chart templating to
conditionally add the labels.

**Description**:
Fixes #4608 

**Testing done**:
Unit tests added for the new `installCRDs` function
The demo also succeeds 
image

**Affected area**:
| Install                    | [x] |
| Other                      | [] |


keithmattix:
@nojnhuh I was going to write a bash script initially that looked similar to what you suggested, but @shashankram said that the team tends to favor Go binaries

shashankram:
> @nojnhuh I was going to write a bash script initially that looked similar to what you suggested, but @shashankram said that the team tends to favor Go binaries

@keithmattix There seems to be a misunderstanding here. What I meant was that for the reconciliation part, it makes sense to keep the code within the Go application, and not script such functionality. If it's just a matter of running a few commands, doing so from the shell should work just fine. In fact, we do such things in the cleanup hooks and control plane init containers.

keithmattix:
Ah, yes I misunderstood; I wasn't planning on re-writing the reconciliation stuff in shell. I'll adjust the PR to modify the labels via shell

cmd/osm-bootstrap/crds/config_multicluster_service.tpl
```diff
@@ -0,0 +1,167 @@
+# Custom Resource Definition (CRD) for OSM's multicluster config specification.
```
nojnhuh, line 1: These templates can all be removed, right?

keithmattix: Ah, yes they can. Undoing a rename in git isn't as intuitive as I thought :)

nojnhuh (approved):

snehachhabria (approved):

keithmattix:
@shashankram Done

charts/osm/templates/osm-bootstrap-deployment.yaml
```diff
@@ -57,10 +57,14 @@ spec:
...
-          - -f
-          - /osm-crds
+          command:
+          - sh
+          - -c
+          - >
+            kubectl apply -f /osm-crds;
+            {{- if .Values.osm.enableReconciler }}
+            kubectl label -f /osm-crds openservicemesh.io/reconcile=true;
+            {{- end }}
```
trstringer, line 67: If this is the plan moving forward we should probably change the [osm-crds Dockerfile](https://github.com/keithmattix/osm/blob/6b3da7b8b82da6c081e991904d847de2d8982960/dockerfiles/Dockerfile.osm-crds#L10) to no longer have `kubectl` as the entrypoint.

trstringer (changes requested):
Added a single comment for consideration and using this to block merging prior to discussing/deciding. // @shashankram @nojnhuh @snehachhabria

trstringer:
Looks like my review didn't get in before merging. We can discuss and create a new issue/PR if changes are needed.

charts/osm/templates/osm-bootstrap-deployment.yaml
nojnhuh: Defining a `command` for a container in a pod overrides `ENTRYPOINT` in the Dockerfile, so this still works the way we want. I can see marginal value in removing it from the Dockerfile for the sake of general cleanliness, but I don't see it causing any issues as long as that image stays an implementation detail and is mostly unexposed to users.

trstringer: Yeah, agreed. I should have specified, it wasn't from a functional standpoint but likely we should have the Dockerfile reflecting how the image should be used.

keithmattix: I found out the osm-crds image is apparently also used for some of the e2e tests in GitHub actions, so I'm wondering if changing the `ENTRYPOINT` will break that workflow.

nojnhuh: Based on where I see us using it in the chart I don't think changing it will break anything since both places define a `command`:
https://github.com/openservicemesh/osm/blob/8bae12f7186d3bf7709951ba8c37e68e569f30c8/charts/osm/templates/cleanup-hook.yaml#L77
https://github.com/openservicemesh/osm/blob/8bae12f7186d3bf7709951ba8c37e68e569f30c8/charts/osm/templates/osm-bootstrap-deployment.yaml#L58

I can't think of any other place where that image is run.

PR #4625, opened 2022-03-31
feat(log): enable Envoy logging on inbound & outbound TCP listeners

### LINKED ISSUE #4530: Add HTTPS Egress Logs
Add logging for egress for HTTPS calls when cluster-wide egress is disabled and egress policies are in use.  When following [this tutorial](https://release-v1-0.docs.openservicemesh.io/docs/demos/egress_policy/) in the HTTPS Egress section, these connections are not logged to stdout of the Envoy sidecar like they are for HTTP.  

This appears to be because the filter config on the outbound-listener passed to the sidecar doesn't have an access_log property set and the HTTP ones do.  

- Egress                     [x]
- Logging                    [x]

**Possible use cases**
General logging improvements it would also be useful as most traffic is HTTPS (not HTTP) now so the HTTP logging is almost redundant, and it would also match the feature which is offered in other service mesh products.

allenlsy:

**Description**:

Add Envoy access log configuration to inbound & outbound listener for TCP traffic.

Resolves #4530 

**Testing done**:

✅  updated unit tests.

Dev build of OSM is deployed and set up following the egress demo documentation. It is observed that:

* `outbound-listener` has proper `access_log` settings 
* Envoy logs egress HTTPS connection process. Screenshot below shows Envoy logging HTTPS connection to httpbin.org on port 443

Screen Shot 2022-03-30 at 8 58 55 PM

**Affected area**:

No

2. Is this a breaking change?

No

Not needed

### DIFF
--- a/pkg/envoy/lds/listener.go
+++ b/pkg/envoy/lds/listener.go
@@ -42,6 +42,7 @@ func (lb *listenerBuilder) newOutboundListener() (*xds_listener.Listener, error)
 				Name: wellknown.OriginalDestination,
 			},
 		},
+		AccessLog: envoy.GetAccessLog(),
 	}
 
 	// Create a default passthrough filter chain when global egress is enabled.
@@ -122,6 +123,7 @@ func newInboundListener() *xds_listener.Listener {
 				Name: wellknown.OriginalDestination,
 			},
 		},
+		AccessLog: envoy.GetAccessLog(),
 	}
 }
 
--- a/pkg/envoy/lds/listener_test.go
+++ b/pkg/envoy/lds/listener_test.go
@@ -43,6 +43,7 @@ var _ = Describe("Construct inbound listeners", func() {
 		It("Tests the inbound listener config", func() {
 			listener := newInboundListener()
 			Expect(listener.Address).To(Equal(envoy.GetAddress(constants.WildcardIPAddr, constants.EnvoyInboundListenerPort)))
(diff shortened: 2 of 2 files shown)


pkg/envoy/lds/listener.go
```diff
@@ -58,6 +58,8 @@ func (lb *listenerBuilder) newOutboundListener() (*xds_listener.Listener, error)
 	}
 
 	if featureflags := lb.cfg.GetFeatureFlags(); featureflags.EnableEgressPolicy {
+		listener.AccessLog = envoy.GetAccessLog()
```
shashankram, line 61: Please enable this by default on the listener and not just when Egress policy is enabled. This will be required for global passthrough egress and also for in-mesh TCP traffic.

shashankram (changes requested):

pkg/envoy/lds/listener.go
allenlsy: I'm curious if there is any specific reason why the log wasn't added for TCP listener initially? If it was ignored, then sure we can add it now.

allenlsy: Updated. Added log for both inbound and outbound TCP listeners

shashankram: Thanks. Could you please update the PR && commit description and title to reflect the changes made.

shashankram (approved):

nojnhuh (approved):

PR #4629, opened 2022-04-04
fix(ci): fix lint

nojnhuh:
**Description**:
This change updates the version of `golangci-lint` used in the Makefile
and CI to the floating `latest` version so we don't have to periodically
update this manually.

Other changes are made based on linter errors thrown by the updated
version of the linter.

**Testing done**:
- CI
- Did a manual smoke test of `make go-lint`

**Affected area**:

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -47,11 +47,10 @@ jobs:
       - name: go build deps
         run: make embed-files-test
       - name: golangci-lint
-        uses: golangci/golangci-lint-action@v2
+        uses: golangci/golangci-lint-action@v3
         with:
-          version: v1.41.1
+          version: latest
           skip-pkg-cache: true
-          args: --timeout 10m --skip-dirs pkg/gen --skip-files "zz_generated.deepcopy.go$"
 
   codegen:
     name: Codegen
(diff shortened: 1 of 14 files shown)


.github/workflows/main.yml
```diff
@@ -47,11 +47,10 @@ jobs:
       - name: go build deps
         run: make embed-files-test
       - name: golangci-lint
-        uses: golangci/golangci-lint-action@v2
+        uses: golangci/golangci-lint-action@v3
         with:
-          version: v1.41.1
+          version: latest
```
draychev, line 52: Wouldn't it be better to pin to the actual latest version vs using `latest` to avoid breakages?

Thank you for upgrading the linter!!

trstringer: We shouldn't care so much about a sliding window tag for something like a linter. A breakage in this scenario is just a failed workflow that likely should be investigated. Keeping up with incremental versions for something as trivial as linting is likely something we don't want to do. My vote would to keep this with @nojnhuh's current change with `latest`.

shashankram (approved):

jaellio (approved):

Issue #4634, opened 2022-04-07
Implement troubleshooting tooling within OSM CLI


shashankram:
**Please describe the Improvement and/or Feature Request**
Implement a troubleshooting tool within OSM CLI to detect, verify, and troubleshoot the correctness of the system for various scenarios.

Previously, a proof-of-concept for this was built as a part of the [osm-health](https://github.com/openservicemesh/osm-health) project. However, there are several limitations with osm-health and it needs to be heavily refactored to be usable, for the following reasons:
1. osm-health was developed independently outside the core osm project repo. The code and design have not been reviewed by all the required stakeholders of the osm projects, and there is a disconnect in terms of how we currently expect the troubleshooting tool to operate. Some of these are highlighted below.
2. Many of the checks implemented within osm-health make assumptions that should not be made, e.g. regarding traffic type (HTTP, TCP, etc.), image tags, error logs, etc. As a result, the tool is unusable beyond a very narrow scenario which makes it unusable in production environments.
3. The tool runs a k8s controller (informer + caches) which is completely unnecessary. Simple API calls via client-go should suffice. Running a k8s controller makes the code more complex than necessary.
4. The output on the console is not concise. Given the use of k8s controller within the tool, control plane logs are emitted with commands. While this can be suppressed, it's best to write to stdout/stderr in a console-friendly manner as common with CLIs.
5. Having the tool within osm CLI simplifies the development, testing, and release workflows.
6. Single tool for OSM management and troubleshooting - OSM CLI

I am proposing to reimplement the necessary tooling as a part of OSM CLI while borrowing certain ideas from osm-health. Integrating this tooling into OSM CLI also gives us the opportunity to think about the structure of the tooling project in code versus simply retrofitting the osm-health POC, which uses a directory structure not suitable within the osm repo, into the core osm repo.

**Scope (please mark with X where applicable)**
- New Functionality          [X]
- CLI Tool                   [X]

**Possible use cases**
Troubleshooting tool for OSM.

**Tasks**
- [x] Pod-to-pod connectivity configuration verification
- [x] Pod-to-external (egress) connectivity configuration verification
- [x] External-to-pod (ingress) connectivity configuration verification
- [x] Control plane health

steeling:
I have some thoughts in this area, but don't have the time on my plate currently. This seems large enough scope to warrant a small design doc prior to moving forward with implementation

steeling:
Yup! I see 3 ways to get data to inspect:

[...]

PR #4639, opened 2022-04-08
cli: add verify command

shashankram:
**Description**:
Adds `osm verify` command to verify the correctness
of the system for specific scenarios.

This is a POC for the idea described in #4634 
to implement a troubleshooting tool. The 
existing checks are incomplete but demonstrate
the intention of the tool. 

The goal of this tool is to help troubleshoot
commonly reported issues related to app
connectivity, policies, etc. It borrows some ideas
from the osm-health tool, but reimplements 
the code for reasons described in #4634.

Each verification check is implemented the `Verifier` interface.
A verification can include other verification checks thereby
allowing checks to be reused.

Sample usage:
```
 $ osm verify connectivity --from-pod bookbuyer/bookbuyer-5c6bc7b74d-9t6p8 --to-pod bookstore/bookstore-v2-5bbbdc9495-6v7zr
[+] Context: Verify if pod "bookbuyer/bookbuyer-5c6bc7b74d-9t6p8" can access pod "bookstore/bookstore-v2-5bbbdc9495-6v7zr" for app protocol "http"
Status: Success

```

```
$ osm verify connectivity --from-pod bookbuyer/bookbuyer-5c6bc7b74d-9t6p8 --to-pod bookthief/bookthief-5b79c95c7c-zszpk
[+] Context: Verify if pod "bookbuyer/bookbuyer-5c6bc7b74d-9t6p8" can access pod "bookthief/bookthief-5b79c95c7c-zszpk" for app protocol "http"
Status: Failure
Reason: A verification step failed
Suggestion: Please follow the suggestions listed in the failed steps below to resolve the issue

[+] Context: Verify if namespace "bookbuyer" is monitored
Status: Success

[+] Context: Verify if namespace "bookthief" is monitored
Status: Failure
Reason: Missing label "openservicemesh.io/monitored-by" on namespace "bookthief"
Suggestion: Add label "openservicemesh.io/monitored-by" on namespace "bookthief" to include it in the mesh and restart the app

```

**Affected area**:

3. Has documentation corresponding to this change been updated in the [osm-docs](https://github.com/openservicemesh/osm-docs) repo (if applicable)? `no`


steeling:
I'm curious what future commands/subcommands will look like, and if we can take a more kubectl-esque approach. 

ie: maybe `osm verify connectivity --from [<pod>|<kind>/<name>] --to [<pod>|<kind>/<name>] [-n|--namespace <namespace>]`

shashankram:
> I'm curious what future commands/subcommands will look like, and if we can take a more kubectl-esque approach.
> 
> ie: maybe `osm verify connectivity --from [<pod>|<kind>/<name>] --to [<pod>|<kind>/<name>] [-n|--namespace <namespace>]`

The immediate scenarios to address are listed [here](https://docs.google.com/document/d/1-HPGNtnch6ahYkcrjPv84UKr7AVeTO4X0Ykb1b2-zGo/edit?usp=sharing).

Regarding flags, it comes down to preference. I like using positional args for mandatory arguments, but see how using required flags can make the arguments more explicit.

pkg/cli/verifier/connectivity_pod_to_pod.go
```diff
@@ -0,0 +1,68 @@
...
+		kubeClient:  kubeClient,
+		srcPod:      srcPod,
+		dstPod:      dstPod,
+		appProtocol: appProtocol,
+		meshName:    meshName,
+	}
+}
+
+// Run executes the pod connectivity verifier
+func (v *PodConnectivityVerifier) Run() VerificationResult {
```
steeling, line 39: +1 I like the way you set this up! 

A couple nits, in comments below, feel free to ignore.

pkg/cli/verifier/connectivity_pod_to_pod.go
```diff
@@ -0,0 +1,68 @@
...
+	result := VerificationResult{
```
steeling, line 40: This whole block seems relatively generic, including the failure status/suggestion in the for loop. You could type alias []Verifier and add a Run command to that.

pkg/cli/verifier/connectivity_pod_to_pod.go
```diff
@@ -0,0 +1,68 @@
...
+		res := verification.Run()
+		if res.Status == Failure {
+			result.Status = Failure
+			result.Reason = "A verification step failed"
+			result.Suggestion = "Please follow the suggestions listed in the failed steps below to resolve the issue"
+		}
+		result.NestedResults = append(result.NestedResults, &res)
+	}
+
+	if result.Status != Failure && result.Status != Unknown {
```
steeling, line 63: I don't think result.Status == Unknown is ever currently possible?

cmd/cli/verify_connectivity.go
```diff
@@ -0,0 +1,25 @@
...
+`
+
+func newVerifyConnectivityCmd(config *action.Configuration, stdout io.Writer, stderr io.Writer) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "connectivity",
+		Short: "verify connectivity configurations",
+		Long:  verifyConnectivityDescription,
+		Args:  cobra.NoArgs,
+	}
+	cmd.AddCommand(newVerifyConnectivityPodToPodCmd(config, stdout, stderr))
```
steeling, line 22: Should we signify in some way that this isn't complete yet? Or is the plan to just get feedback and continue implementation prior to merge?

shashankram: Yes this PR is to just gather feedback on the idea. It cannot be merged in its current state hence still a draft PR.

pkg/cli/verifier/verifier.go
```diff
@@ -0,0 +1,74 @@
...
+
+// Color returns a color coded string for the verification status
+func (s VerificationStatus) Color() string {
+	if s == Success {
+		return color.GreenString("%s", s)
+	} else if s == Failure {
+		return color.RedString("%s", s)
+	} else {
+		return color.YellowString("%s", s)
+	}
```
nojnhuh, line 73: `switch` might be a better fit here.

cmd/cli/verify_connectivity_pod_to_pod.go
```diff
@@ -0,0 +1,87 @@
...
+				return errors.Errorf("Could not access Kubernetes cluster, check kubeconfig: %s", err)
+			}
+			verifyConnectPodToPodCmd.kubeClient = clientset
+			return verifyConnectPodToPodCmd.run()
+		},
+	}
+
+	f := cmd.Flags()
+	f.StringVar(&verifyConnectPodToPodCmd.appProtocol, "app-protocol", constants.ProtocolHTTP, "Application protocol")
+	f.StringVar(&verifyConnectPodToPodCmd.meshName, "mesh-name", "osm", "Mesh name")
```
nojnhuh, line 72: The default mesh name is already defined here: https://github.com/openservicemesh/osm/blob/15aa2f095d698bcd3ec2cfbb6de7123364cd9364/cmd/cli/install.go#L55

shashankram:
@nojnhuh This PR isn't ready for a thorough review, see https://github.com/openservicemesh/osm/pull/4639#discussion_r846393634. I'd like to get feedback on the idea of the tool first before addressing any of the code-specific comments, as this PR is very much a draft at the moment. Hence the reason for unused variables, redundant printing etc.

Once we agree this is the right step forward, I will polish this PR to make it reviewable from an implementation perspective.

cmd/cli/verify_connectivity.go
steeling, line 22: > > I'm curious what future commands/subcommands will look like, and if we can take a more kubectl-esque approach.

> > ie: maybe osm verify connectivity --from [<pod>|<kind>/<name>] --to [<pod>|<kind>/<name>] [-n|--namespace <namespace>]

> The immediate scenarios to address are listed [here](https://docs.google.com/document/d/1-HPGNtnch6ahYkcrjPv84UKr7AVeTO4X0Ykb1b2-zGo/edit?usp=sharing).

> Regarding flags, it comes down to preference. I like using positional args for mandatory arguments, but see how using required flags can make the arguments more explicit.

Personally, I prefer using deployment/sts since it allows me to not have to look up & copy/paste pod each time. ie: `kubectl logs deployment/bookbuyer -n bookbuyer`. Having something like that would be a slightly cleaner API.

I think it would also help since it allows us to permute the different variations. ie: To/From
* service
* service identity
* pod
* deployment
* statefulset
* etc

So maybe with that in mind we switch to `osm verify connectivity --from-pod <namespace/pod> --to-service <namespace/service>`.

That loses the kubectl-likeness I was shooting for originally, but allows these permutations

keithmattix: I think deployments/sts/service could be a shorthand, but I think the base functionality should allow users to specify a given pod. I've run into several scenarios where one pod in particular has an issue getting XDS updates and I need to exec and and test curls to/from that pod. Furthermore, each OSM workload cert has a CN with a uuid corresponding to the pod name, so that increases the weird scenarios customers could get into.

Again, I think having deployment/sts/service(maybe we should generalize to FQDN?) shorthand is still very useful. I just want to make sure we're not excluding the pod to pod connectivity as the base case.

pkg/cli/verifier/connectivity_pod_to_pod.go
keithmattix: +1 I like the pluggable interface model. I can imagine this being run nightly or as part of CI 🚀

pkg/cli/verifier/connectivity_pod_to_pod.go
keithmattix: +1 `type VerificationSet []Verifier` perhaps?

keithmattix (approved):
I like the design and think this'll be really useful. Great work!

pkg/cli/verifier/verifier.go
shashankram: Will leave it as is.

pkg/cli/verifier/connectivity_pod_to_pod.go
```diff
@@ -0,0 +1,68 @@
...
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/kubernetes"
+)
+
+var _ Verifier = (*PodConnectivityVerifier)(nil)
+
+// PodConnectivityVerifier implements the Verifier interface for pod connectivity
+type PodConnectivityVerifier struct {
+	stdout      io.Writer
+	stderr      io.Writer
```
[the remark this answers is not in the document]
shashankram: These get passed to `pkg/cli/verifier`, which will write output to it.

pkg/cli/verifier/connectivity_pod_to_pod.go
jaellio: +1 I like this approach too

pkg/cli/verifier/connectivity_pod_to_pod.go
shashankram: At the moment yes, but it needs to be used when the verifier cannot determine the result, for e.g. when an API call fails. As I add more verifications, this will become evident.

cmd/cli/verify_connectivity.go
shashankram: I dropped the `pod-to-pod` subcommand in favor of flags. For now, both `from-pod` and `to-pod` are mandatory till we allow additional permutations, which won't happen anytime soon.

pkg/cli/verifier/verifier.go
```diff
@@ -0,0 +1,93 @@
...
+	Unknown Status = "Unknown"
+)
+
+// Result defines the result returned by a Verifier instance
+type Result struct {
+	Context       string
+	Status        Status
+	Reason        string
+	Suggestion    string
+	NestedResults []*Result
```
keithmattix, line 30: I assume this is a slice of pointers to reduce excessive copying?

jaellio (approved):

pkg/cli/verifier/namespace.go
```diff
@@ -0,0 +1,63 @@
...
+		stdout:     stdout,
+		stderr:     stderr,
+		kubeClient: kubeClient,
+		namespace:  namespace,
+		meshName:   meshName,
+	}
+}
+
+// Run executes the namespace monitor verification
+func (v *NamespaceMonitorVerifier) Run() Result {
```
jaellio, line 35: Should we add unit tests for `Run()`?

pkg/cli/verifier/connectivity_pod_to_pod.go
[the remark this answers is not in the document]
nojnhuh: But in `pkg/cli/verifier` these aren't being used besides copying them around. The results are only being printed here which doesn't use these fields on the verifier:

https://github.com/openservicemesh/osm/blob/8eb583ffb26756b12ba3e294d197a5457f0adef2/cmd/cli/verify_connectivity.go#L95-L97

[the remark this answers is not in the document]
shashankram: I see what you mean. The verifier will use this for verbose logging to console. I haven't added those to this PR but they will be used. It didn't make sense changing all the args back and forth given the urgency of this tool.

pkg/cli/verifier/verifier.go
shashankram: Yes, although this could very well have been a copy given the struct size is small.

pkg/cli/verifier/namespace.go
shashankram: Yes I plan to. Didn't want to do it will the overall implementation is decided as I'd have to refactor the tests back and forth.

nojnhuh (approved):
Let's err on the side of adding tests to soon rather than too late.

pkg/cli/verifier/connectivity_pod_to_pod.go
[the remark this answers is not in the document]
shashankram: Forgot to mention these are used by the `Print()` function in `verifier.go`.

shashankram:
> Let's err on the side of adding tests to soon rather than too late.

I added a simple test in `pkg/verifier`. At this moment it doesn't make sense to add tests under `cmd/cli` as they would only test the output format. I'll continue adding tests to `pkg/cli/verifier` as I make changes there.

PR #4640, opened 2022-04-11
IngressBackend UpstreamTrafficSetting validations

### LINKED ISSUE #4638: Add stricter validations for IngressBackend and UpstreamTrafficSetting resources
Add validations in the validating webhook to ensure duplicate resources targeting the same destination is not permitted, as these would result in policy conflicts.

- For IngressBackend, ensure multiple IngressBackends cannot specify the same backend service
- For UpstreamTrafficSetting, ensure multiple UpstreamTrafficSettings cannot specify the same upstream host.

- Validations          [X]

**Possible use cases**
To prevent user misconfiguration.

keithmattix:
**Description**:
Fixes #4638. Adds stricter validatingWebhook validations for IngressBackend and UpstreamTrafficSetting (duplicate backends and hosts respectively). This PR also changes MeshService -> IngressBackend retrieval process to match on port as well as service name. This appears to have been an accidental omission from the original design since one would expect to be able to create an IngressBackend that allows access to the same service on different ports.

**Testing done**:
Additional unit tests added in the appropriate Golang packages

**Affected area**:

### DIFF
--- a/pkg/policy/client.go
+++ b/pkg/policy/client.go
@@ -163,7 +163,7 @@ func (c client) GetIngressBackendPolicy(svc service.MeshService) *policyV1alpha1
 		// using a validating webhook.
 		for _, backend := range ingressBackend.Spec.Backends {
 			// we need to check ports to allow ingress to multiple ports on the same svc
-			if backend.Name == svc.Name && backend.Port.Number == int(svc.Port) && backend.Port.Protocol == svc.Protocol {
+			if backend.Name == svc.Name && backend.Port.Number == int(svc.TargetPort) && backend.Port.Protocol == svc.Protocol {
 				return ingressBackend
 			}
 		}
@@ -208,6 +208,10 @@ func (c client) GetUpstreamTrafficSetting(options UpstreamTrafficSettingGetOpt) 
 	// Filter by MeshService
 	for _, resource := range c.caches.upstreamTrafficSetting.List() {
 		upstreamTrafficSetting := resource.(*policyV1alpha1.UpstreamTrafficSetting)
+
+		if upstreamTrafficSetting.Spec.Host == options.Host {
+			return upstreamTrafficSetting
+		}
 
 		if upstreamTrafficSetting.Namespace == options.MeshService.Namespace &&
 			upstreamTrafficSetting.Spec.Host == options.MeshService.FQDN() {
--- a/pkg/policy/client.go
+++ b/pkg/policy/client.go
@@ -163,7 +163,7 @@ func (c client) GetIngressBackendPolicy(svc service.MeshService) *policyV1alpha1
 		// using a validating webhook.
 		for _, backend := range ingressBackend.Spec.Backends {
 			// we need to check ports to allow ingress to multiple ports on the same svc
-			if backend.Name == svc.Name && backend.Port.Number == int(svc.TargetPort) && backend.Port.Protocol == svc.Protocol {
+			if backend.Name == svc.Name && backend.Port.Number == int(svc.TargetPort) {
 				return ingressBackend
 			}
 		}
@@ -191,7 +191,7 @@ func (c client) ListRetryPolicies(source identity.K8sServiceAccount) []*policyV1
 
 // GetUpstreamTrafficSetting returns the UpstreamTrafficSetting resource that matches the given options
 func (c client) GetUpstreamTrafficSetting(options UpstreamTrafficSettingGetOpt) *policyV1alpha1.UpstreamTrafficSetting {
-	if options.MeshService == nil && options.NamespacedName == nil {
+	if options.MeshService == nil && options.NamespacedName == nil && options.Host == "" {
 		log.Error().Msgf("No option specified to get UpstreamTrafficSetting resource")
 		return nil
 	}
--- a/pkg/policy/client.go
+++ b/pkg/policy/client.go
@@ -162,7 +162,8 @@ func (c client) GetIngressBackendPolicy(svc service.MeshService) *policyV1alpha1
 		// Multiple IngressBackend policies for the same backend will be prevented
 		// using a validating webhook.
 		for _, backend := range ingressBackend.Spec.Backends {
-			if backend.Name == svc.Name {
+			// we need to check ports to allow ingress to multiple ports on the same svc
+			if backend.Name == svc.Name && backend.Port.Number == int(svc.Port) && backend.Port.Protocol == svc.Protocol {
 				return ingressBackend
 			}
 		}
--- a/pkg/validator/server.go
+++ b/pkg/validator/server.go
@@ -17,6 +17,7 @@ import (
 	"github.com/openservicemesh/osm/pkg/certificate"
 	"github.com/openservicemesh/osm/pkg/constants"
 	"github.com/openservicemesh/osm/pkg/errcode"
+	policyClientset "github.com/openservicemesh/osm/pkg/gen/client/policy/clientset/versioned"
 	"github.com/openservicemesh/osm/pkg/webhook"
 )
 
@@ -35,7 +36,7 @@ type validatingWebhookServer struct {
 }
 
 // NewValidatingWebhook returns a validatingWebhookServer with the defaultValidators that were previously registered.
-func NewValidatingWebhook(webhookConfigName, osmNamespace, osmVersion, meshName string, enableReconciler, validateTrafficTarget bool, port int, certManager certificate.Manager, kubeClient kubernetes.Interface, stop <-chan struct{}) error {
+func NewValidatingWebhook(webhookConfigName, osmNamespace, osmVersion, meshName string, enableReconciler, validateTrafficTarget bool, port int, certManager certificate.Manager, kubeClient kubernetes.Interface, policyClient policyClientset.Interface, stop <-chan struct{}) error {
 	// This is a certificate issued for the webhook handler
 	// This cert does not have to be related to the Envoy certs, but it does have to match
 	// the cert provisioned with the ValidatingWebhookConfiguration
@@ -46,9 +47,13 @@ func NewValidatingWebhook(webhookConfigName, osmNamespace, osmVersion, meshName 
 		return errors.Errorf("Error issuing certificate for the validating webhook: %+v", err)
 	}
 
+	kv := &policyValidator{
+		policyClient: policyClient,
+	}
+
 	v := &validatingWebhookServer{
 		validators: map[string]validateFunc{
-			policyv1alpha1.SchemeGroupVersion.WithKind("IngressBackend").String(): ingressBackendValidator,
+			policyv1alpha1.SchemeGroupVersion.WithKind("IngressBackend").String(): kv.ingressBackendValidator,
 			policyv1alpha1.SchemeGroupVersion.WithKind("Egress").String():         egressValidator,
 			smiAccess.SchemeGroupVersion.WithKind("TrafficTarget").String():       trafficTargetValidator,
 		},
--- a/pkg/validator/server.go
+++ b/pkg/validator/server.go
@@ -14,10 +14,11 @@ import (
 	"k8s.io/client-go/kubernetes"
 
 	policyv1alpha1 "github.com/openservicemesh/osm/pkg/apis/policy/v1alpha1"
+	"github.com/openservicemesh/osm/pkg/policy"
+
 	"github.com/openservicemesh/osm/pkg/certificate"
 	"github.com/openservicemesh/osm/pkg/constants"
 	"github.com/openservicemesh/osm/pkg/errcode"
-	policyClientset "github.com/openservicemesh/osm/pkg/gen/client/policy/clientset/versioned"
 	"github.com/openservicemesh/osm/pkg/webhook"
 )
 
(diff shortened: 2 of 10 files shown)


pkg/validator/types.go
```diff
@@ -2,6 +2,10 @@ package validator
 
 import (
 	"github.com/openservicemesh/osm/pkg/logger"
+	"github.com/pkg/errors"
 )
 
 var log = logger.New(ValidatorWebhookSvc)
+
+var ErrIngressBackendDuplicateBackends = errors.New("error: duplicate backends detected")
+var ErrUpstreamTrafficSettingDuplicateHosts = errors.New("error: duplicate hosts detected")
```
jaellio, line 11:
```suggestion
var (
  ErrIngressBackendDuplicateBackends      = errors.New("error: duplicate backends detected")
  ErrUpstreamTrafficSettingDuplicateHosts = errors.New("error: duplicate hosts detected")
)
```

jaellio: In most errors we create we don't use the format "error: \<err msg\>". Instead, we do something like the following `var ErrCouldNotOpen = errors.New("could not open")`

pkg/policy/client.go
```diff
@@ -162,7 +162,8 @@ func (c client) GetIngressBackendPolicy(svc service.MeshService) *policyV1alpha1
 		// Multiple IngressBackend policies for the same backend will be prevented
 		// using a validating webhook.
 		for _, backend := range ingressBackend.Spec.Backends {
-			if backend.Name == svc.Name {
+			// we need to check ports to allow ingress to multiple ports on the same svc
+			if backend.Name == svc.Name && backend.Port.Number == int(svc.Port) && backend.Port.Protocol == svc.Protocol {
```
shashankram, line 166: The TargetPort check is performed in catalog/ingress.go, but makes sense to be done here. I can't recollect why I did not add it here.
```suggestion
			if backend.Name == svc.Name && backend.Port.Number == int(svc.TargetPort) && backend.Port.Protocol == svc.Protocol {
```

pkg/validator/types.go
keithmattix: Ah, thanks for the heads up! Fixed

pkg/validator/server.go
```diff
@@ -17,6 +17,7 @@ import (
 	"github.com/openservicemesh/osm/pkg/certificate"
 	"github.com/openservicemesh/osm/pkg/constants"
 	"github.com/openservicemesh/osm/pkg/errcode"
+	policyClientset "github.com/openservicemesh/osm/pkg/gen/client/policy/clientset/versioned"
```
shashankram, line 20: Group this with the API imports and leave a line after it.

```
stdlib

3rd party

OSM APIs

local pkg
```

pkg/validator/types.go
```diff
@@ -2,6 +2,10 @@ package validator
 
 import (
 	"github.com/openservicemesh/osm/pkg/logger"
+	"github.com/pkg/errors"
 )
 
 var log = logger.New(ValidatorWebhookSvc)
+
+var errIngressBackendDuplicateBackends = errors.New("duplicate backends detected")
+var errUpstreamTrafficSettingDuplicateHosts = errors.New("duplicate hosts detected")
```
shashankram, line 11: Let's directly return the errors instead of statically defining this so we can add more context to the error message.

pkg/validator/validators.go
```diff
@@ -76,13 +86,25 @@ func trafficTargetValidator(req *admissionv1.AdmissionRequest) (*admissionv1.Adm
...
 	if err := json.NewDecoder(bytes.NewBuffer(req.Object.Raw)).Decode(ingressBackend); err != nil {
 		return nil, err
 	}
 
+	type backendCacheKey struct {
+		name string
+		port int
+	}
+
+	backends := make(map[backendCacheKey]struct{}, len(ingressBackend.Spec.Backends))
```
shashankram, line 100: Use `mapset` here. If `mapset.Add()` returns true, it implies the item was added and does not contain a duplicate.

pkg/validator/validators.go
```diff
@@ -136,9 +160,52 @@ func ingressBackendValidator(req *admissionv1.AdmissionRequest) (*admissionv1.Ad
...
 	}
 
+	err := kc.checkDuplicateBackends(ingressBackend)
+	if errors.Is(err, errIngressBackendDuplicateBackends) {
+		return nil, err
+	}
+
+	// Fail open for transient errors, but log
+	if err != nil {
+		log.Error().Err(err).Msg("transient error encountered checking duplicate IngressBackend backends")
```
shashankram, line 170: This log message seems ambiguous. Could you reword this to remove the usage of the word `transient`.

pkg/validator/validators.go
```diff
@@ -136,9 +160,52 @@ func ingressBackendValidator(req *admissionv1.AdmissionRequest) (*admissionv1.Ad
...
+		}
+	}
+
+	if conflictString.Len() != 0 {
+		return fmt.Errorf("%w\n%s", errIngressBackendDuplicateBackends, conflictString.String())
+	}
+
+	return nil
+}
+
```
shashankram, line 208: The logic here can be simplified by checking for duplicates inline while processing the backend. Can something like this be done:
When you see a backend, check the cache to see if the backend is already referenced in some other ingress resource. If it is, return an error. Otherwise, the backend is new.

Why do we need to match the entire IngressBackend object with all other IngressBackend objects?

shashankram (changes requested):
Thanks for adding these. Left a few comments, otherwise looks good.

pkg/validator/types.go
keithmattix: These error vars are wrapped [later on](https://github.com/openservicemesh/osm/pull/4640/files#diff-ca09fc1c4167db01ac42e4c120198cbd797d234b280b9986266ac0440b016289R203) so as to add more context. The main reason for explicitly creating types here is so that we can use `errors.Is()` in order to [decide](https://github.com/openservicemesh/osm/pull/4640/files#diff-ca09fc1c4167db01ac42e4c120198cbd797d234b280b9986266ac0440b016289R164) whether or not  to fail the validating webhook

pkg/validator/validators.go
shashankram: The error path here should be deterministic. I see why it's not the case now as you are using `List()` below, but we could simply rely on the cache of objects already present in `policy/client.go`.

pkg/validator/validators.go
```diff
@@ -2,18 +2,24 @@ package validator
...
+	policyClientset "github.com/openservicemesh/osm/pkg/gen/client/policy/clientset/versioned"
+	"github.com/openservicemesh/osm/pkg/policy"
 	"github.com/pkg/errors"
 	smiAccess "github.com/servicemeshinterface/smi-sdk-go/pkg/apis/access/v1alpha3"
 	smiSpecs "github.com/servicemeshinterface/smi-sdk-go/pkg/apis/specs/v1alpha4"
 	admissionv1 "k8s.io/api/admission/v1"
 
 	configv1alpha2 "github.com/openservicemesh/osm/pkg/apis/config/v1alpha2"
 	policyv1alpha1 "github.com/openservicemesh/osm/pkg/apis/policy/v1alpha1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
jaellio, line 22:
```suggestion
	admissionv1 "k8s.io/api/admission/v1"
	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"

	configv1alpha2 "github.com/openservicemesh/osm/pkg/apis/config/v1alpha2"
	policyv1alpha1 "github.com/openservicemesh/osm/pkg/apis/policy/v1alpha1"
```

keithmattix: Fixed!

pkg/validator/validators.go
keithmattix: Fixed as we discussed offline

pkg/validator/types.go
keithmattix: Removed since validators were made to be deterministic

pkg/validator/validators.go
```diff
@@ -76,13 +86,24 @@ func trafficTargetValidator(req *admissionv1.AdmissionRequest) (*admissionv1.Adm
...
+		port int
+	}
+
+	backends := mapset.NewSet()
 	for _, backend := range ingressBackend.Spec.Backends {
+		unique := backends.Add(setEntry{backend.Name, backend.Port.Number})
+		if !unique {
+			return nil, errors.Errorf("Duplicate backends detected with service name: %s and port: %d", backend.Name, backend.Port.Number)
+		}
+
```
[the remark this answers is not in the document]
keithmattix: Good catch - added to ingress backend but not upstreamtrafficsetting for readability

pkg/validator/validators.go
```diff
@@ -136,9 +156,53 @@ func ingressBackendValidator(req *admissionv1.AdmissionRequest) (*admissionv1.Ad
...
+	if err != nil {
+		return nil, err
+	}
+
 	return nil, nil
 }
 
+// checkDuplicateBackends checks to see if the backends present in the given
+// ingressBackend conflict with existing ingressBackends
+func (kc *policyValidator) checkDuplicateBackends(candidateIngressBackend *policyv1alpha1.IngressBackend) error {
```
shashankram, line 169: We shouldn't need this if we use the API in policy.Controller to lookup an existing IngressBackend matching the same backend?

pkg/validator/validators.go
```diff
@@ -182,6 +246,54 @@ func egressValidator(req *admissionv1.AdmissionRequest) (*admissionv1.AdmissionR
...
+	}
+
+	err := kc.checkDuplicateHosts(upstreamTrafficSetting)
+	if errors.Is(err, errUpstreamTrafficSettingDuplicateHosts) {
+		return nil, err
+	}
+
+	if err != nil {
+		return nil, err
+	}
```
shashankram, line 268: Let's replace this with a check `GetUpstreamTrafficSetting(opt)` where opt is of type UpstreamTrafficSettingGetOpt and supports lookups in the case based on host. This means you need to add a new field `Host` to the `UpstreamTrafficSettingGetOpt` type.

Once done, this can be simplified to:
```
if obj := policyController.GetUpstreamTrafficSetting(opt); obj != nil {
    // duplicate
}
```

pkg/validator/validators.go
keithmattix: Ah yep, I was duplicating the loop; I see now where `ingressBackendValidator` already loops through the list of backends. The end result is much simpler 👍🏾

pkg/validator/validators.go
```diff
@@ -76,13 +86,46 @@ func trafficTargetValidator(req *admissionv1.AdmissionRequest) (*admissionv1.Adm
...
+		}
+
+		fakeMeshSvc := service.MeshService{
+			Name:       backend.Name,
+			TargetPort: uint16(backend.Port.Number),
+			Protocol:   backend.Port.Protocol,
+		}
+
+		if matchingPolicy := kc.policyClient.GetIngressBackendPolicy(fakeMeshSvc); matchingPolicy != nil {
+			// we've found a duplicate
```
shashankram, line 116: Another check is required here to ensure `matchingPolicy.Name != ingressBackend.Name`, which would be the case for an `Update` API validation.

shashankram (commented):
Looks great, just 1 major comment to address.

pkg/validator/validators.go
keithmattix: Ah, fantastic catch - I'd forgotten that validation webhooks are run on update as well. I fixed these and added test cases to confirm that they work as expected 👍🏾

pkg/policy/client.go
```diff
@@ -162,7 +162,8 @@ func (c client) GetIngressBackendPolicy(svc service.MeshService) *policyV1alpha1
 		// Multiple IngressBackend policies for the same backend will be prevented
 		// using a validating webhook.
 		for _, backend := range ingressBackend.Spec.Backends {
-			if backend.Name == svc.Name {
+			// we need to check ports to allow ingress to multiple ports on the same svc
+			if backend.Name == svc.Name && backend.Port.Number == int(svc.TargetPort) && backend.Port.Protocol == svc.Protocol {
```
shashankram, line 166: Let's drop the protocol check here as it breaks existing behavior. The protocol on a service represents its protocol for east-west traffic (pod to pod), not north-south (external to pod). So a service with `http` as its internal mesh protocol can still use `https` as the protocol with ingress.

pkg/policy/client.go
```diff
@@ -208,6 +209,10 @@ func (c client) GetUpstreamTrafficSetting(options UpstreamTrafficSettingGetOpt)
 	for _, resource := range c.caches.upstreamTrafficSetting.List() {
 		upstreamTrafficSetting := resource.(*policyV1alpha1.UpstreamTrafficSetting)
 
+		if upstreamTrafficSetting.Spec.Host == options.Host {
```
shashankram, line 212: Need a corresponding check in L194.

pkg/validator/validators.go
```diff
@@ -76,13 +86,46 @@ func trafficTargetValidator(req *admissionv1.AdmissionRequest) (*admissionv1.Adm
...
+		port int
+	}
 
+	backends := mapset.NewSet()
+	var conflictString strings.Builder
+	conflictingIngressBackends := mapset.NewSet()
 	for _, backend := range ingressBackend.Spec.Backends {
+		if unique := backends.Add(setEntry{backend.Name, backend.Port.Number}); !unique {
+			return nil, errors.Errorf("Duplicate backends detected with service name: %s and port: %d", backend.Name, backend.Port.Number)
+		}
```
shashankram, line 107: Why do we need this? Can't this be simplified to returning early in case of a duplicate:
```
for ingressBackend in ingressBackendList:
  if duplicate detected:
      return error(...)
```

shashankram (changes requested):

pkg/validator/validators.go
```diff
@@ -108,6 +151,10 @@ func ingressBackendValidator(req *admissionv1.AdmissionRequest) (*admissionv1.Ad
 		}
 	}
 
+	if conflictString.Len() != 0 {
+		return nil, fmt.Errorf("duplicate backends detected\n%s", conflictString.String())
+	}
+
```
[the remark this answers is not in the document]
keithmattix: @shashankram This and your previous comment seem to be talking about the same thing, so I'm just going to respond to it once. My thought process was that a more detailed error message that tells the customer all of their duplicates at once is a better developer experience than having to apply, fix, apply, fix, etc. I don't personally feel like the code is too complicated, but I'm naturally biased. I'm not too tied to the implementation either way, so if you feel strongly about it, I can change it.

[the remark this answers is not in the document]
shashankram: Not a blocker. Feel free to ignore.

shashankram (approved):

jaellio (approved):

PR #4645, opened 2022-04-12
feat(certificates) rework cert manager, integrate rotor

schristoff:
**Description**:
Builds on the work done in https://github.com/openservicemesh/osm/pull/4580 - this removes the old `Manager` interface, and exports the new `manager` interface, adds `fake_manager.go` to help with testing where needed - as the previous mockCertManager could not be used anymore. Rotor was moved into `manager.go` and is started whenever `NewManager()` is called. 

-->
**Affected area**:
| Certificate Management     | [X ] |

no

2. Is this a breaking change?
no


steeling (approved):

pkg/certificate/manager.go
```diff
@@ -16,30 +17,84 @@ var (
...
+	m.stopRotation = done
+	go func() {
+		for {
+			select {
+			case <-done:
+				ticker.Stop()
+				return
+			default:
+				m.checkAndRotate()
+				<-ticker.C
```
nojnhuh, line 57: As-is, it looks like this will always block until the next tick after `checkAndRotate()`, whereas this would see a done signal immediately after `checkAndRotate()`:

```suggestion
			case <-ticker.C:
				m.checkAndRotate()
```

If this needs to happen before any ticks come through like it would now, we could add a `m.checkAndRotate()` before this for loop.

pkg/certificate/manager.go
```diff
@@ -16,30 +17,84 @@ var (
...
+	return m, nil
+}
 
-	// TODO(#4533) start the cert rotation here.
+func (m *Manager) start(checkInterval time.Duration) {
+	// iterate over the list of certificates
+	// when a cert needs to be rotated - call RotateCertificate()
+	ticker := time.NewTicker(checkInterval)
+	done := make(chan bool)
+	m.stopRotation = done
```
nojnhuh, line 48: I don't think this is doing quite what you want. I would expect this channel to already exist and be passed in to `NewManager` vs. being created in this function. Then the caller can `close` the channel when it wants this to stop. Otherwise I don't see where this is can be triggered to stop besides the uncalled `stop()` method.

pkg/certificate/manager.go
```diff
@@ -16,30 +17,84 @@ var (
...
+	m.start(checkCertificateExpirationInterval)
+	return m, nil
+}
 
-	// TODO(#4533) start the cert rotation here.
+func (m *Manager) start(checkInterval time.Duration) {
+	// iterate over the list of certificates
+	// when a cert needs to be rotated - call RotateCertificate()
+	ticker := time.NewTicker(checkInterval)
+	done := make(chan bool)
```
nojnhuh, line 47: For things that are just plain signals without any other data associated, let's use a `chan struct{}` to make that more obvious. Otherwise it's not clear that `false` on this channel also happens to mean "stop."

nojnhuh (changes requested):

pkg/certificate/manager.go
```diff
@@ -44,24 +46,21 @@ func (m *Manager) start(checkInterval time.Duration) {
...
-				<-ticker.C
 			}
 		}
 	}()
 }
 
 func (m *Manager) stop() {
-	m.stopRotation <- true
+	<-m.certRotation
 }
```
nojnhuh, line 64: When this needs to stop, callers can `close()` the channel passed in when this was initialized, so we don't need a separate function for this.

pkg/certificate/manager.go
```diff
@@ -24,7 +24,8 @@ func NewManager(
 	ca *Certificate,
 	client client,
 	serviceCertValidityDuration time.Duration,
-	msgBroker *messaging.Broker) (*Manager, error) {
+	msgBroker *messaging.Broker,
+	certRotation chan struct{}) (*Manager, error) {
```
nojnhuh, line 28: To indicate this will never send anything on the channel:
```suggestion
	certRotation <-chan struct{}) (*Manager, error) {
```

pkg/certificate/manager.go
```diff
@@ -16,30 +17,79 @@ var (
...
 	}
+	m.start(checkCertificateExpirationInterval)
+	return m, nil
+}
 
-	// TODO(#4533) start the cert rotation here.
+func (m *Manager) start(checkInterval time.Duration) {
+	// iterate over the list of certificates
+	// when a cert needs to be rotated - call RotateCertificate()
+	ticker := time.NewTicker(checkInterval)
```
steeling, line 48: piggybacking jon's suggestion let's have a call to m.checkAndRotate to start this off

steeling: you can put it in the goroutine, prior to the for loop

pkg/certificate/manager_test.go
```diff
@@ -88,8 +88,6 @@ var _ = Describe("Test Rotor", func() {
 			close(done)
 
 			fmt.Printf("It took %+v to rotate certificate %s\n", time.Since(start), cn)
-			//stop rotation
-			certManager.stop()
```
nojnhuh: If cert rotation still needs to stop here, that would look like
```go
close(certManager.certRotation)
```

Or if the existing `stop` channel is used to initialize this, then that will already be closed so you don't need to explicitly stop it here.

pkg/certificate/manager_test.go
```diff
@@ -68,14 +56,55 @@ var _ = Describe("Test Certificate Manager", func() {
...
+	cn := CommonName("foo")
+
+	Context("Testing rotating expiring certificates", func() {
+
+		validityPeriod := -1 * time.Hour // negative time means this cert has already expired -- will be rotated asap
+
+		stop := make(chan struct{})
+		defer close(stop)
+		msgBroker := messaging.NewBroker(stop)
+		certManager, err := NewManager(&Certificate{}, &fakeIssuer{}, validityPeriod, msgBroker, make(chan struct{}))
```
nojnhuh, line 70: You could reuse the channel created earlier in this test to stop the message broker and cert rotation at the same time, unless it's important their lifetimes should be different (which I have a feeling isn't the case):
```suggestion
		certManager, err := NewManager(&Certificate{}, &fakeIssuer{}, validityPeriod, msgBroker, stop)
```

pkg/certificate/fake_manager.go
```diff
@@ -0,0 +1,39 @@
...
+}
+
+//FakeCertManager is a testing helper that returns a *certificate.Manager
+func FakeCertManager() (*Manager, error) {
+	cm, err := NewManager(
+		caCert,
+		&fakeIssuer{},
+		validity,
+		nil,
+		make(chan struct{}),
```
nojnhuh, line 33: Creating the channel here means nothing calling this from outside the package is able to access it to close it and stop the goroutine rotating certs. I'm not sure exactly what happens when you create this in a test and don't stop cert rotation, but I imagine it probably doesn't matter unless we're creating 1000s of these in a single test which is definitely not the case.

pkg/certificate/providers/config.go
```diff
@@ -313,18 +313,17 @@ func (c *Config) getHashiVaultOSMCertificateManager(options VaultOptions) (certi
 		return nil, nil, fmt.Errorf("error getting Vault Root Certificate, got: %w", err)
 	}
 
-	certManager, err := certificate.NewManager(vaultCert, vaultClient, c.cfg.GetServiceCertValidityPeriod(), c.msgBroker)
+	certManager, err := certificate.NewManager(vaultCert, vaultClient, c.cfg.GetServiceCertValidityPeriod(), c.msgBroker, make(chan struct{}))
```
nojnhuh, line 316: Same as before, creating the channel here makes it impossible to access later to stop cert rotation.

Overall, I think in each `main()` where we create a cert manager, `Start` should be called right after that and then `Start` should take the channel in as a parameter so `NewCertManager` doesn't need to accept one in cases cert rotation isn't needed.

To be overly specific, I mean something like this:
```diff
diff --git a/cmd/osm-bootstrap/osm-bootstrap.go b/cmd/osm-bootstrap/osm-bootstrap.go
index 509997780..e736e4ff1 100644
--- a/cmd/osm-bootstrap/osm-bootstrap.go
+++ b/cmd/osm-bootstrap/osm-bootstrap.go
@@ -9,6 +9,7 @@ import (
 	"flag"
 	"fmt"
 	"os"
+	"time"
 
 	"github.com/pkg/errors"
 	"github.com/spf13/pflag"
@@ -182,6 +183,8 @@ func main() {
 		events.GenericEventRecorder().FatalEvent(err, events.InvalidCertificateManager,
 			"Error initializing certificate manager of kind %s", certProviderKind)
 	}
+	// watch for certificate rotation
+	certManager.Start(5*time.Second, stop)
 
 	// Initialize the crd conversion webhook server to support the conversion of OSM's CRDs
 	crdConverterConfig.ListenPort = constants.CRDConversionWebhookPort
diff --git a/pkg/certificate/manager.go b/pkg/certificate/manager.go
index d936228f4..4a3a576b8 100644
--- a/pkg/certificate/manager.go
+++ b/pkg/certificate/manager.go
@@ -22,8 +22,7 @@ func NewManager(
 	ca *Certificate,
 	client client,
 	serviceCertValidityDuration time.Duration,
-	msgBroker *messaging.Broker,
-	certRotation <-chan struct{}) (*Manager, error) {
+	msgBroker *messaging.Broker) (*Manager, error) {
 	if ca == nil {
 		return nil, errNoIssuingCA
 	}
@@ -34,17 +33,16 @@ func NewManager(
 		client:                      client,
 		serviceCertValidityDuration: serviceCertValidityDuration,
 		msgBroker:                   msgBroker,
-		certRotation:                certRotation,
 	}
 	return m, nil
 }
 
 // Start takes an interval to check if the certificate
 // needs to be rotated
-func (m *Manager) Start(checkInterval time.Duration) {
+func (m *Manager) Start(checkInterval time.Duration, certRotation <-chan struct{}) {
 	// iterate over the list of certificates
 	// when a cert needs to be rotated - call RotateCertificate()
-	if m.certRotation == nil {
+	if certRotation == nil {
 		log.Error().Msgf("Cannot start certificate rotation, certRotation is nil")
 		return
 	}
@@ -52,7 +50,7 @@ func (m *Manager) Start(checkInterval time.Duration) {
 	go func() {
 		for {
 			select {
-			case <-m.certRotation:
+			case <-certRotation:
 				ticker.Stop()
 				return
 			case <-ticker.C:
diff --git a/pkg/certificate/providers/config.go b/pkg/certificate/providers/config.go
index ef5edf31d..4d807ac96 100644
--- a/pkg/certificate/providers/config.go
+++ b/pkg/certificate/providers/config.go
@@ -246,12 +246,10 @@ func (c *Config) getTresorOSMCertificateManager() (*certificate.Manager, debugge
 		return nil, nil, errors.Errorf("Failed to instantiate Tresor as a Certificate Manager")
 	}
 
-	tresorCertManager, err := certificate.NewManager(rootCert, tresorClient, c.cfg.GetServiceCertValidityPeriod(), c.msgBroker, make(chan struct{}))
+	tresorCertManager, err := certificate.NewManager(rootCert, tresorClient, c.cfg.GetServiceCertValidityPeriod(), c.msgBroker)
 	if err != nil {
 		return nil, nil, fmt.Errorf("error instantiating osm certificate.Manager for Tresor cert-manager : %w", err)
 	}
-	// watch for certificate rotation
-	tresorCertManager.Start(checkCertificateExpirationInterval)
 	return tresorCertManager, tresorCertManager, nil
 }
```

pkg/certificate/manager_test.go
```diff
@@ -68,14 +54,55 @@ var _ = Describe("Test Certificate Manager", func() {
...
+
+	Context("Testing rotating expiring certificates", func() {
+
+		validityPeriod := -1 * time.Hour // negative time means this cert has already expired -- will be rotated asap
+
+		stop := make(chan struct{})
+		defer close(stop)
+		msgBroker := messaging.NewBroker(stop)
+		certManager, err := NewManager(&Certificate{}, &fakeIssuer{}, validityPeriod, msgBroker)
+		certManager.Start(5*time.Second, stop)
```
nojnhuh, line 69: If we're not running `checkAndRotate` at the beginning of the loop this kicks off, then this 5s is basically a lower bound on how fast the test can run. I'd either do a `checkAndRotate` immediately in the loop or make this timeout (only in this test here) something way smaller, like 10ms.

pkg/certificate/manager_test.go
```diff
@@ -68,14 +54,55 @@ var _ = Describe("Test Certificate Manager", func() {
...
+
+			fmt.Printf("It took %+v to rotate certificate %s\n", time.Since(start), cn)
+			newCert, err := certManager.IssueCertificate(cn, validityPeriod)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(newCert.GetExpiration()).ToNot(Equal(certA.GetExpiration()))
+			Expect(newCert).ToNot(Equal(certA))
+		})
+	})
+
+})
```
nojnhuh, line 96: Since I was playing with this anyway, here's where I ended up with this test:
```go
func TestRotor(t *testing.T) {
	assert := tassert.New(t)

	cn := CommonName("foo")
	validityPeriod := -1 * time.Hour // negative time means this cert has already expired -- will be rotated asap

	stop := make(chan struct{})
	defer close(stop)
	msgBroker := messaging.NewBroker(stop)
	certManager, err := NewManager(&Certificate{}, &fakeIssuer{}, validityPeriod, msgBroker)
	certManager.Start(5*time.Second, stop)
	assert.NoError(err)

	certA, err := certManager.IssueCertificate(cn, validityPeriod)
	assert.NoError(err)
	certRotateChan := msgBroker.GetCertPubSub().Sub(announcements.CertificateRotated.String())

	start := time.Now()
	// Wait for one certificate rotation to be announced and terminate
	<-certRotateChan

	fmt.Printf("It took %+v to rotate certificate %s\n", time.Since(start), cn)
	newCert, err := certManager.IssueCertificate(cn, validityPeriod)
	assert.NoError(err)
	assert.NotEqual(certA.GetExpiration(), newCert.GetExpiration())
	assert.NotEqual(certA, newCert)
}
```

pkg/certificate/manager_test.go
```diff
@@ -68,14 +54,55 @@ var _ = Describe("Test Certificate Manager", func() {
...
+
+		certA, err := certManager.IssueCertificate(cn, validityPeriod)
+		certRotateChan := msgBroker.GetCertPubSub().Sub(announcements.CertificateRotated.String())
+
+		It("issued a new certificate", func() {
+			Expect(err).ToNot(HaveOccurred())
+		})
+
+		It("rotates certificate", func() {
+			certManager.Start(5*time.Second, stop)
```
nojnhuh, line 80: Rotation was already started earlier in this test.

pkg/certificate/manager_test.go
schristoff: Oh I thought I fixed this but I stashed it - adding this outside the for loop 🤦‍♀️

cmd/osm-bootstrap/osm-bootstrap.go
```diff
@@ -178,10 +179,13 @@ func main() {
 		caBundleSecretName, tresorOptions, vaultOptions, certManagerOptions, msgBroker)
 
 	certManager, _, err := certProviderConfig.GetCertificateManager()
+	certManager.Start(5*time.Second, stop)
```
nojnhuh, line 182: This doesn't need to be started twice, so I'd remove this one.

nojnhuh (approved):

pkg/certificate/fake_manager.go
```diff
@@ -0,0 +1,38 @@
...
+	caCert = &Certificate{
+		CommonName: "Test CA",
+		Expiration: time.Now().Add(time.Hour * 24),
+	}
+	validity = time.Hour
+)
+
+type fakeIssuer struct{}
+
+//IssueCertificate is a testing helper to satisfy the certificate client interface
```
jaellio, line 18: Total nit:
```suggestion
// IssueCertificate is a testing helper to satisfy the certificate client interface
```

pkg/certificate/manager_test.go
```diff
@@ -1,38 +1,26 @@
...
 	tassert "github.com/stretchr/testify/assert"
-)
 
-var (
-	caCert = &Certificate{
-		CommonName: "Test CA",
-		Expiration: time.Now().Add(time.Hour * 24),
-	}
+	"github.com/openservicemesh/osm/pkg/announcements"
+	"github.com/openservicemesh/osm/pkg/messaging"
```
jaellio, line 13:
```suggestion
	
	"github.com/openservicemesh/osm/pkg/announcements"
	"github.com/openservicemesh/osm/pkg/messaging"
```

pkg/certificate/fake_manager.go
```diff
@@ -0,0 +1,38 @@
...
+
+// IssueCertificate is a testing helper to satisfy the certificate client interface
+func (i *fakeIssuer) IssueCertificate(cn CommonName, validityPeriod time.Duration) (*Certificate, error) {
+	return &Certificate{
+		CommonName: cn,
+		Expiration: time.Now().Add(validityPeriod),
+	}, nil
+}
+
+//FakeCertManager is a testing helper that returns a *certificate.Manager
```
jaellio, line 26: nit: Use a space following `//`
```suggestion
// FakeCertManager is a testing helper that returns a *certificate.Manager
```

jaellio (approved):

PR #4650, opened 2022-04-13
fix potential bug in where we copy a lock value

steeling:
My linter just picked up the issue that we're passing around a proto by value vs pointer. This has the issue that protos have locks embedded in them which we shouldn't copy.

https://github.com/golang/tools/blob/master/go/analysis/passes/copylock/copylock.go


nojnhuh (commented):
I already made these changes in #4629: https://github.com/openservicemesh/osm/pull/4629/files#diff-c7eba6dc1d911d815481aa030ad94011e2aea15d63cf66810b4c3f4fbf37c371

PR #4655, opened 2022-04-15
Abstract ca root storage

steeling:
Abstract the storage portion of storing/retrieving the CA cert from k8s secrets in prep for the AKV integration.

Issue #4501, opened 2022-02-02
Integrate with Azure Key Vault


snehachhabria:
**Please describe the Improvement and/or Feature Request**
Integrate OSM with Azure Key vault : https://docs.microsoft.com/en-us/azure/key-vault/general/basic-concepts

The tasks include:
- [ ] Research how AKV can be used with OSM
- [ ] Design doc (a Google doc linked to this item)
- [ ] Implementation
- [ ] Unit and e2e tests
- [ ] Documentation: how-to guide
- [ ] Documentation: demo

**Scope (please mark with X where applicable)**
- New Functionality          [X]

steeling:
Looking at the code base, I've identified some refactoring opportunities prior to adding a new cert manager:

- [ ] [Remove certificate.Certificater](https://github.com/openservicemesh/osm/issues/4532)
- [ ] [Leverage a new certificate.Administrator type](https://github.com/openservicemesh/osm/issues/4533)
- [ ] [Switch to options interface, remove provider.Config, factory-style code](https://github.com/openservicemesh/osm/issues/4534)

steeling:
closing this for now, as the recommendation would be to use cert-manager or CSI driver

PR #4657, opened 2022-04-15
refactor k8s root ca secret access

steeling:
Refactor the k8s secret access for root ca's. This is done strictly to make subsequent PR's smaller (#4655), and easier to review.

This is essentially just moving code around for the purpose of avoiding cyclic dependencies, with a couple renames, but almost 0 functionality change.

The one change is abstracting NewFromPem, to work for both tresor and certmanager to accept a nil privatekey. It is now up to the caller to validate whether the private key is present if it is required.


steeling:
part of #4501

steeling:
Let's get #4645 in before this, and I can do any rebasing on my end

cmd/osm-bootstrap/osm-bootstrap.go
```diff
@@ -174,10 +174,8 @@ func main() {
 	cfg := configurator.NewConfigurator(configClient, stop, osmNamespace, osmMeshConfigName, msgBroker)
 
 	// Intitialize certificate manager/provider
-	certProviderConfig := providers.NewCertificateProviderConfig(kubeClient, kubeConfig, cfg, providers.Kind(certProviderKind), osmNamespace,
+	certManager, _, _, err := providers.GenerateCertificateManager(kubeClient, kubeConfig, cfg, providers.Kind(certProviderKind), osmNamespace,
```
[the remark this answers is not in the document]
steeling: That's a good point! I ended up with the rename since the ProviderConfig went away, and theres' already a `certificate.NewCertManager` function.

I'd probably ere on the side of keeping it to remove that confusion, with the added caveat that this function is also going to go away completely in the near future.

pkg/certificate/castorage/k8s/k8s.go
```diff
@@ -0,0 +1,97 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/openservicemesh/osm/pkg/certificate"
+	"github.com/openservicemesh/osm/pkg/constants"
+	"github.com/openservicemesh/osm/pkg/errcode"
+	"github.com/openservicemesh/osm/pkg/version"
+)
+
+// GetCertFromKubernetes is a helper function that loads a certificate from a Kubernetes secret
+// The function returns an error only if a secret is found with invalid data.
```
keithmattix, line 20: nit: this comment is a bit misleading. This function also returns an error if a secret could not be retrieved

steeling: good catch, fixed!

jaellio (approved):

shashankram (approved):

PR #4664, opened 2022-04-19
refactor(cmd/cli): update uninstall cmd

### LINKED ISSUE #4613: Uninstall command should suggest existing mesh to uninstall

Right now `osm uninstall mesh` will show the meshes present in the cluster and defaults to suggesting to uninstall mesh name 'osm' in namespace 'osm-system' - this gives the impression that there is such a mesh in the 'osm-system' namespace even when there is not
image
Instead it should suggest to uninstall the existing mesh(es) if the user runs `osm uninstall mesh` instead of having to add the tags `--osm-namespace` and `--mesh-name` and the tags would only be used if the user knows the exact mesh they'd like to uninstall
image

- CLI Tool                   [x ]

**Possible use cases**

shalier:
**Description**:
Updates `uninstall mesh` command so that user is prompted on whether they want to uninstall an existing mesh. Also, adds a warning for `delete-cluster-wide-resources` if there are remaining meshes in the cluster, before deleting cluster resources. Resolves #4613

**Testing done**:
Uninstall mesh now prompts with existing mesh:
prompts_existing_mesh
Can also uninstall specific mesh with `mesh-name` flag:
specify_mesh_with_mesh_name
Informs when mesh doesn't exist (including when trying to force uninstall):
doesntExist_plus_force_flag
`delete-namespace` flag:
delete_namespace_flag
`delete-cluster-wide-resources` flag:
delete_cluster_wide_resources
warning when deleting cluster resources and there are still meshes in the cluster (updated):
uninstallCmd_cluster_wide_resources
`force` flag uninstalls existing mesh [only one mesh in cluster]:
single_mesh_force
`force` flag uninstalls the default mesh+namespace (osm+osm-system) unless `mesh-name` and `osm-namespace` are specified [multiple meshes]:
multi_mesh_force

**Affected area**:
| CLI Tool                   | [ x] |

3. Has documentation corresponding to this change been updated in the [osm-docs] (https://github.com/openservicemesh/osm-docs) repo (if applicable)?

### DIFF
--- a/cmd/cli/uninstall_mesh.go
+++ b/cmd/cli/uninstall_mesh.go
@@ -94,7 +94,7 @@ func newUninstallMeshCmd(config *action.Configuration, in io.Reader, out io.Writ
 	}
 
 	f := cmd.Flags()
-	f.StringVar(&uninstall.meshName, "mesh-name", defaultMeshName, "Name of the service mesh")
+	f.StringVar(&uninstall.meshName, "mesh-name", "", "Name of the service mesh")
 	f.BoolVarP(&uninstall.force, "force", "f", false, "Attempt to uninstall the osm control plane instance without prompting for confirmation.")
 	f.BoolVarP(&uninstall.deleteClusterWideResources, "delete-cluster-wide-resources", "a", false, "Cluster wide resources (such as osm CRDs, mutating webhook configurations, validating webhook configurations and osm secrets) are fully deleted from the cluster after control plane components are deleted.")
 	f.BoolVar(&uninstall.deleteNamespace, "delete-namespace", false, "Attempt to delete the namespace after control plane components are deleted")
@@ -118,9 +118,9 @@ func (d *uninstallMeshCmd) run() error {
 			fmt.Fprintf(d.out, "No OSM control planes found\n")
 			return nil
 		}
-		specifiedMeshFound := false
 		// Searches for the mesh specified by the mesh-name flag if specified
-		if d.meshName != defaultMeshName {
+		specifiedMeshFound := false
+		if d.meshName != "" {
 			specifiedMeshFound = d.findMesh(meshInfoList)
 			if !specifiedMeshFound {
 				return nil
@@ -174,6 +174,8 @@ func (d *uninstallMeshCmd) run() error {
 				if !d.deleteClusterWideResources && !d.deleteNamespace {
 					return err
 				}
+
+				fmt.Fprintf(d.out, "Error %v when trying to uninstall mesh name [%s] in namespace [%s] - continuing to deleteClusterWideResources and/or deleteNamespace\n", err, m.name, m.namespace)
 			}
 
 			if err == nil {
--- a/cmd/cli/uninstall_mesh.go
+++ b/cmd/cli/uninstall_mesh.go
@@ -33,7 +33,7 @@ The command will not delete:
 (1) the namespace the mesh was installed in unless specified via the
 --delete-namespace flag.
 (2) the cluster-wide resources (i.e. CRDs, mutating and validating webhooks and
-secrets) unless specified via via the --delete-cluster-wide-resources (or -a) flag
+secrets) unless specified via the --delete-cluster-wide-resources (or -a) flag
 
 Be careful when using this command as it is destructive and will
 disrupt traffic to applications left running with sidecar proxies.
@@ -53,6 +53,7 @@ type uninstallMeshCmd struct {
 	localPort                  uint16
 	deleteClusterWideResources bool
 	extensionsClientset        extensionsClientset.Interface
+	actionConfig               *action.Configuration
 }
 
 func newUninstallMeshCmd(config *action.Configuration, in io.Reader, out io.Writer) *cobra.Command {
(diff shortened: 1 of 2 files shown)


cmd/cli/uninstall_mesh_test.go
```diff
@@ -196,11 +201,9 @@ func TestUninstallCmd(t *testing.T) {
...
 			}
 
-			if test.meshExists {
-				rel := release.Mock(&release.MockReleaseOptions{Name: test.meshName})
-				err := store.Create(rel)
-				Expect(err).To(BeNil())
-			}
+			rel := release.Mock(&release.MockReleaseOptions{Name: testMeshName})
+			err := store.Create(rel)
+			Expect(err).To(BeNil())
```
nojnhuh, line 206: I know this is how it was before, but let's stick to using `testify` in non-ginkgo tests.

cmd/cli/uninstall_mesh_test.go
```diff
@@ -196,11 +201,9 @@ func TestUninstallCmd(t *testing.T) {
 				mem.SetNamespace(test.meshNamespace)
 			}
 
-			if test.meshExists {
```
nojnhuh, line 199: I'd lean towards keeping this the way it was since with this change, there's nothing stopping you from defining a test case with a mesh name that isn't `testMeshName` and `meshExists: true` (or vice versa) which won't actually be the case if whether or not the mesh actually exists is based solely on the mesh name.

cmd/cli/uninstall_mesh.go
```diff
@@ -105,92 +107,129 @@ func newUninstallMeshCmd(config *action.Configuration, in io.Reader, out io.Writ
...
+			err := d.actionConfig.Init(settings.RESTClientGetter(), m.namespace, "secret", debug)
+			if err != nil {
 				return err
 			}
-		}
+			_, err = d.client.Run(m.name)
+			if err != nil && errors.Cause(err) == helmStorage.ErrReleaseNotFound && !d.deleteClusterWideResources && !d.deleteNamespace {
+				fmt.Fprintf(d.out, "Did not find mesh [%s] in namespace [%s]\n", d.meshName, d.meshNamespace)
+				return err
+			}
```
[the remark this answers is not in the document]
shalier: Changed it to:
```
if err != nil {
	if errors.Cause(err) == helmStorage.ErrReleaseNotFound {
		fmt.Fprintf(d.out, "No OSM control plane with mesh name [%s] found in namespace [%s]\n", m.name, m.namespace)
	}

	if !d.deleteClusterWideResources && !d.deleteNamespace {
		return err
	}
```

cmd/cli/uninstall_mesh.go
```diff
@@ -105,92 +107,140 @@ func newUninstallMeshCmd(config *action.Configuration, in io.Reader, out io.Writ
...
-		err = d.uninstallValidatingWebhookConfigurations()
-		if err != nil {
-			failedDeletions = append(failedDeletions, "ValidatingWebhookConfigurations")
-		}
-
-		err = d.uninstallSecrets()
-		if err != nil {
-			failedDeletions = append(failedDeletions, "Secrets")
+		if len(meshInfoList) != 0 {
+			fmt.Fprintf(d.out, "Deleting cluster resources will affect current mesh(es) in cluster:\n")
```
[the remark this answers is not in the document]
shalier: @jaellio atm, we don't support multiple meshes in the cluster, but if in the future we do choose to support multiple meshes then we should

jaellio (approved):

cmd/cli/uninstall_mesh.go
```diff
@@ -105,92 +107,140 @@ func newUninstallMeshCmd(config *action.Configuration, in io.Reader, out io.Writ
...
 				return err
 			}
-		}
 
-		_, err := d.client.Run(d.meshName)
-		if err != nil && errors.Cause(err) == helmStorage.ErrReleaseNotFound {
-			fmt.Fprintf(d.out, "No OSM control plane with mesh name [%s] found in namespace [%s]\n", d.meshName, d.meshNamespace)
+			_, err = d.client.Run(m.name)
+			if err != nil {
+				if errors.Cause(err) == helmStorage.ErrReleaseNotFound {
```
nojnhuh, line 170: It looks like this is still silently passing over errors other than "not found" when the `--delete-cluster-wide-resources` or `--delete-namespace` flags are passed.

cmd/cli/uninstall_mesh_test.go
```diff
@@ -542,6 +567,10 @@ func TestUninstallClusterWideResources(t *testing.T) {
 				Log:          func(format string, v ...interface{}) {},
 			}
 
+			fakeClientSet := fake.NewSimpleClientset(existingKubeClientsetObjects...)
+			_, err = addDeployment(fakeClientSet, "osm-controller-1", testMeshName, testNamespace, osmTestVersion, true)
+			Expect(err).NotTo(HaveOccurred())
```
nojnhuh, line 572: Let's stick with any new checks in these tests using `testify` instead. Then the `RegistryFailHandler` line above can be removed too.

nojnhuh (approved):

PR #4667, opened 2022-04-19
tests: move fakes to own sub-package

shashankram:
**Description**:
Currently, fake implementations used for testing
reside in the actual pkg being tested. This is
problematic because fakes and their imports
get imported when the real pkg is imported, which
is undesirable. The side effect of this is that
some fakes (e.g. catalog fake) import Ginkgo
which has an `init()` function that executes
as soon as the pkg is imported, which should
not be a part of the actual binary built. The
result of this is that Ginkgo command line
flags get registered to the binary.

This change moves fakes to their own sub-package
under the parent pkg. It also fixes a test in
catalog/debugger to not rely on fakes to prevent
circular imports, and removes redundant tests
that are already covered by other tests.

**Testing done**:
Unit tests

**Affected area**:


trstringer (approved):
Oh wow, good catch.

jaellio (approved):

PR #4669, opened 2022-04-19
Feature/envoy eth0 forwarding

### LINKED ISSUE #4653: (*bug) Envoy connections fail if app container binds to eth0
**Bug description**:

OSM's Envoy sidecar currently proxies communication with sibling containers over localhost (i.e. 127.0.0.1). This means that backend applications that explicitly bind to the pod IP refuse connections from the Envoy proxy. According to Kubernetes conventions, applications should bind to the pod IP and not localhost; therefore, OSM should proxy traffic via the eth0 interface and not localhost.

**Expected behavior**:

Applications that bind to eth0 should be able to receive traffic from the Envoy sidecar proxy.

1. Install OSM on a cluster
2. Enable OSM in a namespace
3. deploy a webserver Deployment + Service that binds to eth0 (i.e. pod IP)
4. Try to access that service (either through LB, ingress, or portforward)

**How was OSM installed?**:
N/A

keithmattix:
**Description**:
Fixes #4653. Allows users to configure the OSM sidecar to communicate with the backend application via the pod IP instead of localhost. For now, this is accomplished via a DNAT iptables rule, but the implementation may change to utilize [Envoy Original DST clusters](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/service_discovery#original-destination) in the future. To reviewers: I'm sorry, it's probably more efficient to avoid commit [WIP: Batch rename configv1alpha2 to configv1alpha3](https://github.com/openservicemesh/osm/commit/37d98dd4ca904c29262f0e342f266878acca60eb). Please reach out to me if you'd prefer to walk through this synchronously due to the size. 

**Testing done**:
Update Unit tests & added e2e test case

**Affected area**:


shashankram (changes requested):
@keithmattix thanks for this change. The idea looks good.
I'd really like for the change to the MeshConfig API version to happen first in a separate PR as it's large enough, and also would need to handle conversion logic that is best reviewed independently. Moreover, given that the CRD versions are alpha, it is worth considering whether we can simply update v1alpha2 without needing to bump the version. While in practice it's best to bump versions, it seems to not be a requirement for alpha APIs. Could you please split these into 2 PRs so they are easier to review and merge as smaller units?

keithmattix:
Yeah good call out; I'll see if I can work some cherry-pick magic to split this up. Let's discuss the version bump in the separate PR

keithmattix:
Closing in favor of #4671 4671

PR #4673, opened 2022-04-19
add the last applied annotation to mesh config

steeling:
On both creation, and if it doesn't exist.

This will allow users to use declarative config management of their mesh config.


cmd/osm-bootstrap/osm-bootstrap.go
```diff
@@ -240,19 +245,24 @@ func (b *bootstrap) createDefaultMeshConfig() error {
...
+	config, err := b.meshConfigClient.ConfigV1alpha2().MeshConfigs(b.namespace).Get(context.TODO(), meshConfigName, metav1.GetOptions{})
 	if apierrors.IsNotFound(err) {
 		// create a default mesh config since it was not found
-		if err = b.createDefaultMeshConfig(); err != nil {
+		return b.createDefaultMeshConfig()
+	} else if err != nil {
+		return err
+	}
+
+	if _, exists := config.Annotations[corev1.LastAppliedConfigAnnotation]; !exists {
```
nojnhuh, line 256: client-go has `Apply()` methods for all the resources it deals with: https://pkg.go.dev/k8s.io/client-go@v0.22.1/kubernetes/typed/core/v1#PodInterface

Is there an easy way for us to add those to our types vs. rolling this by hand?

go.mod
```diff
@@ -52,13 +52,13 @@ require (
 	gorm.io/driver/mysql v1.1.1
 	gorm.io/gorm v1.21.12
 	helm.sh/helm/v3 v3.7.1
-	k8s.io/api v0.22.2
+	k8s.io/api v0.23.5
```
nojnhuh, line 55: I'm certainly not opposed to bumping the k8s.io dependencies, but could you add your reasons why to the PR description/commit message? ("General upkeep" is a perfectly valid reason if that's the case, but just so we know what the implications are if we need to cherry-pick or otherwise move this commit around later.)

go.mod
```diff
@@ -52,13 +52,13 @@ require (
...
 	k8s.io/apiextensions-apiserver v0.22.1
```
nojnhuh, line 56: We might as well update this too just to keep things in sync as best we can.

go.mod
steeling: The previous CI run was complaining about `go mod tidy` so I ran that which must have bumped this. I'm guessing it was required to reconcile the new kubectl package?

I'll add a comment

cmd/osm-bootstrap/osm-bootstrap.go
steeling: that apply is different somehow.. not quite sure what that's for but I've been playing around with it, and it doesn't add that annotation.

It also requires us hand-rolling those MeshConfigApplyConfiguration types, which would be cumbersome.

Even then, we'd still have to use this method for migrating meshconfig with no annotation to with annotation, as that wouldn't add it.

FWIW this is how kubectl is doing it as well.

go.mod
steeling: done!

steeling:
friendly ping on this

shashankram (approved):

nojnhuh (approved):

PR #4671, opened 2022-04-19
config/meshConfig: New localProxyMode field

keithmattix:
**Description**:
Adds Go types & CRD manifests for v1alpha3 of MeshConfig. v1alpha3 adds the new spec.sidecar.localProxyConfig field that will allow users to modify the way in which the sidecar communicates with the backend application

**Testing done**:
Update Unit tests

**Affected area**:
| Functional Area            |     |
| -------------------------- | --- |
| Config                 | [X] |

Please answer the following questions with yes/no.

1. Does this change contain code from or inspired by another project? no
    -   Did you notify the maintainers and provide attribution? no

2. Is this a breaking change? no

3. Has documentation corresponding to this change been updated in the [osm-docs](https://github.com/openservicemesh/osm-docs) repo (if applicable)?

PR #4681, opened 2022-04-21
config/MeshConfig: Update control plane to v1alpha3

keithmattix:
Updates the OSM control plane to use the new v1alpha3 version, and
add a couple of convenience methods. Also, add install capabilities
by updating the Helm chart

Part of #4653 

**Description**:

**Testing done**:
Update units to reference new control plane version. Actual functionality will be tested in the next PR

**Affected area**:

3. Has documentation corresponding to this change been updated in the [osm-docs](https://github.com/openservicemesh/osm-docs) repo (if applicable)? Not yet applicable

### DIFF
--- a/charts/osm/README.md
+++ b/charts/osm/README.md
@@ -138,6 +138,7 @@ The following table lists the configurable parameters of the osm chart and their
 | osm.injector.replicaCount | int | `1` | Sidecar injector's replica count (ignored when autoscale.enable is true) |
 | osm.injector.resource | object | `{"limits":{"cpu":"0.5","memory":"64M"},"requests":{"cpu":"0.3","memory":"64M"}}` | Sidecar injector's container resource parameters |
 | osm.injector.webhookTimeoutSeconds | int | `20` | Mutating webhook timeout |
+| osm.localProxyMode | string | `"Localhost"` | Proxy mode for the Envoy proxy sidecar. Acceptable values are ['Localhost', 'PodIP'] |
 | osm.maxDataPlaneConnections | int | `0` | Sets the max data plane connections allowed for an instance of osm-controller, set to 0 to not enforce limits |
 | osm.meshName | string | `"osm"` | Identifier for the instance of a service mesh within a cluster |
 | osm.multicluster | object | `{"gatewayLogLevel":"error"}` | OSM multicluster feature configuration |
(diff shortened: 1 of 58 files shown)


steeling:
Is there a bug linked to this? Can you explain why we need a whole new version API? If we're just adding a field I would expect us to be able to add that to the existing API with no backwards incompatible changes.

We would just need to have our init container switch from a Create to an Apply when generating the CRD's

keithmattix:
@steeling adding the related issue; this is a multi-part PR. The reason I opted for a new version is because this is coming after a release (v1.1) and we're likely going to be releasing this feature as a patch. I think the explicitness is helpful even though the change would be backward-compatible

steeling:
> @steeling adding the related issue; this is a multi-part PR. The reason I opted for a new version is because this is coming after a release (v1.1) and we're likely going to be releasing this feature as a patch. I think the explicitness is helpful even though the change would be backward-compatible

There's a lot of overhead in maintaining multiple versions. And reasoning about version change compatibility can be tricky. I'd push for a policy of only bumping versions on non-backwards compatible changes, as is [standard for CRD's](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api_changes.md#adding-a-field).

Reducing the number of API's we need to maintain is going to pay dividends down the road

keithmattix:
Fair point; I wasn't aware of that guidance in sig-architecture. I'd love a centralized guide for these conventions, but that's a separate matter.

So since #4671 has already been merged; what do you suggested we do? cc @shashankram

shashankram:
> > @steeling adding the related issue; this is a multi-part PR. The reason I opted for a new version is because this is coming after a release (v1.1) and we're likely going to be releasing this feature as a patch. I think the explicitness is helpful even though the change would be backward-compatible
> 
> There's a lot of overhead in maintaining multiple versions. And reasoning about version change compatibility can be tricky. I'd push for a policy of only bumping versions on non-backwards compatible changes, as is [standard for CRD's](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api_changes.md#adding-a-field).
> 
> Reducing the number of API's we need to maintain is going to pay dividends down the road

In general, I agree with reducing the burden by not bumping API versions for backward compatible fields. When we bumped the version from v1alpha1->v1alpha2 previously, it was done because there were several new fields, and also because we wanted to exercise the conversion webhook path, so it made sense back then.

I am not opposed to bumping the version here, but this should be accompanied by completely removing `v1alpha1` types from the code and docs, we we don't have an ever increasing list of versions to maintain.

shashankram:
> Fair point; I wasn't aware of that guidance in sig-architecture. I'd love a centralized guide for these conventions, but that's a separate matter.
> 
> So since #4671 has already been merged; what do you suggested we do? cc @shashankram

@keithmattix since this seems to be well documented in the API convention, let's revert #4671 and add the field to the v1alpha2 version. @steeling, thanks for sharing the API guide regarding this.

keithmattix:
Closing due to discussion.

PR #4683, opened 2022-04-22
test((benchmark): add Golang benchmark test cases

### LINKED ISSUE #4622: [Root] OSM performance and scalability test in v1.2

This is the root Github issue for tracking the progress and plan of OSM performance benchmarking and scalability test at the release of v1.2. The scripts o the test resides in the [osm-perftest](https://github.com/openservicemesh/osm-perftest) repo.

### Setup & Automation

- [x] [load generator send requests to multiple endpoints](https://github.com/openservicemesh/osm-perftest/issues/28)
- [x] [[load generator] support setting replicas](https://github.com/openservicemesh/osm-perftest/issues/27)
- [x] [Gather Envoy metrics](https://github.com/openservicemesh/osm-perftest/issues/18)
- [x] [Gather full test reports with one click](https://github.com/openservicemesh/osm-perftest/issues/7)
- [x] [Run the same perf test scenario using other service mesh products.](https://github.com/openservicemesh/osm-perftest/issues/42)

### Performance

- [x] [Profiling OSM control plane performance](https://github.com/openservicemesh/osm-perftest/issues/31)
- [ ] [[WIP PR](https://github.com/openservicemesh/osm-docs/pull/367)] [Data plane performance change with number of connections](https://github.com/openservicemesh/osm/issues/2993)

### Scalability

- [ ] [ [WIP PR](https://github.com/openservicemesh/osm-docs/pull/367) ] [Additional scalability testing scenarios](https://github.com/openservicemesh/osm-perftest/issues/34)

### Profiling

- [x] Offline profiling: benchmark key process performance
- [x] [Document online profiling setup and process](https://github.com/openservicemesh/osm/issues/4694) [ [PR](https://github.com/openservicemesh/osm/pull/4744) ] 
- [x] Online profiling process

allenlsy:
**Description**:

Offline performance profiling is part of issue #4622. This PR adds benchmark test cases that profiles CPU usage, memory usage and program timeline.

* Added `make go-benchmark` make task. It runs all registered packages to be profiled.
* Each `go-benchmark` execution generates profile results which are saved in the `bm_profiles/<execution-timestamp>` folder, which is Git ignored. 
* For the ease of access the results, a symbol link to the latest results is created as `bm_profiles/latest`

TODO after this PR:

* Create a CI job to run benchmark before and after changes and output the performance change. This can be done with [`benchstat` ](https://pkg.go.dev/golang.org/x/perf/cmd/benchstat)

Partially resolve #4622 

**Testing done**:

**Affected area**:
| Performance                | [X] |

No

2. Is this a breaking change?

No

No

### DIFF
--- a/pkg/envoy/ads/response_benchmark_test.go
+++ b/pkg/envoy/ads/response_benchmark_test.go
@@ -1,0 +1,193 @@
+package ads
+
+import (
+	"context"
+	"fmt"
+	"testing"
+
+	mapset "github.com/deckarep/golang-set"
+	xds_discovery "github.com/envoyproxy/go-control-plane/envoy/service/discovery/v3"
+	"github.com/google/uuid"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	k8sClientFake "k8s.io/client-go/kubernetes/fake"
+
+	configv1alpha2 "github.com/openservicemesh/osm/pkg/apis/config/v1alpha2"
+	"github.com/openservicemesh/osm/pkg/catalog"
+	"github.com/openservicemesh/osm/pkg/endpoint"
+	"github.com/openservicemesh/osm/pkg/k8s/informers"
+	"github.com/openservicemesh/osm/pkg/logger"
+	"github.com/openservicemesh/osm/pkg/messaging"
+	"github.com/openservicemesh/osm/pkg/policy"
+	"github.com/openservicemesh/osm/pkg/providers/kube"
+	"github.com/openservicemesh/osm/pkg/signals"
+	"github.com/openservicemesh/osm/pkg/smi"
+
+	"github.com/openservicemesh/osm/pkg/certificate"
+
+	"github.com/openservicemesh/osm/pkg/configurator"
+	"github.com/openservicemesh/osm/pkg/constants"
+	"github.com/openservicemesh/osm/pkg/envoy"
+	"github.com/openservicemesh/osm/pkg/envoy/registry"
+	configFake "github.com/openservicemesh/osm/pkg/gen/client/config/clientset/versioned/fake"
+	policyFake "github.com/openservicemesh/osm/pkg/gen/client/policy/clientset/versioned/fake"
+	"github.com/openservicemesh/osm/pkg/k8s"
+	"github.com/openservicemesh/osm/pkg/service"
+	"github.com/openservicemesh/osm/pkg/tests"
+)
+
+var (
+	proxy           *envoy.Proxy
+	server          xds_discovery.AggregatedDiscoveryService_StreamAggregatedResourcesServer
+	osmConfigurator *configurator.Client
+	adsServer       *Server
+)
+
+func setupTestServer(b *testing.B) {
+	stop := signals.RegisterExitHandlers()
+	msgBroker := messaging.NewBroker(stop)
+	kubeClient := k8sClientFake.NewSimpleClientset()
+	configClient := configFake.NewSimpleClientset()
+	policyClient := policyFake.NewSimpleClientset()
+	informerCollection, err := informers.NewInformerCollection(tests.MeshName, stop,
+		informers.WithKubeClient(kubeClient),
+		informers.WithConfigClient(configClient, tests.OsmMeshConfigName, tests.OsmNamespace),
+	)
+	if err != nil {
+		b.Fatalf("Failed to create informer collection: %s", err)
+	}
+	kubeController := k8s.NewKubernetesController(informerCollection, policyClient, msgBroker)
+	policyController := policy.NewPolicyController(informerCollection, kubeController, msgBroker)
+	osmConfigurator = configurator.NewConfigurator(informerCollection, tests.OsmNamespace, tests.OsmMeshConfigName, msgBroker)
+	kubeProvider := kube.NewClient(kubeController, osmConfigurator)
+
+	meshConfig := configv1alpha2.MeshConfig{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: "config.openservicemesh.io",
+			Kind:       "MeshConfig",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace: tests.OsmNamespace,
+			Name:      tests.OsmMeshConfigName,
+		}, Spec: configv1alpha2.MeshConfigSpec{
+			Certificate: configv1alpha2.CertificateSpec{
+				CertKeyBitSize:              2048,
+				ServiceCertValidityDuration: "1h",
+			},
+			Traffic: configv1alpha2.TrafficSpec{
+				EnableEgress:                      true,
+				EnablePermissiveTrafficPolicyMode: true,
+			},
+			Observability: configv1alpha2.ObservabilitySpec{
+				EnableDebugServer: false,
+				Tracing: configv1alpha2.TracingSpec{
+					Enable: false,
+				},
+			},
+			FeatureFlags: configv1alpha2.FeatureFlags{
+				EnableWASMStats:    false,
+				EnableEgressPolicy: false,
+			},
+		},
+	}
+	_, err = configClient.ConfigV1alpha2().MeshConfigs(tests.OsmNamespace).Create(context.Background(), &meshConfig, metav1.CreateOptions{})
+	if err != nil {
+		b.Fatalf("Failed to create mesh config: %v", err)
+	}
+
+	certManager, err := certificate.FakeCertManager()
+	if err != nil {
+		b.Fatalf("Failed to create fake cert manager: %v", err)
+	}
+
+	// --- setup
+	namespace := tests.Namespace
+	proxyService := service.MeshService{
+		Name:      tests.BookstoreV1ServiceName,
+		Namespace: namespace,
+	}
+	proxySvcAccount := tests.BookstoreServiceAccount
+
+	certPEM, _ := certManager.IssueCertificate(proxySvcAccount.ToServiceIdentity().String(), certificate.Service)
+	cert, _ := certificate.DecodePEMCertificate(certPEM.GetCertificateChain())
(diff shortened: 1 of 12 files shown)


shashankram (commented):
Can the perf benchmarking be done on actual end-to-end scenarios, perhaps as a part of the e2e tests instead of unit tests?

keithmattix:
I think there's value in starting with the unit test cases first since there are fewer externalities that may skew results. Eventually, e2e perf benchmarking would be nice, but starting from unit tests seems acceptable

allenlsy:
@shashankram It is possible to do it with e2e test, but we haven't explored the option yet. We expect the process to be quite similar. The reason I didn't start with e2e test is because with UT, it is easier to focus on the core logic and repeat the function calls to get an averaged result. And UT is slightly easier to setup and execute.

@keithmattix No matter whether the benchmark is in UT or e2e test, it's difficult to exclude extra factors that affect the result. With the current setup, it always requires manual research to pick out the data we concern from the results.

allenlsy:
Ping here

pkg/envoy/ads/response_benchmark_test.go
```diff
@@ -0,0 +1,117 @@
...
+	certManager := fakeTresor.NewFake(nil)
+	certCommonName := envoy.NewXDSCertCommonName(proxyUUID, envoy.KindSidecar, proxySvcAccount.Name, proxySvcAccount.Namespace)
+	certDuration := 1 * time.Hour
+	certPEM, _ := certManager.IssueCertificate(certCommonName, certDuration)
+	cert, _ := certificate.DecodePEMCertificate(certPEM.GetCertificateChain())
+	server, _ := tests.NewFakeXDSServer(cert, nil, nil)
+	kubectrlMock := k8s.NewMockController(mockCtrl)
+	certSerialNumber := certificate.SerialNumber("123456")
+	proxy, _ := envoy.NewProxy(certCommonName, certSerialNumber, nil)
+
```
keithmattix, line 91: I'm wary about mocking function calls in benchmarks. Could we just use a fake clientset with real clients?

keithmattix (changes requested):
Good set of benchmarking test cases to start off with; just one concern about mocking

pkg/envoy/ads/response_benchmark_test.go
allenlsy: This won't be an issue here. On line 109, the benchmark timer is reset, thus the setup time is not included.

steeling: I think keith’s concern is more that when the timer starts the mocks and fakes use a lot of bogus code paths. 

We should be able to benchmark everything , only use fake client sets to avoid network calls

allenlsy: Replacing mocks with actual components. Please check again

pkg/validator/server_benchmark_test.go
```diff
@@ -0,0 +1,46 @@
...
+	"net/http/httptest"
+	"strings"
+	"testing"
+
+	"github.com/openservicemesh/osm/pkg/logger"
+	"github.com/openservicemesh/osm/pkg/webhook"
+)
+
+func BenchmarkDoValidation(b *testing.B) {
+	b.StopTimer()
```
keithmattix, line 15: Is this line necessary?

pkg/tests/fixtures.go
```diff
@@ -26,6 +26,15 @@ import (
 var ErrDecodingPEMBlock = errors.New("failed to decode PEM block containing certificate")
 
 const (
+	// MeshName is the name of the OSM mesh
+	MeshName = "osm"
```
keithmattix, line 30: Do these really need to be exported?

keithmattix (approved):
Two nit questions, but great work overall

keithmattix (approved):

pkg/envoy/ads/response_benchmark_test.go
```diff
@@ -0,0 +1,170 @@
...
+		b.Fatalf("Failed to create service: %v", err)
+	}
+
+	proxy := envoy.NewProxy(envoy.KindSidecar, proxyUUID, proxySvcAccount.ToServiceIdentity(), nil)
+
+	s := NewADSServer(mc, proxyRegistry, true, tests.Namespace, osmConfigurator, certManager, kubeController, nil)
+
+	// Set subscribed resources for SDS
+	proxy.SetSubscribedResources(envoy.TypeSDS, mapset.NewSetWith("service-cert:default/bookstore", "root-cert-for-mtls-inbound:default/bookstore"))
+
```
steeling, line 161: can you offload everything above this line into a `setupTestServer` func? Makes it a bit more clear to see what we're testing

pkg/envoy/ads/response_benchmark_test.go
```diff
@@ -0,0 +1,170 @@
...
+	"github.com/openservicemesh/osm/pkg/envoy/registry"
+	configFake "github.com/openservicemesh/osm/pkg/gen/client/config/clientset/versioned/fake"
+	policyFake "github.com/openservicemesh/osm/pkg/gen/client/policy/clientset/versioned/fake"
+	"github.com/openservicemesh/osm/pkg/k8s"
+	"github.com/openservicemesh/osm/pkg/logger"
+	"github.com/openservicemesh/osm/pkg/service"
+	"github.com/openservicemesh/osm/pkg/tests"
+)
+
+func BenchmarkSendXDSResponse(b *testing.B) {
```
steeling, line 38: You're actually only testing send response for SDS here due to the `SetSubscribedResources` call.

Can we rename this so that it's clear this is only SDS, or you can add a loop over XDSResponseOrder to respond to all?

Alternatively we could have a single test for each so we know the individual benchmarks. Then with the comment below about separating out the server creation the tests themselves are much smaller

pkg/envoy/bootstrap/envoy_config_benchmark_test.go
```diff
@@ -0,0 +1,66 @@
...
+		b.Logf("Failed to set log level to error: %s", err)
+	}
+
+	probes := models.HealthProbes{
+		Liveness:  &models.HealthProbe{Path: "/liveness", Port: 81, IsHTTP: true},
+		Readiness: &models.HealthProbe{Path: "/readiness", Port: 82, IsHTTP: true},
+		Startup:   &models.HealthProbe{Path: "/startup", Port: 83, IsHTTP: true},
+	}
+	cert := tresorFake.NewFakeCertificate()
+	config := EnvoyBootstrapConfigMeta{
```
steeling, line 35: some small changes required here after a refactor

pkg/validator/server_benchmark_test.go
```diff
@@ -0,0 +1,46 @@
...
+		"metadata": {
+			"uid": "some-uid"
+		},
+		"request": {}
+	}`)),
+	}
+
+	w := httptest.NewRecorder()
+	s := &validatingWebhookServer{
+		validators: map[string]validateFunc{},
```
steeling, line 34: Looks like there's no validations here meaning no actual work will be done. Let's instantiate the server the same way we normally do

allenlsy: Validations added.

pkg/envoy/ads/response_benchmark_test.go
allenlsy: Added more xDS server types to be benchmarked individually.

pkg/envoy/ads/response_benchmark_test.go
```diff
@@ -0,0 +1,211 @@
...
+			},
+			FeatureFlags: configv1alpha2.FeatureFlags{
+				EnableWASMStats:    false,
+				EnableEgressPolicy: false,
+			},
+		},
+	}
+	_, err = configClient.ConfigV1alpha2().MeshConfigs(tests.OsmNamespace).Create(context.Background(), &meshConfig, metav1.CreateOptions{})
+	if err != nil {
+		b.Fatalf("Failed to create mesh config: %v", err)
```
steeling, line 95: nit: might be a little more "idiomatic go" to return an error and let the test itself determine if it's a fatal error or not

pkg/envoy/ads/response_benchmark_test.go
```diff
@@ -0,0 +1,211 @@
...
+		b.Fatalf("Failed to create service: %v", err)
+	}
+
+	proxy = envoy.NewProxy(envoy.KindSidecar, proxyUUID, proxySvcAccount.ToServiceIdentity(), nil)
+
+	adsServer = NewADSServer(mc, proxyRegistry, true, tests.Namespace, osmConfigurator, certManager, kubeController, nil)
+}
+
+func BenchmarkSendXDSResponse(b *testing.B) {
+	testCases := []struct {
```
steeling, line 164: nit: normally I approve of this style of test cases, but in this case the name matches the xdsType, so could be more simple to just iterate over `envoy.XDSResponseOrder`

pkg/envoy/ads/response_benchmark_test.go
```diff
@@ -0,0 +1,211 @@
...
+		b.Run(tc.name, func(b *testing.B) {
+			setupTestServer(b)
+
+			// Set subscribed resources for SDS
+			proxy.SetSubscribedResources(tc.xdsType, mapset.NewSetWith("service-cert:default/bookstore", "root-cert-for-mtls-inbound:default/bookstore"))
+
+			b.ResetTimer()
+			b.StartTimer()
+			for i := 0; i < b.N; i++ {
+				if err := adsServer.sendResponse(proxy, &server, nil, osmConfigurator, envoy.XDSResponseOrder...); err != nil {
```
steeling, line 204: instead of passing in envoy.XDSResponseOrder, you should pass in tc.xdsType for more clarity

steeling (approved):

pkg/envoy/ads/response_benchmark_test.go
allenlsy: Somehow EDS does not work with the test setup. That's why I'm not testing it here. I can add it as a TODO item.

PR #4686, opened 2022-04-22
config/meshConfig: New localProxyMode field

keithmattix:
Add new `spec.sidecar.localProxyMode` field for user to
control how mesh traffic gets proxied

**Description**:
Part of the fix for #4653 

**Testing done**:

**Affected area**:

### DIFF
--- a/charts/osm/values.schema.json
+++ b/charts/osm/values.schema.json
@@ -706,7 +706,7 @@
                     "type": "string",
                     "title": "The localProxyMode schema",
                     "description": "Proxy mode for the Envoy proxy sidecar. Acceptable values are ['Localhost', 'PodIP'].",
-                    "pattern": "^(Localhost|PodIP)$",
+                    "enum": ["Localhost","PodIP"],
                     "examples": [
                         "Localhost"
                     ]
--- a/charts/osm/values.schema.json
+++ b/charts/osm/values.schema.json
@@ -701,6 +701,16 @@
                         "error"
                     ]
                 },
+                "localProxyMode": {
+                    "$id": "#/properties/osm/properties/localProxyMode",
+                    "type": "string",
+                    "title": "The localProxyMode schema",
+                    "description": "Proxy mode for the Envoy proxy sidecar. Acceptable values are ['Localhost', 'PodIP'].",
+                    "pattern": "^(Localhost|PodIP)$",
+                    "examples": [
+                        "Localhost"
+                    ]
+                },
                 "controllerLogLevel": {
                     "$id": "#/properties/osm/properties/controllerLogLevel",
                     "type": "string",
--- a/pkg/configurator/methods.go
+++ b/pkg/configurator/methods.go
@@ -123,18 +123,6 @@ func (c *client) GetEnvoyLogLevel() string {
 	return constants.DefaultEnvoyLogLevel
 }
 
-// GetLocalProxyMode returns the local proxy mode
-func (c *client) GetLocalProxyMode() configv1alpha2.LocalProxyMode {
-	proxyConfig := c.getMeshConfig().Spec.Sidecar.LocalProxyMode
-	switch proxyConfig {
-	case configv1alpha2.LocalProxyModeLocalhost, configv1alpha2.LocalProxyModePodIP:
-		return proxyConfig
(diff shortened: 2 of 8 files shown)


pkg/configurator/methods.go
```diff
@@ -123,6 +123,18 @@ func (c *client) GetEnvoyLogLevel() string {
...
 }
 
+// GetLocalProxyMode returns the local proxy mode
+func (c *client) GetLocalProxyMode() configv1alpha2.LocalProxyMode {
+	proxyConfig := c.getMeshConfig().Spec.Sidecar.LocalProxyMode
+	switch proxyConfig {
+	case configv1alpha2.LocalProxyModeLocalhost, configv1alpha2.LocalProxyModePodIP:
+		return proxyConfig
+	default:
+		// unrecognized proxy mode; return the default: ProxyModeLocalhost
```
nojnhuh, line 133: Since the API rejects any other value up front anyway and the default is defined in the CRD, I have a feeling enforcing the default again here isn't worth it since any other theoretical implementation of this interface (including the mock used in tests) would presumably need to do the same thing which may not be obvious. Then we wouldn't necessarily need to add this method to the interface and `GetMeshConfig().Spec.Sidecar.LocalProxyMode` could be used directly instead.

keithmattix: I just followed the approach `envoyLogLevel` did. I added the function `GetLocalProxyMode` because it allows us to explicitly define this behavior on the `Configurator` interface, rather than an implementation detail of `MeshConfig`. You're right though, at the moment, it's not likely this codepath gets hit

steeling:
Do we want this to be a mesh-wide option? Are there situations where a user might want to use a mix (particularly if they use other OSS services outside of their control)?

Would it make more sense for this to be a pod annotation?

keithmattix:
@steeling I think a mesh-wide approach is best for a few reasons:
1. It's simplest
2. Making teams choose mesh-wide lowers the possibility of customers wanting us to support the old way when...
3. ...we're eventually going to deprecate localhost proxy mode.

nojnhuh (approved):

charts/osm/values.schema.json
```diff
@@ -701,6 +701,16 @@
                         "error"
                     ]
                 },
+                "localProxyMode": {
+                    "$id": "#/properties/osm/properties/localProxyMode",
+                    "type": "string",
+                    "title": "The localProxyMode schema",
+                    "description": "Proxy mode for the Envoy proxy sidecar. Acceptable values are ['Localhost', 'PodIP'].",
+                    "pattern": "^(Localhost|PodIP)$",
```
shashankram, line 709: Can this be an enum?

pkg/configurator/methods.go
```diff
@@ -123,6 +123,18 @@ func (c *client) GetEnvoyLogLevel() string {
 	return constants.DefaultEnvoyLogLevel
 }
 
+// GetLocalProxyMode returns the local proxy mode
+func (c *client) GetLocalProxyMode() configv1alpha2.LocalProxyMode {
```
shashankram, line 127: Let's drop these methods in the pkg and use `GetMeshConfig().LocalProxyMode` directly where applicable.

keithmattix: I can remove it; I don't have any strong opinions, but is there a definitive reason why?

shashankram: So initially these methods were added as helpers to parse a ConfigMap instead of a MeshConfig resource, so it made sense to have these helpers convert the data type from the string type. With MeshConfig, we can directly access the fields as `GetMeshConfig().<field>`. The refactoring of `pkg/configurator` to migrate from ConfigMap backing to MeshConfig didn't clean up the existing code, so we see this inconsistency. For new fields, doesn't make sense to create a method per field in the interface.

keithmattix: Ah makes perfect sense. Thanks for the context!

charts/osm/values.schema.json
keithmattix: We didn't use it anywhere else in the helm schema so I didn't know it was a thing; added 👍🏾

PR #4691, opened 2022-04-25
cli/verifier: derive appProtocol from service

shashankram:
**Description**:
Uses the same mechanism as the controller to derive
the app protocol for a service port. This is required
for a service with multiple ports, both serving
different protocols.

The `AppProtocol` in the `TrafficAttribute` struct
will be used for ingress/egress instead.

Part of #4634

**Testing done**:
```
$ osm verify connectivity --from-pod curl/curl-7bb5845476-5b7wr --to-pod httpbin/httpbin-69dc7d545c-hbc6d --service httpbin 
[+] Context: Verify if pod "curl/curl-7bb5845476-5b7wr" can access pod "httpbin/httpbin-69dc7d545c-hbc6d" for service "httpbin/httpbin"
Status: Success

```

**Affected area**:

### DIFF
--- a/pkg/cli/verifier/envoy_config.go
+++ b/pkg/cli/verifier/envoy_config.go
@@ -147,15 +147,13 @@ func (v *EnvoyConfigVerifier) verifySource() Result {
 
 	// Retrieve route configs
 	var routeConfigs []*xds_route.RouteConfiguration
-	if v.configAttr.trafficAttr.AppProtocol == constants.ProtocolHTTP {
-		configs := config.Routes.GetDynamicRouteConfigs()
-		for _, r := range configs {
-			routeConfig := &xds_route.RouteConfiguration{}
-			//nolint: errcheck
-			//#nosec G104: Errors unhandled
-			r.GetRouteConfig().UnmarshalTo(routeConfig)
-			routeConfigs = append(routeConfigs, routeConfig)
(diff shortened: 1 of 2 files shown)


nojnhuh (approved):

pkg/cli/verifier/envoy_config.go
```diff
@@ -147,15 +147,13 @@ func (v *EnvoyConfigVerifier) verifySource() Result {
...
-			//#nosec G104: Errors unhandled
-			r.GetRouteConfig().UnmarshalTo(routeConfig)
-			routeConfigs = append(routeConfigs, routeConfig)
-		}
+	configs := config.Routes.GetDynamicRouteConfigs()
+	for _, r := range configs {
+		routeConfig := &xds_route.RouteConfiguration{}
+		//nolint: errcheck
+		//#nosec G104: Errors unhandled
+		r.GetRouteConfig().UnmarshalTo(routeConfig)
```
keithmattix, line 155: Why not check the error here and return an unknown result?

shashankram: I couldn't find a scenario where the unmarshaling fails. Wanted to avoid unnecessary checks around this if we can't test it.

jaellio (approved):

PR #4685, opened 2022-04-22
feat: add pyroscope profiling

allenlsy:
**Description**:

Pyroscope is a continuous profiling software. By installing it in the Kubernetes cluster, we can periodically profile OSM control plane and explore performance optimizations. 

This PR is part of https://github.com/openservicemesh/osm/issues/4622 

**Testing done**:

**Affected area**:
| Functional Area            |     |
| -------------------------- | --- |
| New Functionality          | [ ] |
| CI System                  | [ ] |
| CLI Tool                   | [ ] |
| Certificate Management     | [ ] |
| Control Plane              | [ ] |
| Demo                       | [ ] |
| Documentation              | [ ] |
| Egress                     | [ ] |
| Ingress                    | [ ] |
| Install                    | [ ] |
| Networking                 | [ ] |
| Observability              | [ ] |
| Performance                | [ ] |
| SMI Policy                 | [ ] |
| Security                   | [ ] |
| Sidecar Injection          | [ ] |
| Tests                      | [ ] |
| Upgrade                    | [ ] |
| Other                      | [ ] |

Please answer the following questions with yes/no.

1. Does this change contain code from or inspired by another project?
    -   Did you notify the maintainers and provide attribution?

2. Is this a breaking change?

3. Has documentation corresponding to this change been updated in the [osm-docs](https://github.com/openservicemesh/osm-docs) repo (if applicable)?

PR #4698, opened 2022-04-27
Optionally install continuous profiling tools in osm-system namespace

allenlsy:
**Description**:

This PR allows use optionally enable the continuous profiling tool Pyroscope in OSM control plane, by passing the `profiling.enabled=true` value to `osm install` command.

To support this, osm control plane components needs to add Pyroscole Golang client and connect to the server.  A new data field `spec.observability.enableProfiling` is added to the MeshConfig. Hence, MeshConfig is migrated from v1alpha2 to v1alpha3.

**Testing done**:

* Updated unit tests
* Performed dev installation on k8s cluster. Verified that Pyroscope is installed cocrrectly and collects profiling results from osm-controller and osm-injector.

Screen Shot 2022-04-26 at 6 48 22 PM

Screen Shot 2022-04-26 at 6 48 32 PM

Resolves #4685 

**Affected area**:
| Performance                | [X] |

No

2. Is this a breaking change?

No

No. The instruction about enabling Pyroscope should be added.


steeling:
I'm in favor of making OSS tooling work seamlessly with OSM, but I wonder if we should be in the business of including them directly. There's a lot of tooling that exists out there, and it may become burdensome to attempt to package and enable everything through the OSM cli. One alternative I see is through blog posts on the OSM blog page. 

What I do think we should do, is instrument OSM code with open standards such that other tools can leverage them. We do this, ie: with prometheus metrics, by using go's net/pprof, etc. If there's anything else we can do in that realm I'd be all for it.

These are just my opinions, so curious on others' thoughts here!

keithmattix:
I agree with @steeling here; I think this would be an excellent OSM blog post or video where @allenlsy can walk the viewer through profiling their own OSM install. However, bundling a bunch of tools with OSM probably isn't the best idea long-term and is only suitable for demos IMO

allenlsy:
Thanks @steeling @keithmattix for the opinions here. We did installed and profiled OSM even before this PR and manipulating MeshConfig. The process is pretty simple, just need to follow Pyroscope documentation. A blog about how to do the actual performance profiling is even more meaningful. We will definitely work on it. 

But before deciding if we drop this change, I want to hear more from other people.

allenlsy:
@jaellio Thanks for the reference. I've done some research on that command. Running `pyroscope exec` still requires us to add ways to enable pyroscope and update Docker images, but no Golang code need to be updated.

I've found a better solution following Pyroscope documentation here: https://pyroscope.io/docs/golang-pull-mode/

For a Golang project, Pyroscope also can leverage Golang pprof web API to do pull-mode profiling, and OSM happens to have it. I've just tested that by just enabling OSM debug server and adding essential annotations, Pyroscope can pull data from `/debug/pprof` path. No code change is needed. I can move on to write a documentation on this process.

allenlsy:
Closing the PR. Pyroscope will be installed without modifying OSM.

PR #4700, opened 2022-04-27
feat(injector): add list of ignored network interfaces

### LINKED ISSUE #4546: Add support for pods with multiple network interfaces

Support for pods with multiple network interfaces, where only the default network interfaces (eth0) should be subject to service mesh interception

In our use case, multi-network k8s cluster (multus) is used, which runs into some problems with OSM (since OSM programs some Linux kernel rules which impact the additional network interface / network namespaces, on which OSM is not intended to be running).

In particular, OSM should program the following rules:
`iptables -t nat -I PROXY_OUTPUT ! -o eth0 -j RETURN`
`iptables -t nat -I PROXY_INBOUND ! -i eth0 -j RETURN`

**Possible use cases**

Support high performance networking workloads, where there may be multiple independent networks in play, or where there may be additional network interfaces.

nojnhuh:
**Description**:
This change adds a new configurable list of ignored network interface
names. Traffic received from and sent to those interfaces is not
forwarded to the sidecar container by iptables. When this list is empty
(the default), osm-injector produces the exact same iptables commands as
it did before.

The list is configured in the chart at `osm.networkInterfaceExclusionList`
and in the MeshConfig at `spec.traffic.networkInterfaceExclusionList`.

Fixes #4546

**Testing done**:
Updated unit tests

<details>
<summary>Tested manually e2e</summary>

- Installed [Multus CNI](https://github.com/k8snetworkplumbingwg/multus-cni) on an AKS cluster: `kubectl apply -f https://raw.githubusercontent.com/k8snetworkplumbingwg/multus-cni/master/deployments/multus-daemonset-thick-plugin.yml`
- Created a new network interface `net1`:
  ```yaml
  apiVersion: "k8s.cni.cncf.io/v1"
  kind: NetworkAttachmentDefinition
  metadata:
    name: macvlan-conf
  spec:
    config: '{
        "cniVersion": "0.3.0",
        "type": "macvlan",
        "master": "eth0",
        "mode": "bridge",
        "ipam": {
          "type": "host-local",
          "subnet": "192.168.1.0/24",
          "rangeStart": "192.168.1.200",
          "rangeEnd": "192.168.1.216",
          "routes": [
            { "dst": "0.0.0.0/0" }
          ],
          "gateway": "192.168.1.1"
        }
      }'
  ```
- Created 2 ubuntu pods from deployments named `client` and `server` each running containers defined by:
  ```dockerfile
  FROM ubuntu
  RUN apt update && apt install -y iproute2 curl python3
  ```

  ```yaml
  apiVersion: apps/v1
  kind: Deployment
  metadata:
    name: client
  spec:
    replicas: 1
    selector:
      matchLabels:
        app: client
    template:
      metadata:
        labels:
          app: client
        annotations:
          k8s.v1.cni.cncf.io/networks: macvlan-conf
      spec:
        nodeName: aks-nodepool2-29838616-vmss000000
        containers:
        - name: client
          command: ["/bin/sh", "-c", "trap : TERM INT; sleep infinity & wait"]
          image: jondev.azurecr.io/nettest
  ```
  - Both pods must be running on the same node
- `kubectl exec -it deploy/server -c server -- bash -c 'ip a && python3 -m http.server'`
  - Note the IP address of the `net1` interface, e.g. 192.168.1.200, which changes each time the pod is recreated
- `kubectl exec -it deploy/client -c client -- curl -I 192.168.1.200:8000`
  ```
    HTTP/1.0 200 OK
    Server: SimpleHTTP/0.6 Python/3.8.10
    Date: Wed, 27 Apr 2022 16:13:02 GMT
    Content-type: text/html; charset=utf-8
    Content-Length: 1018
    ```
- Installed OSM
- `osm namespace add default`
- `kubectl rollout restart deploy client server`
- `kubectl exec -it deploy/server -c server -- bash -c 'ip a && python3 -m http.server'`
- `kubectl exec -it deploy/client -c client -- curl -I 192.168.1.200:8000`
  ```
    curl: (7) Failed to connect to 192.168.1.213 port 8000: Connection refused
    command terminated with exit code 7
    ```
- `kubectl patch meshconfig -n osm-system osm-mesh-config -p '{"spec": {"traffic": {"networkInterfaceExclusionList": ["net1"]}}}' --type=merge`
- `kubectl rollout restart deploy client server`
- `kubectl exec -it deploy/server -c server -- bash -c 'ip a && python3 -m http.server'`
  - Note the IP address of the `net1` interface, e.g. 192.168.1.200
- `kubectl exec -it deploy/client -c client -- curl -I 192.168.1.200:8000`
  ```
    HTTP/1.0 200 OK
    Server: SimpleHTTP/0.6 Python/3.8.10
    Date: Wed, 27 Apr 2022 16:13:02 GMT
    Content-type: text/html; charset=utf-8
    Content-Length: 1018
    ```
- `/stats` Envoy endpoint on either pod didn't show any of that traffic going through Envoy.
</details>

**Affected area**:

3. Has documentation corresponding to this change been updated in the [osm-docs](https://github.com/openservicemesh/osm-docs) repo (if applicable)? Yes, https://github.com/openservicemesh/osm-docs/pull/361

### DIFF
--- a/pkg/injector/iptables.go
+++ b/pkg/injector/iptables.go
@@ -60,7 +60,7 @@ var iptablesInboundStaticRules = []string{
 }
 
 // generateIptablesCommands generates a list of iptables commands to set up sidecar interception and redirection
-func generateIptablesCommands(outboundIPRangeExclusionList []string, outboundIPRangeInclusionList []string, outboundPortExclusionList []int, inboundPortExclusionList []int) string {
+func generateIptablesCommands(outboundIPRangeExclusionList []string, outboundIPRangeInclusionList []string, outboundPortExclusionList []int, inboundPortExclusionList []int, networkInterfaceExclusionList []string) string {
 	var rules strings.Builder
 
 	fmt.Fprintln(&rules, `# OSM sidecar interception rules
@@ -73,6 +73,11 @@ func generateIptablesCommands(outboundIPRangeExclusionList []string, outboundIPR
 
 	// 1. Create inbound rules
 	cmds = append(cmds, iptablesInboundStaticRules...)
+
+	// Ignore inbound traffic on specified interfaces
+	for _, iface := range networkInterfaceExclusionList {
+		cmds = append(cmds, fmt.Sprintf("-I OSM_PROXY_INBOUND -i %s -j RETURN", iface))
+	}
 
 	// 2. Create dynamic inbound ports exclusion rules
 	if len(inboundPortExclusionList) > 0 {
@@ -87,6 +92,11 @@ func generateIptablesCommands(outboundIPRangeExclusionList []string, outboundIPR
 
 	// 3. Create outbound rules
 	cmds = append(cmds, iptablesOutboundStaticRules...)
+
+	// Ignore outbound traffic in specified interfaces
+	for _, iface := range networkInterfaceExclusionList {
+		cmds = append(cmds, fmt.Sprintf("-A OSM_PROXY_OUTBOUND -o %s -j RETURN", iface))
+	}
 
 	//
 	// Create outbound exclusion and inclusion rules.
--- a/pkg/injector/iptables.go
+++ b/pkg/injector/iptables.go
@@ -76,6 +76,8 @@ func generateIptablesCommands(outboundIPRangeExclusionList []string, outboundIPR
 
 	// Ignore inbound traffic on specified interfaces
 	for _, iface := range networkInterfaceExclusionList {
+		// *Note: it is important to use the insert option '-I' instead of the append option '-A' to ensure the
+		// exclusion of traffic to the network interface happens before the rule that redirects traffic to the proxy
 		cmds = append(cmds, fmt.Sprintf("-I OSM_PROXY_INBOUND -i %s -j RETURN", iface))
 	}
 
--- a/charts/osm/README.md
+++ b/charts/osm/README.md
@@ -142,6 +142,7 @@ The following table lists the configurable parameters of the osm chart and their
 | osm.meshName | string | `"osm"` | Identifier for the instance of a service mesh within a cluster |
 | osm.multicluster | object | `{"gatewayLogLevel":"error"}` | OSM multicluster feature configuration |
 | osm.multicluster.gatewayLogLevel | string | `"error"` | Log level for the multicluster gateway |
+| osm.networkInterfaceExclusionList | list | `[]` | Specifies a global list of network interface names to exclude for inbound and outbound traffic interception by the sidecar proxy. |
 | osm.osmBootstrap.podLabels | object | `{}` | OSM bootstrap's pod labels |
 | osm.osmBootstrap.replicaCount | int | `1` | OSM bootstrap's replica count |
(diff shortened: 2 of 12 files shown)


trstringer (approved):

pkg/injector/iptables.go
```diff
@@ -74,6 +74,11 @@ func generateIptablesCommands(outboundIPRangeExclusionList []string, outboundIPR
 	// 1. Create inbound rules
 	cmds = append(cmds, iptablesInboundStaticRules...)
 
+	// Ignore inbound traffic on specified interfaces
+	for _, iface := range networkInterfaceExclusionList {
+		cmds = append(cmds, fmt.Sprintf("-I OSM_PROXY_INBOUND -i %s -j RETURN", iface))
```
shashankram, line 79: What's the reason the inbound rule uses -I (insert before) and outbound rule uses -A (append)?

If ordering is important, kindly document this in the code.

shashankram (changes requested):
Looks good, just 1 blocking comment regarding rule ordering.

pkg/injector/iptables.go
nojnhuh: These inbound rules only work when they're added before the static rules, either with `-I` or by adding them to the slice before with `-A`. I would need help figuring out an exact reason why that's the case.

shashankram: Iptable rules are executed in order. So it makes sense to use `-I` here so that the exclusions get applied first before other rules that can match the traffic. Could you add a comment here and also for `-I OSM_PROXY_INBOUND -p tcp --match multiport --dports %s -j RETURN` which is missing a comment.

I am looking into the code to see if the ordering of outbound rules is correct.

nojnhuh: I narrowed it down to needing these commands at least before the last static rule: 
```go
	// Redirect remaining inbound traffic to Envoy
	"-A OSM_PROXY_INBOUND -p tcp -j OSM_PROXY_IN_REDIRECT",
```

shashankram: So this is how the rules should be ordered.

INBOUND:
exclusions
inclusions
static rules that don't redirect to proxy
redirect to proxy

OUTBOUND
exclusions
inclusions
status rules that don't redirect to proxy
redirect to proxy

The reason `-I` is used with inbound is because the `redirect to proxy rule: ...-A OSM_PROXY_INBOUND -p tcp -j OSM_PROXY_IN_REDIRECT` is a part of the static rule set. While for outbound, `redirect to proxy rule: ... -A OSM_PROXY_OUTBOUND -j OSM_PROXY_OUT_REDIRECT` is added only after the exclusions/inclusions rules because the redirection rule OSM_PROXY_OUTBOUND -> OSM_PROXY_OUT_REDIRECT is different when inclusions are applied.

Your change looks good, could you add the following comment next to the inbound rule added:
```
// *Note: it is important to use the insert option '-I' instead of the append option '-A' to ensure the 
// exclusion of traffic to the network interface happens before the rule that redirects traffic to the proxy 
```

shashankram (approved):
Non blocking comment https://github.com/openservicemesh/osm/pull/4700#discussion_r860146389

PR #4701, opened 2022-04-27
pkg/injector: Enable podIP proxying via meshconfig setting

### LINKED ISSUE #4653: (*bug) Envoy connections fail if app container binds to eth0

keithmattix:
Fixes #4653. Allows customers to choose how their traffic gets proxied

**Description**:

**Testing done**:
Unit and E2E tests

**Affected area**:
| Networking                 | [X ] |

3. Has documentation corresponding to this change been updated in the [osm-docs](https://github.com/openservicemesh/osm-docs) repo (if applicable)? yes


pkg/injector/init_container.go
```diff
@@ -10,7 +10,7 @@ import (
 func getInitContainerSpec(containerName string, cfg configurator.Configurator, outboundIPRangeExclusionList []string,
 	outboundIPRangeInclusionList []string, outboundPortExclusionList []int,
 	inboundPortExclusionList []int, enablePrivilegedInitContainer bool, pullPolicy corev1.PullPolicy, networkInterfaceExclusionList []string) corev1.Container {
-	iptablesInitCommand := generateIptablesCommands(outboundIPRangeExclusionList, outboundIPRangeInclusionList, outboundPortExclusionList, inboundPortExclusionList, networkInterfaceExclusionList)
+	iptablesInitCommand := generateIptablesCommands(cfg, outboundIPRangeExclusionList, outboundIPRangeInclusionList, outboundPortExclusionList, inboundPortExclusionList, networkInterfaceExclusionList)
```
shashankram, line 13: There is no need to pass the cfg object here. For consistency with other args here, passing `cfg.GetMeshConfig().Spec.Sidecar.LocalProxyMode` should suffice. This simplifies the code and tests to not depend on configurator/mocks for iptables rule generation.

pkg/injector/iptables.go
```diff
@@ -5,34 +5,50 @@ import (
...
-	// Traffic to the Proxy Admin port flows to the Proxy -- not redirected
-	fmt.Sprintf("-A OSM_PROXY_OUT_REDIRECT -p tcp --dport %d -j ACCEPT", constants.EnvoyAdminPort),
+	localProxyMode := cfg.GetMeshConfig().Spec.Sidecar.LocalProxyMode
 
-	// For outbound TCP traffic jump from OUTPUT chain to OSM_PROXY_OUTBOUND chain
-	"-A OUTPUT -p tcp -j OSM_PROXY_OUTBOUND",
+	if localProxyMode == configv1alpha2.LocalProxyModePodIP {
+		// For envoy -> local service container proxying, send traffic to pod IP instead of localhost
+		iptablesOutboundStaticRules = append(iptablesOutboundStaticRules, fmt.Sprintf("-A OUTPUT -p tcp -o lo -d 127.0.0.1/32 -m owner --uid-owner %d -j DNAT --to-destination $POD_IP", constants.EnvoyUID))
+	}
```
shashankram, line 29: Can this not be added as a part of the dynamic rule set similar to other "dynamic" rules that are configurable?

tests/e2e/e2e_pod_client_server_test.go
```diff
@@ -25,26 +26,33 @@ var _ = OSMDescribe("Test HTTP traffic from 1 pod client -> 1 pod server",
...
 		Context("Test traffic flowing from client to server without a Kubernetes Service for the Source: HTTP", func() {
 			// Prior iterations of OSM required that a source pod belong to a Kubernetes service
 			// for the Envoy proxy to be configured for outbound traffic to some remote server.
 			// This test ensures we test this scenario: client Pod is not associated w/ a service.
 			withSourceKubernetesService := false
-			testTraffic(withSourceKubernetesService)
+			testTraffic(withSourceKubernetesService, PodCommandDefault)
+		})
+
+		Context("Test traffic flowing from client to a server with a podIP bind", func() {
```
shashankram, line 40: Let's just modify one of the existing tests to bind to the pod IP vs adding another test here. This gives us the coverage we need without increasing the CI time. This test is also not the best test to change, as it is already testing an additional scenario of client with/without k8s service. Could you update a different test, say `e2e_deployment_client_server_test` to simply bind to the pod IP.

tests/framework/common.go
```diff
@@ -328,13 +328,21 @@ nodeRegistration:
 	return nil
 }
 
+// WithLocalProxyMode sets the LocalProxyMode for OSM
+func WithLocalProxyMode(mode configv1alpha2.LocalProxyMode) InstallOsmOpt {
+	return func(opts *InstallOSMOpts) {
+		opts.LocalProxyMode = mode
+	}
+}
+
```
shashankram, line 337: This deviates from how other MeshConfig attributes are configured. Could you follow the same patter, and directly update MeshConfig in tests as:
```
installOpts := Td.GetOSMInstallOpts()
installOpts.LocalProxyMode = ...
Expect(Td.InstallOSM(installOpts)).To(Succeed())
```

tests/framework/common_apps.go
```diff
@@ -273,6 +273,17 @@ func (td *OsmTestData) SimplePodApp(def SimplePodAppDef) (corev1.ServiceAccount,
...
+						{
+							Name: "POD_IP",
+							ValueFrom: &corev1.EnvVarSource{
+								FieldRef: &corev1.ObjectFieldSelector{
+									APIVersion: "v1",
+									FieldPath:  "status.podIP",
+								},
+							},
+						},
+					},
```
shashankram, line 286: Why is this needed when the init-container adds this during sidecar injection?

shashankram (changes requested):

tests/framework/common_apps.go
keithmattix: This is so that the test httpbin pod can bind to the pod IP

tests/e2e/e2e_pod_client_server_test.go
keithmattix: Binding to the pod IP is an explicit behavior that I would think we'd want to test explicitly in a separate test rather than silently attaching it to another test. I picked this test because it's testing networking scenarios: client-server with service and client-server without service. Does it not make sense to test other networking behavior here in the simplest case (i.e. 2 pods)?

tests/framework/common.go
keithmattix: To do that, I'd have to create a separate `It()` block. My goal was to re-use the existing testing setup for pod_client_server and just add a separate case since they're related (see my comment above). I feel like the option pattern is an unobtrusive approach that maintains existing contracts while allowing further flexibility.

pkg/injector/iptables.go
keithmattix: No because this rule must occur before line 33; we have to short-circuit envoy outbound traffic to localhost before it gets caught in the OSM_PROXY_OUTBOUND chain

keithmattix: Oh I wasn't aware of that option! TIL - thanks!

tests/e2e/e2e_pod_client_server_test.go
keithmattix: Initially, I was going to make this a completely separate test, but when I saw there was already a `testTraffic` function generic enough for my use-case, I decided to use it. Regardless of `e2e_pod_client_server_test.go`'s original intent, the Describe block's title is: "Test HTTP traffic from 1 pod client -> 1 pod server". That's pretty generic, no? There's no other test that's this simple (from what I can tell), so I think it makes sense to test pod-to-pod connectivity in this way.

Also, just because the code change is related to iptables, doesn't necessarily mean that's where the e2e test should be located. The whole point of an integration/e2e test is to think from a customer's perspective. When I think of how a bug in this area would present itself, pod-to-pod connectivity is what I come up with (as you and I have both seen from the IcM)

shashankram: Just to share more context regarding this test, it started out as a test to verify HTTP traffic between pods. Then we extended this to test the scenario where a client can have a service associated with it, just so we don't add more tests that have an extremely narrow scope. At the very least if you want to extend this test to further connectivity scenarios, please rename it to something such as `e2e_client_server_connectivity_test.go`. Previously, this test was differentiating itself from `e2e_deployment_client_server_test` by creating Pods rather than Deployment objects, so the meaning of this test is being changed here, which is why I am suggesting making it explicit. Otherwise, it seems like adding one more scenario to an unrelated test (which is something I suggested to reduce the CI time).

keithmattix: Ah, given your context on the evolution of the test file, I now understand your position. I think renaming the file makes sense given its semantics. 👍🏾

shashankram (approved):

nojnhuh (approved):

PR #4703, opened 2022-04-28
bugreport: collect more ingress & control plane info

shashankram:
**Description**:
Collects more information pertaining to ingress by
avoiding early termination, and fetches control
plane namespace scoped resources.

**Testing done**:
```
$ osm support bug-report --app-namespaces httpingress --app-deployments httpingress/httpbin --ingress  --osm-namespace kube-system
[+] Created staging dir /tmp/osm_bug_report_1586314353 to generate bug report


[+] Collecting information about meshed namespaces
✓ Finished generating report for meshed namespaces


[+] Collecting information from individual app namespaces
✓ Collected report from Namespace "httpingress"
✓ Finished generating report for individual app namespaces


[+] Collecting information from individual app pods
✓ Collected report for Pod "httpingress/httpbin-69dc7d545c-brbcf"
✓ Finished generating report for individual app pods


[+] Collecting information from control plane
✓ Collected report for Pod kube-system/osm-bootstrap-55ff95f96b-cxm67
✓ Collected report for Pod kube-system/osm-controller-5bc989994c-kcbx2
✓ Collected previous report for Pod kube-system/osm-controller-5bc989994c-kcbx2
✓ Collected report for Pod kube-system/osm-controller-5bc989994c-mcmxt
✓ Collected previous report for Pod kube-system/osm-controller-5bc989994c-mcmxt
✓ Collected report for Pod kube-system/osm-injector-7dfcd68fd4-7qjmx
✓ Collected report for Pod kube-system/osm-injector-7dfcd68fd4-d9zkr
✓ Finished generating report for control plane


[+] Collecting ingress information
✓ Collected report for ingress httpbin/httpbin
✓ Collected report for ingress backend httpbin/httpbin
✓ Collected report for ingress controller ingress-nginx/ingress-nginx-controller-867c9b9ddb-5frmj
✓ Finished generating report for ingress


[+] Collecting information from individual app namespaces
✓ Bug report successfully archived to /tmp/36131790_osm-bug-report.tar.gz
```

**Affected area**:

### DIFF
--- a/pkg/bugreport/controlplane.go
+++ b/pkg/bugreport/controlplane.go
@@ -11,7 +11,7 @@ import (
 	"github.com/openservicemesh/osm/pkg/constants"
 )
 
-func (c *Config) collectControlPlaneLogs() error {
+func (c *Config) collectControlPlaneLogs() {
 	pods, err := c.KubeClient.CoreV1().Pods(c.ControlPlaneNamepace).
 		List(context.Background(), v1.ListOptions{
 			LabelSelector: fmt.Sprintf(
@@ -21,14 +21,16 @@ func (c *Config) collectControlPlaneLogs() error {
 			),
 		})
 	if err != nil {
-		return fmt.Errorf("error getting control plane pods: %w", err)
+		c.completionFailure("Error getting control plane pods: %s", err)
+		goto afterPodLogCollection
 	}
 
 	for _, pod := range pods.Items {
 		cmd := []string{"kubectl", "logs", "-n", pod.Namespace, pod.Name}
 		outPath := path.Join(c.rootNamespaceDirPath(), pod.Namespace, rootPodDirName, pod.Name, commandsDirName, strings.Join(cmd, "_"))
 		if err := runCmdAndWriteToFile(cmd, outPath); err != nil {
-			return fmt.Errorf("error writing control pod logs: %w", err)
+			c.completionFailure("Error writing control pod logs: %s", err)
+			continue
 		}
 		c.completionSuccess("Collected report for Pod %s/%s", pod.Namespace, pod.Name)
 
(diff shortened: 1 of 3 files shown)


pkg/bugreport/controlplane.go
```diff
@@ -21,26 +21,41 @@ func (c *Config) collectControlPlaneLogs() error {
...
 			}
 			c.completionSuccess("Collected previous report for Pod %s/%s", pod.Namespace, pod.Name)
 		}
 	}
 
-	return nil
+afterPodLogCollection:
+	// Collect MeshConfig, secrets in control plane namespace
+	namespaceCmds := [][]string{
+		{"kubectl", "get", "meshconfig", "-n", c.ControlPlaneNamepace, "-A", "-o", "yaml"},
```
nojnhuh, line 51: Was passing both `-n` to scope to one namespace and `-A` to read from all namespaces intentional?

pkg/bugreport/controlplane.go
```diff
@@ -11,7 +11,7 @@ import (
 	"github.com/openservicemesh/osm/pkg/constants"
 )
 
-func (c *Config) collectControlPlaneLogs() error {
+func (c *Config) collectControlPlaneLogs() {
```
nojnhuh, line 14: The main drawback to not returning an error here anymore is that the overall exit status won't reflect if an error was hit here. I wonder if accumulating errors in something like a `multierror` would let us both continue when we hit errors while exposing whether or not errors were hit.

pkg/bugreport/controlplane.go
```diff
@@ -21,26 +21,41 @@ func (c *Config) collectControlPlaneLogs() error {
 			),
 		})
 	if err != nil {
-		return fmt.Errorf("error getting control plane pods: %w", err)
+		c.completionFailure("Error getting control plane pods: %s", err)
+		goto afterPodLogCollection
```
nojnhuh, line 25: Depending on how we feel about `goto`, one alternative I can think of is to put the `for` loop after this in an `else` block.

pkg/bugreport/ingress.go
```diff
@@ -60,28 +65,25 @@ func (c *Config) collectIngressReport() error {
 					strings.Join(cmd, "_"),
 				)
 				if err := runCmdAndWriteToFile(cmd, filename); err != nil {
-					return err
+					c.completionFailure("Error wiring IngressBackend command: %s", err)
```
nojnhuh, line 68:
```suggestion
					c.completionFailure("Error writing IngressBackend command: %s", err)
```

pkg/bugreport/controlplane.go
shashankram: It wasn't, fixed.

pkg/bugreport/controlplane.go
shashankram: This is consistent with log collection for other components - meshes, app namespaces, app pods etc.
The completion success/failure messages indicate when a step succeeds/fails.

pkg/bugreport/controlplane.go
shashankram: I prefer this to avoid unnecessary nesting.

pkg/bugreport/controlplane.go
keithmattix: Agreed; I tend to hesitate when making a change that removes information. A multierror or similar implementation would be what I prefer.

pkg/bugreport/controlplane.go
keithmattix: +1 an else block or a helper function is my vote here...maybe even a `defer`?

pkg/bugreport/controlplane.go
shashankram: This does not remove error info, it correctly gets printed within the function, which was previously not the case due to the early return on error semantics.

pkg/bugreport/controlplane.go
shashankram: I think this is a pretty reasonable use of `goto`, which of course is also used in many Go std libs. I don't see a good reason to: 1. increase the nesting by moving this into an else block, 2. use a defer just to avoid goto.

keithmattix: It's not a bad use of go to; agreed. However, I feel like the use of goto here is equivalent to a finally block in a Java based language, and `defer` feels more appropriate semantically assuming that's the case. It's up to you though; this is not a hill I'll die on

pkg/bugreport/controlplane.go
keithmattix: Ah I see now; that's what completionFailure does. LGTM

jaellio (approved):

PR #4704, opened 2022-04-28
(feat/statefulsets): MeshService API changes for Headless Services

keithmattix:
**Description**:
In order to provide statefulset support, introduce 2 new public 
functions to the MeshService type to accomplish the following:
1. The `ProviderKey` function describes the mapping from a MeshService back to a provider
service (e.g. a Kubernetes service). This decouples the MeshService name
from being a foreign key between the provider's collection of services
and the MeshCatalog's set of services.
2. The `Subdomain` function returns the subdomain for a particular mesh service (e.g. the pod name for a particular pod within a headless service). This allows for proper matching of endpoints when building EDS and CDS responses.

Both of these functions rely on a new convention around MeshService.Name in that the name shall be formatted as (`[subdomain.]providerKey`). The values returned by these functions are calculated once and then memoized via unexported fields. The `MeshService.Equals()` function has been updated to account for these unexported fields.

Because of the semantics of statefulsets and headless services in Kubernetes, it is important to highlight the following guidance re: headless services:

```markdown
Specific statefulset instances (e.g. mysql-0.mysql.namespace.svc.cluser.local) should only be
referenced for traffic routing (e.g. TrafficSplit) and not policy (e.g. TrafficTarget, Retry).
Conversely, the headless service (e.g. mysql.namespace.svc.cluster.local) should only be referenced
for policy and not traffic routing.
```

**Testing done**:

**Affected area**:

2. Is this a breaking change?

### DIFF
--- a/pkg/catalog/inbound_traffic_policies.go
+++ b/pkg/catalog/inbound_traffic_policies.go
@@ -246,7 +246,7 @@ func (mc *MeshCatalog) getUpstreamServicesIncludeApex(upstreamServices []service
 		for _, split := range mc.meshSpec.ListTrafficSplits(smi.WithTrafficSplitBackendService(svc)) {
 			svcName := k8s.GetServiceFromHostname(split.Spec.Service)
 			subdomain := k8s.GetSubdomainFromHostname(split.Spec.Service)
-			apexMeshService := service.NewMeshService(svcName, svc.Namespace, svc.Port, svc.TargetPort, svc.Protocol)
+			apexMeshService := service.NewMeshService(svc.Namespace, svcName, svc.Port, svc.TargetPort, svc.Protocol)
 
 			if subdomain != "" {
 				apexMeshService.Name = fmt.Sprintf("%s.%s", subdomain, svcName)
--- a/pkg/catalog/inbound_traffic_policies.go
+++ b/pkg/catalog/inbound_traffic_policies.go
@@ -244,13 +244,20 @@ func (mc *MeshCatalog) getUpstreamServicesIncludeApex(upstreamServices []service
 		}
 
 		for _, split := range mc.meshSpec.ListTrafficSplits(smi.WithTrafficSplitBackendService(svc)) {
+			svcName := k8s.GetServiceFromHostname(split.Spec.Service)
+			subdomain := k8s.GetSubdomainFromHostname(split.Spec.Service)
 			apexMeshService := service.MeshService{
+				Name:       svcName,
 				Namespace:  svc.Namespace,
-				Name:       k8s.GetServiceFromHostname(split.Spec.Service),
 				Port:       svc.Port,
 				TargetPort: svc.TargetPort,
 				Protocol:   svc.Protocol,
 			}
+
+			if subdomain != "" {
+				apexMeshService.Name = fmt.Sprintf("%s.%s", subdomain, svcName)
+			}
+
 			if newlyAdded := svcSet.Add(apexMeshService); newlyAdded {
 				allServices = append(allServices, apexMeshService)
 			}
--- a/pkg/catalog/inbound_traffic_policies.go
+++ b/pkg/catalog/inbound_traffic_policies.go
@@ -244,8 +244,8 @@ func (mc *MeshCatalog) getUpstreamServicesIncludeApex(upstreamServices []service
 		}
 
 		for _, split := range mc.meshSpec.ListTrafficSplits(smi.WithTrafficSplitBackendService(svc)) {
-			svcName := k8s.GetServiceFromHostname(split.Spec.Service)
-			subdomain := k8s.GetSubdomainFromHostname(split.Spec.Service)
+			svcName := k8s.GetServiceFromHostname(mc.kubeController, split.Spec.Service)
+			subdomain := k8s.GetSubdomainFromHostname(mc.kubeController, split.Spec.Service)
 			apexMeshService := service.MeshService{
 				Namespace:  svc.Namespace,
 				Name:       svcName,
--- a/pkg/catalog/inbound_traffic_policies.go
+++ b/pkg/catalog/inbound_traffic_policies.go
@@ -246,13 +246,7 @@ func (mc *MeshCatalog) getUpstreamServicesIncludeApex(upstreamServices []service
 		for _, split := range mc.meshSpec.ListTrafficSplits(smi.WithTrafficSplitBackendService(svc)) {
 			svcName := k8s.GetServiceFromHostname(split.Spec.Service)
 			subdomain := k8s.GetSubdomainFromHostname(split.Spec.Service)
-			apexMeshService := service.MeshService{
-				Name:       svcName,
-				Namespace:  svc.Namespace,
-				Port:       svc.Port,
-				TargetPort: svc.TargetPort,
-				Protocol:   svc.Protocol,
-			}
+			apexMeshService := service.NewMeshService(svcName, svc.Namespace, svc.Port, svc.TargetPort, svc.Protocol)
 
 			if subdomain != "" {
 				apexMeshService.Name = fmt.Sprintf("%s.%s", subdomain, svcName)
--- a/pkg/catalog/inbound_traffic_policies.go
+++ b/pkg/catalog/inbound_traffic_policies.go
@@ -246,7 +246,13 @@ func (mc *MeshCatalog) getUpstreamServicesIncludeApex(upstreamServices []service
 		for _, split := range mc.meshSpec.ListTrafficSplits(smi.WithTrafficSplitBackendService(svc)) {
 			svcName := k8s.GetServiceFromHostname(split.Spec.Service)
 			subdomain := k8s.GetSubdomainFromHostname(split.Spec.Service)
-			apexMeshService := service.NewMeshService(svc.Namespace, svcName, svc.Port, svc.TargetPort, svc.Protocol)
+			apexMeshService := service.MeshService{
+				Namespace:  svc.Namespace,
+				Name:       svcName,
+				Port:       svc.Port,
+				TargetPort: svc.TargetPort,
+				Protocol:   svc.Protocol,
+			}
 
 			if subdomain != "" {
 				apexMeshService.Name = fmt.Sprintf("%s.%s", subdomain, svcName)
--- a/pkg/cli/verifier/envoy_config.go
+++ b/pkg/cli/verifier/envoy_config.go
@@ -16,7 +16,6 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
-	"k8s.io/utils/pointer"
 
 	configv1alpha2 "github.com/openservicemesh/osm/pkg/apis/config/v1alpha2"
 	"github.com/openservicemesh/osm/pkg/trafficpolicy"
@@ -461,49 +460,18 @@ func (v *EnvoyConfigVerifier) verifyDestination() Result {
 }
 
 func (v *EnvoyConfigVerifier) getDstMeshServicesForSvc(svc corev1.Service) ([]service.MeshService, error) {
-	endpoints, err := v.kubeClient.CoreV1().Endpoints(svc.Namespace).Get(context.Background(), svc.Name, metav1.GetOptions{})
-	if err != nil || endpoints == nil {
+	var err error
+	svcs := k8s.ServiceToMeshServices(svc, func(ms service.MeshService) (*corev1.Endpoints, error) {
+		endpoints, e := v.kubeClient.CoreV1().Endpoints(svc.Namespace).Get(context.Background(), svc.Name, metav1.GetOptions{})
+		err = e
+		return endpoints, nil
+	})
+
+	if err != nil {
 		return nil, err
 	}
 
-	var headless bool
-	if len(svc.Spec.ClusterIP) == 0 || svc.Spec.ClusterIP == corev1.ClusterIPNone {
-		headless = true
-	}
-
-	var meshServices []service.MeshService
-	for _, portSpec := range svc.Spec.Ports {
-		meshSvc := service.MeshService{
-			Namespace: svc.Namespace,
-			Name:      svc.Name,
-			Port:      uint16(portSpec.Port),
-			Protocol:  pointer.StringDeref(portSpec.AppProtocol, constants.ProtocolHTTP),
-		}
-
-		// The endpoints for the kubernetes service carry information that allows
-		// us to retrieve the TargetPort for the MeshService.
-		meshSvc.TargetPort = k8s.GetTargetPortFromEndpoints(portSpec.Name, *endpoints)
-
-		if headless {
-			for _, subset := range endpoints.Subsets {
-				for _, address := range subset.Addresses {
-					if address.Hostname != "" {
-						mSvc := service.MeshService{
-							Namespace:  svc.Namespace,
-							Name:       fmt.Sprintf("%s.%s", address.Hostname, svc.Name),
-							Port:       meshSvc.Port,
-							TargetPort: meshSvc.TargetPort,
-							Protocol:   meshSvc.Protocol,
-						}
-						meshServices = append(meshServices, mSvc)
-					}
-				}
-			}
-		} else {
-			meshServices = append(meshServices, meshSvc)
-		}
-	}
-	return meshServices, nil
+	return svcs, err
 }
 
 func (v *EnvoyConfigVerifier) findIngressFilterChainForService(svc *corev1.Service, filterChains []*xds_listener.FilterChain) error {
--- a/pkg/cli/verifier/envoy_config.go
+++ b/pkg/cli/verifier/envoy_config.go
@@ -483,6 +483,7 @@ func (v *EnvoyConfigVerifier) getDstMeshServicesForSvcPod(svc corev1.Service, po
 
 		if !k8s.IsHeadlessService(svc) {
 			meshServices = append(meshServices, meshSvc)
+			continue
 		}
 
 		for _, subset := range endpoints.Subsets {
--- a/pkg/cli/verifier/envoy_config.go
+++ b/pkg/cli/verifier/envoy_config.go
@@ -498,7 +498,6 @@ func (v *EnvoyConfigVerifier) getDstMeshServicesForSvcPod(svc corev1.Service, po
 					Protocol:   meshSvc.Protocol,
 				}
 				meshServices = append(meshServices, mSvc)
-
 			}
 		}
 	}
--- a/pkg/cli/verifier/envoy_config.go
+++ b/pkg/cli/verifier/envoy_config.go
@@ -188,25 +188,26 @@ func (v *EnvoyConfigVerifier) verifySource() Result {
 	// If the destination service is known, verify it has a matching filter chain and route
 	if v.configAttr.trafficAttr.DstService != nil {
 		dst := v.configAttr.trafficAttr.DstService
+		dstPod := v.configAttr.trafficAttr.DstPod
 		svc, err := v.kubeClient.CoreV1().Services(dst.Namespace).Get(context.Background(), dst.Name, metav1.GetOptions{})
 		if err != nil {
 			result.Status = Failure
 			result.Reason = fmt.Sprintf("Destination service %q not found: %s", dst, err)
 			return result
 		}
-		if err := v.findOutboundFilterChainForService(svc, outboundListener.FilterChains); err != nil {
+		if err := v.findOutboundFilterChainForService(svc, outboundListener.FilterChains, dstPod.Name); err != nil {
 			result.Status = Failure
 			result.Reason = fmt.Sprintf("Did not find matching outbound filter chain for service %q: %s", dst, err)
 			return result
 		}
 
-		if err := v.findHTTPRouteForService(svc, routeConfigs, true); err != nil {
+		if err := v.findHTTPRouteForService(svc, routeConfigs, true, dstPod.Name); err != nil {
 			result.Status = Failure
 			result.Reason = fmt.Sprintf("Did not find matching outbound route configuration for service %q: %s", dst, err)
 			return result
 		}
 
-		if err := v.findClusterForService(svc, clusters, true); err != nil {
+		if err := v.findClusterForService(svc, clusters, true, dstPod.Name); err != nil {
 			result.Status = Failure
 			result.Reason = fmt.Sprintf("Did not find matching outbound cluster for service %q: %s", dst, err)
 			return result
@@ -244,7 +245,7 @@ func (v *EnvoyConfigVerifier) verifySource() Result {
 	return result
 }
 
-func (v *EnvoyConfigVerifier) findOutboundFilterChainForService(svc *corev1.Service, filterChains []*xds_listener.FilterChain) error {
+func (v *EnvoyConfigVerifier) findOutboundFilterChainForService(svc *corev1.Service, filterChains []*xds_listener.FilterChain, podName string) error {
 	if svc == nil {
 		return nil
 	}
@@ -269,7 +270,7 @@ func (v *EnvoyConfigVerifier) findOutboundFilterChainForService(svc *corev1.Serv
 		dstIPRanges.Add(svc.Spec.ClusterIP)
 	}
 
-	meshServices, err := v.getDstMeshServicesForSvc(*svc)
+	meshServices, err := v.getDstMeshServicesForSvc(*svc, podName)
 	if len(meshServices) == 0 || err != nil {
 		return errors.Errorf("endpoints not found for service %s/%s, err: %s", svc.Namespace, svc.Name, err)
 	}
@@ -416,23 +417,24 @@ func (v *EnvoyConfigVerifier) verifyDestination() Result {
 	// Next, if the destination service is known, verify it has a matching filter chain
 	if v.configAttr.trafficAttr.DstService != nil {
 		dst := v.configAttr.trafficAttr.DstService
+		dstPod := v.configAttr.trafficAttr.DstPod
 		svc, err := v.kubeClient.CoreV1().Services(dst.Namespace).Get(context.Background(), dst.Name, metav1.GetOptions{})
 		if err != nil {
 			result.Status = Failure
 			result.Reason = fmt.Sprintf("Destination service %q not found: %s", dst, err)
 			return result
 		}
-		if err := v.findInboundFilterChainForService(svc, inboundListener.FilterChains); err != nil {
+		if err := v.findInboundFilterChainForService(svc, inboundListener.FilterChains, dstPod.Name); err != nil {
 			result.Status = Failure
 			result.Reason = fmt.Sprintf("Did not find matching inbound filter chain for service %q: %s", dst, err)
 			return result
 		}
-		if err := v.findHTTPRouteForService(svc, routeConfigs, false); err != nil {
+		if err := v.findHTTPRouteForService(svc, routeConfigs, false, dstPod.Name); err != nil {
 			result.Status = Failure
 			result.Reason = fmt.Sprintf("Did not find matching inbound route configuration for service %q: %s", dst, err)
 			return result
 		}
-		if err := v.findClusterForService(svc, clusters, false); err != nil {
+		if err := v.findClusterForService(svc, clusters, false, dstPod.Name); err != nil {
 			result.Status = Failure
 			result.Reason = fmt.Sprintf("Did not find matching inbound cluster for service %q: %s", dst, err)
 			return result
@@ -459,7 +461,7 @@ func (v *EnvoyConfigVerifier) verifyDestination() Result {
 	return result
 }
 
-func (v *EnvoyConfigVerifier) getDstMeshServicesForSvc(svc corev1.Service) ([]service.MeshService, error) {
+func (v *EnvoyConfigVerifier) getDstMeshServicesForSvc(svc corev1.Service, podName string) ([]service.MeshService, error) {
 	var err error
 	svcs := k8s.ServiceToMeshServices(svc, func(ms service.MeshService) (*corev1.Endpoints, error) {
 		endpoints, e := v.kubeClient.CoreV1().Endpoints(svc.Namespace).Get(context.Background(), svc.Name, metav1.GetOptions{})
@@ -469,6 +471,11 @@ func (v *EnvoyConfigVerifier) getDstMeshServicesForSvc(svc corev1.Service) ([]se
 
 	if err != nil {
 		return nil, err
+	}
+
+	if k8s.IsHeadlessService(svc) {
+		// filter down meshServices to eliminate subdomains not from this pod
+		svcs = service.FilterMeshServicesBySubdomain(svcs, podName, false)
 	}
 
 	return svcs, err
@@ -529,12 +536,12 @@ func (v *EnvoyConfigVerifier) findIngressFilterChainForService(svc *corev1.Servi
 	return nil
 }
 
-func (v *EnvoyConfigVerifier) findInboundFilterChainForService(svc *corev1.Service, filterChains []*xds_listener.FilterChain) error {
+func (v *EnvoyConfigVerifier) findInboundFilterChainForService(svc *corev1.Service, filterChains []*xds_listener.FilterChain, podName string) error {
 	if svc == nil {
 		return nil
 	}
 
-	meshServices, err := v.getDstMeshServicesForSvc(*svc)
+	meshServices, err := v.getDstMeshServicesForSvc(*svc, podName)
 	if len(meshServices) == 0 || err != nil {
 		return errors.Errorf("endpoints not found for service %s/%s, err: %s", svc.Namespace, svc.Name, err)
 	}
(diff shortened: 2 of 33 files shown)


pkg/service/types.go
```diff
@@ -22,12 +22,19 @@ const (
 	RemoteCluster
 )
 
+// ProviderMapper describes any entity that can be mapped to a service.Provider based on NamespacedKey
+type ProviderMapper interface {
```
steeling, line 26: I fear this may be a little more clever than it needs to be. Can we get rid of this interface? We can still keep the NamespacedKey(), but I don't see a need for an interface, especially since we have no plans to have anything else implement it

pkg/service/types.go
```diff
@@ -43,9 +50,21 @@ type MeshService struct {
 
 	// Protocol is the protocol served by the service's port
 	Protocol string
+
+	// ProviderKey represents the name of the original entity from which this MeshService was created (e.g. a Kubernetes service)
+	ProviderKey string
+
```
steeling, line 56: In that scenario would a providerkey ever be different than the name?

pkg/service/types.go
keithmattix: The only reason I created it was to start a conversation on possibly reducing the coupling to MeshService throughout the code. But I see your point: if we're fine with MeshService continuing to be the contract between providers and mesh configuration, then I'm good with taking a YAGNI approach and removing it

pkg/service/types.go
keithmattix: I'm assuming you mean "what scenario" - headless services are the one that comes to mind. I feel like `MeshService.Name` should have semantic meaning beyond just the original service name in the provider. I think it also should identify a _routeable_ entry within the mesh's catalog (which is why I created ProviderKey). Just because the provider stored the service with one name doesn't mean the mesh catalog entry should have that exact same name.

steeling: I see MeshService is sort of already decoupled, since it's our own custom type, as opposed to something out of K8s, or even Envoy, so I don't think we need to make further abstractions around it.

keithmattix: I'm assuming this is meant for the other thread? If so, that makes sense to me, and I'll remove the interface

pkg/service/types.go
```diff
@@ -43,9 +45,21 @@ type MeshService struct {
...
+	// Subdomain is an optional subdomain for this MeshService
+	Subdomain string
```
steeling, line 53: can you give an example of when we use subdomain?

pkg/service/types.go
steeling: yup, feel free to resolve

pkg/service/types.go
keithmattix: [Endpoint Listing](https://github.com/openservicemesh/osm/pull/4704/files#diff-a3fcf30b91ac74a293c1b1796ebb8b53d9f8bf748472a72a462607cadaeb6d55R58) and [FQDN](https://github.com/openservicemesh/osm/pull/4704/files#diff-428786751bb8848d286e65326ee03834f68486a3897130242ea3175534490389R79) are the two main ones

pkg/cli/verifier/envoy_config.go
```diff
@@ -447,6 +447,11 @@ func (v *EnvoyConfigVerifier) getDstMeshServicesForSvc(svc corev1.Service) ([]se
 		return nil, err
 	}
 
+	var headless bool
+	if len(svc.Spec.ClusterIP) == 0 || svc.Spec.ClusterIP == corev1.ClusterIPNone {
+		headless = true
+	}
```
shashankram, line 453: Let's add a helper in `pkg/k8s/util.go` `IsHeadlessService(...)`

pkg/k8s/client.go
```diff
@@ -346,6 +347,10 @@ func (c client) UpdateStatus(resource interface{}) (metav1.Object, error) {
 // MeshService objects per port.
 func ServiceToMeshServices(c Controller, svc corev1.Service) []service.MeshService {
 	var meshServices []service.MeshService
+	var headless bool
+	if len(svc.Spec.ClusterIP) == 0 || svc.Spec.ClusterIP == corev1.ClusterIPNone {
+		headless = true
+	}
```
shashankram, line 353: Define a helper for this as noted earlier.

pkg/k8s/types.go
```diff
@@ -100,10 +100,10 @@ type Controller interface {
 	ListPods() []*corev1.Pod
 
 	// ListServiceIdentitiesForService lists ServiceAccounts associated with the given service
-	ListServiceIdentitiesForService(service.MeshService) ([]identity.K8sServiceAccount, error)
+	ListServiceIdentitiesForService(svc service.MeshService) ([]identity.K8sServiceAccount, error)
 
 	// GetEndpoints returns the endpoints for a given service, if found
-	GetEndpoints(service.MeshService) (*corev1.Endpoints, error)
+	GetEndpoints(svc service.MeshService) (*corev1.Endpoints, error)
```
shashankram, line 106: Drop the parameter names from the interface.

pkg/k8s/util.go
```diff
@@ -37,15 +37,70 @@ func GetHostnamesForService(svc service.MeshService, localNamespace bool) []stri
 	return hostnames
 }
 
-// GetServiceFromHostname returns the service name from its hostname
-func GetServiceFromHostname(host string) string {
-	// The service name is the first string in the host name for a service.
-	// Ex. service.namespace, service.namespace.cluster.local
-	service := strings.Split(host, ".")[0]
+func splitHostName(host string) (service string, subdomain string) {
```
shashankram, line 40: Add a comment describing what the function does.

pkg/service/types.go
```diff
@@ -27,7 +28,11 @@ type MeshService struct {
 	// If the service resides on a Kubernetes service, this would be the Kubernetes namespace.
 	Namespace string
 
-	// The name of the service
+	// The name of the service. May include instance (e.g. pod) information if the backing service
```
shashankram, line 31: Please add an example for the Name value when this represents a non-headless k8s service, as it is unclear whether the subdomain will be set.

pkg/service/types.go
```diff
@@ -43,9 +48,60 @@ type MeshService struct {
...
+	providerKey string
+
+	subdomain string
+
+	subdomainPopulated bool
+}
+
+// NamespacedKey is the key (i.e. namespace + name) with which to lookup the backing service within the provider
+func (ms MeshService) NamespacedKey() string {
+	return fmt.Sprintf("%s/%s", ms.Namespace, ms.ProviderKey())
```
shashankram, line 61: The comment indicates it should be `namespace/name`, while the code uses `namespace/providerKey`.  This function should either be renamed or the comment needs to be updated.

pkg/service/types.go
```diff
@@ -43,9 +48,60 @@ type MeshService struct {
...
+}
+
+// Subdomain is an optional subdomain for this MeshService
+// It is calculated once based on Name and stored in an unexported field
+// which is why this function has a pointer receiver
+func (ms *MeshService) Subdomain() string {
+	if ms.subdomain == "" && !ms.subdomainPopulated {
```
shashankram, line 68: Let's invert this condition to reduce nesting:
```
if ms.subdomainPopulated != "" {
    return ms.subdomain
}
...
```

pkg/service/types.go
```diff
@@ -43,9 +48,60 @@ type MeshService struct {
...
+		ms.subdomainPopulated = true
+	}
+
+	return ms.subdomain
+}
+
+// ProviderKey represents the name of the original entity from which this MeshService was created (e.g. a Kubernetes service name)
+// It is calculated once based on Name and stored in an unexported field which is why this function has a pointer receiver
+func (ms *MeshService) ProviderKey() string {
+	if ms.providerKey == "" {
```
shashankram, line 85: Let's invert the condition to reduce indent.

shashankram (changes requested):
Thanks for this change. The overall idea looks good, left some comments to enable the reuse of code and readability.

pkg/service/service.go
```diff
@@ -1,16 +1,27 @@
...
+	if !ms.subdomainPopulated {
+		ms.Subdomain()
+	}
+
+	if !service.subdomainPopulated {
+		service.Subdomain()
+	}
+
+	ms.ProviderKey()
+	service.ProviderKey()
```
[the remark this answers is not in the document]
keithmattix: The only way to make sure these values are initialized on creation would be to create a NewMeshService function. We create MeshService literals in so many places, forcing the Equals function to initialize the unexported fields while everything else stayed lazy seemed to be the best option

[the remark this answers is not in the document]
shashankram: That's reasonable.

pkg/k8s/client.go
```diff
@@ -363,8 +368,27 @@ func ServiceToMeshServices(c Controller, svc corev1.Service) []service.MeshServi
 		} else {
 			log.Warn().Msgf("k8s service %s/%s does not have endpoints but is being represented as a MeshService", svc.Namespace, svc.Name)
 		}
-		meshServices = append(meshServices, meshSvc)
+
+		if headless && endpoints != nil {
```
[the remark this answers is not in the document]
keithmattix: Creating a helper for this is non-trivial since both this implementation and the verifier rely on two different data structures for querying: the former uses the `Controller` and the latter uses a base kubeClient. I'll whip something up though and you tell me if you think it's worth it

pkg/cli/verifier/envoy_config.go
```diff
@@ -459,7 +464,25 @@ func (v *EnvoyConfigVerifier) getDstMeshServicesForSvc(svc corev1.Service) ([]se
...
+							Namespace:  svc.Namespace,
+							Name:       fmt.Sprintf("%s.%s", address.Hostname, svc.Name),
+							Port:       meshSvc.Port,
+							TargetPort: meshSvc.TargetPort,
+							Protocol:   meshSvc.Protocol,
+						}
+						meshServices = append(meshServices, mSvc)
+					}
+				}
+			}
```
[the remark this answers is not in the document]
keithmattix: Carving this into a function is difficult due to the complexity of the logic (namely the different dependencies on an initial meshService, ports, etc.). This function is almost identical to pkg/k8s.ServiceToMeshServices so it should just use that if possible...the resulting generic function is a bit icky though, let's discuss once I push the changes

pkg/service/types.go
keithmattix: Good catch!

shashankram (approved):

pkg/k8s/util.go
```diff
@@ -37,15 +38,72 @@ func GetHostnamesForService(svc service.MeshService, localNamespace bool) []stri
...
+		service = serviceComponents[0]
+		subdomain = ""
+	case l == 2:
+		// e.g. service.namespace, mysql-0.service
+		// NOTE: There's no way to tell which component is the service or not, so we're going to keep the default behavior
+		// Thus, we have a requirement that users who want to reference statefulsets must include the namespace in the host name
+		service = serviceComponents[0]
+		subdomain = ""
+	case l == 3:
+		if strings.Contains(serviceComponents[l-1], "svc") {
```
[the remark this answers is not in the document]
keithmattix: The contains is there so that if there's a port at the end (e.g. `mysql-0.service.svc:9090`), the condition still evaluates to true. I see what you mean though; I'll write some more logic to account for that use-case. 

As to your other question, I don't even think Kubernetes supports a namespace called `svc` due to how it resolves DNS.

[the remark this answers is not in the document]
nojnhuh: Following the example here to spin up a basic StatefulSet in a namespace called `svc` in a kind cluster: https://kubernetes.io/docs/tutorials/stateful-application/basic-stateful-set/

`nslookup` successfully resolves all of the following:
```
web-0.nginx
web-0.nginx.svc
web-0.nginx.svc.svc
web-0.nginx.svc.svc.cluster.local
```

From another namespace also with a service named `nginx`, I see `nslookup nginx.svc` and `nslookup web-0.nginx.svc` consistently resolve to only the `nginx` in the `svc` namespace instead of the `nginx` in the same namespace. I don't think I've seen that exact behavior documented so I'm not sure exactly what we should expect there, but it's probably worth making sure the behavior here matches that.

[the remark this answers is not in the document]
keithmattix: Wow, I wouldn't have expected that; thanks for digging in. This does become more difficult though since there's no way to distinguish between `web-0.nginx.svc` where svc is the namespace and `web-0.nginx.svc` where web-0 is the actual service name and svc is the TLD.

pkg/catalog/inbound_traffic_policies.go
```diff
@@ -244,13 +244,20 @@ func (mc *MeshCatalog) getUpstreamServicesIncludeApex(upstreamServices []service
 		}
 
 		for _, split := range mc.meshSpec.ListTrafficSplits(smi.WithTrafficSplitBackendService(svc)) {
+			svcName := k8s.GetServiceFromHostname(split.Spec.Service)
```
steeling, line 247: This leaks k8s primitives into the catalog. We already do this in a bunch of places, but I think we should avoid it going forward.

What do you think about adding this method to an interface (service.Provider?), and calling that? Same for GetSubdomainFromHostname

pkg/envoy/registry/services.go
```diff
@@ -60,7 +62,9 @@ func (k *KubeProxyServiceMapper) ListProxyServices(p *envoy.Proxy) ([]service.Me
 
 func kubernetesServicesToMeshServices(kubeController k8s.Controller, kubernetesServices []v1.Service) (meshServices []service.MeshService) {
 	for _, svc := range kubernetesServices {
-		meshServices = append(meshServices, k8s.ServiceToMeshServices(kubeController, svc)...)
+		meshServices = append(meshServices, k8s.ServiceToMeshServices(svc, func(meshSvc service.MeshService) (*v1.Endpoints, error) {
```
steeling, line 65: same comment

pkg/k8s/client.go
```diff
@@ -344,7 +345,7 @@ func (c client) UpdateStatus(resource interface{}) (metav1.Object, error) {
 
 // ServiceToMeshServices translates a k8s service with one or more ports to one or more
 // MeshService objects per port.
-func ServiceToMeshServices(c Controller, svc corev1.Service) []service.MeshService {
+func ServiceToMeshServices(svc corev1.Service, endpointsGetter func(service.MeshService) (*corev1.Endpoints, error)) []service.MeshService {
```
steeling, line 348: A getter feels like an unnecessary layer of indirection. We're already operating on a k8s corev1.Service, so this should know how to translate between headless and non-headless. 

It likely makes sense to add this as a method on client instead

pkg/envoy/registry/services.go
```diff
@@ -50,6 +50,8 @@ func (k *KubeProxyServiceMapper) ListProxyServices(p *envoy.Proxy) ([]service.Me
 	}
 
 	meshServices := kubernetesServicesToMeshServices(k.KubeController, services)
+	// filter down meshServices to eliminate subdomains not from this pod (keeping empty subdomains)
+	meshServices = service.FilterMeshServicesBySubdomain(meshServices, pod.GetName(), true)
```
steeling, line 54: Instead of this, I think we can get 2 birds one stone?

Both `kubernetesServicesToMeshServices`, and `listServicesForPod` only exist in one place. Can we instead have `listServicesForPod` return meshServices, appropriately filtered already?

steeling (commented):
Can we add an E2E test  in this PR as well?

pkg/envoy/registry/services.go
keithmattix: The problem is that a single headless service can spawn N MeshServices where N = # of endpoints in that headless service. So from the k8s service perspective, there's no way to pre-filter

pkg/k8s/client.go
keithmattix: The reason the getter was added was because the verifier package needs to call this method without initializing an entire controller client. See [this comment](https://github.com/openservicemesh/osm/pull/4704#discussion_r869666034) for context

pkg/catalog/inbound_traffic_policies.go
keithmattix: Oooh good catch!! I agree, we shouldn't leak this 👍🏾

keithmattix:
@steeling An E2E test for all statefulset functionality will be added in subsequent PRs. This PR is only for the MeshService API changes and their downstream effects

pkg/k8s/client.go
steeling: I'd like to push back against that more still. Can we refactor so that it does have a client?

Regarding that comment, and as a general rule, I think we should avoid pushing changes we think are currently code smells with the excuse that we will refactor soon. Too often things get left untouched, and if we can refactor soon, I think we can refactor now.

if it makes the PR too large, maybe we can get a refactoring PR in now, and then merge/rebase the changes here

pkg/catalog/inbound_traffic_policies.go
steeling: FWIW a refactor like this, where we're already following this pattern a lot, I think we *can* do in a later PR :)

pkg/envoy/registry/services.go
keithmattix: Ah, I'm with you now; that makes sense 👍🏾

pkg/k8s/client.go
keithmattix: I agree with you in general. The complexity with this particular subject is that the verifier shouldn't need a whole k8s controller with informers and such. At this point, I truly wonder if de-duplication is worth it; yes the logic is the same, but without using indirection such as this getter function, it's impossible to generalize due to the various locations data is coming from (i.e. controller vs raw client-go client)

steeling (changes requested):
Instead of all the init/partial stuff, if the issue is unexported fields, it's likely a code smell that we're using the wrong data structure, ie: instead of a hashset let's just use a map

pkg/k8s/util.go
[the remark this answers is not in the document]
keithmattix: I've been thinking about this some more, and I still can't see a away to distinguish between these two cases. Maybe we just have to make it a known limitation that namespaces named `svc` may result in unexpected behavior?

pkg/k8s/util.go
```diff
@@ -37,15 +38,80 @@ func GetHostnamesForService(svc service.MeshService, localNamespace bool) []stri
...
+	// Ex. mysql-0.service.namespace, mysql-0.service.namespace.svc.cluster.local, mysql-0.service.namespace.svc.cluster.local
+	switch l := len(serviceComponents); {
+	case l == 1:
+		// e.g. service
+		service = serviceComponents[0]
+		subdomain = ""
+	case l == 2:
+		// e.g. service.namespace, mysql-0.service
+		// NOTE: There's no way to tell which component is the service or not, so we're going to keep the default behavior
+		// Thus, we have a requirement that users who want to reference statefulsets must include the namespace in the host name
```
steeling, line 58: How would k8s resolve this? they likely have an order preference for search, and we could do something similar. Particularly by bringing these methods onto the individual clients/controllers, then we can correlate with the state of the world.

pkg/k8s/client.go
keithmattix: Spoke offline with Sean, decided to circle back to this later. For now, I'm going to err on the side of the [WET principle](https://betterprogramming.pub/when-dry-doesnt-work-go-wet-6befda0444bf) and duplicate these for loops

pkg/catalog/inbound_traffic_policies.go
keithmattix: Yeah, this is probably going to be a bigger de-coupling initiative that we should capture in a standalone issue. Would you mind creating that @steeling? I think you've got some good context to share that may help prioritize it

pkg/k8s/util.go
steeling: CoreDNS is the standard implementation. I'd imagine they're deterministic. I'd avoid another dependency on that, but we could probably implement the same ordering.

The thing is, there's probably a negligible amount of users that have overlap between statefulsets where a conflict like this arises, so my thought is that allowing people to use the same semantics is a much better user experience. For the case where there is a conflict, we can try our best to have the same ordering as coreDNS, and for anybody else, we can add a caveat that they should expect our ordering.

I'd imagine this would leave us with near 0 users that attempt to use this feature and achieve an unexpected lookup.

steeling: Another way of saying this, I'd imagine the percentage of users that try to use <pod>.<service_name> is much higher than the percentage of users that have a <pod>.<service_name> that conflicts with an existing <service_name>.<namespace>, so lets try to accommodate the former, and do our best for the latter

keithmattix: Conflicts aren't really the issue here; remember, this function just takes a host name and says here's the subdomain and service name. This behavior has to remain consistent for all clusters, so there's no way to give best effort in the latter case. It's as if CoreDNS didn't ever support <service_name>.<namespace>; a DNS server has the benefit of just following NS records and returning an NXDOMAIN if it goes down the wrong path. In our case, if we said that `x.y` always meant `x = podName` and `y = service name`, then no OSM user would be able to use `service.namespace` and Envoy configs would break. 

The only way to do what you're suggesting is to either: 
1. do an actual DNS lookup or
2. add an additional heuristic to this function's parameters

steeling: let me see if i can try to clarify what i'm saying with a small snippet

```
// l == 2
p1, p2 := strings.Cut(".")
// query for pods == p1, backed by service == p2.
validPodRef := k8sController.PodWithServiceExists(p1, p2)
// query for 
validServiceNSRef := k8sController.GetService(p1, p2)

switch {
case p1 && p2:
    // conflict resolution
case p1:
   ...
case p2:
   ....
}

```

does that make sense? I am essentially saying we should do a DNS lookup, but through the K8s API by querying for services/pods, not by adding a new dependency or performing an actual DNS query. I may be missing something else, so please let me  know!

keithmattix: Ohhh I thought you were suggesting we could somehow do this with just strings; yeah, for this to work we'll need to pull in something else. I'll do that 👍🏾

pkg/k8s/util.go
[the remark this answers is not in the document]
keithmattix: @nojnhuh I ended up taking Sean's approach and used a Controller to make this split more accurae

pkg/envoy/registry/services_test.go
```diff
@@ -63,13 +70,55 @@ var _ = Describe("Test Proxy-Service mapping", func() {
...
+				Subsets: []v1.EndpointSubset{
+					{
+						Addresses: []v1.EndpointAddress{
+							{
+								IP:       "8.8.8.9",
+								Hostname: podName,
+							},
+							{
+								IP:       "8.8.8.9",
+								Hostname: podName2,
```
nojnhuh, line 116: Is there any issue with not actually creating the pod with this name in the test here if it's not referred to otherwise?

pkg/envoy/registry/services_test.go
```diff
@@ -63,13 +70,55 @@ var _ = Describe("Test Proxy-Service mapping", func() {
...
+			expectedSvc2.Subdomain()
 			expectedList := []service.MeshService{expectedSvc1, expectedSvc2}
 
-			mockKubeController.EXPECT().GetEndpoints(gomock.Any()).Return(nil, nil).Times(2)
+			mockKubeController.EXPECT().GetEndpoints(gomock.Any()).Return(&v1.Endpoints{
+				ObjectMeta: metav1.ObjectMeta{
+					Namespace: tests.Namespace,
+					Name:      expectedSvc1.Name,
+				},
+				Subsets: []v1.EndpointSubset{
```
nojnhuh, line 88: It looks like `Subsets` doesn't need to be defined at all for this to pass.

pkg/envoy/registry/services_test.go
```diff
@@ -63,13 +70,55 @@ var _ = Describe("Test Proxy-Service mapping", func() {
...
+			mockKubeController.EXPECT().GetEndpoints(gomock.Any()).Return(&v1.Endpoints{
+				ObjectMeta: metav1.ObjectMeta{
+					Namespace: tests.Namespace,
+					Name:      expectedSvc2.Name,
+				},
+				Subsets: []v1.EndpointSubset{
+					{
+						Addresses: []v1.EndpointAddress{
+							{
+								IP:       "8.8.8.9",
```
nojnhuh, line 111: If the IPs here aren't relevant to the test, I'd remove them.

pkg/envoy/registry/services_test.go
```diff
@@ -111,9 +160,11 @@ var _ = Describe("Test Proxy-Service mapping", func() {
 				Port:      tests.ServicePort,
 				Protocol:  "http",
 			}
+			expected.Subdomain()
```
nojnhuh, line 163: Same as above re: not needed anymore since this isn't memoized.

pkg/envoy/registry/services_test.go
```diff
@@ -150,6 +201,36 @@ var _ = Describe("Test Proxy-Service mapping", func() {
...
+					{
+						Addresses: []v1.EndpointAddress{
+							{
+								IP: "8.8.8.8",
+							},
+						},
+					},
+				},
+			}, nil).Times(1)
+			mockKubeController.EXPECT().GetEndpoints(gomock.Any()).Return(&v1.Endpoints{
```
nojnhuh, line 219: Should this resource be different somehow from the one right above it?

pkg/envoy/registry/services_test.go
```diff
@@ -150,6 +201,36 @@ var _ = Describe("Test Proxy-Service mapping", func() {
 
 			pod := tests.NewPodFixture(namespace, "pod-name", tests.BookstoreServiceAccountName, tests.PodLabels)
 			mockKubeController.EXPECT().ListServices().Return(services)
+			mockKubeController.EXPECT().GetEndpoints(gomock.Any()).Return(&v1.Endpoints{
+				ObjectMeta: metav1.ObjectMeta{
+					Namespace: namespace,
+					Name:      service1Name,
+				},
+				Subsets: []v1.EndpointSubset{
```
nojnhuh, line 209: Like above, if these aren't headless services, then it seems like `Subsets` doesn't matter here.

pkg/cli/verifier/envoy_config.go
```diff
@@ -478,8 +480,28 @@ func (v *EnvoyConfigVerifier) getDstMeshServicesForSvc(svc corev1.Service) ([]se
 		// The endpoints for the kubernetes service carry information that allows
 		// us to retrieve the TargetPort for the MeshService.
 		meshSvc.TargetPort = k8s.GetTargetPortFromEndpoints(portSpec.Name, *endpoints)
-		meshServices = append(meshServices, meshSvc)
+
+		if !k8s.IsHeadlessService(svc) {
+			meshServices = append(meshServices, meshSvc)
```
nojnhuh, line 485: Should there be a `continue` after this to match pkg/k8s?
https://github.com/openservicemesh/osm/blob/b3d0c7ff9b02b8253c96f142415709544cbb1fc2/pkg/k8s/client.go#L368-L371

pkg/envoy/registry/services.go
```diff
@@ -92,7 +86,15 @@ func listServicesForPod(pod *v1.Pod, kubeController k8s.Controller) []v1.Service
...
 	}
 
-	return serviceList
+	if len(serviceList) == 0 {
+		return nil
+	}
+
+	meshServices := kubernetesServicesToMeshServices(kubeController, serviceList)
+	// filter down meshServices to eliminate subdomains not from this pod (keeping empty subdomains)
+	meshServices = service.FilterMeshServicesBySubdomain(meshServices, pod.GetName(), true)
```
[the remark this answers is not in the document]
keithmattix: That still requires iterating through the loop an extra time. The code above feels a bit too clever IMO; do you feel like there's a benefit to multiple invocations vs. a parameter?

pkg/tests/fixtures.go
```diff
@@ -607,6 +608,12 @@ func NewServiceFixture(serviceName, namespace string, selectors map[string]strin
 			Selector: selectors,
 		},
 	}
+
+	if headless {
+		svc.Spec.ClusterIP = corev1.ClusterIPNone
+	}
```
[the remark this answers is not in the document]
keithmattix: Good call; I agree

pkg/envoy/registry/services.go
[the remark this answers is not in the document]
steeling: `kubernetesServicesToMeshServices` is another method called in one place, and the code in FilterMeshServiceBySubdomain is trivial, let's duplicate it..


```go
	var meshServices []service.MeshService
	for _, svc := range kubernetesServices {
		if svc.Subdomain() == subdomain || servie.Subdomain() == "" {
			meshServices = append(meshServices, k8s.ServiceToMeshServices(kubeController, svc)...)
 		}
	}
	return meshServices
```

pkg/k8s/util_test.go
```diff
@@ -72,6 +74,8 @@ func TestGetServiceFromHostname(t *testing.T) {
 		name            string
 		hostnames       []string
 		expectedService string
+		withController  bool
+		namespaceFound  bool
```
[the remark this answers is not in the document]
keithmattix: Added more test cases where the variables are set 👍🏾

pkg/k8s/util_test.go
```diff
@@ -187,3 +204,76 @@ func TestNamespacedNameFrom(t *testing.T) {
...
 	}
 }
+
+func TestGetSubdomainFromHostname(t *testing.T) {
+	testCases := []struct {
+		name              string
+		hostnames         []string
+		expectedSubdomain string
+		withController    bool
+		foundNamespace    bool
```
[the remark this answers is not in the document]
keithmattix: Same with this one

pkg/k8s/client_test.go
```diff
@@ -748,6 +760,61 @@ func TestK8sServicesToMeshServices(t *testing.T) {
...
+					ObjectMeta: metav1.ObjectMeta{
+						// Should match svc.Name and svc.Namespace
+						Namespace: "ns1",
+						Name:      "s1",
+					},
+					Subsets: []corev1.EndpointSubset{
+						{
+							Addresses: []corev1.EndpointAddress{
+								{
+									IP:       "10.1.0.1",
```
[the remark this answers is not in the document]
keithmattix: The IP address is used in the function being tested to mark these services as not being headless services

pkg/envoy/registry/services_test.go
keithmattix: There's technically no pod being persisted in a cluster; it's just allocated in memory and returned as the result of a mock function (e.g. `ListPods`)

pkg/envoy/registry/services_test.go
keithmattix: The IPs are also used down the callstack in ServiceToMeshServices

pkg/cli/verifier/envoy_config.go
keithmattix: Yes good catch

pkg/envoy/registry/services.go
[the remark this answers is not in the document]
keithmattix: Done and filter function removed 👍🏾

pkg/k8s/client_test.go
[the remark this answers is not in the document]
nojnhuh: Isn't the IP address that determines whether or not a service is headless different?
https://github.com/openservicemesh/osm/blob/191a9cfe0a593c4dc8f6f923e0b016d6d9f9d2d3/pkg/k8s/util.go#L177-L180

The main thing I notice is that the test still passes if I comment this out, so it doesn't seem necessary.

pkg/envoy/registry/services_test.go
nojnhuh: I mean it looks like the test can sidestep even creating the mocked pod as long as the name is still defined here. That would remove some code from the test that otherwise isn't particularly relevant.

pkg/envoy/registry/services_test.go
nojnhuh: Using `.Times(2)` on the single resource then might be a more clear and concise way to express that then.

pkg/envoy/registry/services_test.go
nojnhuh: My personal rule of thumb is "if I can comment it out and the test still passes, remove it." I find trying to make things as realistic as possible in unit tests is more work than it's worth and makes the tests harder to reason about later. If `Subsets` does come into play later for non-headless services, then the tests will almost certainly need to be updated to account for that anyway.

pkg/envoy/registry/services_test.go
keithmattix: Similar to https://github.com/openservicemesh/osm/pull/4704#discussion_r877265445. Even though this particular endpoints object doesn't need endpoints, I feel like it raises more questions since the next endpoints object does have IPs

pkg/envoy/registry/services_test.go
nojnhuh: Yes, but only for the proxy that's being passed to `ListProxyServices`. The second pod is only ever needed in the endpoints here. That goes back to the whole "realistic vs. minimal" thing though.

keithmattix: Wait, correct me if I'm wrong...but `listPodsForService` isn't ever used (even on the main branch). If that's the case, you're right, there's no need for the full pod manifest. I'll remove the function and clean this up

keithmattix: Well, turns out that's not where `ListPods` was being called, but I'll just defer to you on this one 😄

pkg/envoy/registry/services_test.go
keithmattix: I ended up adding a Headless service to the test, so I needed to split them out again

nojnhuh (approved):

Issue #4502, opened 2022-02-02
Automated Root Certificate rotation in OSM


snehachhabria:
#**Please describe the Improvement and/or Feature Request**
As of today OSM's root certificate once created cannot be automatically rotated.  This issue will track the root certificate rotation feature in OSM. 

The tasks include:
- [x] Proposal doc (a Google doc linked to this item)
- [x] Design doc (a Google doc linked to this item)
- [ ] Phase 1 Implementation
  - [x] Support Envoy xDS certificate rotation 
    - #4635 
  - [x] Implement MeshRootCertificate API
    - #4687 
    - #4677
    - #4736  
  - [x] MeshRootCertificate informer client
    - #4721 
  - [x] Create or load MeshRootCertificate on install #4712 
  - [ ] Update validating and mutating webhook to support the MeshRootCertificate resource 
    - [ ] Validating webhook #4723
    - [x] Mutating webhook
  - [x] Refactor Manager and Certificate to support Auto Root Certificate Rotation
    - [x] Update Manager to support multiple clients
      - #4705
    - [x] Issue certificates based on client and certificate settings 
      - #4743 
    - [x] Refactor IssueCertificate #4810 
      - #4800 
  - [x] Create Issuer from MeshRootCertificate #4713 
    - #4718 
    - #4742 
    - #4781 
    - [x] Implement MRC Watch and List
      - #4816 
  - [x] Update TLS configs for OSM webhooks and ADS server to dynmaically select certificate used for establishing mTLS #4819
  - [ ] Implement rotation stages
    - [ ] Update manager struct #4815 
    - [x] Update webhooks configurations #4817  
    - [x] Update bootstrap secret #4818 
  - [ ] CLI Tooling
    - [ ] osm cert CLI cmd
      -  osm cert status
      -  osm cert rotate
- [ ] Phase 2 Implementation
  - [ ] Implement detection strategy 
  - [ ] Support automatic movement between rotation stages
  - [ ] Support automatic detection of expiring root certificate, creation of new root certificate, and rotation initiation
- [ ] Unit and e2e tests
  - #4835
- [ ] Documentation: how-to guide
- [ ] Documentation: demo

**Scope (please mark with X where applicable)**
- New Functionality          [X]

jaellio:
[Proposal doc](https://docs.google.com/document/d/19-rmjK6WiemecHr0htpNUXsdaOXpBWHTWngDCCcETtA/edit?usp=sharing)

[...]

PR #4705, opened 2022-04-28
ref(cert): update Manager to support mult clients

jaellio:
**Description**:

Supports multiple clients in Manager struct. Refactoring
as a part of the root cert rotation work. During root cert
rotation, there will be 2 CertManagers - the CertManager
being rotated in and the CertManager being rotated out as
specified in the MeshRootCertificates.

The ca is moved from the Manager to the CertManagers
for each cert provider.

Part of #4502

**Testing done**:
- CI

**Affected area**:

### DIFF
--- a/pkg/certificate/providers/config.go
+++ b/pkg/certificate/providers/config.go
@@ -242,7 +242,7 @@ func (c *Config) getTresorOSMCertificateManager() (*certificate.Manager, debugge
 		return nil, nil, errors.Errorf("Failed to instantiate Tresor as a Certificate Manager")
 	}
 
-	tresorCertManager, err := certificate.NewManager(rootCert, tresorClient, c.cfg.GetServiceCertValidityPeriod(), c.msgBroker)
+	tresorCertManager, err := certificate.NewManager(tresorClient, c.cfg.GetServiceCertValidityPeriod(), c.msgBroker)
 	if err != nil {
 		return nil, nil, fmt.Errorf("error instantiating osm certificate.Manager for Tresor cert-manager : %w", err)
 	}
@@ -302,12 +302,7 @@ func (c *Config) getHashiVaultOSMCertificateManager(options VaultOptions) (*cert
 		return nil, nil, fmt.Errorf("error instantiating Hashicorp Vault as a Certificate Manager: %w", err)
 	}
 
-	vaultCert, err := vaultClient.GetRootCertificate()
-	if err != nil {
-		return nil, nil, fmt.Errorf("error getting Vault Root Certificate, got: %w", err)
-	}
-
-	certManager, err := certificate.NewManager(vaultCert, vaultClient, c.cfg.GetServiceCertValidityPeriod(), c.msgBroker)
+	certManager, err := certificate.NewManager(vaultClient, c.cfg.GetServiceCertValidityPeriod(), c.msgBroker)
 	if err != nil {
 		return nil, nil, fmt.Errorf("error instantiating osm certificate.Manager for Vault cert-manager : %w", err)
 	}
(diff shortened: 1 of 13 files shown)


pkg/certificate/providers/vault/certificate_manager.go
```diff
@@ -68,8 +88,8 @@ func (cm *CertManager) IssueCertificate(cn certificate.CommonName, validityPerio
 	return newCert(cn, secret, time.Now().Add(validityPeriod)), nil
 }
 
-// GetRootCertificate returns the root certificate.
-func (cm *CertManager) GetRootCertificate() (*certificate.Certificate, error) {
+// GetRootCA returns the root certificate.
+func (cm *CertManager) GetRootCA() (*certificate.Certificate, error) {
```
steeling, line 92: I would probably leave this unexported

pkg/certificate/providers/vault/certificate_manager.go
```diff
@@ -30,6 +30,21 @@ const (
 
 // New constructs a new certificate client using Vault's cert-manager
 func New(vaultAddr, token, role string) (*CertManager, error) {
```
steeling, line 32: not a big deal, but you could keep these as one function

pkg/certificate/providers/tresor/fake/fake.go
```diff
@@ -25,7 +25,7 @@ func NewFake(msgBroker *messaging.Broker) *certificate.Manager {
 	if err != nil {
 		return nil
 	}
-	tresorCertManager, err := certificate.NewManager(ca, tresorClient, 1*time.Hour, msgBroker)
```
steeling, line 28: let's keep this syntax for now, it will make subsequent PR's easier, and we can adapt as needed

pkg/certificate/providers/config.go
```diff
@@ -242,7 +242,7 @@ func (c *Config) getTresorOSMCertificateManager() (*certificate.Manager, debugge
 		return nil, nil, errors.Errorf("Failed to instantiate Tresor as a Certificate Manager")
 	}
 
-	tresorCertManager, err := certificate.NewManager(rootCert, tresorClient, c.cfg.GetServiceCertValidityPeriod(), c.msgBroker)
+	tresorCertManager, err := certificate.NewManager(tresorClient, c.cfg.GetServiceCertValidityPeriod(), c.msgBroker)
```
steeling, line 245: This one can also remain unchanged

pkg/certificate/providers/config.go
```diff
@@ -351,7 +346,7 @@ func (c *Config) getCertManagerOSMCertificateManager(options CertManagerOptions)
 		return nil, nil, errors.Errorf("Error instantiating Jetstack cert-manager client: %+v", err)
 	}
 
-	certManager, err := certificate.NewManager(rootCert, cmClient, c.cfg.GetServiceCertValidityPeriod(), c.msgBroker)
+	certManager, err := certificate.NewManager(cmClient, c.cfg.GetServiceCertValidityPeriod(), c.msgBroker)
```
steeling, line 349: as well as this one

steeling (approved):

pkg/certificate/providers/vault/certificate_manager.go
jaellio: I split this up in order to make the client creation (without adding the root certificate) testable. Do you have any suggestions for how to keep this a single function without requiring too much refactoring of the tests?

pkg/certificate/providers/vault/certificate_manager.go
jaellio: Changed to be unexported

pkg/certificate/providers/tresor/fake/fake.go
jaellio: Reverted changes to the NewManager method's signature. The ca will not be used in NewManager, is that okay?

pkg/certificate/providers/vault/certificate_manager.go
steeling: oh i see. Ya you can leave as is.

Alternatively you could do something like

```
// package level
type getRootCAFunc func(c *CertManager)
var getRootCA = func() {...} // fill in with current getRootCa (not a function not a method)
```

then in tests, before running:

```
getRootCA = func(){...} // overridden for testing
```

Up to you!

steeling (approved):

shashankram (approved):

Issue #4707, opened 2022-05-02
OSM NoInstall Nightly Job failing for e2e_client_server_connectivity test. 


nshankar13:
**Bug description**: NoInstall nightly job is failing the e2e_client_server_connectivity_test, likely due to an error with how the meshconfig is set. 

**Affected area (please mark with X where applicable)**:
- Install                    [ ]
- SMI Traffic Access Policy  [ ]
- SMI Traffic Specs Policy   [ ]
- SMI Traffic Split Policy   [ ]
- Permissive Traffic Policy  [ ]
- Ingress                    [ ]
- Egress                     [ ]
- Envoy Control Plane        [ ]
- CLI Tool                   [ ]
- Metrics                    [ ]
- Certificate Management     [ ]
- Sidecar Injection          [ ]
- Logging                    [ ]
- Debugging                  [ ]
- Tests                      [X]
- Demo                       [ ]
- CI System                  [X]

**Expected behavior**:

**Steps to reproduce the bug (as precisely as possible)**:

**How was OSM installed?**: NoInstall

**Anything else we need to know?**:

**Bug report archive**:

**Environment**:
- OSM version (use `osm version`):
- Kubernetes version (use `kubectl version`):
- Size of cluster (number of worker nodes in the cluster):
- Others:

PR #4708, opened 2022-05-02
Fix e2e_client_server_connectivity_test for NoInstall

nshankar13:

**Description**: Fix e2e_client_server_connectivity_test for NoInstall by explictly setting LocalProxyMode in setMeshConfigToDefault Method. 

Resolves https://github.com/openservicemesh/osm/issues/4707

**Testing done**: Tested locally on AKS cluster with OSM already installed. 

**Affected area**:

2. Is this a breaking change? 

### DIFF
--- a/tests/framework/common.go
+++ b/tests/framework/common.go
@@ -442,6 +442,7 @@ func setMeshConfigToDefault(instOpts InstallOSMOpts, meshConfig *configv1alpha2.
 	meshConfig.Spec.Sidecar.LogLevel = instOpts.EnvoyLogLevel
 	meshConfig.Spec.Sidecar.MaxDataPlaneConnections = 0
 	meshConfig.Spec.Sidecar.ConfigResyncInterval = "0s"
+	meshConfig.Spec.Sidecar.LocalProxyMode = instOpts.LocalProxyMode
 
 	meshConfig.Spec.Certificate.ServiceCertValidityDuration = instOpts.CertValidtyDuration.String()
 	meshConfig.Spec.Certificate.CertKeyBitSize = instOpts.CertKeyBitSize


tests/framework/common.go
```diff
@@ -442,6 +442,7 @@ func setMeshConfigToDefault(instOpts InstallOSMOpts, meshConfig *configv1alpha2.
 	meshConfig.Spec.Sidecar.LogLevel = instOpts.EnvoyLogLevel
 	meshConfig.Spec.Sidecar.MaxDataPlaneConnections = 0
 	meshConfig.Spec.Sidecar.ConfigResyncInterval = "0s"
+	meshConfig.Spec.Sidecar.LocalProxyMode = instOpts.LocalProxyMode
```
shashankram, line 445: @keithmattix isn't this set as the default? Wondering why this needs to be explicitly set without looking further at the test code.

nshankar13: LocalProxyMode is changed for the "Test traffic flowing from client to a server with a podIP bind" part of the test to v1alpha2.LocalProxyModePodIP, the previous 2 ginkgo.contexts use the default Proxy Mode. 

The way the test was before, the new value for LocalProxyMode was only being set for fresh installs. For the change to also be made for NoInstall, the value needs to be explicitly changed in setMeshConfigToDefault.

keithmattix: Because we don't have a helper function to access the LocalProxyMode field, there's no default being set outside of a Helm install.

shashankram: @keithmattix Why don't we default its value in the CRD schema?

shashankram (approved):

shashankram:
@nshankar13 please add a commit description to share more context regarding the issue being fixed..

tests/framework/common.go
keithmattix: [We do](https://github.com/openservicemesh/osm/blob/main/cmd/osm-bootstrap/crds/config_meshconfig.yaml#L125). But unless we're using an openapi unmarshaler, those defaults won't take effect

shashankram: We could add a default in the json unmarshaler, but that also means we would need to change the API version if we change those defaults. This seems more like a test issue where the control plane is reused between tests.

keithmattix: Yep, that's one of the downsides to adding not bumping a new version. This will also happen when users bump their OSM version, but because LocalProxyConfig is an opt-in behavior, we're fine with the 3rd case (unset) being the default.

PR #4734, opened 2022-05-11
cli/verifier: add control-plane-health command

shashankram:
**Description**:
Adds a `verify control-plane-health` command.
Currently this check is limited to verifying
that the control plane pods are running.
This will be extended to perform additional
checks pertaining to the health of the
control plane, such as looking for fatal/error
events reported, probe failures, etc.

Part of #4634

**Testing done**:
When all control plane pods are running
```
$ osm verify control-plane-health
[+] Context: Verify the health of OSM control plane
Status: Success

```

When the pod is not found
```
$ osm verify control-plane-health
checking pods
[+] Context: Verify the health of OSM control plane
Status: Failure
Reason: A verification step failed
Suggestion: Please follow the suggestions listed in the failed steps below to resolve the issue

[++] Context: Verify status of pod for app osm-system/osm-controller
Status: Failure
Reason: No pods found for app osm-system/osm-controller

```

**Affected area**:

### DIFF
--- a/pkg/cli/verifier/pod.go
+++ b/pkg/cli/verifier/pod.go
@@ -1,0 +1,72 @@
+package verifier
+
+import (
+	"context"
+	"fmt"
+	"io"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/kubernetes"
+
+	"github.com/openservicemesh/osm/pkg/constants"
+)
+
+// PodStatusVerifier implements the Verifier interface for control plane health
+type PodStatusVerifier struct {
+	stdout     io.Writer
+	stderr     io.Writer
+	kubeClient kubernetes.Interface
+	app        types.NamespacedName
+}
+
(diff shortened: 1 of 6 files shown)


pkg/cli/verifier/pod.go
```diff
@@ -0,0 +1,72 @@
...
+
+	"github.com/openservicemesh/osm/pkg/constants"
+)
+
+// PodStatusVerifier implements the Verifier interface for control plane health
+type PodStatusVerifier struct {
+	stdout     io.Writer
+	stderr     io.Writer
+	kubeClient kubernetes.Interface
+	app        types.NamespacedName
```
nojnhuh, line 22: It seems a bit unintuitive to me to have the `.Name` of this correspond to the `app` label instead of the pod's actual name. Could this be a plain label selector instead, or even a plain list of pods that's constructed elsewhere? I think that would be clearer and more widely applicable.

shashankram: The idea was to not use pod names here, as there may be many of them. The tool's job is to figure out the pods (which it can) and determine whether they are running. I didn't want to expose additional labels as they wouldn't be used. But the variable is called `app`, so I personally didn't find this ambiguous. I also used the NamespacedName type to avoid passing an additional namespace argument to the inner functions.

nojnhuh (approved):

jaellio (approved):

PR #4736, opened 2022-05-11
ref(certs): use secretRef for Vault token in MRC

jaellio:
**Description**:

Updates the MeshRootCertifcate CRD to use secretRefs for the 
Tresor CA and a secretKeyRef for the vault token. Currently, the
Vault token would be set as a string in the MeshRootCertificate.
The token needs to be stored securely and not inline in the MRC.

This change also implies that users will be able to reference secrets
containing the CA for Tresor in any namespace. Currently, the 
kuberenetes secret (by default called osm-ca-bundle) is assumed 
to be in the OSM control plane namespace. If we want to enforce 
this, we could validate the MRC or only allow the secretRef's
name to be set and the namespace could be populated by OSM.

**Testing done**:
- CI

**Affected area**:

2. Is this a breaking change?


pkg/certificate/providers/config.go
```diff
@@ -64,6 +64,11 @@ func NewCertificateManager(kubeClient kubernetes.Interface, kubeConfig *rest.Con
 		},
 	}
 
+	// TODO(#4745): Remove after deprecating the osm.vault.token option.
+	if vaultOption, ok := options.(VaultOptions); ok {
+		mrcClient.MRCProviderGenerator.VaultToken = vaultOption.VaultToken
+	}
+
```
steeling, line 71: you shouldn't need this here, will leave more comments with explanation below :)

pkg/certificate/providers/config.go
```diff
@@ -128,7 +133,7 @@ func (c *MRCProviderGenerator) getHashiVaultOSMCertificateManager(mrc *v1alpha2.
 	vaultAddr := fmt.Sprintf("%s://%s:%d", provider.Protocol, provider.Host, provider.Port)
```
steeling, line 133: if c.VaultToken is empty, we should query the secret ref

cmd/osm-bootstrap/crds/config_mesh_root_certificate.yaml
```diff
@@ -106,16 +106,48 @@ spec:
...
-                          type: string
+                          type: object
+                          required:
+                            - secretRef
+                          properties:
+                            secretRef:
+                              description: Reference to the kubernetes secret storing the vault token
+                              type: object
+                              required:
+                                - name
```
steeling, line 117: you also need the key to know how to access the data.

 Tresor does not need the key

pkg/certificate/providers/types.go
```diff
@@ -80,4 +80,7 @@ type MRCProviderGenerator struct {
 
 	// TODO(#4502): move these to the compat client once we have added these fields to the MRC.
 	KeyBitSize int
+
+	// TODO(#4745): Remove after deprecating the osm.vault.token option.
+	VaultToken string
```
steeling, line 85: Let's call this DefaultVaultToken, since if there's a secret ref on the MRC we should use that.

steeling (changes requested):

pkg/certificate/providers/config.go
jaellio: I'm not sure I understand. Where should the Vault token be set on the MRCProviderGenerator?

jaellio: I'm not sure I understand. Where should the Vault token be set on the MRCProviderGenerator?

pkg/certificate/providers/config.go
jaellio: Agreed, I wasn't going to include querying the secret ref as a part of this change. I was focussing on updating the MRC CRD while also supporting the existing requirement to set the Vault token on install. I am happy to include this as a part of the change if you think that would make more sense.

pkg/certificate/providers/config.go
steeling: ah.. i thought i deleted this comment before sending the review sorry!

cmd/osm-bootstrap/crds/config_mesh_root_certificate.yaml
steeling: Assuming that the tresor cert is self-created, although it may be a good user experience to not have that requirement. In reality it should probably not be an opaque secret but instead be a tls cert type.

For vault, ya we can do that, although if a user already has the token as a secret it might be nice for them to reuse it.

pkg/certificate/providers/config.go
jaellio: No worries! I was hoping I could get rid of it, so I was excited :)

cmd/osm-bootstrap/crds/config_mesh_root_certificate.yaml
jaellio: The Tresor cert can be created by the user or OSM. Currently, the `osm-ca-bundle` is an opaque secret unless we plan to no longer create an `osm-ca-bundle`.

pkg/certificate/providers/config.go
jaellio: I can add a TODO here

pkg/certificate/providers/types.go
jaellio: Done

cmd/osm-bootstrap/crds/config_mesh_root_certificate.yaml
jaellio: Updated the Vault token secretRef to be a secretKeyRef

nojnhuh (approved):

pkg/certificate/providers/types.go
schristoff: I'm a little confused with `DefaultVaultToken` 
Is this a Default OSM sets? That's typically how I read the word "Default" as in, the software has it's own defaults it provides if something is nil or not specified.
If this is the deprecated or user supplied Vault Token can we use a name such as `DeprecatedVaultToken` or something similar. Using the term Deprecated/Deprecation in var names will let other engineers know not to use that variable - and it will make it easy for cleanup later :)

PR #4743, opened 2022-05-13
Certificate issuance state management

steeling:
State management for certificate rotation

This PR implements certificate state management for rotation, by maintaining 2 `certificate.Issuer`'s (previously `certificate.Provider`). It adds an ID to each issuer, to be extracted from the MRC, to delineate between a prior issued cert, and the current intended issuer.

We then add a combine method on a certificate, which will add in a `pem.RootCertificate`, such that `GetIssuingCA` will now return the combined set of certificates for validation.

### DIFF
--- a/pkg/certificate/manager.go
+++ b/pkg/certificate/manager.go
@@ -74,7 +74,7 @@ func (m *Manager) checkAndRotate() {
 
 		if shouldRotate {
 			// Remove the certificate from the cache of the certificate manager
-			newCert, err := m.RotateCertificate(cert.GetCommonName())
+			newCert, err := m.rotateCertificate(cert.GetCommonName())
 			if err != nil {
 				// TODO(#3962): metric might not be scraped before process restart resulting from this error
 				log.Error().Err(err).Str(errcode.Kind, errcode.GetErrCodeWithMetric(errcode.ErrRotatingCert)).
@@ -144,7 +144,7 @@ func (m *Manager) ReleaseCertificate(cn CommonName) {
 }
 
 // RotateCertificate implements Manager and rotates an existing
-func (m *Manager) RotateCertificate(cn CommonName) (*Certificate, error) {
+func (m *Manager) rotateCertificate(cn CommonName) (*Certificate, error) {
 	start := time.Now()
 
 	oldObj, ok := m.cache.Load(cn)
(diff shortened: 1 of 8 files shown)


jaellio:
@steeling could you please update the PR description to elaborate on the changes made in this PR?

pkg/certificate/types.go
```diff
@@ -56,6 +57,9 @@ type Certificate struct {
 
 	// Certificate authority signing this certificate
 	IssuingCA pem.RootCertificate
+
+	keyIssuer string
+	pubIssuer string
```
jaellio, line 62:
```suggestion
	keyIssuerID string
	pubIssuerID string
```

pkg/certificate/manager.go
```diff
@@ -106,16 +105,34 @@ func (m *Manager) getFromCache(cn CommonName) *Certificate {
...
-	}
+			cert = cert.newMergedWithRoot(pubCert.GetIssuingCA())
+		}
 
-	// TODO(#4502): determine client(s) to use based on the rotation stage(s) of the client(s)
-	cert, err := m.clients[0].IssueCertificate(cn, validityPeriod)
-	if err != nil {
-		return cert, err
+		cert.keyIssuer = keyIssuer.ID
+		cert.pubIssuer = m.pubIssuer.ID
```
jaellio, line 134: Why is `m.pubIssuer.ID` vs `pubIssuer.ID`

steeling: good catch! fixed

pkg/certificate/manager.go
```diff
@@ -99,16 +104,32 @@ func (m *Manager) getFromCache(cn CommonName) *Certificate {
 
 // IssueCertificate implements Manager and returns a newly issued certificate from the given client.
 func (m *Manager) IssueCertificate(cn CommonName, validityPeriod time.Duration) (*Certificate, error) {
+	// var additionalRoot pem.RootCertificate
```
jaellio, line 107:
```suggestion
```

pkg/certificate/manager.go
```diff
@@ -59,6 +55,9 @@ func (m *Manager) Start(checkInterval time.Duration, certRotation <-chan struct{
 }
 
 func (m *Manager) checkAndRotate() {
+	// NOTE: checkAndRotate can reintroduce a certificate that has been released, thereby creating an unbounded cache.
```
jaellio, line 58: Should we create an issue to track this so it can be revisted?

steeling: I'm tempted not to, since I don't think it's very critical and happens in a few edge cases. 

For instance there's a ring buffer in the k8s informer  that is unbounded and a much larger concern for OOM's. My thought is this would get created then closed by the issue bot before it got prioritized for anybody to work on it. I'll leave it up to others though, if somebody else wants to create it!

steeling:
@nojnhuh @shashankram @trstringer friendly ping on this

shashankram (approved):

PR #4752, opened 2022-05-19
Feature/statefulsets

### LINKED ISSUE #3477: Support for Statefulsets
OSM should support routing to specific statefulset pods within a mesh.

- New Functionality          [x ]

**Possible use cases**
This is required for many statefulset applications like Kafka, MongoDB, cassandra, or other sharded/replicated databases.

Statefulsets use headless services to route to each pod, meaning an Endpoint is not created, but a DNS entry for each Pod is used. Pods in a statefulset have a predictable naming format, so it would be as simple as adding a domain entry for each statefulset, that maps to the Pod endpoint, or allowing DNS lookups to be used

keithmattix:
**Description**:
This PR fixes k8s service port protocol detection and a headless service example to the demo. Fixes #3477

**Testing done**:

**Affected area**:


demo/run-osm-demo.sh
```diff
@@ -37,6 +37,7 @@ ENABLE_FLUENTBIT="${ENABLE_FLUENTBIT:-false}"
 DEPLOY_PROMETHEUS="${DEPLOY_PROMETHEUS:-false}"
 DEPLOY_WITH_SAME_SA="${DEPLOY_WITH_SAME_SA:-false}"
 ENVOY_LOG_LEVEL="${ENVOY_LOG_LEVEL:-debug}"
+LOCAL_PROXY_MODE="${LOCAL_PROXY_MODE:-Localhost}"
```
shashankram, line 40: Please update `.env.example` on its usage.

pkg/constants/constants.go
```diff
@@ -239,3 +239,7 @@ const (
 	// multicluster field value for context field.
 	LogContextMulticluster = "multicluster"
 )
+
+var (
+	SupportedProtocols = []string{ProtocolHTTP, ProtocolHTTPS, ProtocolTCP, ProtocolGRPC, ProtocolTCPServerFirst}
```
shashankram, line 244: HTTPS is not a supported protocol for in-mesh traffic.

demo/deploy-zookeeper.sh
```diff
@@ -0,0 +1,60 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
keithmattix: Running MySQL replication is a bit more complicated, and consensus doesn't happen automatically. The zookeeper helm chart configures ZK nodes to automatically find each other and perform leader election. This isn't meant to be a first-class demo feature yet; just a short-cut for folks who want to prove Statefulsets work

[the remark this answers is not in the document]
shashankram: It seems this would be more appropriate to be an e2e test, followed by a standalone stateful set demo at https://release-v1-1.docs.openservicemesh.io/docs/demos/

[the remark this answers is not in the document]
keithmattix: E2E test [just landed](https://github.com/openservicemesh/osm/blob/168101f52c7f4f6df8b18cf07a9b798b9981b8f6/tests/e2e/e2e_statefulsets_test.go). Yeah, I definitely need to add a demo in there; should I add the demo before or after this PR?

Is there any harm in keeping this bash script in the repo though? It seems similar to the `deploy-vault.sh` script that's in the `main` branch: it doesn't get deployed in the regular demo, but it's there for convenience and displaying an OSM feature/integration.

demo/run-osm-demo.sh
keithmattix: Example is added: https://github.com/openservicemesh/osm/pull/4752/files#diff-a3046da0d15a27e89f2afe639b25748a7ad4d9290af3e7b1b6c1a5533c8f0a8cR187

demo/deploy-zookeeper.sh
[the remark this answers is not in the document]
shashankram: Yes having the script is fine. The demo can be added once the feature is ready.
My initial comment pertained to whether we could leverage the existing mysql stateful set. Alternatively, we could swap the mysql app with zookeeper in the default demo, if that's a better choice.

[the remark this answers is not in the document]
keithmattix: If we really wanted to, we could invest the time to make MySQL truly HA for the current demo. I just didn't think that was the best move for right now since zookeeper is a simpler demo. Using Zookeeper instead of MySQL would be weird since Zookeeper isn't a DBMS, but I wouldn't be opposed to a pubsub demo with Kafka (which uses zookeeper) at a later date

pkg/k8s/client.go
```diff
@@ -353,9 +354,21 @@ func ServiceToMeshServices(c Controller, svc corev1.Service) []service.MeshServi
...
 
+		// attempt to parse protocol from port name
+		protocol := constants.ProtocolHTTP
+		portNameProtocol := strings.Split(portSpec.Name, "-")[0]
+		for _, p := range constants.SupportedProtocols {
+			if p == portNameProtocol {
+				protocol = p
+				break
+			}
+		}
```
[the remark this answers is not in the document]
keithmattix: Added 👍🏾

pkg/constants/constants.go
keithmattix: Removed 👍🏾

tests/e2e/e2e_statefulsets_test.go
```diff
@@ -0,0 +1,153 @@
...
+
+				cond := Td.WaitForRepeatedSuccess(func() bool {
+
+					results := map[string]error{}
+					for _, po := range pods.Items {
+						stdout, stderr, err := Td.RunRemote(testNS, po.GetName(), "zookeeper", strings.Fields(cmd))
+
+						if err != nil {
+							// Error codes from the execution come through err
+							// Curl 'Connection refused' err code = 7
```
nojnhuh, line 126: If this isn't particularly relevant to zookeeper, I'd remove it.

tests/e2e/e2e_statefulsets_test.go
```diff
@@ -0,0 +1,153 @@
...
+
+						Td.T.Logf("> (%s) Stdout %s", po.GetName(), stdout)
+
+						results[po.GetName()] = err
+					}
+
+					hadErr := false
+					for podName, err := range results {
+						if err != nil {
+							Td.T.Logf("> (%s) ZK status check failed: expected nil err, got %s", podName, err)
```
nojnhuh, line 138: FWIW it looks like these errors are being logged twice.

nojnhuh (changes requested):

pkg/k8s/client.go
```diff
@@ -353,9 +354,25 @@ func ServiceToMeshServices(c Controller, svc corev1.Service) []service.MeshServi
...
 
+		// attempt to parse protocol from port name
+		// Order of Preference is:
+		// 1. port.appProtocol field
+		// 2. protocol prefixed to port name (e.g. tcp-my-port)
+		// 3. default to http
+		protocol := constants.ProtocolHTTP
+		portNameProtocol := strings.Split(portSpec.Name, "-")[0]
+		for _, p := range constants.SupportedProtocolsInMesh {
+			if p == portNameProtocol {
```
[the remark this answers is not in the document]
keithmattix: Oooh good catch; I forgot about tcp-server-first. I'll apply these fixes 👍🏾

tests/e2e/e2e_statefulsets_test.go
keithmattix: These are being logged per pod (there are 3 replicas). Are you seeing further duplication?

nojnhuh: This is the first place I see them logged:
https://github.com/openservicemesh/osm/blob/f3b661f9dc368912bac1e741b4312c3aec0d58c6/tests/e2e/e2e_statefulsets_test.go#L125-L127

keithmattix: Ah I see what you're saying. Code has been modified to only log err once 👍🏾

nojnhuh (approved):

shashankram (approved):

Issue #4754, opened 2022-05-23
Feature: Custom Trust Domains


steeling:
We have received feature requests from customers that require custom trust domains for their services. This means that certificates issued should have a common name of the format:

`<identity-name>.<namespace>.<custom-trust-domain>`

Additionally, we should add the DNS entry with the custom trust domain in `k8s.GetHostnamesForService`.

With the new MRC, we can build this in such a way that we can migrate from one trust domain to another with 0 downtime. To do so, we will add a new immutable field to the MRC, "trustDomain".

This issue will track the sub-issues for implementation:

- [x] Decouple service identity from common name, by stripping the trust domain from service identity
- [x] Add the trust domain field to the MRC, and plumb through to the Getter described in the step above.
- [x] Add a validation that the trust domain field cannot change to the validating webhook
- [x] Add the required business logic to actually use the new trust domain

steeling:
Brief proposal https://docs.google.com/document/d/1nLI4PcfB7-hdPvJipQwQ9dw28f8G7p77iqv3VzsCRFI/edit#

steeling:
Kind of. That is actually already possible although the service certs won’t have the custom trust domain. 

This is for users that use a PKI with a requirement that all certs have a specific trust domain

jsturtevant:
Synced with @steeling, this work is complete but dependent on cert-rotation work to be released

keithmattix:
@steeling If the work for this is complete, can we close this? I think we've captured cert-rotation work elsewhere

PR #4755, opened 2022-05-23
migrate sds secret names that contain a service account to use identity

steeling:
This PR prepares for custom trust domains, by migrating SDS Secrets' names from service account to Service Identity.

Currently there are 3 types of SDS Secrets:
```
"service-cert:<namespace>/<identity>"
"root-cert-for-mtls-outbound:<namespace>/<service>"
"root-cert-for-mtls-inbound:<namespace/identity>"
```

To support multiple trust domains, which will have different certificates, we need to make the names unique. The simplest way to do this is to include the trust domain itself, which is currently present in the identity. (Note that we'll also need to include it in the service, but we will tackle that in a future PR).

Instead of using the single type `SDSCert` to contain this information, with methods that only work if the underlying name/type is matched correctly, we instead break out SDSCert into an interface, backed by a type for each. We bring the error checking into the parsing method, which reduces the need for additional error checking throughout the code, and lastly remove/restructure a couple methods to remove additional boilerplate.

If you notice tests are deleted, rest-assured the test cases are maintained, but are instead moved to test the exported functions, instead of unit-testing unexported functions.

This is part of #4754


shashankram:
> To support multiple trust domains, which will have different certificates, we need to make the names unique

Could you explain with an example when we would have multiple trust domains for the same service/root?

steeling:
> > To support multiple trust domains, which will have different certificates, we need to make the names unique
> 
> Could you explain with an example when we would have multiple trust domains for the same service/root?

During a migration from 1 to another. This also will likely tee us up for multicluster use cases

shashankram:
> > > To support multiple trust domains, which will have different certificates, we need to make the names unique
> > 
> > 
> > Could you explain with an example when we would have multiple trust domains for the same service/root?
> 
> During a migration from 1 to another. This also will likely tee us up for multicluster use cases

I don't see a requirement for migration/multicluster in #4754. 

Given the nature of this change, could you kindly draft a design doc outlining the changes being proposed to add support for configurable trust domains before proceeding with this implementation, thanks

steeling:
> > > > To support multiple trust domains, which will have different certificates, we need to make the names unique
> > > 
> > > 
> > > Could you explain with an example when we would have multiple trust domains for the same service/root?
> > 
> > 
> > During a migration from 1 to another. This also will likely tee us up for multicluster use cases
> 
> I don't see a requirement for migration/multicluster in #4754.

Multicluster is not a requirement, but it's a nice to have. However 0 downtime migration is explicitly called out in that issue.

> 
> Given the nature of this change, could you kindly draft a design doc outlining the changes being proposed to add support for configurable trust domains before proceeding with this implementation, thanks

Design doc feels like overkill, since there's not a lot of ambiguity on how to implement this.

I think we should strive to provide zero-downtime migrations for users whenever possible. Is there a reason we shouldn't implement it this way?

FWIW I think we should be able to hash out discussion on the right approach in this PR

shashankram:
> > > > > To support multiple trust domains, which will have different certificates, we need to make the names unique
> > > > 
> > > > 
> > > > Could you explain with an example when we would have multiple trust domains for the same service/root?
> > > 
> > > 
> > > During a migration from 1 to another. This also will likely tee us up for multicluster use cases
> > 
> > 
> > I don't see a requirement for migration/multicluster in #4754.
> 
> Multicluster is not a requirement, but it's a nice to have. However 0 downtime migration is explicitly called out in that issue.
> 
> > Given the nature of this change, could you kindly draft a design doc outlining the changes being proposed to add support for configurable trust domains before proceeding with this implementation, thanks
> 
> Design doc feels like overkill, since there's not a lot of ambiguity on how to implement this.
> 

This feature makes fundamental changes to the control plane in a manner that affects how secrets are provisioned and used. I disagree that a design doc is an overkill. It will help understand the concrete requirements and evaluate the overall scope of the work involed and the proposed approach. 

Since you brought up things like multicluster and migration, which falls outside the scope of the issue referenced, I'd like to see something more concrete in the form of a doc that outlines how these pieces tie together.

steeling:
> > > > > > To support multiple trust domains, which will have different certificates, we need to make the names unique
> > > > > 
> > > > > 
> > > > > Could you explain with an example when we would have multiple trust domains for the same service/root?
> > > > 
> > > > 
> > > > During a migration from 1 to another. This also will likely tee us up for multicluster use cases
> > > 
> > > 
> > > I don't see a requirement for migration/multicluster in #4754.
> > 
> > 
> > Multicluster is not a requirement, but it's a nice to have. However 0 downtime migration is explicitly called out in that issue.
> > > Given the nature of this change, could you kindly draft a design doc outlining the changes being proposed to add support for configurable trust domains before proceeding with this implementation, thanks
> > 
> > 
> > Design doc feels like overkill, since there's not a lot of ambiguity on how to implement this.
> 
> This feature makes fundamental changes to the control plane in a manner that affects how secrets are provisioned and used. I disagree that a design doc is an overkill. It will help understand the concrete requirements and evaluate the overall scope of the work involed and the proposed approach.

A requirements doc sounds like more what you're looking for? If so, I'd be happy to put that together, as it would be significantly less effort than a DD.

> 
> Since you brought up things like multicluster and migration, which falls outside the scope of the issue referenced, I'd like to see something more concrete in the form of a doc that outlines how these pieces tie together.

Please ignore multicluster then.. I mentioned it in passing as a nice to have -- again it's not a hard requirement.
Migration however is definitively scoped in the referenced issue.

PR #4758, opened 2022-05-24
remove unused code paths

steeling:
Remove dead code


shashankram (commented):
This change seems to be doing more than simply removing dead code. Did you mean to include this refactor, and if yes could you update the commit title and description. Also, note that the CI is failing on this change.

steeling:
> This change seems to be doing more than simply removing dead code. Did you mean to include this refactor, and if yes could you update the commit title and description. Also, note that the CI is failing on this change.

Updated comment

shashankram (approved):

nojnhuh:
@steeling Could you explain in a bit more detail why these code paths are unreachable? That's not immediately obvious to me.

steeling:
> @steeling Could you explain in a bit more detail why these code paths are unreachable? That's not immediately obvious to me.

Sure thing!

The changed method is in `getRootCert`, which is only called in `getSDSSecrets`, and only in the case where the type is RootCert for inbound or outbound.

Now `getServiceIdentitiesFromCert` is only called in getRootCert, but it occurs after line154, which returns early if it is an Inbound cert, leaving only the case for outbound.

Yet `getServiceIdentitiesFromCert`, has a case for Inbound, which is unreachable.

After removing that case, the function is brought down to a few lines, and is better placed directly within `getRootCert`

PR #4759, opened 2022-05-24
Decouple certificate common name from various components

steeling:
In preparation for custom trust domains, I am beginning the process of decoupling the equality between certificate.CommonName, and identity.ServiceIdentity (they will remain tightly coupled, but not strictly equal)

There's a bunch of places in the code that currently relies on CertificateCommonName, particularly with respect to proxies, proxy maps (registries), and xdsLog. 

In these cases I prefer using either the UUID, when mapping directly to the proxy OR a new "proxy name" comprised of the uuid and service identity for the xdsLog to maintain the usefulness of having both the identity and uuid.

This PR also removes some dead code as a result of the above changes.


steeling:
Part of #4754

pkg/debugger/types.go
```diff
@@ -50,5 +50,5 @@ type MeshCatalogDebugger interface {
 // XDSDebugger is an interface providing debugging server with methods introspecting XDS.
 type XDSDebugger interface {
 	// GetXDSLog returns a log of the XDS responses sent to Envoy proxies.
-	GetXDSLog() *map[certificate.CommonName]map[envoy.TypeURI][]time.Time
```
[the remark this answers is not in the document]
steeling: My guess was a misunderstanding with maps vs structs because this one is being copied.

pkg/envoy/ads/profile.go
```diff
@@ -29,23 +28,23 @@ func xdsPathTimeTrack(startedAt time.Time, typeURI envoy.TypeURI, proxy *envoy.P
 		Observe(elapsed.Seconds())
 }
 
-func (s *Server) trackXDSLog(cn certificate.CommonName, typeURL envoy.TypeURI) {
+func (s *Server) trackXDSLog(proxyID string, typeURL envoy.TypeURI) {
 	s.withXdsLogMutex(func() {
-		if _, ok := s.xdsLog[cn]; !ok {
-			s.xdsLog[cn] = make(map[envoy.TypeURI][]time.Time)
+		if _, ok := s.xdsLog[proxyID]; !ok {
+			s.xdsLog[proxyID] = make(map[envoy.TypeURI][]time.Time)
```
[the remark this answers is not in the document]
steeling: It was actually using the bootstrap CN which contains the proxy uiid so it was per proxy before as well

pkg/envoy/ads/secrets.go
```diff
@@ -1,73 +0,0 @@
-package ads
-
-import (
-	xds_discovery "github.com/envoyproxy/go-control-plane/envoy/service/discovery/v3"
```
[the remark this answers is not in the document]
steeling: Yup

pkg/envoy/ads/server.go
```diff
@@ -51,12 +50,10 @@ func NewADSServer(meshCatalog catalog.MeshCataloger, proxyRegistry *registry.Pro
 		osmNamespace:   osmNamespace,
 		cfg:            cfg,
 		certManager:    certManager,
-		xdsMapLogMutex: sync.Mutex{},
-		xdsLog:         make(map[certificate.CommonName]map[envoy.TypeURI][]time.Time),
+		xdsLog:         make(map[string]map[envoy.TypeURI][]time.Time),
 		workqueues:     workerpool.NewWorkerPool(workerPoolSize),
 		kubecontroller: kubecontroller,
 		cacheEnabled:   cfg.GetFeatureFlags().EnableSnapshotCacheMode,
-		configVerMutex: sync.Mutex{},
```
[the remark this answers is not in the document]
steeling: It’s a struct so it’s zero value is the same

shashankram:
I understand what this PR is doing, but couldn't understand how this change helps #4754. Could you describe in the commit message why this decoupling is necessary? 
Thanks

steeling:
> I understand what this PR is doing, but couldn't understand how this change helps #4754. Could you describe in the commit message why this decoupling is necessary?
> Thanks

Good question! That’s definitely not obvious on first pass

1. A proxy’s CN can change with rotated trust domains, so we want to keep these mappings stable
2. After we decouple service identity and from common name, we can no longer derive the CN without access to the cert manager instance, and knowing which trust domain to use

pkg/debugger/types.go
```diff
@@ -50,5 +50,5 @@ type MeshCatalogDebugger interface {
...
+	GetXDSLog() map[string]map[envoy.TypeURI][]time.Time
```
shashankram, line 53: An additional comment regarding the key in the map being returned would improve readability since it has been changed to a generic string type.

pkg/envoy/proxy.go
```diff
@@ -168,6 +170,11 @@ func (p *Proxy) PodMetadataString() string {
 	return fmt.Sprintf("UID=%s, Namespace=%s, Name=%s, ServiceAccount=%s", p.PodMetadata.UID, p.PodMetadata.Namespace, p.PodMetadata.Name, p.PodMetadata.ServiceAccount.Name)
 }
 
+// GetName returns a unique name for this proxy based on the identity and uuid.
+func (p *Proxy) GetName() string {
+	return fmt.Sprintf("%s-%s", p.Identity.String(), p.UUID.String())
```
shashankram, line 175: Since `-` is a part of the UUID, I would use a different delimiter for easier parsing.
```suggestion
	return fmt.Sprintf("%s:%s", p.Identity.String(), p.UUID.String())
```

pkg/envoy/registry/announcement_handlers.go
```diff
@@ -33,13 +33,14 @@ func (pr *ProxyRegistry) ReleaseCertificateHandler(certManager certificateReleas
 				continue
 			}
 
-			podUID := deletedPodObj.GetObjectMeta().GetUID()
-			if podIface, ok := pr.podUIDToCN.Load(podUID); ok {
-				endpointCN := podIface.(certificate.CommonName)
-				log.Warn().Msgf("Pod with UID %s found in proxy registry; releasing certificate %s", podUID, endpointCN)
-				certManager.ReleaseCertificate(endpointCN)
+			uuid := deletedPodObj.Labels[constants.EnvoyUniqueIDLabelName]
```
shashankram, line 36:
```suggestion
			proxyUUID := deletedPodObj.Labels[constants.EnvoyUniqueIDLabelName]
```

pkg/envoy/registry/registry.go
```diff
@@ -19,27 +17,16 @@ func NewProxyRegistry(mapper ProxyServiceMapper, msgBroker *messaging.Broker) *P
...
-	// If this proxy object is on a Kubernetes Pod - it will have an UID
-	if proxy.HasPodMetadata() {
-		podUID := types.UID(proxy.PodMetadata.UID)
-
-		// Create a PodUID to Certificate CN map so we can easily determine the CN from the PodUID
-		pr.podUIDToCN.Store(podUID, proxy.GetCertificateCommonName())
-
-		// Create a PodUID to Cert Serial Number so we can easily look-up the SerialNumber of the cert issued to a proxy for a given Pod.
-		pr.podUIDToCertificateSerialNumber.Store(podUID, proxy.GetCertificateSerialNumber())
-	}
```
shashankram, line 36: I'd also note in the commit description that the change removes dead code arising from the refactor.

pkg/envoy/proxy.go
```diff
@@ -245,6 +252,7 @@ func NewProxy(certCommonName certificate.CommonName, certSerialNumber certificat
 		xDSCertificateCommonName:   certCommonName,
 		xDSCertificateSerialNumber: certSerialNumber,
 		UUID:                       cnMeta.ProxyUUID,
+		Identity:                   cnMeta.ServiceIdentity,
```
shashankram, line 255: Could you also add an example of the Identity string here to improve readability. It would be convenient to know what this is expected to look like.

pkg/envoy/ads/profile.go
```diff
@@ -29,23 +28,23 @@ func xdsPathTimeTrack(startedAt time.Time, typeURI envoy.TypeURI, proxy *envoy.P
 		Observe(elapsed.Seconds())
 }
 
-func (s *Server) trackXDSLog(cn certificate.CommonName, typeURL envoy.TypeURI) {
+func (s *Server) trackXDSLog(proxyID string, typeURL envoy.TypeURI) {
```
nojnhuh, line 31: If the proxy's "name" is what's being passed here, I'd rename this accordingly since the proxy also has a plain "(UU)ID" which might make this confusing.

pkg/envoy/ads/response_test.go
```diff
@@ -97,30 +96,6 @@ var _ = Describe("Test ADS response functions", func() {
 		Expect(err).ToNot(HaveOccurred())
 	})
 
-	Context("Test makeRequestForAllSecrets()", func() {
```
nojnhuh, line 100: Since some changes here look the same as #4758, please update this branch with the recent changes from `main` (ideally with a merge, not a rebase) so we don't have to review the same changes again.

pkg/envoy/rds/response.go
```diff
@@ -18,13 +18,6 @@ import (
 
 // NewResponse creates a new Route Discovery Response.
 func NewResponse(cataloger catalog.MeshCataloger, proxy *envoy.Proxy, discoveryReq *xds_discovery.DiscoveryRequest, cfg configurator.Configurator, _ *certificate.Manager, proxyRegistry *registry.ProxyRegistry) ([]types.Resource, error) {
-	proxyIdentity, err := envoy.GetServiceIdentityFromProxyCertificate(proxy.GetCertificateCommonName())
-	if err != nil {
-		log.Error().Err(err).Str(errcode.Kind, errcode.GetErrCodeWithMetric(errcode.ErrGettingServiceIdentity)).
```
nojnhuh, line 23: Looks like `errcode.ErrGettingServiceIdentity` is no longer used at all.

pkg/envoy/proxy.go
steeling: I'm hesitant to add this, as I think Identity should be treated opaque. I'd even go as far to say that clients/callers shouldn't even know that it's just an aliased string, and treat it as an opaque object.

pkg/envoy/ads/profile.go
steeling: Good call! Note that I'm only having a name, to maintain the usefulness of the CommonName which included both the service identity and uuid, to ensure that all the information is included for customers.

I'd rather just key it by UUID, but not sure if that reduces readability/usefulness. Curious on your thoughts here. Adding @shashankram to comment as well

pkg/envoy/rds/response.go
steeling: done!

pkg/debugger/types.go
steeling: done!

pkg/envoy/proxy.go
steeling: good call! fixed!

pkg/envoy/registry/announcement_handlers.go
steeling: done!

pkg/envoy/ads/response_test.go
steeling: done!

pkg/envoy/proxy.go
shashankram: Not documenting a sample value hinders readability for me. Even if it's opaque, a sample value in a comment is enough to clarify what can be expected here so it's easier to reason about it.

pkg/envoy/ads/profile.go
shashankram: In the past, logging by UUID was rather inconvenient to debug the concerned proxy by looking at the logs. Since the CommonName used k8s primitives such as service account and namespace, it made parsing these logs easier. So I would say let's continue to add additional context in the logs and not just its UUID.

pkg/envoy/proxy.go
steeling: ok, added

shashankram (approved):

pkg/envoy/ads/profile.go
steeling: yup, that's what i figured. Modified the field name!

nojnhuh: Yeah, I figured if this is ultimately shown as debug info then including some more meaningful info like the service name/namespace is valuable here. Figuring out what's what based only on UUIDs would be a lot more involved.

nojnhuh (approved):

pkg/debugger/proxy.go
```diff
@@ -20,8 +20,8 @@ func (ds DebugConfig) getProxies() http.Handler {
 	// the type (map) required by the printProxies function.
 	listConnected := func() map[certificate.CommonName]time.Time {
 		proxies := make(map[certificate.CommonName]time.Time)
-		for cn, proxy := range ds.proxyRegistry.ListConnectedProxies() {
-			proxies[cn] = (*proxy).GetConnectedAt()
+		for _, proxy := range ds.proxyRegistry.ListConnectedProxies() {
+			proxies[proxy.GetCertificateCommonName()] = (*proxy).GetConnectedAt()
```
jaellio, line 24: Should we be calling `GetName()` instead of `GetCertificateCommonName()`?
```suggestion
			proxies[proxy.GetName()] = (*proxy).GetConnectedAt()
```

pkg/envoy/ads/types.go
```diff
@@ -27,10 +27,11 @@ var (
...
 // Server implements the Envoy xDS Aggregate Discovery Services
 type Server struct {
-	catalog        catalog.MeshCataloger
-	proxyRegistry  *registry.ProxyRegistry
-	xdsHandlers    map[envoy.TypeURI]func(catalog.MeshCataloger, *envoy.Proxy, *xds_discovery.DiscoveryRequest, configurator.Configurator, *certificate.Manager, *registry.ProxyRegistry) ([]types.Resource, error)
-	xdsLog         map[certificate.CommonName]map[envoy.TypeURI][]time.Time
+	catalog       catalog.MeshCataloger
+	proxyRegistry *registry.ProxyRegistry
+	xdsHandlers   map[envoy.TypeURI]func(catalog.MeshCataloger, *envoy.Proxy, *xds_discovery.DiscoveryRequest, configurator.Configurator, *certificate.Manager, *registry.ProxyRegistry) ([]types.Resource, error)
+	// xdsLog is a map of key proxy.GetName(), which is of the form <identity>/<uuid> to map xds TypeURI to a slice of timestamps
```
jaellio, line 33:
```suggestion
	// xdsLog is a map of key proxy.GetName(), which is of the form <identity>:<uuid> to map xds TypeURI to a slice of timestamps
```

pkg/debugger/proxy.go
steeling: on first glance you would think so, but this is actually creating a new map from the old map, keyed by common name. I have some future changes that are removing the GetCommonName method, so that will change in the future, and we can just reuse the existing map.

jaellio: Thank you for clarifying!

PR #4763, opened 2022-05-25
Decouple certificate common name from proxy registry

steeling:
Use proxy UUID for the key in the proxy registry, as opposed to certificate common.

This is part of the effort to decouple service identity and certificate common name.

A proxy's certificate common name may change, when the trust domain is rotated. Maintaining a stable cache, and a stable identity requires this decoupling.

Part of #4754

### DIFF
--- a/pkg/debugger/proxy.go
+++ b/pkg/debugger/proxy.go
@@ -62,18 +62,14 @@ func (ds DebugConfig) getConfigDump(uuid string, w http.ResponseWriter) {
 	if !ok {
 		msg := fmt.Sprintf("Proxy for UUID %s not found, may have been disconnected", uuid)
 		log.Error().Msg(msg)
-		if _, err := w.Write([]byte(msg)); err != nil {
-			log.Error().Err(err).Msg("Error writing debugger response")
-		}
+		http.Error(w, msg, http.StatusNotFound)
 		return
 	}
 	pod, err := envoy.GetPodFromCertificate(proxy.GetCertificateCommonName(), ds.kubeController)
 	if err != nil {
 		msg := fmt.Sprintf("Error getting Pod from certificate with CN=%s", proxy.GetCertificateCommonName())
 		log.Error().Err(err).Msg(msg)
-		if _, err := w.Write([]byte(msg)); err != nil {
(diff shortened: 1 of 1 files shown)


pkg/debugger/proxy.go
```diff
@@ -6,71 +6,99 @@ import (
...
+			idx+1, proxy.Identity, ts, time.Since(ts), proxyURL, configDumpURL)
 	}
 	_, _ = fmt.Fprint(w, `</table>`)
 }
 
-func (ds DebugConfig) getConfigDump(cn certificate.CommonName, w http.ResponseWriter) {
-	pod, err := envoy.GetPodFromCertificate(cn, ds.kubeController)
+func (ds DebugConfig) getConfigDump(uuid string, w http.ResponseWriter) {
+	proxy, ok := ds.proxyRegistry.GetConnectedProxy(uuid)
+	if !ok {
```
nojnhuh, line 62: Should we have a `http.Error()` or otherwise make sure the HTTP status is set to something non-2xx in these kinds of cases?

steeling: good call, fixed

pkg/envoy/registry/registry.go
```diff
@@ -15,22 +13,40 @@ func NewProxyRegistry(mapper ProxyServiceMapper, msgBroker *messaging.Broker) *P
...
+}
+
 // UnregisterProxy unregisters the given proxy from the catalog.
 func (pr *ProxyRegistry) UnregisterProxy(p *envoy.Proxy) {
 	pr.connectedProxies.Delete(p.UUID.String())
 	log.Debug().Msgf("Unregistered proxy %s", p.String())
 }
 
 // GetConnectedProxyCount counts the number of connected proxies
+// TODO(steeling): switch to a regular map with mutex so we can get the count without iterating the entire list.
```
trstringer, line 38: Can you create an issue to track this future work and add it to vFuture?

trstringer (approved):
Looks good, just one small request that isn't blocking the PR!

pkg/envoy/registry/registry.go
steeling: done! and squashed to 1 commit

PR #4767, opened 2022-05-26
add trust domain to mesh root certificate

steeling:
Add the trust domain field to the MRC

### DIFF
--- a/cmd/osm-bootstrap/crds/config_mesh_root_certificate.yaml
+++ b/cmd/osm-bootstrap/crds/config_mesh_root_certificate.yaml
@@ -52,6 +52,10 @@ spec:
               required:
                 - provider
               properties:
+                trustDomain:
+                  description: Trust Domain to use in common name for certificates, ie "example.com"
+                  type: string
+                  default: cluster.local
                 provider:
                   description: Certificate provider used by the mesh control plane
                   type: object
--- a/cmd/osm-bootstrap/crds/config_mesh_root_certificate.yaml
+++ b/cmd/osm-bootstrap/crds/config_mesh_root_certificate.yaml
@@ -19,7 +19,7 @@ kind: CustomResourceDefinition
 metadata:
   name: meshrootcertificates.config.openservicemesh.io
   labels:
-    app.kubernetes.io/name : "openservicemesh.io"
+    app.kubernetes.io/name: "openservicemesh.io"
 spec:
   group: config.openservicemesh.io
   scope: Namespaced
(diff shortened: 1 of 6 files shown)


cmd/osm-bootstrap/crds/config_mesh_root_certificate.yaml
```diff
@@ -52,6 +52,10 @@ spec:
               required:
                 - provider
               properties:
+                trustDomain:
+                  description: Trust Domain to use in common name for certificates, ie "example.com"
```
nojnhuh, line 56: nit:
```suggestion
                  description: Trust Domain to use in common name for certificates, e.g. "example.com"
```

nojnhuh (approved):

pkg/apis/config/v1alpha2/meshrootcertificate.go
```diff
@@ -30,6 +30,9 @@ type MeshRootCertificate struct {
 type MeshRootCertificateSpec struct {
 	// Provider specifies the mesh certificate provider
 	Provider ProviderSpec `json:"provider"`
+
+	// TrustDomain is the trust domain to use as a suffix in Common Names for new certificates.
+	TrustDomain string `json:"trustDomain"`
```
jaellio, line 35: Have you thought about making the TrustDomain a pointer? That way we could distinguish between the trust domain being set to an empty string vs not being set at all

jaellio: If it is a required field can it be added under `required` in `config_mesh_root_certificate.yaml`?

steeling: default and required are mutually exclusive. It's explicitly not required, in that it allows a user to omit it, in which case the default value (cluster.local) will be used.

steeling: from [openapi docs](https://swagger.io/docs/specification/describing-parameters/#common-mistakes-10):


> Common Mistakes
> There are two common mistakes when using the default keyword:
> Using default with required parameters or properties, for example, with path parameters. This does not make sense – if a value is required, the client must always send it, and the default value is never used.
> Using default to specify a sample value. This is not intended use of default and can lead to unexpected behavior in some Swagger tools. Use the example or examples keyword for this purpose instead. See [Adding Examples](https://swagger.io/docs/specification/adding-examples/).
>

jaellio: I missed the default somehow :)

jaellio (approved):

PR #4768, opened 2022-05-26
ref(e2e): move k8s version test config to CI

nojnhuh:
**Description**: There are 2 sets of changes here that I made in separate commits to make them maybe easier to review since they overlap a bit but are strictly dependent. The commit messages have more details, but it's roughly:
- (refactor) remove PR/push distinction in e2e tests
- (the actual fix) move k8s version test config to CI

**Testing done**:
- verified the test is using the right version of Kubernetes in CI:
  - Default: https://github.com/nojnhuh/osm/runs/6615268990?check_suite_focus=true#step:7:33
  - v1.22.7: https://github.com/nojnhuh/osm/runs/6615269545?check_suite_focus=true#step:7:33
  - v1.21.10: https://github.com/nojnhuh/osm/runs/6615269445?check_suite_focus=true#step:7:33
- Tested that failed bucket and k8s_version e2e jobs both have artifacts uploaded successfully:
  - https://github.com/nojnhuh/osm/actions/runs/2392290219
  - https://github.com/nojnhuh/osm/actions/runs/2392300420

**Affected area**:


.github/workflows/main.yml
```diff
@@ -202,7 +202,16 @@ jobs:
...
+        k8s_version: [""]
+        focus: [""]
         bucket: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
+        include:
+          - k8s_version: v1.21.10
+            focus: "Test traffic flowing from client to server with a Kubernetes Service for the Source: HTTP"
+            bucket: ".*"
+          - k8s_version: v1.22.7
+            focus: "Test traffic flowing from client to server with a Kubernetes Service for the Source: HTTP"
+            bucket: ".*"
```
trstringer, line 214: Couldn't this be abbreviated with the following?

```yaml
matrix:
  k8s_version: ["v1.21.10", "v1.22.7"]
  focus: ["Test traffic flowing from client to server with a Kubernetes Service for the Source: HTTP"]
  bucket: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
```

I don't think there's anything gained by expanding the matrix config here with `include`.

shashankram (approved):

.github/workflows/main.yml
```diff
@@ -229,46 +238,32 @@ jobs:
...
-        run: go test ./tests/e2e -test.v -ginkgo.v -ginkgo.progress -installType=KindCluster -test.timeout 0 -test.failfast -ginkgo.failFast -ginkgo.focus='\[Bucket ${{ matrix.bucket }}\]'
-        continue-on-error: true
-      - name: Upload Push test logs
-        if: ${{ steps.push_test.conclusion != 'skipped' }}
+      - name: Set Logs name
+        if: ${{ steps.test.conclusion != 'skipped' }}
+        id: logname
+        run: |
+          if [[ -n "${{ matrix.k8s_version }}" ]]; then
+            echo "::set-output name=name::test_logs_k8s_version_${{ matrix.k8s_version }}"
```
trstringer, line 252: I think passing data between different steps is better served with appending the `GITHUB_ENV` file, although this does work with the workflow command `set-output` (which is more typically used to set output values for the workflow itself, for instance in the context of reusable workflows). I think this changed at some point for passing data between steps. But this is personal preference and not a required change.

If you do stay with `set-output`, can you rename the output variable from `name` to something more descriptive?

```suggestion
            echo "::set-output name=K8S_VERSION::test_logs_k8s_version_${{ matrix.k8s_version }}"
```

Or something to that effect. Note if you do make this change it'll need to be reflected below where it is referenced.

.github/workflows/main.yml
```diff
@@ -229,46 +238,32 @@ jobs:
...
+          else
+            echo "::set-output name=name::test_logs_bucket_${{ matrix.bucket }}"
```
trstringer, line 254:
```suggestion
            echo "::set-output name=BUCKET::test_logs_bucket_${{ matrix.bucket }}"
```

Same as above, rename `name` to something more descriptive.

trstringer (changes requested):

jaellio (approved):

.github/workflows/main.yml
trstringer: Ah I see, thanks for the clarification.

trstringer (approved):
Looks like DCO needs to be fixed prior to merging.

PR #4775, opened 2022-05-31
fix(certs): update checkAndRotate to use validity duration from MeshConfig

jaellio:
**Description**:

Updates the checkAndRotate method to only rotate service
certificates. Currently, the method will rotate any certificate
that expires and set the validity period of that certificate to
the service cert validity duration. Therefore, if any webhook,
ads, or ingress cert expired the validity would be set to 24 hrs
rather than the default 10 years. Additionally, the checkAndRotate
method referenced the service cert validity period directly rather
than through the cached config. If the validity was updated in
the mesh config, once certificates expired they would be reissued
with the original duration.

Change notes:
- Added a Certificate type to the Certificate struct. This is used
only by the OSM control plane to determine rotation
responsibility. This will reduce complexity when adding support
for rotating the webhook and ads certs during root cert rotation.
- IssueCertificate returns a rotated bool. This is used to
determine if cert rotation occurred when calling IssueCertificate
or if the cert was just returned from the cache.

**Testing done**:
- Verified expiring certs are re-issued with the current `serviceCertValidityDuration` as defined in the OSM MeshConfig. Currently, expiring service certs are only re-issued with the original `serviceCertValidityDuration`.
  - Verified the bug on the main branch by running the osm demo with the `serviceCertValidityDuration` set to 5m. After OSM install, I patched the MeshConfig and updated the validity duration to 4m. Expiring certs were reissued using the original 5m validity duration.
  - Verified the bug is no longer present with this change by following the same steps as above
- Verified only service certs are rotated in checkAndRotateServiceCerts

**Affected area**:


steeling:
> **Description**:
> 
> Updates the checkAndRotate method to only rotate service certificates. Currently, the method will rotate any certificate that expires and set the validity period of that certificate to the service cert validity duration. Therefore, if any webhook, ads, or ingress cert expired the validity would be set to 24 hrs rather than the default 10 years. Additionally, the checkAndRotate method referenced the service cert validity period directly rather than through the cached config. If the validity was updated in the mesh config, once certificates expired they would be reissued with the original duration.

Why not rotate them though? Wouldn't rotating be a better experience, even if it brings down the duration to something much smaller, as opposed to letting them expire? Now that we can rotate these, should we even keep them at a decade?

If we really do want to  keep these long lived certs, should we instead annotate the certificate with the duration, instead of the "type"?

pkg/certificate/manager.go
```diff
@@ -12,7 +13,7 @@ import (
 )
 
 // NewManager creates a new CertManager with the passed CA and CA Private Key
-func NewManager(mrcClient MRCClient, serviceCertValidityDuration time.Duration, msgBroker *messaging.Broker) (*Manager, error) {
+func NewManager(mrcClient MRCClient, getServiceCertValidityPeriod func() time.Duration, msgBroker *messaging.Broker) (*Manager, error) {
```
steeling, line 16: Could we move the validity period to the MRC itself? Since there's already a mechanism for updating that

pkg/certificate/manager.go
```diff
@@ -41,42 +42,47 @@ func NewManager(mrcClient MRCClient, serviceCertValidityDuration time.Duration,
...
 	for _, cert := range m.ListIssuedCertificates() {
-		shouldRotate := cert.ShouldRotate()
+		if cert.GetCertificateType() != Service {
+			log.Debug().Msgf("Skipping check and rotate of cert %s of type %s since it is not of type Service",
+				cert.GetCommonName(),
+				cert.GetCertificateType())
+			continue
+		}
+
+		newCert, rotated, err := m.IssueCertificate(cert.GetCommonName(), m.serviceCertValidityDuration(), Service)
```
steeling, line 70: Looks like the rotated bool is only used here. Instead of modifying the API and exposing that info when callers don't seem to care, what if we did:

```go
rotated := cert != newCert
```

And kept the old method signature

jaellio:
> > **Description**:
> > Updates the checkAndRotate method to only rotate service certificates. Currently, the method will rotate any certificate that expires and set the validity period of that certificate to the service cert validity duration. Therefore, if any webhook, ads, or ingress cert expired the validity would be set to 24 hrs rather than the default 10 years. Additionally, the checkAndRotate method referenced the service cert validity period directly rather than through the cached config. If the validity was updated in the mesh config, once certificates expired they would be reissued with the original duration.
> 
> Why not rotate them though? Wouldn't rotating be a better experience, even if it brings down the duration to something much smaller, as opposed to letting them expire? Now that we can rotate these, should we even keep them at a decade?

Hi Sean! Thanks for your feedback! This is still a draft PR so I am still  We can rotate certs other than service certificates. Being able to rotate certs due to expiration or root certificate rotation is the goal. Currently, even though we "checkAndRotate" all certificates that have been issued by OSM, the rotation only impacts service certs, since all other certs are directly referenced and therefore do not support rotation. 

> If we really do want to keep these long lived certs, should we instead annotate the certificate with the duration, instead of the "type"?

I don't think we should store the certificate duration in the certificate to be used when reissuing a certificate. The durations could be

PR #4789, opened 2022-06-03
feat(retry): Add validating webhook for retry policy

### LINKED ISSUE #4778: Add validation webhook for retry policy

- New Functionality          [ x]

**Possible use cases**

Implement validation webhook for retry and remove checks here - https://github.com/openservicemesh/osm/blob/main/pkg/envoy/rds/route/route_config.go#L299-L316

shalier:
**Description**:
Adds validating webhook for retry policy and sets PerTryTimeout, RetryBackoffBaseInterval,
 and NumRetries to optional fields.
Resolves #4778

**Testing done**:
Added unit test for retryValidator function

**Affected area**:

2. Is this a breaking change?

### DIFF
--- a/pkg/validator/validators.go
+++ b/pkg/validator/validators.go
@@ -7,15 +7,18 @@ import (
 	"net"
 	"strconv"
 	"strings"
+	"time"
 
 	mapset "github.com/deckarep/golang-set"
 	"github.com/pkg/errors"
 	smiAccess "github.com/servicemeshinterface/smi-sdk-go/pkg/apis/access/v1alpha3"
 	smiSpecs "github.com/servicemeshinterface/smi-sdk-go/pkg/apis/specs/v1alpha4"
 	admissionv1 "k8s.io/api/admission/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/validation/field"
 
 	configv1alpha2 "github.com/openservicemesh/osm/pkg/apis/config/v1alpha2"
+	"github.com/openservicemesh/osm/pkg/apis/policy/v1alpha1"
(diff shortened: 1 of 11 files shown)


nojnhuh (approved):
Left some nits over tests, but LGTM

pkg/validator/validators_test.go
```diff
@@ -786,6 +786,181 @@ func TestIngressBackendValidator(t *testing.T) {
...
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			assert := tassert.New(t)
+
+			resp, err := retryValidator(tc.input)
+			assert.Equal(tc.expResp, resp)
+			if err != nil {
+				assert.Equal(tc.expErrStr, err.Error())
+			}
```
nojnhuh, line 959: This could pass when (I would argue) it should fail when the test defines a non-empty `tc.expErrStr` but a nil error is actually returned. One way to get around that could be with something like:
```suggestion
			errStr := ""
			if err != nil {
				errStr = err.Error()
			}
			assert.Equal(tc.expErrStr, errStr)
```

pkg/validator/validators.go
```diff
@@ -186,6 +189,47 @@ func (kc *policyValidator) ingressBackendValidator(req *admissionv1.AdmissionReq
 	return nil, nil
 }
 
+// retryValidator validates the Retry custom resource
+func retryValidator(req *admissionv1.AdmissionRequest) (*admissionv1.AdmissionResponse, error) {
```
jaellio, line 193: If the source and destination of a retry policy are the same as an existing retry policy with a different name should it be rejected?

shalier: You're right! But will hold off on this PR (put into draft mode), to work on "Retries: consolidate retries into UpstreamTrafficSetting API" #4799 for now.

PR #4791, opened 2022-06-03
Remove unused code paths and switch the policy object to a policy builder

steeling:
Remove large swaths of code by removing unused code paths and merging some of the resulting small functions into their callers.

This PR adds the following simplifications (with justification provided) around building RBAC rules from the policy object:

* Remove the method `AddRule`, which is only called in a single instance, where there are no pre-existing rules (permissive mode).
* Modify `getTrafficSpecName` from a method to a function. This provides additional clarity to the reader since the reader knows the method getTrafficSpecName only operates on its parameters, and not on the fields of the method's instance.
* Support only OR rules for Principals. Note that this changes from nesting inside of a single Top-level "or Rule", to simply listing all of the principals, which uses OR semantics by default.
* Both OR and AND are supported for permissions, however OR rules are similarly moved to top level permissions, which also use OR by default
* Simplify the wildcard scenario to result in a single principal of `ANY`, whereas before it could have both `ANY`, and specific principals.
* Remove code paths associated with `allowPartialHostnamesMatch`, since only a single value was ever provided.


pkg/envoy/rbac/policy.go
```diff
@@ -1,136 +1,71 @@
...
+}
 
-	// Construct the Principals ------------------------
-	var finalPrincipals []*xds_rbac.Principal
+// Build constructs an RBAC policy for the policy object on which this method is called
+func (p *PolicyBuilder) Build() *xds_rbac.Policy {
+	policy := &xds_rbac.Policy{}
 
 	// Each RuleList follows OR semantics with other RuleList in the list of RuleList
-	for _, principalRuleList := range p.Principals {
```
shashankram, line 19: Why is this being changed?

pkg/envoy/rbac/policy.go
```diff
@@ -1,136 +1,71 @@
...
-		}
+	return policy
+}
 
-		finalPermissions = append(finalPermissions, currentPermission)
+// AddPrincipal adds a principal to the list of allowed principals
+func (p *PolicyBuilder) AddPrincipal(principal string) {
+	// We need this extra defense in depth because it is currently possible to configure a wildcard principal
+	// in addition to specific principals. We should instead not allow this, preventing this misconfiguration from
+	// happening through validating webhooks.
```
shashankram, line 50: I am unsure what this implies. Is there a problem with the existing code?

pkg/trafficpolicy/trafficpolicy.go
```diff
@@ -61,26 +60,6 @@ func (rwc *RouteWeightedClusters) TotalClustersWeight() int {
 	return totalWeight
 }
 
-// AddRule adds a Rule to an InboundTrafficPolicy based on the given HTTP route match, weighted cluster, and allowed service account
-//	parameters. If a Rule for the given HTTP route match exists, it will add the given service account to the Rule. If the the given route
-//	match is not already associated with a Rule, it will create a Rule for the given route and service account.
-func (in *InboundTrafficPolicy) AddRule(route RouteWeightedClusters, allowedServiceIdentities identity.ServiceIdentity) {
```
shashankram, line 67: Why is this being removed?

pkg/trafficpolicy/trafficpolicy.go
```diff
@@ -116,22 +95,15 @@ func (out *OutboundTrafficPolicy) AddRoute(httpRouteMatch HTTPRouteMatch, retryP
 // 2. when a policy having its hostnames from a host header needs to be merged with other inbound traffic policies
 // in either of these cases the will be only a single hostname and there is a possibility that this hostname is part of an existing traffic policy
 // hence the rules need to be merged
-func MergeInboundPolicies(allowPartialHostnamesMatch bool, original []*InboundTrafficPolicy, latest ...*InboundTrafficPolicy) []*InboundTrafficPolicy {
```
shashankram, line 119: Why is this being changed?

shashankram (changes requested):
I see too many parts of the code being changed without a clear motivation being described in the commit message. It's hard for me to understand why significant parts of the RBAC policy builder + traffic policy helpers are being refactored.

My request to you is:
1. Clearly describe the motivation of the PR in the commit and PR descriptions
2. Break this PR into smaller independent units (removal of unused code, traffic policy, RBAC etc.)
3. Open a Github issue for larger refactors so they can be tracked and discussed

steeling:
> I see too many parts of the code being changed without a clear motivation being described in the commit message. It's hard for me to understand why significant parts of the RBAC policy builder + traffic policy helpers are being refactored.
> 
> My request to you is:
> 
> 1. Clearly describe the motivation of the PR in the commit and PR descriptions
> 2. Break this PR into smaller independent units (removal of unused code, traffic policy, RBAC etc.)
> 3. Open a Github issue for larger refactors so they can be tracked and discussed

A net 650 lines of code removed is a big benefit in my opinion. All of the refactored code is due to unused code paths, or inlining functions that are called in one place (particularly after portions of that function that are not used are removed).

The net result is a significant improvement in readability, simplicity

pkg/trafficpolicy/trafficpolicy.go
steeling: it's only ever used in one context, with allowPartialHostnamesMatch set to true

pkg/envoy/rbac/policy.go
steeling: having both a wildcard and a principal reduces to the wildcard, and specifying extra is unnecessary. 

If you can think of a more clear way to word the comment, please let me know and I'll change it :)

shashankram:
> > I see too many parts of the code being changed without a clear motivation being described in the commit message. It's hard for me to understand why significant parts of the RBAC policy builder + traffic policy helpers are being refactored.
> > My request to you is:
> > 
> > 1. Clearly describe the motivation of the PR in the commit and PR descriptions
> > 2. Break this PR into smaller independent units (removal of unused code, traffic policy, RBAC etc.)
> > 3. Open a Github issue for larger refactors so they can be tracked and discussed
> 
> A net 650 lines of code removed is a big benefit in my opinion. All of the refactored code is due to unused code paths, or inlining functions that are called in one place (particularly after portions of that function that are not used are removed).
> 
> The net result is a significant improvement in readability, simplicity, with an added benefit of less bytes used in the envoy config itself :)

I am not disputing the removal of unused code, rather pointing out that there's not enough context to review these changes effectively. For example, the RBAC policy code is being refactored, yet there isn't any indication on why this is necessary or if there are existing bugs. It would be a lot easier to review by breaking this change up into smaller units, each potentially addressing a localized problem.

steeling:
> > > I see too many parts of the code being changed without a clear motivation being described in the commit message. It's hard for me to understand why significant parts of the RBAC policy builder + traffic policy helpers are being refactored.
> > > My request to you is:
> > > 
> > > 1. Clearly describe the motivation of the PR in the commit and PR descriptions
> > > 2. Break this PR into smaller independent units (removal of unused code, traffic policy, RBAC etc.)
> > > 3. Open a Github issue for larger refactors so they can be tracked and discussed
> > 
> > 
> > A net 650 lines of code removed is a big benefit in my opinion. All of the refactored code is due to unused code paths, or inlining functions that are called in one place (particularly after portions of that function that are not used are removed).
> > The net result is a significant improvement in readability, simplicity, with an added benefit of less bytes used in the envoy config itself :)
> 
> I am not disputing the removal of unused code, rather pointing out that there's not enough context to review these changes effectively. For example, the RBAC policy code is being refactored, yet there isn't any indication on why this is necessary or if there are existing bugs. It would be a lot easier to review by breaking this change up into smaller units, each potentially addressing a localized problem.

It's mostly for readability. I have to make a couple small changes here with respect to the trust domain, and ended up spending a lot more time than I anticipated in simply understanding the code. In my experience this understanding will last for about a month, and then I'll need to refresh myself.

The goal here is purely around simplicity. I can add more context, but I'd expect that this PR is simple enough to remain as a single PR vs splitting it up

shashankram:
> Remove And/OR rules from permissions and principals in RBAC, since OR is the only flavor we ever use. We now no longer need to nest multiple lists of principals in the principals array, with OR's, since multiple instances in the top-level list are applied with OR semantics anyways.

I suggest not removing what is already supported. There is a reason that And/Or rules was implemented for permissions and principals, so that we could support more fine grained RBAC on attributes other than just service-account, eg. source/destination IP/port, L7 headers etc.

I am not in favor of refactoring out this functionality for the purpose of integrating the trust domain work. Instead, I'd like to see the trust domain transparently plumbed using the existing code.

keithmattix (commented):
I love the transition to a builder pattern here, but I'm hesitant to remove the code powering AND semantics. Just because SMI doesn't really support it now doesn't mean customers don't want/need that policy crafting tool

steeling:
FWIW i wrote this PR in < a day. We read code much more than we write it, so I think removing unused features, when it would be trivial to add them back, would make sense.

Especially since the and/or rules, as written are limited, in that they don't support nesting. If we do need AND later, there's no promise that the current implementation would even support the need.

I'd love to get more thoughts here, especially around the idea of maintaining unused code, particularly since I'll be adding on top of this, so essentially we're adding code on top of tech debt.

@trstringer  @nojnhuh @jaellio

shashankram:
> FWIW i wrote this PR in < a day, and I'd estimate re-adding that functionality to take a similar amount of time. However we read code much more than we write it, so I think removing unused features, when it would be trivial to add them back, makes sense.
> 
> Especially since the and/or rules, as written are limited, in that they don't support nesting. If we do need AND later, there's no promise that the current implementation would even support the need.
> 
> I'd love to get more thoughts here, especially around the idea of maintaining unused code, particularly since I'll be adding on top of this, so essentially we're adding code on top of tech debt.
> 
> @trstringer @nojnhuh @jaellio

A lot of effort went into ensuring the given RBAC policy generator supports basic AND/OR semantics to accommodate future use cases beyond just the service principal. We didn't need complex nesting semantics, so it wasn't an immediate priority back then.

I am not in favor of refactoring this code and removing core functionality just to plumb the trust domain work, which I don't believe requires significant refactoring of this code path. If this refactor was done without removing the generic AND/OR support for readability, then it would be more appealing to me. But as it stands, I am not in favor of this change.

steeling:
> > FWIW i wrote this PR in < a day, and I'd estimate re-adding that functionality to take a similar amount of time. However we read code much more than we write it, so I think removing unused features, when it would be trivial to add them back, makes sense.
> > Especially since the and/or rules, as written are limited, in that they don't support nesting. If we do need AND later, there's no promise that the current implementation would even support the need.
> > I'd love to get more thoughts here, especially around the idea of maintaining unused code, particularly since I'll be adding on top of this, so essentially we're adding code on top of tech debt.
> > @trstringer @nojnhuh @jaellio
> 
> A lot of effort went into ensuring the given RBAC policy generator supports basic AND/OR semantics to accommodate future use cases beyond just the service principal. We didn't need complex nesting semantics, so it wasn't an immediate priority back then.
> 
> I am not in favor of refactoring this code and removing core functionality just to plumb the trust domain work, which I don't believe requires significant refactoring of this code path. If this refactor was done without removing the generic AND/OR support for readability, then it would be more appealing to me. But as it stands, I am not in favor of this change.

I'd say YAGNI should be a leading principle here. Since this seems to be relatively subjective I'd like to open it up to others to get opinions before reverting if that's ok

steeling:
@nojnhuh @trstringer @jaellio  pinging other maintainers for opinions on above

steeling:
> > > FWIW i wrote this PR in < a day, and I'd estimate re-adding that functionality to take a similar amount of time. However we read code much more than we write it, so I think removing unused features, when it would be trivial to add them back, makes sense.
> > > Especially since the and/or rules, as written are limited, in that they don't support nesting. If we do need AND later, there's no promise that the current implementation would even support the need.
> > > I'd love to get more thoughts here, especially around the idea of maintaining unused code, particularly since I'll be adding on top of this, so essentially we're adding code on top of tech debt.
> > > @trstringer @nojnhuh @jaellio
> > 
> > 
> > A lot of effort went into ensuring the given RBAC policy generator supports basic AND/OR semantics to accommodate future use cases beyond just the service principal. We didn't need complex nesting semantics, so it wasn't an immediate priority back then.
> > I am not in favor of refactoring this code and removing core functionality just to plumb the trust domain work, which I don't believe requires significant refactoring of this code path. If this refactor was done without removing the generic AND/OR support for readability, then it would be more appealing to me. But as it stands, I am not in favor of this change.
> 
> I'd say YAGNI should be a leading principle here. Since this seems to be relatively subjective I'd like to open it up to others to get opinions before reverting if that's ok

Adding to this argument, I have some proposals to remove most of the logic from the trafficpolicy objects like Rules. Instead, the meshcatalog should present more normalized data, and allow builders (like policy.Builder), to specify the rules.

So instead of

```
builder.SetRules(rules)
```

Where rules contains business logic on how the rules were applied (or at least needed a lot of business logic to construct), we would instead do:


```go
builder.SetApplyPermissionsWithAnd(true)
// or 
builder.SetAndRules(rules)
```

This allows for that flexible nesting that is currently not possible, removes this "double" translation logic that is currently happening, simplifies the code base, and is trivial to implement.

With that system I double down that we don't need to maintain unused AND logic when we don't use it. If for nothing else that

1) It would be incredibly trivial to implement with the above style
2) If I implement it now, I have little to no confidence it would actually support our hypothetical future use cases
3) If I implement it now, I'm spending time to implement it... why not spend this time when it is needed
4) YAGNI

Again pinging the rest of the maintainers for there thoughts here

@nojnhuh, @trstringer, @jaellio

steeling:
As a final point of data, why does the mesh catalog know that a port translates into a rule? We should follow the rule of single responsbility, where the mesh catalog is aware about the state of the mesh, not how that translates to an envoy config (which is the responsbility of the builders).

Instead, it should be supplying available ports, not a set of AND/OR rules. That places envoy config logic in multiple places.

Again, happy to implement on the policy builder side but I strongly believe adding this logic within the mesh catalog would take us in the wrong direction

shashankram:
> As a final point of data, why does the mesh catalog know that a port translates into a rule? We should follow the rule of single responsbility, where the mesh catalog is aware about the state of the mesh, not how that translates to an envoy config (which is the responsbility of the builders).
> 
> Instead, it should be supplying available ports, not a set of AND/OR rules. That places envoy config logic in multiple places.
> 
> Again, happy to implement on the policy builder side but I strongly believe adding this logic within the mesh catalog would take us in the wrong direction

My primary concern pertains to removing the capability to remove AND/OR rule semantics from the existing RBAC code. I am not against refactoring this code to use a data model. As we are adding more complex features, I can imagine us needing to use AND/OR semantics for RBAC rules, as the API is complex due to nesting of protos. I also believe we are not going to expose nested APIs to users, so top level AND/OR support should be good enough (e.g. filter on source IP AND destination port, or filter on source IP OR source principal, etc.).

Further, I understand that you are proposing a design change to the MeshCatalog. I am not against this either. I want to point out that the existing design of MeshCatalog is to translate k8s config to objects that can easily be translated to XDS objects, vs needing to call into k8s APIs from `pkg/envoy`, hence the code is structured this way. And will also note that the MeshCatalog and route building code has been heavily refactored multiple times by different engineers, following a new pattern each time. Obviously, this is not ideal but was done in the interest of time. So maybe a concise doc on how we want MeshCatalog to look could be a starting point. Without having an underlying philosphy documented, these changes are hard to reason about.

keithmattix:
In general, I'm in favor of docs or other formats as a method of detailing technical goals maintainers & other community members have for a project's code base. These kinds of docs serve as a guiding light to folks working on the project, and when a group of people have an explicit, common goal, accomplishing that goal tends to be smoother.

steeling:
> > As a final point of data, why does the mesh catalog know that a port translates into a rule? We should follow the rule of single responsbility, where the mesh catalog is aware about the state of the mesh, not how that translates to an envoy config (which is the responsbility of the builders).
> > Instead, it should be supplying available ports, not a set of AND/OR rules. That places envoy config logic in multiple places.
> > Again, happy to implement on the policy builder side but I strongly believe adding this logic within the mesh catalog would take us in the wrong direction
> 
> My primary concern pertains to removing the capability to remove AND/OR rule semantics from the existing RBAC code. I am not against refactoring this code to use a data model. As we are adding more complex features, I can imagine us needing to use AND/OR semantics for RBAC rules, as the API is complex due to nesting of protos. I also believe we are not going to expose nested APIs to users, so top level AND/OR support should be good enough (e.g. filter on source IP AND destination port, or filter on source IP OR source principal, etc.).
> 
> Further, I understand that you are proposing a design change to the MeshCatalog. I am not against this either. I want to point out that the existing design of MeshCatalog is to translate k8s config to objects that can easily be translated to XDS objects, vs needing to call into k8s APIs from `pkg/envoy`, hence the code is structured this way. And will also note that the MeshCatalog and route building code has been heavily refactored multiple times by different engineers, following a new pattern each time. Obviously, this is not ideal but was done in the interest of time. So maybe a concise doc on how we want MeshCatalog to look could be a starting point. Without having an underlying philosphy documented, these changes are hard to reason about.

I think we're on the same page :) I just added a commit that includes the AND semantics to the policy portion, PTAL.

Agreed, I have a doc I'll be sharing with the team that lays out some founding principals to help make decision on how this code should be structured going forward

pkg/catalog/traffictarget_test.go
```diff
@@ -507,7 +507,7 @@ func TestListInboundTrafficTargetsWithRoutes(t *testing.T) {
 					},
 					TCPRouteMatches: []trafficpolicy.TCPRouteMatch{
 						{
-							Ports: []int{8000, 9000},
+							Ports: []uint32{8000, 9000},
```
shashankram, line 510: Can these be uint16 instead as port ranges are from 0-65535

pkg/envoy/lds/rbac.go
```diff
@@ -75,41 +68,20 @@ func (lb *listenerBuilder) buildInboundRBACPolicies() (*xds_network_rbac.RBAC, e
...
-	var permissionRuleList []rbac.RulesList
 	for _, tcpRouteMatch := range trafficTarget.TCPRouteMatches {
 		// Matching ports have an OR relationship
-		var orPortRules []rbac.Rule
 		for _, port := range tcpRouteMatch.Ports {
-			portRule := rbac.Rule{
-				Attribute: rbac.DestinationPort, Value: strconv.Itoa(port),
-			}
-			orPortRules = append(orPortRules, portRule)
+			pb.AddAllowedDestinationPort(port)
```
shashankram, line 82: How do we know from the API whether these rules are using AND or OR? Previously this was clear by the variable name `orPortRules`

pkg/envoy/lds/rbac_test.go
```diff
@@ -45,27 +44,10 @@ func TestBuildRBACPolicyFromTrafficTarget(t *testing.T) {
...
-					},
-					{
-						Identifier: &xds_rbac.Principal_OrIds{
-							OrIds: &xds_rbac.Principal_Set{
-								Ids: []*xds_rbac.Principal{
-									rbac.GetAuthenticatedPrincipal("sa-3.ns-3.cluster.local"),
-								},
-							},
-						},
-					},
```
shashankram, line 65: Are we no longer using OR by default? Could you refactor this in such a way that we continue to use OR based rules as before, and then remove the need for OR based rules in a separate change if necessary? Otherwise, it's a bit hard to know if this change is regressing the existing behavior as it involes test changes too.

pkg/envoy/lds/rbac_test.go
```diff
@@ -80,58 +62,25 @@ func TestBuildRBACPolicyFromTrafficTarget(t *testing.T) {
...
 					{
-						Ports: []int{3000},
+						Ports: []uint32{3000},
 					},
 				},
 			},
 
 			expectedPolicy: &xds_rbac.Policy{
 				Permissions: []*xds_rbac.Permission{
-					{
```
shashankram, line 93: Same comment as above.

pkg/envoy/rbac/policy.go
```diff
@@ -1,136 +1,84 @@
...
+	if principal == "*" {
+		p.AllowAnyPrincipal()
 	}
+	if !p.allowAllPrincipals {
+		// TODO(4754) return the trust domain of the first MRC, ***while grabbing the lock***
```
shashankram, line 68: I don't understand this comment. 
1. Why does this code have anything to do with the trust domain, which should be a part of the opaque principal passed to the builder
2. Expand MRC

pkg/trafficpolicy/types.go
```diff
@@ -42,7 +42,7 @@ type HTTPRouteMatch struct {
 
 // TCPRouteMatch is a struct to represent a TCP route matching based on ports
 type TCPRouteMatch struct {
-	Ports []int `json:"ports:omitempty"`
+	Ports []uint32 `json:"ports:omitempty"`
```
shashankram, line 45: Can this be a uint16 instead, or was this done because Envoy protos use uint32?

shashankram (changes requested):
I am still not comfortable with the changes made to the RBAC code. Here's a suggestion, let me know if that works.
1. Refactor the existing policy code in a way where the test expectations remain the same. Currently, I see a lot of test code being updated, so it's extremely hard to spot whether the rules are applied exactly as before (defaultin to OR). If this is not the case, then this PR title is misleading. It would not be removing unusued code paths, but rather changing the behavior of the existing RBAC policies, which I believe should be a separate change in itself.
2. In a follow-up change, migrate from OR rules to top-level rules. 

This way, there is an easier path to making the refactoring without changing the underlying behavior.

pkg/envoy/rbac/policy.go
steeling: resolving after discussing in person

shashankram:
Discussed this offline with @steeling. Although this change moves from OR based fields to top-level fields for both principals and permissions, by default Envoy uses OR based rules for top-level fields: https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/rbac/v3/rbac.proto#envoy-v3-api-msg-config-rbac-v3-policy

pkg/envoy/lds/rbac.go
steeling: in case this is referring to teh API on the builder, it is the responsbility of the caller. So AddAllowedDestinationPort, would be called, and how they are configured depends on if the caller also calls `UseAndForPermissions`. 

It may not be perfect, but my hope is its clear enough, while being simple enough we can make future changes if confusion arises

steeling: resolving, feel free to re-open if more discussion is necessary!

pkg/envoy/lds/rbac_test.go
steeling: added comments on how OR is used by default

pkg/envoy/rbac/policy.go
```diff
@@ -1,136 +1,84 @@
...
-	policy := &xds_rbac.Policy{}
+// PolicyBuilder is a utility for constructing *xds_rbac.Policy's
+type PolicyBuilder struct {
+	allowedPorts       []uint32
+	allowedPrincipals  []string
+	allowAllPrincipals bool
 
-	// Construct the Principals ------------------------
-	var finalPrincipals []*xds_rbac.Principal
+	applyPermissionsAsAnd bool
```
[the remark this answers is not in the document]
steeling: added a comment to clarify

pkg/envoy/rbac/policy.go
```diff
@@ -1,136 +1,84 @@
...
 	}
-	if len(p.Permissions) == 0 {
-		// No permissions specified for this policy, allow ANY
-		finalPermissions = append(finalPermissions, getAnyPermission())
+
+	return policy
+}
+
+// UseAndForPermissions will apply all permissions with AND semantics.
+func (p *PolicyBuilder) UseAndForPermissions(val bool) {
```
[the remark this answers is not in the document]
steeling: done!

pkg/trafficpolicy/types.go
steeling: it was uint32 for envoy, but ya i can do uint16 and make the translation downstream.

I think that furthers the decoupling in a reasonable way

shashankram: Not a problem. It's fine.

shashankram (approved):

nojnhuh (approved):

Issue #2018, opened 2020-11-10
add rate limiting functionality


michelleN:
**Please describe the Improvement and/or Feature Request**
Design and implement rate limting.

- [x] Local rate limiting (Phase1)
    - [x] [Design doc](https://docs.google.com/document/d/1VTdte9v0JpX7odMtXlGsLlGwDlnXBAEZAf8PMJ_EyY4/edit?usp=sharing)
    - [x] Implementation
    - [x] Documentation
    
- [ ] Global rate limiting (Phase 2)
    - [x] [Design doc](https://docs.google.com/document/d/1fNfDvOV0p93DxP0vLGJmP4FTqd4ZgggI9PnA-sZWnWI/edit?usp=sharing)
    - [x] Implementation
    - [x] Documentation

**Possible use cases**
Users should be able to specify rate limiting policies for traffic.

shashankram:
The core functionality is now complete.

PR #4796, opened 2022-06-07
apis: add local rate limiting to UpstreamTrafficSetting

shashankram:
**Description**:
Adds the local rate limiting API to the UpstreamTrafficSetting
spec.

The API allows enforcing connection and request level
rate limiting for TCP and HTTP traffic respectively. 

Design doc can be found [here](https://docs.google.com/document/d/1VTdte9v0JpX7odMtXlGsLlGwDlnXBAEZAf8PMJ_EyY4/edit#heading=h.ljwfw0m22gcd)

Part of #2018

**Testing done**:

**Affected area**:

3. Has documentation corresponding to this change been updated in the [osm-docs](https://github.com/openservicemesh/osm-docs) repo (if applicable)? API doc will be updated once the API is available for general use.


pkg/apis/policy/v1alpha1/upstreamtrafficsetting.go
```diff
@@ -99,6 +113,119 @@ type HTTPConnectionSettings struct {
...
+// per HTTP route.
+type HTTPPerRouteRateLimitSpec struct {
+	// Local defines the local rate limiting specification
+	// applied per HTTP route.
+	Local *HTTPPerRouteLocalRateLimitSpec `json:"local,omitempty"`
+}
+
+// HTTPPerRouteLocalRateLimitSpec defines the local rate limiting
+// specification per HTTP route.
+type HTTPPerRouteLocalRateLimitSpec struct {
```
steeling, line 223: Can we get rid of this type?

pkg/apis/policy/v1alpha1/upstreamtrafficsetting.go
```diff
@@ -38,6 +38,20 @@ type UpstreamTrafficSettingSpec struct {
...
+	// If HTTP rate limiting is specified, the rate limiting is applied
+	// at the VirtualHost level applicable to all routes within the
+	// VirtualHost.
+	// +optional
+	RateLimit *RateLimitSpec `json:"rateLimit,omitempty"`
+
+	// HTTPRoutes defines the list of HTTP routes settings
+	// for the upstream host. Settings are applied at a per
+	// route level.
+	// +optional
```
keithmattix, line 53: To clarify, this doesn't actually do anything does it? It's purely informational?

pkg/apis/policy/v1alpha1/upstreamtrafficsetting.go
```diff
@@ -99,6 +113,111 @@ type HTTPConnectionSettings struct {
...
+	// Burst defines the number of requests above the baseline
+	// rate that are allowed in a short period of time.
+	// +optional
+	Burst uint32 `json:"burst,omitempty"`
+
+	// ResponseStatusCode defines the HTTP status code to use for responses
+	// to rate limited requests. Code must be in the 400-599 (inclusive)
+	// error range. If not specified, a default of 429 (Too Many Requests) is used.
+	// +optional
+	ResponseStatusCode uint32 `json:"responseStatusCode,omitempty"`
```
keithmattix, line 186: Does it make sense for this to be a pointer so that presence is easily validated?

pkg/apis/policy/v1alpha1/upstreamtrafficsetting.go
```diff
@@ -99,6 +113,111 @@ type HTTPConnectionSettings struct {
...
+
+	// ResponseHeadersToAdd defines the list of HTTP headers that should be
+	// added to each response for requests that have been rate limited.
+	// +optional
+	ResponseHeadersToAdd []HTTPHeaderValue `json:"responseHeadersToAdd,omitempty"`
```
keithmattix, line 191: Is there any value in this not just being a `map[string]string`?

pkg/apis/policy/v1alpha1/upstreamtrafficsetting.go
shashankram: From a code perspective, it doesn't. But it renders the field as `optional` in the docs. E.g. https://release-v1-1.docs.openservicemesh.io/docs/api_reference/policy/v1alpha1/

image

pkg/apis/policy/v1alpha1/upstreamtrafficsetting.go
shashankram: I thought about this, and since Envoy will default to 429 anyway for values < 400, keeping this an uint32 should be ok.

steeling: I think we should opt more for default values in the CRD definition. That way when a user performs a Get on an object, they see the value.

This is what K8s recommends for default values, and ie: what they use to populate objects without a specified namespace

shashankram: In some cases, it helps to let Envoy use its default when fields are unset, and we can simply document these instead of encoding Envoy defaults throughout the code. I understand your point though, where setting defaults in the CRD may be more intuitive.

trstringer (approved):
Looks good to me. Not merging yet because I see unresolved conversations.

pkg/apis/policy/v1alpha1/upstreamtrafficsetting.go
keithmattix: Oh that is VERY good to know; I had no idea. Will be doing this more often 👍🏾

pkg/apis/policy/v1alpha1/upstreamtrafficsetting.go
steeling: that makes sense to!

keithmattix: Yeah I can see both sides. I think if we're going to allow Envoy defaults, we just need to document them and specify that Envoy will handle it.

pkg/apis/policy/v1alpha1/upstreamtrafficsetting.go
keithmattix: Ah, good call. Do we need to worry about this one too?

> Do not use unsigned integers, due to inconsistent support across languages and libraries. Just validate that the integer is non-negative if that's the case.

shashankram: I think it's fine here, as we aren't really in the business of supporting multiple clients for different languages similar to the k8s project. We are only publishing the Golang APIs for our own control plane to consume. I have seen uint being used similarly across other projects.

keithmattix (approved):

jaellio (approved):

Issue #4711, opened 2022-05-02
Define certKeyBitSize and serviceCertValidityDuration certificate settings in the MeshRootCertificate


jaellio:
**Please describe the Improvement and/or Feature Request**

Part of #4502
Dependent on #4710 

I propose removing the spec.certificate fields from the OSM MeshConfig CRD and specifying the settings in the MeshRootCertificate CRD. This change would combine all OSM certificate settings into a single resource. A new version of the MeshConfig CRD, v1alpha3, would be created. 

Changes to the new fields in the MeshRootCertificate resource would not trigger a root certificate rotation or modify the resource's status. OSM would watch for changes to these fields and make updates accordingly. 

High-level changes:
- support conversion between v1alpha2 <-> v1alpha3 versions of the osm MeshConfig
- update MeshRootCertificate CRD and MRC types
- watch new MRC fields 

**Scope (please mark with X where applicable)**
- New Functionality          [x]
- Install [x]
- Certificate Management     [x]

steeling:
we also need to add the Vault Port to the VaultSpec

jaellio:
Should changes to certKeyBitSize and serviceCertValidity duration cause all existing, non-expired certificates to be rotated or should it only be updated for all future certs?

steeling:
> 

Hmm good question!

I was actually surprised to see that we don't use certKeyBitSize for the CA generation for tresor. If we switch that, then that means a certKeyBitSize change for tresor would require a full rotation of the CA.

The other certificate manager's don't face that problem. From a user perspective, if I made those changes, I would expect them to occur sooner than later, but I think it's a subjective answer, and we have some freedom to make a choice here!

jaellio:
This is a nice to have for v1.3 but not a requirement. Removing from v1.3 milestone.

PR #4800, opened 2022-06-08
fix(certs): updates checkAndRotate to use current durations

jaellio:
**Description**:

Updates the `checkAndRotate` method to update certs with the current
validity duration specified in the MeshConfig for their respective
cert type. Adds a `certType` to the certificate struct that is used
to lookup validity duration when calling `IssuingCertificate`.
Updates `shouldRotate` to check if the cert needs to be rotated due
to an in-progress root certificate rotation. `GetFromCache` will
return nil if the cert is not found in the cache, expiring, or
needs to be rotated because of a root cert rotation.

**Testing done**:
- Verified expiring certs are re-issued with the current serviceCertValidityDuration as defined in the OSM MeshConfig. Currently, expiring service certs are only re-issued with the original serviceCertValidityDuration.
  - Verified the bug on the main branch by running the osm demo with the serviceCertValidityDuration set to 5m. After OSM install, I patched the MeshConfig and updated the validity duration to 4m. Expiring certs were reissued using the original 5m validity duration.
  - Verified the bug is no longer present with this change by following the same steps as above
- CI

**Affected area**:


pkg/certificate/manager.go
```diff
@@ -63,29 +66,56 @@ func (m *Manager) GetTrustDomain() string {
...
+			c.GetCommonName(),
+			time.Until(c.GetExpiration()),
+			renewBefore)
+		return true
+	}
+
+	m.mu.RLock()
+	validatingIssuer := m.validatingIssuer
+	signingIssuer := m.signingIssuer
+	m.mu.RUnlock()
```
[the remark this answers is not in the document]
steeling: not quite, since these are addresses, and we are changing the address, not the content of the address, the old address (which is different than a newly changed address) is still valid.

See https://go.dev/play/p/QIW1jMcor2j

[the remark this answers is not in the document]
steeling: i should clarify here, this may indeed grab an old issuer, and begin signing with that, which is ok, since we'll achieve eventual consistency

the lock is strictly to prevent data races with access to the field itself, not to guarantee that the issuer is still current

pkg/certificate/manager.go
```diff
@@ -1,18 +1,20 @@
...
 	"github.com/openservicemesh/osm/pkg/announcements"
+	"github.com/openservicemesh/osm/pkg/constants"
 	"github.com/openservicemesh/osm/pkg/errcode"
 	"github.com/openservicemesh/osm/pkg/k8s/events"
 	"github.com/openservicemesh/osm/pkg/messaging"
 )
 
 // NewManager creates a new CertManager with the passed CA and CA Private Key
-func NewManager(mrcClient MRCClient, serviceCertValidityDuration time.Duration, msgBroker *messaging.Broker) (*Manager, error) {
+func NewManager(mrcClient MRCClient, getServiceCertValidityPeriod func() time.Duration, getIngressCertValidityDuration func() time.Duration, msgBroker *messaging.Broker) (*Manager, error) {
```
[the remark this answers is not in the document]
jaellio: I am still in favor of moving the service cert validity duration to the MRC. Was just blocked on some other changes so decided to leave it out of this PR. Moving this setting from the MeshCondig to the MRC is tracked in #4711.

pkg/certificate/manager.go
```diff
@@ -56,29 +59,56 @@ func (m *Manager) Start(checkInterval time.Duration, stop <-chan struct{}) {
...
+		// Get existing or issue new certificate
+		newCert, err := m.IssueCertificate(cert.GetCommonName(), cert.certType)
+		if err != nil {
+			// TODO(#3962): metric might not be scraped before process restart resulting from this error
+			log.Error().Err(err).Str(errcode.Kind, errcode.GetErrCodeWithMetric(errcode.ErrRotatingCert)).
+				Msgf("Error rotating cert SerialNumber=%s", cert.GetSerialNumber())
+			continue
+		}
+		if newCert != cert {
+			// Certificate was rotated
```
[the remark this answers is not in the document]
jaellio: Moved publishing to the cert pub-sub to `IssueCertificate` to ensure a message is published whenever a certificate is rotated (including outside of `checkAndRotate`).

pkg/certificate/manager.go
```diff
@@ -90,47 +120,62 @@ func (m *Manager) checkAndRotate() {
...
-			return nil, err
-		}
-
-		// if we have different signing and validating issuers,
-		// create the cert's trust context
-		if validatingIssuer.ID != signingIssuer.ID {
-			cert = cert.newMergedWithRoot(validatingIssuer.CertificateAuthority)
-		}
+	validityDuration := m.getValidityDurationForCertType(ct)
+	cert, err = signingIssuer.IssueCertificate(cn, validityDuration)
```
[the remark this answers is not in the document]
jaellio: Updated IssueCertificate so all calls to issue certificates are included in `group.Do`.

pkg/certificate/manager.go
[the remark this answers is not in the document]
jaellio: @nojnhuh what are your thoughts on this?

This is how I understand it. Once we create an issuer, the issuer itself should never be modified. We will only ever create a new issuer and update the pointer to the address of that new issuer. So while we aren't doing a deep copy of the issuer, it is guaranteed that the issuer at that address will never be updated. Therefore we don't need locks for reads. We do need locks when we update the validating and signing issuer pointers. @steeling is that on track?

pkg/configurator/methods.go
```diff
@@ -162,6 +165,22 @@ func (c *Client) GetServiceCertValidityPeriod() time.Duration {
 	return validityDuration
 }
 
+// GetIngressGatewayCertValidityPeriod returns the validity duration for ingress gateway certificates, and a default in case of unspecified or invalid duration
+func (c *Client) GetIngressGatewayCertValidityPeriod() time.Duration {
```
shashankram, line 169: What's the use case to use a default if unspecified? I'd rather make this field required in the CRD spec vs assuming user intent.

jaellio: If GetIngressGatewaCertValidityPeriod() is somehow called when the ingressGateway field isn't specified should it return an error? Currently, If the ingressGateway ValidityDuration field is updated and the duration is invalid the ingressGatewayCert won't be updated. On rotation of the ingressGateway cert we'll attempt to get the duration from the meshConfig, but it will be invalid. Do you think it makes sense to return a default duration in this case?

jaellio: If an ingressGateway is specified in the MeshConfig then the duration is required. We could specify a default instead of making it required? I guess ideally we would validate the MeshConfig and prevent invalid durations before they are applied.

shashankram: If it's a required property, why do we need a default? Is there a default to simplify the code path?

shashankram: If the ingressGateway field is removed/unset, the behavior as implemented in pkg/ingress/gateway.go is to call ReleaseCertificate(). So rotation should not happen for such a scenario?

jaellio: > If it's a required property, why do we need a default? Is there a default to simplify the code path?

I am not sure what you mean. Are you asking if I am using a default value to make things simpler?

shashankram: Yes. Additionally,

> On rotation of the ingressGateway cert we'll attempt to get the duration from the meshConfig, but it will be invalid.

I am wondering if we ever want to use a default during rotation for the ingress cert. I don't see a need to rotate the ingress gateway cert when the state of the gateway cert spec is deemed invalid. Would it be easier to prevent a misconfig? 

This is not a blocking comment. I can understand if there are a few corner cases being guarded against through defensive programming here.

shashankram: Also I would imagine, if the updated value is invalid, we should continue to use the older value encoded in the previous cert and not a default?

jaellio: If the ingress gateway cert's validity duration is set in the meshConfig to an invalid duration the cert will not be reissued but the invalid duration will remain set in the meshConfig. When the ingress gateway cert expires should the previous validity duration be used? Prior to my change, the service cert validity duration would have been used for the ingress gateway cert on expiration. I could add a validity duration field to the `Certificate` struct or I could add validation for the MeshConfig, but I think that is beyond the scope of this PR.

jaellio: Could this be addressed in a follow-up PR? I think this could be addressed when we add validation for the MeshConfig. I am also happy to make the changes to include the validity duration as a field in `Certificate` or make those changes in a follow-up PR as well.

pkg/certificate/options.go
```diff
@@ -20,19 +18,6 @@ func (o *issueOptions) formatCN(prefix, trustDomain string) CommonName {
...
 }
 
-func defaultOptions(validityPeriod time.Duration) *issueOptions {
-	return &issueOptions{
-		validityPeriod: validityPeriod,
-	}
-}
-
-// WithValidityPeriod modifies the expiration on a newly issued certificate.
-func WithValidityPeriod(validityPeriod time.Duration) IssueOption {
```
jaellio, line 30: @steeling I removed this because I didn't want to make the cert type/ validity duration optional and assume a default. With my current design for cert rotation and issuing it is required that I have a cert type or something similar to determine whether to call getServiceCertValidityDuration, getIngressGatewayCertValidityDuration, or use the OSM internal cert duration. Do you have any suggestions for how to incorporate the validity period options with my changes or take a different approach? Thanks!!

pkg/configurator/methods.go
shashankram: Yes, sure. This comment is not blocking in nature.

shashankram (approved):

nojnhuh (approved):

pkg/certificate/manager.go
```diff
@@ -15,10 +17,13 @@ import (
 	"github.com/openservicemesh/osm/pkg/messaging"
 )
 
+var group singleflight.Group
+
 // NewManager creates a new CertificateManager with the passed MRCClient and options
-func NewManager(ctx context.Context, mrcClient MRCClient, serviceCertValidityDuration time.Duration, msgBroker *messaging.Broker, checkInterval time.Duration) (*Manager, error) {
+func NewManager(ctx context.Context, mrcClient MRCClient, getServiceCertValidityPeriod func() time.Duration, getIngressCertValidityDuration func() time.Duration, msgBroker *messaging.Broker, checkInterval time.Duration) (*Manager, error) {
```
steeling, line 23: In the future, it may be a better plan to work more linearly towards the goal, so that we don't need these intermediate hacks to get us there.

That is, we could guarantee the creation of the MRC in a prior PR to putting this in

pkg/certificate/manager.go
```diff
@@ -15,10 +17,13 @@ import (
 	"github.com/openservicemesh/osm/pkg/messaging"
 )
 
+var group singleflight.Group
```
steeling, line 20: i'd consider scoping this to the certmanager, in case there are ever multiple

pkg/certificate/manager.go
```diff
@@ -152,75 +157,116 @@ func (m *Manager) GetTrustDomain() string {
...
 }
 
-// IssueCertificate implements Manager and returns a newly issued certificate from the given client.
-func (m *Manager) IssueCertificate(prefix string, opts ...IssueOption) (*Certificate, error) {
-	var err error
-	cert := m.getFromCache(prefix) // Don't call this while holding the lock
+// IssueCertificate returns a newly issued certificate from the given client
+// or an existing valid certificate from the local cache.
+func (m *Manager) IssueCertificate(prefix string, ct CertType, opts ...IssueOption) (*Certificate, error) {
+	cert, err, _ := group.Do(prefix, func() (interface{}, error) {
```
steeling, line 249: nit: add a comment around singleflight semantics and why we can use it

pkg/certificate/manager.go
```diff
@@ -152,75 +157,116 @@ func (m *Manager) GetTrustDomain() string {
...
+		return nil, err
+	}
+	return cert.(*Certificate), nil
+}
 
-	options := defaultOptions(m.serviceCertValidityDuration)
+func (m *Manager) issueCertificate(prefix string, ct CertType, opts ...IssueOption) (*Certificate, error) {
+	var rotate bool
+	cert := m.getFromCache(prefix) // Don't call this while holding the lock
+	if cert != nil {
```
steeling, line 261: nit: put the nil check in should rotate to reduce nesting

steeling (changes requested):

pkg/certificate/manager.go
jaellio: Resolved offline. Marking as resolved

pkg/certificate/manager.go
jaellio: I think I prefer the nesting the call to `shouldRotate`. It makes sense to only call `shouldRotate` when the cert is not nil. Happy to make the change if you think moving the nil check to `shouldRotate` improves readability.

pkg/certificate/manager.go
jaellio: Good idea. I think I might leave this as-is for this PR unless this is a blocking comment.

steeling: I would say it’s a bug not to, since if 2 Cert manager objects are created the cache would not be populated correctly. 

It’s a 1-2 line fix, so I would push for this in this pr

jaellio: Updated

steeling (approved):

Issue #4551, opened 2022-02-17
[Enhancement] Push all k8s clients into pkg/k8s


steeling:
Right now there are 5 separate clients for issuing requests to Kubernetes, including:

pkg/smi
pkg/configurator
pkg/policy
pkg/k8s
pkg/config

We should push all of these into the pkg/k8s library, and restructure the code to have a single client that interacts with the k8s master. This creates a more intuitive client/server relationship, and keeps all k8s constructs in a single location, which helps decouple OSM from k8s

steeling:
closing in favor of https://github.com/openservicemesh/osm/issues/4863

PR #4801, opened 2022-06-09
k8s/informers: centralize informers to simplify code

keithmattix:
Instead of creating different objects (from 5 different packages) that
each manage their own informer collections, create a single object that
runs a set of informers. A pointer to this object is then shared with
all objects that need it.

Addresses #4551

This PR only modifies the k8s client; once this is merged, the other 4 clients (SMI, config, configurator, and policy) will be modified to use `InformerCollection` as well

**Description**:

**Testing done**:

**Affected area**:

### DIFF
--- a/pkg/k8s/client.go
+++ b/pkg/k8s/client.go
@@ -11,40 +11,33 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/labels"
-	"k8s.io/client-go/informers"
-	"k8s.io/client-go/kubernetes"
-	"k8s.io/client-go/tools/cache"
 	"k8s.io/utils/pointer"
 
+	"github.com/openservicemesh/osm/pkg/announcements"
 	policyv1alpha1 "github.com/openservicemesh/osm/pkg/apis/policy/v1alpha1"
 	"github.com/openservicemesh/osm/pkg/envoy"
 	"github.com/openservicemesh/osm/pkg/errcode"
 	policyv1alpha1Client "github.com/openservicemesh/osm/pkg/gen/client/policy/clientset/versioned"
 	"github.com/openservicemesh/osm/pkg/messaging"
 
-	"github.com/openservicemesh/osm/pkg/announcements"
 	"github.com/openservicemesh/osm/pkg/constants"
 	"github.com/openservicemesh/osm/pkg/identity"
+	osminformers "github.com/openservicemesh/osm/pkg/k8s/informers"
 	"github.com/openservicemesh/osm/pkg/service"
 )
 
 // NewKubernetesController returns a new kubernetes.Controller which means to provide access to locally-cached k8s resources
-func NewKubernetesController(kubeClient kubernetes.Interface, policyClient policyv1alpha1Client.Interface, meshName string,
-	stop <-chan struct{}, msgBroker *messaging.Broker, selectInformers ...InformerKey) (Controller, error) {
-	return newClient(kubeClient, policyClient, meshName, stop, msgBroker, selectInformers...)
+func NewKubernetesController(informerCollection *osminformers.InformerCollection, policyClient policyv1alpha1Client.Interface, msgBroker *messaging.Broker, selectInformers ...InformerKey) Controller {
+	return newClient(informerCollection, policyClient, msgBroker, selectInformers...)
 }
 
-func newClient(kubeClient kubernetes.Interface, policyClient policyv1alpha1Client.Interface, meshName string,
-	stop <-chan struct{}, msgBroker *messaging.Broker, selectInformers ...InformerKey) (*client, error) {
+func newClient(informerCollection *osminformers.InformerCollection, policyClient policyv1alpha1Client.Interface, msgBroker *messaging.Broker, selectInformers ...InformerKey) *client {
 	// Initialize client object
 	c := &client{
-		kubeClient:   kubeClient,
-		policyClient: policyClient,
-		meshName:     meshName,
-		informers:    informerCollection{},
+		informers:    informerCollection,
 		msgBroker:    msgBroker,
+		policyClient: policyClient,
 	}
 
 	// Initialize informers
@@ -65,35 +58,18 @@ func newClient(kubeClient kubernetes.Interface, policyClient policyv1alpha1Clien
 		informerInitHandlerMap[informer]()
 	}
 
-	if err := c.run(stop); err != nil {
-		log.Error().Err(err).Msg("Could not start Kubernetes Namespaces client")
-		return nil, err
-	}
-
-	return c, nil
+	return c
 }
 
 // Initializes Namespace monitoring
 func (c *client) initNamespaceMonitor() {
-	monitorNamespaceLabel := map[string]string{constants.OSMKubeResourceMonitorAnnotation: c.meshName}
-
-	labelSelector := fields.SelectorFromSet(monitorNamespaceLabel).String()
-	option := informers.WithTweakListOptions(func(opt *metav1.ListOptions) {
-		opt.LabelSelector = labelSelector
-	})
-
-	informerFactory := informers.NewSharedInformerFactoryWithOptions(c.kubeClient, DefaultKubeEventResyncInterval, option)
-
-	// Add informer
-	c.informers[Namespaces] = informerFactory.Core().V1().Namespaces().Informer()
-
 	// Add event handler to informer
 	nsEventTypes := EventTypes{
 		Add:    announcements.NamespaceAdded,
 		Update: announcements.NamespaceUpdated,
 		Delete: announcements.NamespaceDeleted,
 	}
-	c.informers[Namespaces].AddEventHandler(GetEventHandlerFuncs(nil, nsEventTypes, c.msgBroker))
+	c.informers.AddEventHandler(osminformers.InformerKeyNamespace, GetEventHandlerFuncs(nil, nsEventTypes, c.msgBroker))
 }
 
 // Function to filter K8s meta Objects by OSM's isMonitoredNamespace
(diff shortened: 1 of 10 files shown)


pkg/k8s/informers/informers.go
```diff
@@ -0,0 +1,167 @@
...
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/cache"
+
+	configClientset "github.com/openservicemesh/osm/pkg/gen/client/config/clientset/versioned"
+	policyClientset "github.com/openservicemesh/osm/pkg/gen/client/policy/clientset/versioned"
+)
+
+type InformerCollectionOpt func(*InformerCollection)
+
+func NewInformerCollection(meshName string, stop <-chan struct{}, kubeClient kubernetes.Interface, smiTrafficSplitClient smiTrafficSplitClient.Interface,
```
steeling, line 17: nit: personally not a fan of variadic args with such a long list of params, as it makes it hard to tell what the variadic args are. Consider separating the constructor, ie: New().WithOptions, or even having the informers themselves passed as options, ie:


`NewInformerCollection("osm", stop, WithK8sClient(k8sClient), WithSMIClient(smiClient))`, etc

Maybe not, just wanted to provoke the line of thinking here

steeling: Then we could have an option, ie: DontRunInformers, or RunAllInformers, as we discussed

steeling, line 17: A final thought, is if we are able to properly nest clients, this could simply take a kubeconfig object, and call the NewConfigOrDie on each set of API's itself, instead of taking a forever growing list of api clients.

pkg/k8s/informers/types.go
```diff
@@ -0,0 +1,97 @@
...
+	errSyncingCaches       = errors.New("failed initial cache sync for informers")
+	errReadFromNilInformer = errors.New("failed to read from a nil informer")
+)
+
+type informer struct {
+	customStore cache.Store
+	informer    cache.SharedIndexInformer
+}
+
+type informerCollection map[InformerKey]*informer
```
steeling, line 64: I'm not sure type aliasing this is providing much more flexibility

steeling (changes requested):
A lot of questions intended as thought experiment, with some concrete suggestions as well

pkg/k8s/informers/informers.go
```diff
@@ -0,0 +1,167 @@
...
+		InformerKeyMeshRootCertificate: ic.initMeshRootCertificateMonitor,
+
+		// Policy
+		InformerKeyEgress:                 ic.initEgressMonitor,
+		InformerKeyIngressBackend:         ic.initIngressBackendMonitor,
+		InformerKeyUpstreamTrafficSetting: ic.initUpstreamTrafficSettingMonitor,
+		InformerKeyRetry:                  ic.initRetryMonitor,
+	}
+
+	if len(ic.selectedInformers) == 0 {
```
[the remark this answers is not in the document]
keithmattix: I feel like the informer collection can probably assume that the informers should be used rather than raw client calls.

pkg/k8s/informers/informers.go
keithmattix: Hmm that would be easier for sure. It does get weirder if someone wants to subselect informers within a particular client though

pkg/k8s/informers/informers.go
```diff
@@ -0,0 +1,167 @@
...
+// WithCustomStore sets the shared store for the informerCollection to reference.
+// This store will be used instead of each informer's store. This should typically
+// only be used in tests
+func WithCustomStores(stores map[InformerKey]cache.Store) InformerCollectionOpt {
+	return func(ic *InformerCollection) {
+		ic.customStores = stores
+	}
+}
+
+func WithSelectedInformers(selectedInformers ...InformerKey) InformerCollectionOpt {
```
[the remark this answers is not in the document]
keithmattix: Hmm that would be easier for sure. It does get weirder if someone wants to subselect informers within a particular client though

pkg/k8s/informers/informers.go
keithmattix: Good point; I feel like that's probably a later iteration of this work; this set of PRs won't collapse all the various clients; just provide a shared informer collection for them to use

pkg/k8s/informers/informers.go
```diff
@@ -0,0 +1,191 @@
...
+
+// WithSelectedInformers sets the selected informers for the InformerCollection
+func WithSelectedInformers(selectedInformers ...InformerKey) InformerCollectionOption {
+	return func(ic *InformerCollection) {
+		ic.selectedInformers = selectedInformers
+	}
+}
+
+// WithKubeClient sets the kubeClient for the InformerCollection
+func WithKubeClient(kubeClient kubernetes.Interface) InformerCollectionOption {
```
steeling, line 107: Is there ever a place where we pass in a client but don't run the informer? If not, I'd say this could be the signal.

If there is, maybe we could add a bool here, runInformer? Simplifying the API otherwise you could specify an informer to run where the client was never passed in, which would be a bug.

pkg/k8s/informers/informers.go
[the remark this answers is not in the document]
steeling: One of the issues is that (I think) calling Add on the store, only adds the object to the local store, and not to the k8s master. IDK a scenario in which we would actually want to expose that. Removing footguns would be a benefit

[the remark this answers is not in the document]
keithmattix: So with this proposed implementation, the store is never actually exposed unless you opt-in by passing in a custom store. The previous implementations kept the store exposed for testing purposes (which I definitely think was the right tradeoff given the complexity of timing some of these tests with fake informers), but I agree that removing that possible footgun is preferable which is why I added that custom store option

pkg/k8s/informers/informers.go
```diff
@@ -0,0 +1,167 @@
...
+		return nil, err
+	}
+
+	return ic, nil
+}
+
+// WithCustomStore sets the shared store for the informerCollection to reference.
+// This store will be used instead of each informer's store. This should typically
+// only be used in tests
+func WithCustomStores(stores map[InformerKey]cache.Store) InformerCollectionOpt {
```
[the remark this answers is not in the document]
keithmattix: ```// WithCustomStores provides the InformerCollection an set of `cache.Store`s indexed
// by InformerKey. This functionality was added for the express purpose of testing
// flexibility since the alternative often leads to flaky tests and race conditions
// between the time an object is added to a fake clientset and when that object
// is actually added to the informer `cache.Store`.```

pkg/k8s/client.go
```diff
@@ -65,35 +58,18 @@ func newClient(kubeClient kubernetes.Interface, policyClient policyv1alpha1Clien
...
-	c.informers[Namespaces] = informerFactory.Core().V1().Namespaces().Informer()
-
 	// Add event handler to informer
 	nsEventTypes := EventTypes{
 		Add:    announcements.NamespaceAdded,
 		Update: announcements.NamespaceUpdated,
 		Delete: announcements.NamespaceDeleted,
 	}
-	c.informers[Namespaces].AddEventHandler(GetEventHandlerFuncs(nil, nsEventTypes, c.msgBroker))
+	c.informers.AddEventHandler(osminformers.InformerKeyNamespace, GetEventHandlerFuncs(nil, nsEventTypes, c.msgBroker))
```
shashankram, line 72: nit: we can drop the prefix `InformerKey` from the variable name. Same applies to other informer keys.
```suggestion
	c.informers.AddEventHandler(osminformers.NamespaceKey, GetEventHandlerFuncs(nil, nsEventTypes, c.msgBroker))
```

shashankram (approved):

pkg/k8s/informers/init.go
```diff
@@ -0,0 +1,230 @@
...
+	// Add informer
+	informer := &informer{
+		informer: informerFactory.Core().V1().Namespaces().Informer(),
+	}
+
+	customStore := ic.customStores[InformerKeyNamespace]
+	if customStore != nil {
+		informer.customStore = customStore
+	}
+
```
jaellio, line 44: nit: delete for consistency
```suggestion
```

pkg/k8s/client_test.go
```diff
@@ -205,12 +286,23 @@ func TestGetService(t *testing.T) {
...
+			serviceStore := map[string]interface{}{}
+			store := newFakeCacheStore(serviceStore, serviceKeyFunc)
+			stores := map[informers.InformerKey]cache.Store{
+				informers.InformerKeyService: store,
+			}
+			ic, err := informers.NewInformerCollection(testMeshName, nil, informers.WithKubeClient(testclient.NewSimpleClientset()), informers.WithCustomStores(stores), informers.WithSelectedInformers(k8sInformerKeys...))
 			a.Nil(err)
-			_ = c.informers[Services].GetStore().Add(tc.service)
+			c := newClient(ic, nil, nil)
+			_ = store.Add(tc.service)
```
steeling, line 305: So the real store doesn't return the value immediately when doing an add followed by a get?

I thought the only time it wasn't immediate is when doing an add through the actual client, which triggers an async add.

I believe this direct add is synchronous

pkg/k8s/informers/informers.go
steeling: a further suggestion: put the init methods in here and get rid of the initInformerMap

pkg/k8s/informers/informers.go
```diff
@@ -0,0 +1,240 @@
...
+// InformerCollectionOption is a function that modifies an informer collection
+type InformerCollectionOption func(*InformerCollection)
+
+// NewInformerCollection creates a new InformerCollection
+func NewInformerCollection(meshName string, stop <-chan struct{}, opts ...InformerCollectionOption) (*InformerCollection, error) {
+	ic := &InformerCollection{
+		meshName:  meshName,
+		informers: map[InformerKey]*informer{},
+	}
+
```
jaellio, line 53: Maybe add a comment here to clarify what is happening in this loop. Something like
```suggestion

// Select specific informers
```

pkg/k8s/informers/informers.go
```diff
@@ -0,0 +1,240 @@
...
+		// Policy
+		InformerKeyEgress:                 ic.initEgressMonitor,
+		InformerKeyIngressBackend:         ic.initIngressBackendMonitor,
+		InformerKeyUpstreamTrafficSetting: ic.initUpstreamTrafficSettingMonitor,
+		InformerKeyRetry:                  ic.initRetryMonitor,
+	}
+
+	if len(ic.selectedInformers) == 0 {
+		// Initialize all informers
```
jaellio, line 87:
```suggestion
		// Select all informers
```

pkg/k8s/informers/informers.go
```diff
@@ -0,0 +1,240 @@
...
+			InformerKeyTCPRoute:               {},
+			InformerKeyMeshConfig:             {},
+			InformerKeyMeshRootCertificate:    {},
+			InformerKeyEgress:                 {},
+			InformerKeyIngressBackend:         {},
+			InformerKeyUpstreamTrafficSetting: {},
+			InformerKeyRetry:                  {},
+		}
+	}
+
```
jaellio, line 106:
```suggestion

// Initialize selected informers
```

pkg/k8s/informers/informers.go
```diff
@@ -0,0 +1,240 @@
...
+		informers: map[InformerKey]*informer{},
+	}
+
+	for _, opt := range opts {
+		if opt != nil {
+			opt(ic)
+		}
+	}
+
+	// Initialize informers
```
jaellio, line 60:
```suggestion
```

pkg/k8s/client.go
keithmattix: This prefix is actually a naming convention used in [client-go](https://pkg.go.dev/k8s.io/api/core/v1#NodeSelectorOperator) and other packages when creating variables for a type alias. It helps remove ambiguity when searching for ambiguous or overloaded vars (like namespace or pod)

pkg/k8s/informers/informers.go
keithmattix: I think the map makes initialization cleaner; the alternatives are to re-declare a subset of the map and loop through the keys again or to have each method called sequentially. Happy to discuss further, but I feel like it's probably a preference

pkg/k8s/informers/informers.go
keithmattix: Not sure I understand this suggestion

pkg/k8s/informers/informers.go
```diff
@@ -0,0 +1,198 @@
...
+	if ic.informers == nil {
+		return errInitInformers
+	}
+
+	for name, informer := range ic.informers {
+		if informer == nil {
+			continue
+		}
+
+		go informer.Run(make(chan struct{}))
```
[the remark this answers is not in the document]
keithmattix: Ooh good catch! I set this when I was debugging some nil panics and missed it in cleanup.

nojnhuh (approved):

pkg/k8s/client_test.go
keithmattix: Because we can't access the real store from this package

jaellio (approved):

pkg/k8s/client_test.go
steeling: I see. adding all of this extra stuff to get around this testing seems like a bit of a code smell IMO. I would recommend 1 of 3 things:

1) Add all the objects to the k8s client, ie `testclient.NewSimpleClientset(tc.namespace, ...)`, then use the internals to ensure that the informer `HasSynced`.
2) Expose an add method on the store itself, so you don't need this fake store.
3) Don't separate these into 2 different packages if you need to rely so heavily on internal implementation details

PR #4802, opened 2022-06-09
Set (empty) trust domain on listener builder

steeling:
This PR has 2 main changes:

1. I set the policy builder's AddPrincipal method to AddIdentity. Following my own advice, the builders should accept inputs in terms of OSM concepts, and return envoy configs. A principal is an envoy concept, and so should be constructed internally, and not leak into its parameters.
2. I set the trust domain (currently an empty string) on the listenerbuilder so that it is used on the policy builder.

The final PR will be to simultaneously remove the trustDomain from service.Identity, while populating the trust domain within the cert manager.

### DIFF
--- a/pkg/certificate/manager.go
+++ b/pkg/certificate/manager.go
@@ -56,6 +56,13 @@ func (m *Manager) Start(checkInterval time.Duration, stop <-chan struct{}) {
 	}()
 }
 
+// GetTrustDomain returns the trust domain from the configured signingkey issuer.
+// Note that the CRD uses a default, so this value will always be set.
+func (m *Manager) GetTrustDomain() string {
+	// TODO(4754): implement
+	return ""
+}
+
 func (m *Manager) checkAndRotate() {
 	// NOTE: checkAndRotate can reintroduce a certificate that has been released, thereby creating an unbounded cache.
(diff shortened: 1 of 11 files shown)


keithmattix:
Is this different from #4782? The names are very similar

nojnhuh (approved):

pkg/certificate/manager.go
```diff
@@ -56,6 +56,13 @@ func (m *Manager) Start(checkInterval time.Duration, stop <-chan struct{}) {
 	}()
 }
 
+// GetTrustDomain returns the trust domain from the configured signingkey issuer.
+// Note that the CRD uses a default, so this value will always be set.
+func (m *Manager) GetTrustDomain() string {
+	// TODO(4754): implement
+	return ""
```
jaellio, line 63: For my own understanding, will this be something like

```
return m.signingIssuer.TrustDomain
```

jaellio (approved):

shashankram (approved):

pkg/certificate/manager.go
steeling: yup! (with the lock grabbed)

jaellio: Cool thanks!

PR #4804, opened 2022-06-09
(k8s/informers): Use InformerCollection for other clients

keithmattix:
Use the new InformerCollection instead of fragmenting informers across
different clients

Part of #4551 

**Description**:

**Testing done**:

**Affected area**:

2. Is this a breaking change?


pkg/k8s/informers/informers.go
```diff
@@ -117,64 +78,60 @@ func NewInformerCollection(meshName string, stop <-chan struct{}, opts ...Inform
...
 
-		// select the k8s informers
-		for _, key := range k8sInformerKeys {
-			ic.selectedInformers[key] = struct{}{}
-		}
+		labelSelector := fields.SelectorFromSet(monitorNamespaceLabel).String()
+		option := informers.WithTweakListOptions(func(opt *metav1.ListOptions) {
+			opt.LabelSelector = labelSelector
+		})
+
```
steeling, line 91: super nit: tiny bit more readable if:

```go
v1api := informerFactory.Core().V1()

ic.informers[InformerKeyNamespace] = v1api.Namespace().Informer()
...
```

```

File: pkg/k8s/informers/informers.go [hunk 2]
```diff
@@ -207,6 +164,36 @@ func (ic *InformerCollection) run(stop <-chan struct{}) error {
 	return nil
 }
 
+// Add is only exported for the sake of tests and requires a testing.T to ensure it's
+// never used in production
+func (ic *InformerCollection) Add(key InformerKey, obj interface{}, t *testing.T) error {
```
steeling, line 169: Can you add the comment about the delay between adding to the k8s client, and checking its existence here?

steeling (approved):
Awesome stuff, a couple nits, but LGTM.

Love to see the large amount of code deleted :)

pkg/configurator/client.go
```diff
@@ -21,44 +17,16 @@ import (
 )
 
 // NewConfigurator implements configurator.Configurator and creates the Kubernetes client to manage namespaces.
-func NewConfigurator(configClient configClientset.Interface, stop <-chan struct{}, osmNamespace, meshConfigName string,
+func NewConfigurator(informerCollection *informers.InformerCollection, osmNamespace, meshConfigName string,
```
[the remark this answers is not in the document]
keithmattix: I actually think we should keep them separate since the unexported function returns a concrete type instead of an interface, allowing tests to call unexported methods on said concrete type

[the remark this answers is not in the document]
steeling: from Rob Pike "accept interfaces return structs", let's just return the struct, there's no need to provide additional restrictions on top of us choosing which methods to export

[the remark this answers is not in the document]
keithmattix: Fair enough!

pkg/configurator/client.go
```diff
@@ -21,44 +17,9 @@ import (
...
-		meshConfig:          meshConfigInformerFactory.Config().V1alpha2().MeshConfigs().Informer(),
-		meshRootCertificate: configInformerFactory.Config().V1alpha2().MeshRootCertificates().Informer(),
-	}
-
-	cacheCollection := cacheCollection{
-		meshConfig:          informerCollection.meshConfig.GetStore(),
-		meshRootCertificate: informerCollection.meshRootCertificate.GetStore(),
-	}
-
+func NewConfigurator(informerCollection *informers.InformerCollection, osmNamespace, meshConfigName string, msgBroker *messaging.Broker) (*client, error) {
```
keithmattix, line 20: @steeling @shashankram  What are your opinions on this pattern? I've seen it used before, but the linter doesn't like it. I don't have strong opinions either way; just curious how y'all feel

steeling: I think we should likely export it, all the fields are unexported so it still has to be intiialized through this mechanism

cmd/osm-controller/osm-controller.go
```diff
@@ -249,7 +244,7 @@ func main() {
 		events.GenericEventRecorder().FatalEvent(err, events.InitializationError, "Error creating Ingress client")
 	}
 
-	policyController, err := policy.NewPolicyController(k8sClient, policyClient, stop, msgBroker)
+	policyController := policy.NewPolicyController(informerCollection, k8sClient, msgBroker)
 	if err != nil {
```
nojnhuh, line 248: Looks like this can be removed too.

nojnhuh (approved):

cmd/osm-controller/osm-controller.go
keithmattix: Good catch!

jaellio (approved):

steeling (approved):

PR #4805, opened 2022-06-09
start with a clean slate for future multicluster work

steeling:
This removes *all* multicluster references so that we can start from scratch.

It's a lot of deletions, and I can peel this out into multiple PR's if that makes it more readable, just let me know!


charts/osm/README.md
```diff
@@ -88,7 +88,6 @@ The following table lists the configurable parameters of the osm chart and their
 | osm.featureFlags.enableEgressPolicy | bool | `true` | Enable OSM's Egress policy API. When enabled, fine grained control over Egress (external) traffic is enforced |
 | osm.featureFlags.enableEnvoyActiveHealthChecks | bool | `false` | Enable Envoy active health checks |
 | osm.featureFlags.enableIngressBackendPolicy | bool | `true` | Enables OSM's IngressBackend policy API. When enabled, OSM will use the IngressBackend API allow ingress traffic to mesh backends |
-| osm.featureFlags.enableMulticlusterMode | bool | `false` | Enable Multicluster mode. When enabled, multicluster mode will be enabled in OSM |
```
keithmattix, line 91: One thought: does removing these options violate our semver support guarantees? I believe all the multi-cluster stuff was alpha, but I do think this does break installation for users

steeling: I'll let others comment on
@trstringer @shashankram

shashankram: @keithmattix how does this break installation?
We can always leave this feature flag behind as a no-op to be extra cautious.

jaellio: I think we could just add this to the release notes as well

keithmattix: Sorry I was thinking more generally about removing options/feature flags from meshconfig

steeling: ya he's right, removing from a CRD warrants a version change. Leaving the field in for now, and marking as deprecated

nojnhuh: Even if the field is kept in the CRD, removing the value here will break users of the chart if they either have existing scripts setting it or they upgrade with `helm upgrade --reuse-values`, either of which they would likely expect to succeed if the chart's major version is the same.

I think it's likely though that the number of people we'd break is literally zero with this change, but it's something to consider whenever making a change like this.

demo/run-osm-demo.sh
```diff
@@ -89,7 +89,9 @@ if [ "$DEPLOY_ON_OPENSHIFT" = true ] ; then
 fi
 
 if [ "$PUBLISH_IMAGES" = true ]; then
+    docker network disconnect kind kind-registry
```
nojnhuh, line 92: I haven't had to do this since #4554 was merged.
more context: https://github.com/openservicemesh/osm/pull/4554#pullrequestreview-891801791

nojnhuh (commented):
It looks like we can remove `KindGateway` in addition since that's no longer referred to. And at that point it might be worth removing `ProxyKind` altogether.

https://github.com/openservicemesh/osm/blob/4c816542eef105873e60702c58c01b6574bbb799/pkg/envoy/types.go#L99-L108

charts/osm/README.md
steeling: Agreed! Maybe we can say that support between versions is non-breaking when using the CLI directly, but if they play around with the underlying helm chart we don't offer guaranteed compatibility. That way we can add additional logic to the CLI.

I also think it could make sense to provide additional steps for upgrades. That is, an upgrade is backwards compatible, but you may have to follow some small set of steps. We should probably also say you can only upgrade from 1 version to +2 minor versions at a time, or something along those lines.

nojnhuh (approved):

steeling:
@nojnhuh tagging you since it won't let me quote reply directly..
> Even if the field is kept in the CRD, removing the value here will break users of the chart if they either have existing scripts setting it or they upgrade with helm upgrade --reuse-values, either of which they would likely expect to succeed if the chart's major version is the same.
> 
> I think it's likely though that the number of people we'd break is literally zero with this change, but it's something to consider whenever making a change like this.

Agreed. Maybe we should consider if direct helm upgrades (as opposed to upgrades via the CLI) are supported by our upgrade policy. I would say no, and we should provide a single upgrade path.

Even then, upgrades might not be a magic one command, they may require a few steps.

allenlsy (approved):

shashankram (approved):

PR #4811, opened 2022-06-13
feat(certs): add validating webhook for MeshRootCertificate

schristoff:
**Description**:

Adds an additional webhook to the validating webhook configuration for resources in the OSM control plane namespace. The new webhook will be used to validate any MeshRootCertificates created in the control plane namespace. The validator for the MRC is responsible for determining if there is an active root certificate rotation, if a status change is allowed, and preventing updates to the provider and trust domain specs. MRC validation is enabled when the binary flag `enableMeshRootCertificate` is set to true.

The validator performs the following check on each admission request type:

On add:
- err with empty trust domain
- accept if `state` status not set
- if the `state` status is set, err if rotation is already occurring. This is when 2 or more MRCs are present in the control plane namespace with `states` other than `error` or `inactive`
- accept if the `state` is an accepted `state` (one of `error`, `inactive`, `active`, `validatingRollout`, `validatingRollbck`, `issuingRollout`, `issuingRollback`)
  - Note: at this time we do not verify if the `state` is valid in combination with an existing MRCs state
 
On update:
- err if the provider was updated
- err if the trust domain was updated
- accept if the `state` hasn't changed
- err if the state transition is invalid
- err if there is a rotation already occurring and the MRC being updated is not the MRC being rolled out or rolled back

At this time we are not performing any validation on delete.

**Testing done**:
- CI
- Demo
  - Manually add MRCs and test the above validation scenarios
 

**Affected area**:


pkg/validator/validators.go
```diff
@@ -283,3 +283,91 @@ func MultiClusterServiceValidator(req *admissionv1.AdmissionRequest) (*admission
...
+	applied *configv1alpha2.MeshRootCertificate) (*admissionv1.AdmissionResponse, error) {
+	//prevent out of order status changes
+	// we can do:
+	// validating -> issuing
+	// issuing -> validating
+	// complete -> none?
+	switch current.Status.RotationStage {
+	//where are the const for these?
+	case constants.MRCStageValidating:
+		if applied.Status.RotationStage == constants.MRCStateComplete {
```
jaellio, line 357: Here we are comparing rotation stage and state. We have to compare the rotation stage with other rotation stages like issuing and validating.

pkg/validator/validators.go
```diff
@@ -283,3 +283,91 @@ func MultiClusterServiceValidator(req *admissionv1.AdmissionRequest) (*admission
...
+
+	return nil, nil
+}
+
+func validateMRCdelete(current *configv1alpha2.MeshRootCertificate,
+	applied *configv1alpha2.MeshRootCertificate) (*admissionv1.AdmissionResponse, error) {
+	//todo(schristoff): will we ever be stuck in these states? how/where do we get out?
+	// if osm gets stuck, will it touch this and get blocked?
+	switch current.Status.RotationStage {
+	case constants.MRCStageIssuing:
```
steeling, line 338: commas work here:

```
case constants.MRCStageIssuing, constants.MRCStageValidating:
....

```

pkg/validator/validators.go
```diff
@@ -283,3 +283,116 @@ func MultiClusterServiceValidator(req *admissionv1.AdmissionRequest) (*admission
...
+}
+
+func (m MRCInfo) validateMRCcreate() bool {
+	return m.countActiveMRCs() < 2
+}
+
+func (m MRCInfo) validateMRCTransition() bool {
+	allowedTransitions := map[string][]string{
+		constants.MRCStateIssuingRollout:     {constants.MRCStateValidatingRollout, constants.MRCStateError},
+		constants.MRCStateValidatingRollout:  {constants.MRCStateActive, constants.MRCStateError},
```
jaellio, line 367:
```suggestion
		constants.MRCStateValidatingRollout:  {constants.MRCStateIssuingRollout, constants.MRCStateError},
```

pkg/validator/validators.go
```diff
@@ -283,3 +283,116 @@ func MultiClusterServiceValidator(req *admissionv1.AdmissionRequest) (*admission
...
+		constants.MRCStateActive:             {constants.MRCStateValidatingRollback, constants.MRCStateError},
+		constants.MRCStateValidatingRollback: {constants.MRCStateIssuingRollback, constants.MRCStateError},
```
jaellio, line 369:
```suggestion
		constants.MRCStateIssuingRollback: {constants.MRCStateValidatingRollback, constants.MRCStateError},
```

pkg/validator/validators.go
```diff
@@ -283,3 +283,116 @@ func MultiClusterServiceValidator(req *admissionv1.AdmissionRequest) (*admission
...
+		constants.MRCStateIssuingRollback:    {constants.MRCStateInactive, constants.MRCStateError},
```
jaellio, line 370:
```suggestion
		constants.MRCStateValidatingRollback:    {constants.MRCStateInactive, constants.MRCStateError},
```

pkg/validator/validators.go
```diff
@@ -283,3 +283,116 @@ func MultiClusterServiceValidator(req *admissionv1.AdmissionRequest) (*admission
...
+
+func (m MRCInfo) validateMRCcreate() bool {
+	return m.countActiveMRCs() < 2
+}
+
+func (m MRCInfo) validateMRCTransition() bool {
+	allowedTransitions := map[string][]string{
+		constants.MRCStateIssuingRollout:     {constants.MRCStateValidatingRollout, constants.MRCStateError},
+		constants.MRCStateValidatingRollout:  {constants.MRCStateActive, constants.MRCStateError},
+		constants.MRCStateActive:             {constants.MRCStateValidatingRollback, constants.MRCStateError},
```
jaellio, line 368:
```suggestion
		constants.MRCStateActive:             {constants.MRCStateIssuingRollback, constants.MRCStateError},
```

pkg/validator/mrc_validators.go
```diff
@@ -0,0 +1,95 @@
...
+	return false
+}
+
+func (m MRCInfo) validateMRCdelete() error {
+	//Delete inactive or error only
+	switch m.storedMRC.Status.State {
+	case constants.MRCStateInactive, constants.MRCStateError:
+		return nil
+	default:
+		return errors.Errorf("cannot delete certificate %v in stage %v", m.storedMRC.Name, m.storedMRC.Status.State)
```
jaellio, line 59:
```suggestion
		return errors.Errorf("cannot delete certificate %v in state %v", m.storedMRC.Name, m.storedMRC.Status.State)
```

pkg/validator/mrc_validators.go
```diff
@@ -0,0 +1,95 @@
...
+func (m MRCInfo) validateMRCTransition() bool {
+	allowedTransitions := map[string][]string{
+		constants.MRCStateValidatingRollout:  {constants.MRCStateIssuingRollout, constants.MRCStateError},
+		constants.MRCStateIssuingRollout:     {constants.MRCStateActive, constants.MRCStateError},
+		constants.MRCStateActive:             {constants.MRCStateValidatingRollback, constants.MRCStateError},
+		constants.MRCStateValidatingRollback: {constants.MRCStateIssuingRollback, constants.MRCStateError},
+		constants.MRCStateIssuingRollback:    {constants.MRCStateInactive, constants.MRCStateError},
+	}
+	//look up storedMRC state key
+	//applied state must be in the values for that key
```
jaellio, line 80: nit
```suggestion
	// look up storedMRC state key
	// applied state must be in the values for that key
```

pkg/validator/mrc_validators.go
```diff
@@ -0,0 +1,95 @@
...
+	if allowedStates, ok := allowedTransitions[m.storedMRC.Status.State]; ok {
+		//if going into active, safety check we have less than two
```
jaellio, line 82: nit
```suggestion
		// if going into active, safety check we have less than two
```

pkg/validator/validators_test.go
```diff
@@ -10,13 +10,17 @@ import (
...
+	"github.com/openservicemesh/osm/pkg/constants"
+	fakeConfigClientset "github.com/openservicemesh/osm/pkg/gen/client/config/clientset/versioned/fake"
 	fakePolicyClientset "github.com/openservicemesh/osm/pkg/gen/client/policy/clientset/versioned/fake"
+
 	"github.com/openservicemesh/osm/pkg/k8s"
 	"github.com/openservicemesh/osm/pkg/k8s/informers"
-
 	"github.com/openservicemesh/osm/pkg/messaging"
 	"github.com/openservicemesh/osm/pkg/policy"
+	corev1 "k8s.io/api/core/v1"
```
jaellio, line 23: nit: group import with other k8s.io imports

pkg/validator/validators_test.go
```diff
@@ -1383,3 +1387,135 @@ func TestUpstreamTrafficSettingValidator(t *testing.T) {
...
+	testCases := []struct {
+		name         string
+		input        *admissionv1.AdmissionRequest
+		expResp      *admissionv1.AdmissionResponse
+		isErr        bool
+		expErrStr    string
+		existingMRCs int
+	}{
+		{
+			name: "MeshRootCertificate",
```
jaellio, line 1401: Let's add some test cases where an error is expected

pkg/validator/validators.go
```diff
@@ -244,25 +254,149 @@ func (kc *policyValidator) upstreamTrafficSettingValidator(req *admissionv1.Admi
...
+	}
+
+	return nil, nil
+}
+
+// activeRootCertificateRotation returns true if there is an active OSM root certificate rotation, false otherwise
+func (cv *configValidator) activeRootCertificateRotation(newMRC *configv1alpha2.MeshRootCertificate) bool {
+	mrcList, err := cv.configClient.ConfigV1alpha2().MeshRootCertificates(cv.osmNamespace).List(context.TODO(), v1.ListOptions{})
+	if err != nil {
+		log.Error().Err(err)
```
steeling, line 359: The log package requires you call `Msg` or `Msgf` to flush the log. You can also just use `log.Err(err).Msg(...)` without the `.Error().Err(err)`

pkg/validator/validators.go
```diff
@@ -244,25 +254,149 @@ func (kc *policyValidator) upstreamTrafficSettingValidator(req *admissionv1.Admi
...
+	constants.MRCStateIssuingRollback:    {constants.MRCStateInactive, constants.MRCStateError},
+	constants.MRCStateInactive:           {},
+	constants.MRCStateError:              {},
+	"":                                   {constants.MRCStateValidatingRollout, constants.MRCStateError},
+}
+
+// meshRootCertificateValidator validates the MeshRootCertificate custom resource
+func (cv *configValidator) meshRootCertificateValidator(req *admissionv1.AdmissionRequest) (*admissionv1.AdmissionResponse, error) {
+	log.Debug().Msg("made it to MRC validator")
+	switch req.Operation {
```
steeling, line 294: we can validate there are no deletions if it is in some form of an active state.

pkg/validator/validators.go
```diff
@@ -244,25 +254,149 @@ func (kc *policyValidator) upstreamTrafficSettingValidator(req *admissionv1.Admi
...
+				newMRC.GetNamespace(), newMRC.GetName(), newMRC.Status.State)
+		}
+
+		if _, ok := validMRCTransitions[newMRC.Status.State]; !ok {
+			return nil, fmt.Errorf("invalid status %s for MRC %s/%s. Must be %s, %s, %s, %s, %s, %s, or %s",
+				newMRC.Status.State, newMRC.GetNamespace(), newMRC.GetName(), constants.MRCStateValidatingRollout,
+				constants.MRCStateIssuingRollout, constants.MRCStateActive, constants.MRCStateIssuingRollback,
+				constants.MRCStateValidatingRollback, constants.MRCStateInactive, constants.MRCStateError)
+		}
+	case admissionv1.Update:
```
steeling, line 321: A thought, although it may be overkill:

we can validate that the combination of the current and new MRC are in the right state

pkg/validator/validators.go
jaellio: Would this prevent deletion on uninstall?

pkg/validator/validators.go
```diff
@@ -244,25 +254,149 @@ func (kc *policyValidator) upstreamTrafficSettingValidator(req *admissionv1.Admi
...
+		newMRC := &configv1alpha2.MeshRootCertificate{}
+		if err := json.NewDecoder(bytes.NewBuffer(req.Object.Raw)).Decode(newMRC); err != nil {
+			return nil, err
+		}
+
+		if newMRC.Spec.TrustDomain == "" {
+			return nil, fmt.Errorf("trustDomain must be non empty for MRC %s/%s", newMRC.GetNamespace(), newMRC.GetName())
+		}
+
+		// if no state is set, MRC will not be considered for rotation
```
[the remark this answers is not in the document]
jaellio: I still go back and forth on what approach makes the most sense. With the proposed manual rotation CLI tool outlined here https://docs.google.com/document/d/1MQZ4QPeKV2pxZ30uaYIxMjVYQUGdgQmK0eLYy1JzpHs/edit# I think it makes sense to have an explicit command for initiating rotation. We could have a flag in the MeshConfig that enables automatic/automated root certificate rotation by updating the mutating webhook configuration to set the status of new mrcs

pkg/validator/validators.go
```diff
@@ -244,25 +254,149 @@ func (kc *policyValidator) upstreamTrafficSettingValidator(req *admissionv1.Admi
...
+		if newMRC.Status.State == "" {
+			return nil, nil
+		}
+
+		if cv.activeRootCertificateRotation(newMRC) {
+			return nil, fmt.Errorf("cannot create MRC %s/%s with status %s. Certificate rotation in progress. At most, only 2 MRCs can be in non error or non inactive states",
+				newMRC.GetNamespace(), newMRC.GetName(), newMRC.Status.State)
+		}
+
+		if _, ok := validMRCTransitions[newMRC.Status.State]; !ok {
```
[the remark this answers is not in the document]
jaellio: I'm not sure I understand your comment, would you mind clarifying? I don't think `if _, ok := validMRCTransitions[newMRC.Status.State]; !ok {` implies that creates have to have no state. I'm still not sure if I am in favor of having the mutating webhook set the initial status of the MRC on create or not.

pkg/validator/validators.go
jaellio: I think I am in favor of this overkill. I'll make the update and we can evaluate if the complexity is worth it

pkg/validator/patch.go
```diff
@@ -18,11 +18,14 @@ const (
...
 
+	// ControlPlaneValidatingWebhookName is the name of the validating webhook for control plane resource.
+	ControlPlaneValidatingWebhookName = "osm-control-plane-validator.k8s.io"
+
 	// ValidatorWebhookSvc is the name of the validator service.
 	ValidatorWebhookSvc = "osm-validator"
 )
 
-func createOrUpdateValidatingWebhook(clientSet kubernetes.Interface, cert *certificate.Certificate, webhookName, meshName, osmNamespace, osmVersion string, validateTrafficTarget bool, enableReconciler bool) error {
+func createOrUpdateValidatingWebhook(clientSet kubernetes.Interface, cert *certificate.Certificate, webhookName, meshName, osmNamespace, osmVersion string, validateTrafficTarget bool, enableReconciler bool, validateMeshRootCertificate bool) error {
```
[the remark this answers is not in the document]
jaellio: We could, I was trying to hide all MRC-related work/validation, but I don't think there are any negative impacts of always performing MRC validation. Happy to remove this boolean if you think that makes more sense

pkg/validator/validators.go
jaellio: Updated. Only performing this check on update for now. I can add it on create as well but the mapping is slightly different

pkg/validator/validators.go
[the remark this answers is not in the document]
jaellio: @keithmattix what are your thoughts on this?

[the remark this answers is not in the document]
keithmattix: Sorry, just seeing this. I think it makes sense to start with a manual rotation process and then think through the implications of automatic rotation as we iterate. Here's my reasoning:

- accidental CI/CD runs happen all the time
- We could always add a mechanism to the MRC where a specific annotation automatically triggers a rotation so that users have a choice
  - In this model, there could be a validating webhook to add the state on creation

pkg/validator/validators.go
steeling: Not sure how that would work actually since helm's ordering would delete the service and pods first anyways

https://github.com/helm/helm/blob/12f1bc0acdeb675a8c50a78462ed3917fb7b2e37/pkg/releaseutil/kind_sorter.go#L72.

Might be worth testing out. If so, maybe we add a custom uninstall portion that removes the webhooks prior to the helm uninstall

pkg/validator/validators.go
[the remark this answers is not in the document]
steeling: here's a question that may help elicit the proper answer:

Can you point to another Kubernetes object where the existance of that object does not signify the intent to create that object? If not, I don't think we should create a snowflake here, and we  should assimilate with the rest of the k8s landscape/ecosystem

pkg/validator/patch.go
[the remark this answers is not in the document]
steeling: seems a bit cleaner IMO :)

pkg/validator/validators.go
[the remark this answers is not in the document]
steeling: ya i misread the code i think on first pass.. don't remember what i was getting at :/ 

I think we should probably be validating that a create comes in as either one of 2 states though, or maybe exactly 1 state is allowed, depending on implementation, vs allowing it to come in at any state?

pkg/validator/validators.go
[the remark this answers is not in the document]
keithmattix: Ah, that makes sense; I agree. I do think we could invert the behavior of the annotation I mentioned, just to give those cautious users an option

pkg/validator/validators.go
jaellio: Will explore and track this in a separate PR/issue

pkg/validator/patch.go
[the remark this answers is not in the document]
jaellio: Removed flag

pkg/validator/validators.go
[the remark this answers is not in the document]
jaellio: Since the MRC's status cannot be set on create, I removed the status check. We should definitely continue the discussion about root certificate rotation initiation, but I don't think it needs to be a blocker for this PR.

pkg/validator/validators.go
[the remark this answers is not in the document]
jaellio: The status validation on create depends on whether or not the status can be successfully set by a mutating webhook. AFAIK there is no other way to set the status of CR on creation.

pkg/validator/validators.go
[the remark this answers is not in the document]
jaellio: Update: I created this [POC](https://github.com/jaellio/osm/pull/new/mutateMRCState) to verify if the status of a custom resource can be set by the mutating webhook and determined it cannot be set. For the POC I added an additional webhook to the mutating webhook configuration that responds to MRC creation events in the control plane namespace. The webhook server sets the status of the MRC and returns the updated MRC in its response. After mutation, the MRC resource has no status. 

This design doc from the kubernetes repo describes the intended behavior of the status subresource https://github.com/kubernetes/design-proposals-archive/blob/main/api-machinery/customresources-subresources.md#status-behavior. 
> If the /status subresource is enabled, the following behaviors change:
> - The main resource endpoint will ignore all changes in the status subpath. (note: it will not reject requests which try to change the status, following the existing semantics of other resources).

Therefore, either the OSM control plane or the user must be responsible for setting the status of an MRC on update. New MRCs (without status) can still be treated as initiating rotation, but the initial statuses cannot be set by a mutating webhook

pkg/validator/validators.go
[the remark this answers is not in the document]
jaellio: See https://github.com/openservicemesh/osm/pull/4811#discussion_r938260427

PR #4816, opened 2022-06-13
Abstract message broker

### LINKED ISSUE #4713: Use MeshRootCertificate resource to configure OSM certificate manager on startup

On injector, bootstrap, and controller start-up read from the MeshRootCertificate to configure the certificate manager.

keithmattix:
**Description**:
Use MRC to initialize certificate manager on startup. Fixes #4713

**Testing done**:

**Affected area**:

3. Has documentation corresponding to this change been updated in the [osm-docs](https://github.com/openservicemesh/osm-docs) repo (if applicable)? no


pkg/certificate/manager.go
```diff
@@ -1,43 +1,105 @@
...
+func (m *Manager) init(mrcs []*v1alpha2.MeshRootCertificate, mrcClient MRCClient) error {
+	for _, mrc := range mrcs {
+		client, ca, clientID, err := mrcClient.GetCertIssuerForMRC(mrc)
+		if err != nil {
+			return err
+		}
+
+		c := &issuer{Issuer: client, ID: clientID, CertificateAuthority: ca}
+		switch {
+		case mrc.Status.State == "errored":
```
steeling, line 73: nit: use a variable instead of hardcoded string "errored"

pkg/certificate/providers/config.go
```diff
@@ -40,47 +42,65 @@ var getCA func(certificate.Issuer) (pem.RootCertificate, error) = func(i certifi
...
-		mrc: &v1alpha2.MeshRootCertificate{
-			ObjectMeta: metav1.ObjectMeta{
-				Name:      "legacy-compat",
-				Namespace: providerNamespace,
-				Annotations: map[string]string{
-					constants.MRCVersionAnnotation: "legacy-compat",
-				},
+	var mrcClient certificate.MRCClient
+	// TODO(4713): Switch out with feature flag or some other mechanism
+	if true {
```
steeling, line 54: I'm not sure we need an extra feature flag toggle. I'd consider just having one or the other, and making the switch at some point in time.

Features are pinned to releases and too many feature flags conflates and confuses users with internal deatils they won't need IMO

pkg/certificate/types.go
```diff
@@ -96,7 +97,37 @@ type Manager struct {
...
+	// The last observed version of the MRC as of the time of this event
+	MRC *v1alpha2.MeshRootCertificate
+	// The previous version of the MRC. Only populated during updates
+	// +optional
+	OldMRC *v1alpha2.MeshRootCertificate
+}
+
+var (
+	// MRCEventAdded is the type of announcement emitted when we observe an addition of a Kubernetes MeshRootCertificate
+	MRCEventAdded MRCEventType = "meshrootcertificate-added"
```
steeling, line 118: Wondering if we might only need update/add, and even then if we need to "know" which event it was.

ie: an add will be a replacement, so knows to move the client into the old stage

pkg/configurator/client.go
```diff
@@ -68,6 +69,20 @@ func (c *Client) getMeshConfig() configv1alpha2.MeshConfig {
 	return meshConfig
 }
 
+func (c *Client) getMeshRootCertificates() ([]configv1alpha2.MeshRootCertificate, error) {
```
steeling, line 72: why not just export this func?

pkg/configurator/methods.go
```diff
@@ -36,6 +36,10 @@ func (c *Client) GetMeshConfig() configv1alpha2.MeshConfig {
 	return c.getMeshConfig()
 }
 
+func (c *Client) GetMeshRootCertificates() ([]configv1alpha2.MeshRootCertificate, error) {
```
steeling, line 39: It looks like this is never used do we need it?

pkg/certificate/manager.go
keithmattix: I hardcoded it since we're merging the statuses and states into 1 + this state doesn't have a car yet. So I agree with you; this is meant to create this same PR comment in the next PR 🙂

pkg/configurator/methods.go
keithmattix: Yeah it's not anymore; this is a relic before we came up with the watch approach

pkg/certificate/fake_manager.go
```diff
@@ -24,6 +25,16 @@ func (c *fakeMRCClient) List() ([]*v1alpha2.MeshRootCertificate, error) {
 	return []*v1alpha2.MeshRootCertificate{{}}, nil
 }
 
+func (c *fakeMRCClient) Watch(ctx context.Context) (<-chan MRCEvent, error) {
+	ch := make(chan MRCEvent)
+	go func() {
```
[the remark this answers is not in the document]
keithmattix: I didn't care that much since it was a fake client. I can though if you think it'll be helpful

pkg/certificate/manager.go
```diff
@@ -1,43 +1,105 @@
...
 	}
 
-	c := &issuer{Issuer: client, ID: clientID, CertificateAuthority: ca}
+	m.init(mrcs, mrcClient)
+
+	// now that the manager is initialized, start a watch
+	// we wait until the manager is initialized so that there's
+	// no lock contention between the event handler and the initial
+	// state of the manager.
+	go func() {
```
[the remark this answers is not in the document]
keithmattix: Ah good call; these can probably be merge. I will say though that this begs the question of if we want to preempt the ticker when an MRC event comes through.

pkg/certificate/types.go
```diff
@@ -96,7 +97,37 @@ type Manager struct {
 // MRCClient is an interface that can watch for changes to the MRC. It is typically backed by a k8s informer.
 type MRCClient interface {
 	List() ([]*v1alpha2.MeshRootCertificate, error)
+	MRCEventBroker
```
jaellio, line 100: Does the MRCEventBroker need to exist in addition to the MRCClient or can the watch method be defined here?

Watch(context.Context) (<-chan MRCEvent, error)

pkg/certificate/providers/mrcclient.go
```diff
@@ -0,0 +1,77 @@
...
+	return mrcs, nil
+}
+
+func (m *MRCClientImpl) Watch(ctx context.Context) (<-chan certificate.MRCEvent, error) {
+	if m.informerCollection == nil {
+		return nil, errNilInformerCollection
+	}
+
+	eventChan := make(chan certificate.MRCEvent)
+	m.informerCollection.AddEventHandler(informers.InformerKeyMeshRootCertificate, cache.ResourceEventHandlerFuncs{
```
jaellio, line 50: Not sure on the exact functionality here, but if there are multiple event handlers for the same type are they called in a specific order? For example, if there are multiple addFuncs defined for the MRC in what order are they invoked? Just curious :)

pkg/certificate/providers/mrcclient.go
```diff
@@ -0,0 +1,77 @@
...
+		UpdateFunc: func(oldObj, newObj interface{}) {
+			oldMRC := oldObj.(*v1alpha2.MeshRootCertificate)
+			mrc := newObj.(*v1alpha2.MeshRootCertificate)
+			eventChan <- certificate.MRCEvent{
+				Type:   certificate.MRCEventUpdated,
+				MRC:    mrc,
+				OldMRC: oldMRC,
+			}
+		},
+		DeleteFunc: func(obj interface{}) {
```
jaellio, line 67: Not sure we need this. But I don't think it hurts

pkg/certificate/manager.go
```diff
@@ -1,43 +1,105 @@
...
+		msgBroker:                   msgBroker,
+	}
+	// We need an MRC in order to get the correct issuers, so handle initial MRC List now
+	mrcs, err := mrcClient.List()
 	if err != nil {
 		return nil, err
 	}
 
-	c := &issuer{Issuer: client, ID: clientID, CertificateAuthority: ca}
+	m.init(mrcs, mrcClient)
```
jaellio, line 35: I think I'd initialize the Manager before watching for mrcEvents.

pkg/certificate/manager.go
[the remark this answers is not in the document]
jaellio: That's a good point Keith. I think I'm in favor of keeping the two processes separate. One is responsible for updating the manager struct and the other is responsible for rotating certificates (due to expiration or root certificate rotation). The ticker is already set for 5 seconds intervals so the certificates will be reissued relatively quickly following a manager update. That being said, I can also see the benefits of combining the processes

pkg/certificate/types.go
keithmattix: We'll, in the final version of this, I'll implement the Transform function for the message broker so that it can fulfill this interface if need be. Having a single type makes that easier

pkg/certificate/manager.go
```diff
@@ -1,43 +1,105 @@
...
 	return m, nil
 }
 
+// init initializes the Manager receiver based on the observed state of the cluster at this
+// instant in time. This is why we pass in a slice of MRCs; we know our Watch call will make us initially
+// consistent, we just need a valid state of the world at this instant to kick things off.
+// This method is meant to be called at initialization time, so we don't grab the lock
+func (m *Manager) init(mrcs []*v1alpha2.MeshRootCertificate, mrcClient MRCClient) error {
+	for _, mrc := range mrcs {
+		client, ca, clientID, err := mrcClient.GetCertIssuerForMRC(mrc)
```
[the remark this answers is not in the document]
keithmattix: I thought about that, but the performance hit didn't seem to outweigh the readability of duplicating the initialization 3+ times

pkg/certificate/providers/config.go
keithmattix: True...what I might do is default to MRCs, but if there are none, use the compat client

pkg/certificate/types.go
keithmattix: I generally err on the side of giving subscribers too much information rather than not enough. Having the context of what action is being taken on the MRC can aid in decisionmaking

pkg/certificate/providers/mrcclient.go
keithmattix: Good question! So this is what the Go documentation for that function says (emphasis mine):
> AddEventHandler adds an event handler to the shared informer using the shared informer's resync period.  Events to a single handler are delivered sequentially, **but there is no coordination between different handlers.**

So there's no coordination between subscribers to these events, but events to a single subscriber are guaranteed to be delivered in-order.

pkg/certificate/providers/mrcclient.go
keithmattix: You have to provide a Delete function otherwise, you'll get a nil pointer exception when the internal informer logic tries to handle the event.

pkg/certificate/providers/mrcclient.go
steeling: To clarify: Are they guaranteed to be in order from the k8s master? ie: if a user does a delete then an update on an object is that guaranteed to be in order?

I'd guess that k8s doesn't place those guarantees. I'd expect that missed updates are also possible (this is probably why they have a resync interval exposed)

keithmattix: They guarantee that received events will be in order, NOT that every state will be enumerated. From the sharedinformer docs:
> A SharedInformer maintains a local cache...of he state of each relevant object.  This cache is eventually consistent with the authoritative state.  This means that, unless prevented by persistent communication problems, if ever a particular object ID X is authoritatively associated with a state S then for every SharedInformer I whose collection includes (X, S) eventually either (1) **I's cache associates X with S or a later state of X**, (2) I is stopped, or (3) the authoritative state service for X terminates.

So we won't get out-of-order updates, but updates may be coalesced

pkg/certificate/manager.go
```diff
@@ -1,43 +1,105 @@
...
+	}
+
+	if m.signingIssuer == nil || m.validatingIssuer == nil {
+		return ErrNoValidMeshRootCertificates
+	}
+
+	return nil
+}
+
+func (m *Manager) handleMRCEvent(event MRCEvent) {
```
steeling, line 92: based on ordering guarantees, and assuming we have proper validations in the validating webhook, we may want to consider simplifying this, and only accept opaque "updates", (where an update is 1 of an add or an update)

We likely don't need to know the details of prior state (we should have that cached, and as mentioned, old updates may be coalesced), nor do we need to know about deletes, as that should be associated with the promotion of another MRC to active

pkg/certificate/providers/mrcclient.go
keithmattix: Oh that's a fantastic point. I don't think we have to worry about that at this layer, but subscribers should be aware.

pkg/certificate/manager.go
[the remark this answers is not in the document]
steeling: i think it's fine to have 2 separate processes, but I think it's an incongruence to implicitly start one and explicitly start the other.

Can we do either both implicit, or both explicit?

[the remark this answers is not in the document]
keithmattix: Ah, I see what you're saying now. So in the former case would the `Start()` function's arguments move to the constructor?

pkg/certificate/manager.go
steeling: ya, but you already have a context, so i'd get rid of the stopchan

pkg/certificate/manager.go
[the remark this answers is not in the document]
steeling: ya, but you already have a context, so i'd get rid of the stopchan

pkg/certificate/manager.go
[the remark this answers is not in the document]
steeling: maybe pull the invalid state out of the switch and to the top of the func

pkg/certificate/manager.go
keithmattix: Hm, yeah I see what you mean. I'm good to YAGNI here

pkg/certificate/manager.go
```diff
@@ -1,59 +1,125 @@
...
 	}()
+
+	return m, nil
+}
+
+// init initializes the Manager receiver based on the observed state of the cluster at this
+// instant in time. This is why we pass in a slice of MRCs; we know our Watch call will make us initially
+// consistent, we just need a valid state of the world at this instant to kick things off.
+// This method is meant to be called at initialization time, so we don't grab the lock
+func (m *Manager) init(mrcs []*v1alpha2.MeshRootCertificate, mrcClient MRCClient) error {
```
nojnhuh, line 80: Is there a way we can get away with not doing this explicitly at startup and handle this just like an MRC added during runtime?

Then it seems like we could remove this constraint and let the informers do their regular list and watch without us having to filter out the initially-listed objects ourselves:
https://github.com/openservicemesh/osm/blob/f3aa02a10078d608c04e63dad20c23306e4c20d2/pkg/certificate/types.go#L129-L130

pkg/certificate/providers/mrc.go
```diff
@@ -0,0 +1,127 @@
...
+			}
+			eventChan <- certificate.MRCEvent{
+				Type: certificate.MRCEventAdded,
+				MRC:  mrc,
+			}
+		},
+		UpdateFunc: func(oldObj, newObj interface{}) {
+			oldMRC := oldObj.(*v1alpha2.MeshRootCertificate)
+			mrc := newObj.(*v1alpha2.MeshRootCertificate)
+			if !checkGeneration(mrcObjectVersionCache, *mrc) {
```
nojnhuh, line 94: (If we find we still need to do this)

Would an update or delete event ever come through that happened before the event handler is added? My understanding is that every existing object comes through as an "add" event to the handler, then any update or delete is based off of those objects which would have to happen after the handler was created.

pkg/certificate/providers/mrc.go
```diff
@@ -0,0 +1,127 @@
...
+				Type:   certificate.MRCEventUpdated,
+				MRC:    mrc,
+				OldMRC: oldMRC,
+			}
+		},
+		DeleteFunc: func(obj interface{}) {
+			mrc, ok := obj.(*v1alpha2.MeshRootCertificate)
+			// Sometimes, deleted objects will be of type DeletedFinalStateUnknown
+			// We don't send these events because we literally can't gain anything
+			// from knowing that an unknown MRC was deleted
```
nojnhuh: Does the object stay in the local cache in that scenario? If so, should we periodically prune the cache ourselves by querying the API so things like `List` remain accurate?

keithmattix: To answer your question, yes the object would still be in the local cache during that scenario. Typically, the way to solve this is by setting a nonzero resync period on the informer OR the event handler. I ended up gutting this function so it's moot for this PR, but good question!

pkg/certificate/manager.go
keithmattix: So, the scenario I'm trying to avoid is when an Envoy proxy or the ADS server needs a certificate and there's no CA configured in the certificate manager at the moment. Especially since the `Start()` method was removed from `Manager`, the idea is that the `Manager` returned from `NewManager` is ready to use (or an error is returned). Yes that means we have to add some logic in the MRCClient, but I think the tradeoff is preferable

pkg/certificate/manager.go
```diff
@@ -1,59 +1,122 @@
...
-// needs to be rotated
-func (m *Manager) Start(checkInterval time.Duration, stop <-chan struct{}) {
+	go func() {
+		for {
+			select {
+			case <-ctx.Done():
+				return
+			case event, open := <-mrcEvents:
+				if !open {
+					// channel was closed; return
```
steeling, line 51: add a log statement for this case, and for done as well. For the ctx.Done(), probably also check ctx.Err

pkg/certificate/manager.go
steeling: I think this code might look almost identical to what we need for `handleMRCEvent` FWIW.

Might make sense to tackle in a single PR. thoughts @jaellio

pkg/certificate/manager.go
```diff
@@ -1,59 +1,122 @@
...
+		case mrc.Status.RotationStage == constants.MRCStageValidating:
+			m.validatingIssuer = c
+		default:
+			m.signingIssuer = c
+			m.validatingIssuer = c
+		}
+	}
+
+	if m.signingIssuer == nil || m.validatingIssuer == nil {
+		return fmt.Errorf("invalid mesh root certificate configuration. Signing issuer was %#v and validating issuer was %#v", m.signingIssuer, m.validatingIssuer)
```
steeling, line 107: nit: if logging these, we could also add a `String() string` method to issuer, to more nicely print. Not a requirement

pkg/certificate/providers/mrc.go
```diff
@@ -0,0 +1,118 @@
...
+// to observe MRCs (via List() and Watch()) as well as generate
+// `certificate.Provider`s from those MRCs
+type MRCComposer struct {
+	informerCollection *informers.InformerCollection
+	MRCProviderGenerator
+}
+
+// List returns the MRCs stored in the informerCollection's store
+func (m *MRCComposer) List() ([]*v1alpha2.MeshRootCertificate, error) {
+	if m.informerCollection == nil {
```
steeling, line 33: this is a non-exported field, so we should use constructors such that this can never be nil, then we don't need the extra check

pkg/certificate/providers/mrc.go
```diff
@@ -0,0 +1,118 @@
...
+		// We don't care about deletes because the only delates that should
+		// happen come from the control plane cleaning up an old MRC. Our
+		// ValdatingWebhookConfiguration should prevent deletes from users
+		DeleteFunc: func(obj interface{}) {},
+	})
+
+	return eventChan, nil
+}
+
+func checkGeneration(mrcObjectVersionCache map[string]int64, mrc v1alpha2.MeshRootCertificate) bool {
```
steeling, line 111: is this checking what we discussed, where the informers guarantee we don't need to do this?

pkg/certificate/providers/mrc.go
```diff
@@ -0,0 +1,118 @@
...
+)
+
+var mrcKeyFunc = func(mrc v1alpha2.MeshRootCertificate) string {
+	return fmt.Sprintf("%s/%s", mrc.GetNamespace(), mrc.GetName())
+}
+
+// MRCComposer is a composer object that allows consumers
+// to observe MRCs (via List() and Watch()) as well as generate
+// `certificate.Provider`s from those MRCs
+type MRCComposer struct {
```
steeling, line 26: Can we get rid of the MRCComposer, and put these methods on the MRC client?

keithmattix: `MRCClient` is an interface; are you suggesting turning it into a struct?

pkg/certificate/types.go
```diff
@@ -96,7 +97,36 @@ type Manager struct {
...
+)
+
+// MRCEventBroker describes any type that allows the caller to Watch() MRCEvents
+type MRCEventBroker interface {
+	// Watch allows the caller to subscribe to events surrounding
+	// MRCs that belong to this particular mesh. Watch returns a channel
+	// that emits events, and an error if the subscription goes awry.
+	// Watch() should only publish events that occur after the invocation
+	// of watch
+	Watch(context.Context) (<-chan MRCEvent, error)
```
steeling, line 131: maybe consider putting this on the mrcclient interface

pkg/certificate/providers/mrc.go
keithmattix: From the informer docs re: event handlers:
```A client added after `Run()` starts gets a startup batch of notifications of additions of the objects existing in the cache at the time that client was added; also, for every update to the SharedInformer's local cache after that client was added, eventually either the SharedInformer is stopped or that client is notified of that update.  Client notifications happen after the corresponding cache update and, in the case of a SharedIndexInformer, after the corresponding index updates. ```

This function makes sure the `certificate.Manager` doesn't try to recompute issuer state from the very beginning. Admittedly, the drift between when the MRC informer starts and when the Manager starts to watch is likely minimal, but this logic also prevents `Watch` from duplicating results retrieved from the manger's initial `List` call

File: pkg/certificate/providers/mrc.go (hunk 21 above)
Comment (steeling) [reply pkg/certificate/providers/mrc.go:26]: err.. sorry i was getting mixed up with the compatclient vs this vs interface.

File: pkg/certificate/providers/mrc.go [hunk 23]
```diff
@@ -0,0 +1,118 @@
+package providers
```
steeling, line 1: thoughts on putting this in `pkg/k8s`, since this is a k8s specific implementation of the MRCClient?

steeling (changes requested):

pkg/certificate/providers/mrc.go
keithmattix: Technically, the informerCollection still could be nil if its parameter higher in the stack is set to nil (e.g. during tests). These checks were honestly a way to get around nil pointer exceptions in a couple of tests, but I don't _think_ it's needed anymore because we've got a check upstream. I'll remove it

pkg/certificate/manager.go
keithmattix: This error formatting is really just to show which issuer is nil. If we log the issuers elsewhere, I agree, we should add a `String()` func

pkg/certificate/providers/mrc.go
steeling: since watch will return the entire set, should we just remove the list call?

would that fix the issues you're describing? if not can you elaborate on the state drift you're referring to

pkg/certificate/types.go
keithmattix: This interface is embedded there. [See my conversation with Jackie](https://github.com/openservicemesh/osm/pull/4816/files/ca6599a0777ab0e48a7f6968e45fd64fa5f7eb60#r896252417)

pkg/certificate/providers/mrc.go
keithmattix: The `MRCProviderGenerator` field/embedding would make that an import cycle. `pkg/certificate/providers` imports `pkg/k8s` which would import `pkg/certificate/providers` where `MRCProviderGenerator is stored

pkg/certificate/providers/mrc.go
keithmattix: That gets into my [conversation with John](https://github.com/openservicemesh/osm/pull/4816#discussion_r898464225). In order to return a ready-to-use `Manager`, my approach was essentially to initialize the two issuers based on the MRC state at one point of time (i.e. a stable state). From there, the manager can receive future updates via the watch. This function ensures that the events sent to the `Manager` are in fact "future updates" and not old events

steeling: i think the thing is that our state machine can be really dumb. In that it can see an active MRC, and know that it is the only active one, and act accordingly.

I'd advocate for simplicity vs flexibility here

keithmattix: But then what happens on a control plane restart during an MRC rotation? There will be 2 MRCs, maybe one validating and one issuing. I can factor out the base case (`len(mrcs) == 1`)  if that's what you're saying, but I feel like allowing users of `MRCBroker` implementations to assume `Watch` events are all relevant is worth some of the complexity. FWIW, this is the part of this PR I'm least comfortable with, so if we can address some of the scenarios I've described, I'll be the first one to rip this out 😄

pkg/certificate/manager.go
```diff
@@ -56,6 +47,103 @@ func (m *Manager) Start(checkInterval time.Duration, stop <-chan struct{}) {
...
+				}
+			}
+
+			if m.signingIssuer != nil && m.validatingIssuer != nil {
+				once.Do(func() {
+					wg.Done()
+				})
+			}
+		}
+	}(&wg, &once)
```
steeling, line 93: nit: no need to pass these in

pkg/certificate/manager.go
```diff
@@ -56,6 +47,103 @@ func (m *Manager) Start(checkInterval time.Duration, stop <-chan struct{}) {
...
+
+	done := make(chan struct{})
+
+	// Wait for WaitGroup to finish and notify select when it does
+	go func() {
```
steeling, line 98: nit: this pattern looks well suited for the sync.Cond type

steeling (approved):

pkg/certificate/manager.go
keithmattix: Oh I didn't know go had a cond data structure! Yeah, I'll try to implement this

pkg/certificate/providers/mrc.go
```diff
@@ -0,0 +1,70 @@
...
+		// We don't really care about the previous version
+		// since the "state machine" of the MRC is well defined
+		UpdateFunc: func(oldObj, newObj interface{}) {
+			mrc := newObj.(*v1alpha2.MeshRootCertificate)
+			eventChan <- certificate.MRCEvent{
+				Type: certificate.MRCEventUpdated,
+				MRC:  mrc,
+			}
+		},
+		// We don't care about deletes because the only delates that should
```
nojnhuh, line 63:
```suggestion
		// We don't care about deletes because the only deletes that should
```

pkg/certificate/providers/mrc.go
```diff
@@ -0,0 +1,70 @@
...
+		AddFunc: func(obj interface{}) {
+			mrc := obj.(*v1alpha2.MeshRootCertificate)
+			eventChan <- certificate.MRCEvent{
+				Type: certificate.MRCEventAdded,
+				MRC:  mrc,
+			}
+		},
+		// We don't really care about the previous version
+		// since the "state machine" of the MRC is well defined
+		UpdateFunc: func(oldObj, newObj interface{}) {
```
nojnhuh, line 56: Selfish suggestion to keep my editor from whining at me about the unused parameter:
```suggestion
		UpdateFunc: func(_, newObj interface{}) {
```

pkg/certificate/providers/config.go
```diff
@@ -40,47 +42,65 @@ var getCA func(certificate.Issuer) (pem.RootCertificate, error) = func(i certifi
...
-		mrc: &v1alpha2.MeshRootCertificate{
-			ObjectMeta: metav1.ObjectMeta{
-				Name:      "legacy-compat",
-				Namespace: providerNamespace,
-				Annotations: map[string]string{
-					constants.MRCVersionAnnotation: "legacy-compat",
-				},
+	var mrcClient certificate.MRCClient
+	if ic == nil || len(ic.List(informers.InformerKeyMeshRootCertificate)) == 0 {
+		// no MRCs detected; use the compat client
```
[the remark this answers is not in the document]
jaellio: Right now a MRC is always created by osm-bootstrap.  If we want to support the option of not using the MRC created on startup to configure the Manager, maybe it would be best to make using the MRC an opt in option on install?

pkg/certificate/manager.go
```diff
@@ -56,6 +47,103 @@ func (m *Manager) Start(checkInterval time.Duration, stop <-chan struct{}) {
...
+	}
+
+	return nil
+}
+
+func (m *Manager) handleMRCEvent(mrcClient MRCClient, event MRCEvent) error {
+	switch event.Type {
+	case MRCEventAdded:
+		mrc := event.MRC
+		if mrc.Status.State == "errored" {
```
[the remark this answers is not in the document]
keithmattix: Ah, that var wasn't there before my previous rebase...the states also changed. I'll have to rebase again to fix merge conflicts and I'll fix this as well

pkg/certificate/manager.go
```diff
@@ -56,6 +47,103 @@ func (m *Manager) Start(checkInterval time.Duration, stop <-chan struct{}) {
...
+		return err
+	}
+
+	wg.Add(1)
+
+	go func(wg *sync.WaitGroup, once *sync.Once) {
+		for {
+			select {
+			case <-ctx.Done():
+				if ctx.Err() != nil {
```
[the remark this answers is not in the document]
keithmattix: Good catch!

pkg/certificate/manager.go
keithmattix: On second thought, this scenario seems a bit too complicated for a cond; we don't have multiple readers or writers. Plus, it's harder to implement a timeout since a cond is based on locks not channels

pkg/certificate/providers/config.go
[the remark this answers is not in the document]
keithmattix: I don't have strong opinions either way. If we ARE making a hard cut though, we're likely going to want some documentation on the change and instructions on how to switch

nojnhuh (approved):
LGTM once conflicts are resolved.

jaellio (approved):

PR #4825, opened 2022-06-16
chore(cmd/cli): Reduce cyclomatic complexity of uninstall command

shalier:
**Description**:
Separate code into functions to reduce the cyclomatic complexity
Helps unblock #4555

Reduces cyclomatic complexity 30->19
image

image

**Testing done**:

**Affected area**:
| CLI Tool                   | [x ] |

2. Is this a breaking change?

### DIFF
--- a/cmd/cli/uninstall_mesh.go
+++ b/cmd/cli/uninstall_mesh.go
@@ -184,27 +184,25 @@ func (d *uninstallMeshCmd) run() error {
 	}
 
 	err := d.deleteClusterResources()
-	if err != nil {
-		return err
-	}
-
-	return nil
+	return err
 }
 
 func (d *uninstallMeshCmd) findFlagSpecifiedMesh(meshInfoList []meshInfo) (bool, bool) {
 	meshFlagSpecified := d.meshName != ""
 	specifiedMeshFound := false
-	if meshFlagSpecified {
-		specifiedMeshFound = d.findMesh(meshInfoList)
-		if !specifiedMeshFound {
-			fmt.Fprintf(d.out, "Did not find mesh [%s] in namespace [%s]\n", d.meshName, d.meshNamespace)
-			// print a list of meshes within the cluster for a better user experience
-			err := d.printMeshes()
-			if err != nil {
-				fmt.Fprintf(d.out, "Unable to list meshes in the cluster - [%v]", err)
-			}
+	if !meshFlagSpecified {
+		return meshFlagSpecified, specifiedMeshFound
+	}
+
+	specifiedMeshFound = d.findMesh(meshInfoList)
+	if !specifiedMeshFound {
+		fmt.Fprintf(d.out, "Did not find mesh [%s] in namespace [%s]\n", d.meshName, d.meshNamespace)
+		// print a list of meshes within the cluster for a better user experience
+		if err := d.printMeshes(); err != nil {
+			fmt.Fprintf(d.out, "Unable to list meshes in the cluster - [%v]", err)
 		}
 	}
+
 	return meshFlagSpecified, specifiedMeshFound
 }
 
@@ -226,16 +224,17 @@ func (d *uninstallMeshCmd) promptMeshUninstall(meshInfoList, meshesToUninstall [
 }
 
 func (d *uninstallMeshCmd) deleteNs(ctx context.Context, ns string) error {
-	if d.deleteNamespace {
-		if err := d.clientSet.CoreV1().Namespaces().Delete(ctx, ns, v1.DeleteOptions{}); err != nil {
-			if k8sApiErrors.IsNotFound(err) {
-				fmt.Fprintf(d.out, "OSM namespace [%s] not found\n", ns)
-				return nil
-			}
-			return errors.Errorf("Error occurred while deleting OSM namespace [%s] - %v", ns, err)
-		}
-		fmt.Fprintf(d.out, "OSM namespace [%s] deleted successfully\n", ns)
+	if !d.deleteNamespace {
+		return nil
 	}
+	if err := d.clientSet.CoreV1().Namespaces().Delete(ctx, ns, v1.DeleteOptions{}); err != nil {
+		if k8sApiErrors.IsNotFound(err) {
+			fmt.Fprintf(d.out, "OSM namespace [%s] not found\n", ns)
+			return nil
+		}
+		return errors.Errorf("Error occurred while deleting OSM namespace [%s] - %v", ns, err)
+	}
+	fmt.Fprintf(d.out, "OSM namespace [%s] deleted successfully\n", ns)
 	return nil
 }
 
--- a/cmd/cli/uninstall_mesh.go
+++ b/cmd/cli/uninstall_mesh.go
@@ -118,8 +118,11 @@ func (d *uninstallMeshCmd) run() error {
 			fmt.Fprintf(d.out, "No OSM control planes found\n")
 			return nil
 		}
+
+		meshFlagSpecified := d.meshName != ""
+
 		// Searches for the mesh specified by the mesh-name flag if specified
-		meshFlagSpecified, specifiedMeshFound := d.findFlagSpecifiedMesh(meshInfoList)
+		specifiedMeshFound := d.findSpecifiedMesh(meshFlagSpecified, meshInfoList)
 		if meshFlagSpecified && !specifiedMeshFound {
 			return nil
 		}
@@ -164,7 +167,7 @@ func (d *uninstallMeshCmd) run() error {
 					return err
 				}
 
-				fmt.Fprintf(d.out, "Error %v when trying to uninstall mesh name [%s] in namespace [%s] - continuing to deleteClusterWideResources and/or deleteNamespace\n", err, m.name, m.namespace)
+				fmt.Fprintf(d.out, "Could not uninstall mesh name [%s] in namespace [%s]- %v - continuing to deleteClusterWideResources and/or deleteNamespace\n", m.name, m.namespace, err)
 			}
 
 			if err == nil {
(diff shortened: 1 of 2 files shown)


cmd/cli/uninstall_mesh.go
```diff
@@ -199,6 +182,64 @@ func (d *uninstallMeshCmd) run() error {
 			fmt.Fprintf(d.out, "OSM namespace CANNOT be deleted in a managed environment\n")
 		}
 	}
+
+	err := d.deleteClusterResources()
+	if err != nil {
+		return err
+	}
+
+	return nil
```
trstringer, line 191:
```suggestion
	err := d.deleteClusterResources()
	return err
```

Or this could even be more simplified to:

```suggestion
	return d.deleteClusterResources()
```

But I think this is likely not the most readable choice.

cmd/cli/uninstall_mesh.go
```diff
@@ -199,6 +182,64 @@ func (d *uninstallMeshCmd) run() error {
...
+		}
+		if confirm {
+			meshesToUninstall = append(meshesToUninstall, mesh)
+		}
+	}
+	return meshesToUninstall, nil
+}
+
+func (d *uninstallMeshCmd) deleteNs(ctx context.Context, ns string) error {
+	if d.deleteNamespace {
```
trstringer, line 229: You can remove a level of indentation by reworking the check.

```suggestion
	if !d.deleteNamespace {
	    return nil
	}
```

trstringer (changes requested):

allenlsy:
LGTM other than Thomas's comments

cmd/cli/uninstall_mesh.go
```diff
@@ -199,6 +182,64 @@ func (d *uninstallMeshCmd) run() error {
...
+		if !specifiedMeshFound {
+			fmt.Fprintf(d.out, "Did not find mesh [%s] in namespace [%s]\n", d.meshName, d.meshNamespace)
+			// print a list of meshes within the cluster for a better user experience
+			err := d.printMeshes()
+			if err != nil {
+				fmt.Fprintf(d.out, "Unable to list meshes in the cluster - [%v]", err)
+			}
+		}
+	}
+	return meshFlagSpecified, specifiedMeshFound
```
jaellio, line 208: nit: provided an alternative. Feel free to keep this method as is!!
```suggestion
	if !meshFlagSpecified {
	        return meshFlagSpecified, specifiedMeshFound
	}
	
	specifiedMeshFound = d.findMesh(meshInfoList)
	if !specifiedMeshFound {
		fmt.Fprintf(d.out, "Did not find mesh [%s] in namespace [%s]\n", d.meshName, d.meshNamespace)
		// print a list of meshes within the cluster for a better user experience
		if err := d.printMeshes(); err != nil {
			fmt.Fprintf(d.out, "Unable to list meshes in the cluster - [%v]", err)
		}
	}
	
	return meshFlagSpecified, specifiedMeshFound
```

cmd/cli/uninstall_mesh.go
```diff
@@ -199,6 +182,63 @@ func (d *uninstallMeshCmd) run() error {
...
+func (d *uninstallMeshCmd) deleteNs(ctx context.Context, ns string) error {
+	if !d.deleteNamespace {
+		return nil
+	}
+	if err := d.clientSet.CoreV1().Namespaces().Delete(ctx, ns, v1.DeleteOptions{}); err != nil {
+		if k8sApiErrors.IsNotFound(err) {
+			fmt.Fprintf(d.out, "OSM namespace [%s] not found\n", ns)
+			return nil
+		}
+		return errors.Errorf("Error occurred while deleting OSM namespace [%s] - %v", ns, err)
```
mudit-01, line 235: Its better to lowercase the error message here.

cmd/cli/uninstall_mesh.go
```diff
@@ -199,6 +182,63 @@ func (d *uninstallMeshCmd) run() error {
 			fmt.Fprintf(d.out, "OSM namespace CANNOT be deleted in a managed environment\n")
 		}
 	}
+
+	err := d.deleteClusterResources()
+	return err
+}
+
+func (d *uninstallMeshCmd) findFlagSpecifiedMesh(meshInfoList []meshInfo) (bool, bool) {
```
steeling, line 190: nit, since the method doesn't "know" that the fields are specified via flag or not, it shouldn't "leak" that info through its name.

`findMesh` seems like a descriptive name for this method

steeling, line 190: how about we hide the double return value, to return a single boolean on whether the mesh was found via the specified info.

cmd/cli/uninstall_mesh.go
```diff
@@ -199,6 +182,63 @@ func (d *uninstallMeshCmd) run() error {
...
+	meshFlagSpecified := d.meshName != ""
+	specifiedMeshFound := false
+	if !meshFlagSpecified {
+		return meshFlagSpecified, specifiedMeshFound
```
steeling, line 194: you know the values here, prefer a return with the direct values not the vars.

`return false, false`

although, with above comment, we can simply return false

cmd/cli/uninstall_mesh.go
```diff
@@ -199,6 +182,63 @@ func (d *uninstallMeshCmd) run() error {
...
+	specifiedMeshFound = d.findMesh(meshInfoList)
+	if !specifiedMeshFound {
+		fmt.Fprintf(d.out, "Did not find mesh [%s] in namespace [%s]\n", d.meshName, d.meshNamespace)
+		// print a list of meshes within the cluster for a better user experience
+		if err := d.printMeshes(); err != nil {
+			fmt.Fprintf(d.out, "Unable to list meshes in the cluster - [%v]", err)
+		}
+	}
+
+	return meshFlagSpecified, specifiedMeshFound
```
steeling, line 206: again it's easier to hard code the value if we know its static, vs tracing to figure it out.

ie: meshFlagSpecified is guaranteed to be true here.

Although with above comment, we should be able to just return specifiedMeshFound

cmd/cli/uninstall_mesh.go
```diff
@@ -119,12 +119,9 @@ func (d *uninstallMeshCmd) run() error {
...
 		}
 		// Searches for the mesh specified by the mesh-name flag if specified
-		specifiedMeshFound := false
-		if d.meshName != "" {
-			specifiedMeshFound = d.findMesh(meshInfoList)
-			if !specifiedMeshFound {
-				return nil
-			}
+		meshFlagSpecified, specifiedMeshFound := d.findFlagSpecifiedMesh(meshInfoList)
+		if meshFlagSpecified && !specifiedMeshFound {
```
steeling, line 123: if these values are just logically compared, we should be able to return a single boolean from the method findFlagSpecifiedMesh

steeling (changes requested):

cmd/cli/uninstall_mesh.go
shalier: I ended up changing the error message to `return errors.Errorf("Could not delete OSM namespace [%s] - %v", ns, err)`

cmd/cli/uninstall_mesh.go
shalier: `findMesh` is already an existing function, I changed the function name to `findSpecifiedMesh` and the function signature to take in the `flagMeshSpecified` bool.

nojnhuh (approved):

draychev (approved):
Who can argue against reduction in cyclomatic complexity!

Thank you for these changes @shalier

cmd/cli/uninstall_mesh.go
```diff
@@ -199,6 +185,61 @@ func (d *uninstallMeshCmd) run() error {
...
+func (d *uninstallMeshCmd) deleteNs(ctx context.Context, ns string) error {
+	if !d.deleteNamespace {
+		return nil
+	}
+	if err := d.clientSet.CoreV1().Namespaces().Delete(ctx, ns, v1.DeleteOptions{}); err != nil {
+		if k8sApiErrors.IsNotFound(err) {
+			fmt.Fprintf(d.out, "OSM namespace [%s] not found\n", ns)
+			return nil
+		}
+		return errors.Errorf("Could not delete OSM namespace [%s] - %v", ns, err)
```
steeling, line 236: nit: errors package is deprecated, use fmt.Errorf

steeling (commented):
a couple small comments, but otherwise LGTM!

steeling (approved):

cmd/cli/uninstall_mesh.go
```diff
@@ -118,13 +118,11 @@ func (d *uninstallMeshCmd) run() error {
...
 		// Searches for the mesh specified by the mesh-name flag if specified
-		specifiedMeshFound := false
-		if d.meshName != "" {
-			specifiedMeshFound = d.findMesh(meshInfoList)
-			if !specifiedMeshFound {
-				return nil
-			}
+		specifiedMeshFound := d.findSpecifiedMesh(meshInfoList)
+		if !specifiedMeshFound {
+			return nil
```
jaellio, line 125: I am not sure this change has the same functionality as before. Previously, if a mesh name was specified, we'd search for the mesh and verify its existence. If it didn't exist we would return nil and not perform an uninstall. With this change, we call `findSpecifiedMesh` regardless of whether or not a mesh name was specified. If a mesh name was not specified `findSpecifiedMesh` will return false and the following condition check with be met and we will return without uninstalling. I think this may be the cause of the failing tests in `e2e_osm_reinstall_test.go`.

jaellio (changes requested):

jaellio (approved):

PR #4833, opened 2022-06-20
Abstract webhook logic to prepare for rotating certificates

steeling:
Abstract the webhook logic so that it shares a common base set of methods for creating the server.

This is in preparation to allow these webhook servers to rotate their certificates

There's a couple things this PR does:

1. Leverage a single, underlying webhook server that starts up the appropriate server with TLS and an ability to rotate certificates.
2. remove ports as flags as these were neither exposed through helm, and there was a bug where if a user *did* specify a different port via the flags, it wouldn't actually be plumbed through properly.
3. Remove inconsistency with health checks across the 3 backend systems, by creating a simple healthcheck on the metrics port.
4. Switch the TLS Config from a hard coded cert to a getter function, to allow for *future* rotation

Part of #4839


cmd/osm-injector/osm-injector.go
```diff
@@ -241,6 +242,7 @@ func main() {
 	}
 
 	<-stop
+	cancel()
```
nojnhuh, line 245: Is there a reason this is necessary on top of the `defer cancel()` where this is created? And it looks like this was already being called at more-or-less this same point in time from within `signals.RegisterExitHandlers`.

pkg/webhook/server.go
```diff
@@ -0,0 +1,102 @@
...
+	// set the certificate once.
+	if err := s.setCert(name, namespace); err != nil {
+		return nil, err
+	}
+	return s, nil
+}
+
+// Run actually starts the server. It blocks until the passed in context is done.
+func (s *Server) Run(ctx context.Context) {
+	log.Info().Msgf("Starting conversion webhook server on: %s", s.server.Addr)
```
jaellio, line 62: nit: similar to Jon's comment below 
```suggestion
	log.Info().Msgf("Starting webhook server on: %s", s.server.Addr)
```

pkg/injector/webhook.go
```diff
@@ -39,103 +37,42 @@ const (
...
-	w.WriteHeader(http.StatusOK)
-	if _, err := w.Write([]byte("Health OK")); err != nil {
-		log.Error().Err(err).Msg("Error writing bytes for mutating webhook health check handler")
+	// Start the MutatingWebhook web server
+	srv, err := webhook.NewServer(constants.OSMInjectorName, osmNamespace, constants.InjectorWebhookPort, certManager, map[string]http.HandlerFunc{
+		webhookCreatePod: http.HandlerFunc(wh.podCreationHandler),
+	},
+		func(cert *certificate.Certificate) error {
+			if err := createOrUpdateMutatingWebhook(kubeClient, cert, webhookTimeout, webhookConfigName, meshName, osmNamespace, osmVersion, enableReconciler); err != nil {
+				return errors.Errorf("Error creating MutatingWebhookConfiguration %s: %+v", webhookConfigName, err)
```
jaellio, line 67:
```suggestion
				return errors.Errorf("Error creating MutatingWebhookConfiguration %s: %+v", webhookConfigName, err)
```
```suggestion
				return errors.Errorf("Error creating or updating MutatingWebhookConfiguration %s: %+v", webhookConfigName, err)
```

pkg/validator/server.go
```diff
@@ -58,12 +47,20 @@ func NewValidatingWebhook(webhookConfigName, osmNamespace, osmVersion, meshName
...
 	}
 
-	// Create the ValidatingWebhook
-	if err := createOrUpdateValidatingWebhook(kubeClient, webhookHandlerCert, webhookConfigName, meshName, osmNamespace, osmVersion, validateTrafficTarget, enableReconciler); err != nil {
-		return errors.Errorf("Error creating ValidatingWebhookConfiguration %s: %+v", webhookConfigName, err)
+	srv, err := webhook.NewServer(ValidatorWebhookSvc, osmNamespace, constants.ValidatorWebhookPort, certManager, map[string]http.HandlerFunc{
+		validationAPIPath: v.doValidation,
+	}, func(cert *certificate.Certificate) error {
+		if err := createOrUpdateValidatingWebhook(kubeClient, cert, webhookConfigName, meshName, osmNamespace, osmVersion, validateTrafficTarget, enableReconciler); err != nil {
+			return errors.Errorf("Error creating ValidatingWebhookConfiguration %s: %+v", webhookConfigName, err)
```
jaellio, line 54:
```suggestion
			return errors.Errorf("Error creating or updating ValidatingWebhookConfiguration %s: %+v", webhookConfigName, err)
```

pkg/webhook/server.go
```diff
@@ -0,0 +1,102 @@
...
+		TLSConfig: &tls.Config{
+			GetCertificate: func(chi *tls.ClientHelloInfo) (*tls.Certificate, error) {
+				s.mu.Lock()
+				defer s.mu.Unlock()
+				return &s.cert, nil
+			},
+			MinVersion: tls.VersionTLS13,
+		},
+	}
+	// set the certificate once.
```
jaellio, line 53: Maybe this comment could clarify the webhook configuration is also being created.

cmd/osm-injector/osm-injector.go
steeling: nope, removed that call

pkg/validator/server_test.go
```diff
@@ -162,15 +162,14 @@ func TestNewValidatingWebhook(t *testing.T) {
 		tassert.NoError(t, err)
 		policyClient := policy.NewPolicyController(informerCollection, nil, broker)
 
-		err = NewValidatingWebhook(webhook.Name, testNamespace, testVersion, testMeshName, enableReconciler, validateTrafficTarget, port, certManager, kube, policyClient, nil)
+		err = NewValidatingWebhook(context.Background(), webhook.Name, testNamespace, testVersion, testMeshName, enableReconciler, validateTrafficTarget, certManager, kube, policyClient)
```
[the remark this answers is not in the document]
steeling: done

pkg/webhook/server.go
```diff
@@ -0,0 +1,102 @@
...
+	}
+	return s, nil
+}
+
+// Run actually starts the server. It blocks until the passed in context is done.
+func (s *Server) Run(ctx context.Context) {
+	log.Info().Msgf("Starting conversion webhook server on: %s", s.server.Addr)
+	go func() {
+		err := s.server.ListenAndServeTLS("", "") // err is always non-nil
+		log.Error().Err(err).Msg("crd-converter webhook HTTP server failed to start")
```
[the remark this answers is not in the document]
steeling: done

charts/osm/templates/osm-bootstrap-deployment.yaml
```diff
@@ -74,8 +74,6 @@ spec:
               containerPort: 9443
             - name: "metrics"
               containerPort: 9091
-            - name: "health"
```
keithmattix, line 77: I get cutting down on unnecessary ports, but using the metrics port for healthchecks feels weird to me

cmd/osm-controller/osm-controller.go
```diff
@@ -255,7 +255,7 @@ func main() {
 
 	clientset := extensionsClientset.NewForConfigOrDie(kubeConfig)
 
-	if err := validator.NewValidatingWebhook(validatorWebhookConfigName, osmNamespace, osmVersion, meshName, enableReconciler, validateTrafficTarget, constants.ValidatorWebhookPort, certManager, kubeClient, policyController, stop); err != nil {
+	if err := validator.NewValidatingWebhook(ctx, validatorWebhookConfigName, osmNamespace, osmVersion, meshName, enableReconciler, validateTrafficTarget, certManager, kubeClient, policyController); err != nil {
```
keithmattix, line 258: +1 for more context usage 😄

pkg/cli/verifier/control_plane.go
```diff
@@ -48,9 +47,9 @@ func NewControlPlaneHealthVerifier(stdout io.Writer, stderr io.Writer, kubeClien
 		bootstrapProber: &podProber{
 			kubeClient: kubeClient,
 			restConfig: restConfig,
-			port:       crdconversion.HealthzPort,
+			port:       constants.CRDConversionWebhookPort,
 			path:       constants.WebhookHealthPath,
-			protocol:   constants.ProtocolHTTP,
+			protocol:   constants.ProtocolHTTPS,
```
keithmattix, line 52: I'm not 100% sure what this is being used for, but OSM doesn't support protocol HTTPS

pkg/webhook/server.go
```diff
@@ -0,0 +1,104 @@
...
+	go func() {
+		err := s.server.ListenAndServeTLS("", "") // err is always non-nil
+		log.Error().Err(err).Msgf("%s webhook HTTP server shutdown")
+	}()
+
+	// Wait on exit signals
+	<-ctx.Done()
+
+	// Stop the servers
+	if err := s.server.Shutdown(ctx); err != nil {
```
keithmattix, line 74: You're passing a canceled context into `Shutdown` here; I'm not sure what the implications of that are.

pkg/webhook/server.go
```diff
@@ -0,0 +1,104 @@
...
+		Addr:    fmt.Sprintf(":%d", port),
+		Handler: mux,
+		// #nosec G402
+		TLSConfig: &tls.Config{
+			GetCertificate: func(*tls.ClientHelloInfo) (*tls.Certificate, error) {
+				s.mu.Lock()
+				defer s.mu.Unlock()
+				return &s.cert, nil
+			},
+			MinVersion: tls.VersionTLS13,
```
keithmattix, line 52: Are we using this minVersion elsewhere? If any client doesn't support TLS 1.3, then the handshake will fail. I'd say 1.2 is probably safer.

pkg/cli/verifier/control_plane.go
steeling: health check and metrics do

pkg/webhook/server.go
steeling: This is what was used previously, I don't think it fits in the scope of this PR to make that change

pkg/webhook/server.go
steeling: good catch! fixed

charts/osm/templates/osm-bootstrap-deployment.yaml
keithmattix: Fair enough; yeah meta or admin makes sense to me, but I won't block this PR on it

nojnhuh (approved):

jaellio (approved):

Issue #682, opened 2020-05-20
Use custom image for Envoy container with necessary debugging tools or a debug sidecar for debugging


shashankram:
For debugging traffic issues, tcpdump/tshark is useful. Add this to the sidecar image to help debug traffic issues without mutating the pods.

phillipgibson:
[The OSM project has been officially archived by the CNCF.](https://github.com/cncf/toc/pull/1044) There will be no more new development on any repo under the OpenServiceMesh organization.

PR #4834, opened 2022-06-20
envoy: update to latest version and fix typed proto usage

shashankram:
**Description**:
envoy: update to latest version and fix typed proto usage

- Updates Envoy to its latest available version
  (v1.22.2 for Linux, v1.22.1 for Windows).
  The latest version includes the latest released
  security fix.
  We could not update Envoy previously due to a blocking bug:
  https://github.com/envoyproxy/envoy/issues/20113

- Updates filter names to custom names as wellknown
  names are deprecated in Envoy (with 1 exception
  for the http.rbac filter). Envoy will use the
  TypeURL in the proto to determine which filter
  to use instead. Wellknown names are not required
  and using them is confusing because not all filters
  are defined in the legacy wellknown pkg (e.g.
  http.local_ratelimit).
  See:
  https://github.com/envoyproxy/envoy/issues/21759
  https://github.com/envoyproxy/envoy/pull/21763
  https://github.com/envoyproxy/go-control-plane/issues/293
  https://github.com/envoyproxy/go-control-plane/issues/552

- Uses the distroless image as the alpine image has been
  discontinued: https://github.com/envoyproxy/envoy/issues/21758

- Updates tests to use custom filter names

- Adds `proto_types.go` to aid dynamic proto resolution
  for typed configs using `any.Any()`. This helps resolve
  protos where dynamic resolution is necessary.

- Also adds the typed proto for rate limiting to the http_filters,
  without which Envoy will not be able to perform rate limiting
  of HTTP traffic.

- Updates Prometheus' ConfigMap to reflect changes to
  Envoy metrics prefixes

- Updates Contour's envoy and control plane image
   so they are compatible.

**Testing done**:
CI, unit tests

**Affected area**:

4. Has documentation corresponding to this change been updated in the [osm-docs](https://github.com/openservicemesh/osm-docs) repo (if applicable)? `will be updated once this change is merged`


keithmattix:
The distroless envoy image scares me a bit; I understand Envoy dropped support for alpine (which is kind of shocking to me tbh), but I wonder about users looking to debug their envoy sidecars who now are left without a shell

shashankram:
> The distroless envoy image scares me a bit; I understand Envoy dropped support for alpine (which is kind of shocking to me tbh), but I wonder about users looking to debug their envoy sidecars who now are left without a shell

That's a fair concern that would be better addressed by resolving https://github.com/openservicemesh/osm/issues/682, similar to [Linkerd's approach](https://linkerd.io/2.10/tasks/using-the-debug-container/).

jaellio (approved):

nojnhuh (approved):

shashankram:
> > > The distroless envoy image scares me a bit; I understand Envoy dropped support for alpine (which is kind of shocking to me tbh), but I wonder about users looking to debug their envoy sidecars who now are left without a shell
> > 
> > 
> > That's a fair concern that would be better addressed by resolving #682, similar to [Linkerd's approach](https://linkerd.io/2.10/tasks/using-the-debug-container/).
> 
> Will #682 be a part of v1.2?

No it won't, but I think that would be the best way to go for debugging purpose, vs bundling all tools into the default Envoy image. Alternatively, we could use the Envoy Ubuntu image, but that goes against the idea to use security hardened images. We already use distroless images for the control planes, so I am fine to use the same for Envoy.

PR #4837, opened 2022-06-21
use friendlier defaults for egress and permissive mode

steeling:
Set enableEgress and permissive mode to default to true in values.yaml


shashankram (approved):

shashankram:
@steeling this may need a rebase as I merged https://github.com/openservicemesh/osm/pull/4834 which also updated the values.yaml file.

PR #4841, opened 2022-06-22
[RFC] feat(k8s): stop sidecars on otherwise completed Jobs

nojnhuh:
**Description**:

This change adds a new long-lived goroutine to osm-controller to explicitly stop sidecar containers on Pods controlled by Jobs. Currently, Jobs never complete because the sidecars run indefinitely. This is the high-level approach:

- For any `pod-updated` event:
  - If the Pod does not belong to a Job, ignore it
  - If the Pod's status indicates the main containers are still running, ignore it
  - For each running sidecar, send `SIGTERM` to PID 1 with the equivalent of `kubectl exec <pod> -c <container> -- kill 1`
    - Based on how https://hub.docker.com/r/zachomedia/kubernetes-sidecar-terminator works

This handles both `envoy` sidecars and `osm-healthcheck` sidecars (for Pods with `tcpSocket` health probes).

The main restriction this introduces is that any sidecar we intend to inject must include a `kill` command. Envoy's images already include busybox which provides a `kill` command. These changes include changing the base image for `osm-healthcheck` to busybox to give it a `kill` command.

Part of #2316

**Testing done**:

**Affected area**:

2. Is this a breaking change?

### DIFF
--- a/pkg/k8s/util.go
+++ b/pkg/k8s/util.go
@@ -1,6 +1,7 @@
 package k8s
 
 import (
+	"bytes"
 	"fmt"
 	"strings"
 
@@ -9,7 +10,14 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/remotecommand"
 
+	"github.com/openservicemesh/osm/pkg/announcements"
+	"github.com/openservicemesh/osm/pkg/constants"
+	"github.com/openservicemesh/osm/pkg/k8s/events"
+	"github.com/openservicemesh/osm/pkg/messaging"
 	"github.com/openservicemesh/osm/pkg/service"
 )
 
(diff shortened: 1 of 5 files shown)


steeling:
What are your thoughts on the wrapping the sidecar proxy (and healthcheck) into a single container. That container could contain a startup script that additionally triggers this, except only watches itself.

This provides a nice platform for us to add more logic into sidecars, hides the healthcheck container, and has a more self-contained mechanism for killing a container.

It scares me a bit that a bug in that code could result in unwanted killing of containers.

pkg/k8s/util.go
```diff
@@ -178,3 +186,79 @@ func NamespacedNameFrom(name string) (types.NamespacedName, error) {
 func IsHeadlessService(svc corev1.Service) bool {
 	return len(svc.Spec.ClusterIP) == 0 || svc.Spec.ClusterIP == corev1.ClusterIPNone
 }
+
+func StopJobSidecars(stop <-chan struct{}, broker *messaging.Broker, config *rest.Config, clientset kubernetes.Interface) {
```
steeling, line 190: nit: we should avoid passing client objects as params when we're not doing so for dependency injection.

Instead, we should maintain a higher level client object (k8s.Controller, or something similar), and place this as a method on that type, which has access to the underlying (dependency injected client).

This maintains a cleaner approach for decoupling from API's and leveraging interfaces.

See https://docs.google.com/document/d/1L_u-XlVU8Fog-uFS3Uf0O2nz8IArQy82cd6Npxm038Y/edit#heading=h.cvqx4tbzt6i7

nojnhuh:
Before putting everything into one sidecar, I got this working as a separate sidecar like osm-healthcheck. I've come across two limitations with this approach that would also affect if everything was in a single sidecar container:
- The Pods' RBAC needs to allow listing/watching Pods to tell when the main containers are done
- Since Envoy redirection is applied to this new sidecar, the sidecar needs to be allowed to communicate with the API server, either by enabling egress or adding an outbound IP range exclusion, or there might be a way to make this more automatic by changing the iptables rules for this specific use case.

We could perhaps get around communicating externally at all if we instead set [`shareProcessNamespace`](https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/pod-v1/#hosts-namespaces) on Pods and inspect the processes directly. The main limitation there is that it would be impossible for the sidecar to know exactly what the "main" processes are it should wait on. Since there's no real coordination of execution among a Pod's containers, it seems possible that the "main" containers' processes could start and finish before the sidecar has a chance to see what's running in the rest of the Pod. Inspecting the container runtime on the node is probably a better version of this approach, but that is probably similar to the proposed changes to upstream Kubernetes mentioned in #2316 and would probably be outside the scope of something like OSM already running on top of Kubernetes.

Overall, I'm getting the sense the approach the changes on this PR show is simpler than these alternatives.

steeling:
> Before putting everything into one sidecar, I got this working as a separate sidecar like osm-healthcheck. I've come across two limitations with this approach that would also affect if everything was in a single sidecar container:
> 
> * The Pods' RBAC needs to allow listing/watching Pods to tell when the main containers are done
> * Since Envoy redirection is applied to this new sidecar, the sidecar needs to be allowed to communicate with the API server, either by enabling egress or adding an outbound IP range exclusion, or there might be a way to make this more automatic by changing the iptables rules for this specific use case.
> 
> We could perhaps get around communicating externally at all if we instead set [`shareProcessNamespace`](https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/pod-v1/#hosts-namespaces) on Pods and inspect the processes directly. The main limitation there is that it would be impossible for the sidecar to know exactly what the "main" processes are it should wait on. Since there's no real coordination of execution among a Pod's containers, it seems possible that the "main" containers' processes could start and finish before the sidecar has a chance to see what's running in the rest of the Pod. Inspecting the container runtime on the node is probably a better version of this approach, but that is probably similar to the proposed changes to upstream Kubernetes mentioned in #2316 and would probably be outside the scope of something like OSM already running on top of Kubernetes.
> 
> Overall, I'm getting the sense the approach the changes on this PR show is simpler than these alternatives.

Ya I agree with your conclusions on leveraging the control plane then.

I wonder if it would be even easier to provide thorough documentation for how users can fix this themselves. Something that just hits the proxy's `quitquitquit` page when their app is done. It is a common enough problem though.

We really just need something to tie us over until sidecars get more love from k8s (could be another couple years though)

keithmattix:
So I implemented the quitquitquit solution with a Linux trap when dealing with a different service mesh. It worked pretty well as long as the main container had curl available. It even had retries and timeouts since curl supports it natively

nojnhuh:
> I wonder if it would be even easier to provide thorough documentation for how users can fix this themselves. Something that just hits the proxy's `quitquitquit` page when their app is done. It is a common enough problem though.

> So I implemented the quitquitquit solution with a Linux trap when dealing with a different service mesh. It worked pretty well as long as the main container had curl available. It even had retries and timeouts since curl supports it natively

Trying out `trap`, it's really simple to set up when `sh` is running everything else. I imagine implementing the equivalent functionality should be almost as easy no matter how the container is built and run. Even without curl and using something like `nc` (which is in busybox) you can hand craft the 2-line HTTP request.

The `osm-healthcheck` container doesn't have an HTTP endpoint to stop it, but that would be easy enough to add so it could be stopped similarly. We'd likely want to move that to a different port from the `/osm-healthcheck` endpoint which kubelet interacts with since that has a hole punched for it in iptables, so then the new quit endpoint would only be accessible from `localhost`.

I think documenting this process for now is a good first step to unblock the use case, so I'll close this PR and follow up in #2316.

PR #4853, opened 2022-06-28
docs: update SUPPORT with OSM and K8s version support

zr-msft:
Signed-off-by: Zach Rhoads <zachary.rhoads@microsoft.com>

**Description**: Provide details on which versions of OSM are supported on which versions of Kubernetes in support file.

**Testing done**: Verified file displays correctly.

**Affected area**:
| Documentation              | [ x ] |

3. Has documentation corresponding to this change been updated in the [osm-docs](https://github.com/openservicemesh/osm-docs) repo (if applicable)? yes

### DIFF
--- a/SUPPORT
+++ b/SUPPORT
@@ -1,3 +1,7 @@
 # Support
 
 [Please search open issues on GitHub](https://github.com/openservicemesh/osm/issues), and if your issue isn't already represented please [open a new one](https://github.com/openservicemesh/osm/issues/new/choose). The OSM project maintainers will respond to the best of their abilities.
+
+## Supported versions of OSM
+
+The OSM project maintainers will respond to the best of their abilities to issues on the current and two previous minor versions of OSM running on Kubernetes v1.22.9 or greater.
--- a/SUPPORT
+++ b/SUPPORT
@@ -4,4 +4,4 @@
 
 ## Supported versions of OSM
 
-The OSM project maintainers will respond to the best of their abilities to issues on the current and two previous minor versions of OSM running on Kubernetes v1.22.9 or greater.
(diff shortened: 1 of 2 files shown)


SUPPORT
```diff
@@ -1,3 +1,7 @@
 # Support
 
-[Please search open issues on GitHub](https://github.com/openservicemesh/osm/issues), and if your issue isn't already represented please [open a new one](https://github.com/openservicemesh/osm/issues/new/choose). The OSM project maintainers will respond to the best of their abilities.
\ No newline at end of file
+[Please search open issues on GitHub](https://github.com/openservicemesh/osm/issues), and if your issue isn't already represented please [open a new one](https://github.com/openservicemesh/osm/issues/new/choose). The OSM project maintainers will respond to the best of their abilities.
+
+## Supported versions of OSM
+
+The OSM project maintainers will respond to the best of their abilities to issues on the current and two previous minor versions of OSM running on Kubernetes v1.22.9 or greater.
```
nojnhuh, line 7: Is there a way we can phrase this so we don't need to have a specific version listed here that we need to manually keep in sync with Chart.yaml?

zr-msft: Ideally yes, @nojnhuh, but this is consistent with the README:
https://github.com/openservicemesh/osm/blob/main/README.md?plain=1#L85

We could phrase both these lines to say something like:
> ... running on [active Kubernetes releases](https://kubernetes.io/releases/).

but if the support strayed from the linked page it would cause issues. For example, we say 1.22.9 but 1.22.1 is listed on the page.

Can we include updating these two MD files as well as the relevant files in osm-docs when the chart is updated? I understand its less than ideal and extra work, but it does allow us to be specific about the K8s versions we support

nojnhuh: Can we point to this in both places? https://github.com/openservicemesh/osm/blob/c8d7559b8303f8df8da52dcf8d050600d7826a3e/charts/osm/Chart.yaml#L24

jsturtevant: makes sense to link to the code here so they don't get out of sync.  @zr-msft any objections?

zr-msft: @jsturtevant @nojnhuh I've updated the PR to incorporate your feedback. If it looks good to you, feel free to merge

keithmattix (approved):

jaellio (approved):

Issue #4818, opened 2022-06-14
Update OSM bootstrap secret on certificate manager updates


jaellio:
**Please describe the Improvement and/or Feature Request**

For each manager update, issue a new certificate and update the OSM bootstrap secret if it hasn't already been updated. 

To do:
- [x] Move envoy bootstrap config creation logic to envoy/bootstrap package https://github.com/openservicemesh/osm/commit/226ee6499208fb2871f77660208b778661db9652
- [x] Implement builder pattern for envoy bootstrap config
- [ ] Subscribe to bootstrap certificate rotations on the cert rotation chan and update the respective envoy bootstrap config secret 

**Scope (please mark with X where applicable)**
- New Functionality          [x]
- Certificate Management     [x]
- Sidecar Injection          [x]

PR #4858, opened 2022-06-29
Feature/bootstrap

steeling:
Remove redundant bootstrap config, and leverage a [builder pattern](https://docs.google.com/document/d/1L_u-XlVU8Fog-uFS3Uf0O2nz8IArQy82cd6Npxm038Y/edit) we plan on implementing.

Part of https://github.com/openservicemesh/osm/issues/4818

### DIFF
--- a/pkg/envoy/bootstrap/config_test.go
+++ b/pkg/envoy/bootstrap/config_test.go
@@ -1,47 +1,34 @@
 package bootstrap
 
 import (
-	"fmt"
-	"io/ioutil"
-	"path"
-	"path/filepath"
-
-	xds_bootstrap "github.com/envoyproxy/go-control-plane/envoy/config/bootstrap/v3"
-	. "github.com/onsi/ginkgo"
-	. "github.com/onsi/gomega"
-
-	"github.com/golang/mock/gomock"
-
 	"testing"
-
-	"github.com/openservicemesh/osm/pkg/apis/config/v1alpha2"
-	"github.com/openservicemesh/osm/pkg/configurator"
 
 	tassert "github.com/stretchr/testify/assert"
 
 	tresorFake "github.com/openservicemesh/osm/pkg/certificate/providers/tresor/fake"
-	"github.com/openservicemesh/osm/pkg/constants"
 	"github.com/openservicemesh/osm/pkg/models"
 	"github.com/openservicemesh/osm/pkg/utils"
 )
 
-func TestBuildFromConfig(t *testing.T) {
+func TestBuild(t *testing.T) {
 	assert := tassert.New(t)
 	cert := tresorFake.NewFakeCertificate()
 
-	config := Config{
+	b := &Builder{
 		NodeID:                cert.GetCommonName().String(),
-		AdminPort:             15000,
-		XDSClusterName:        constants.OSMControllerName,
 		XDSHost:               "osm-controller.osm-system.svc.cluster.local",
-		XDSPort:               15128,
 		TLSMinProtocolVersion: "TLSv1_0",
 		TLSMaxProtocolVersion: "TLSv1_2",
 		CipherSuites:          []string{"abc", "xyz"},
 		ECDHCurves:            []string{"ABC", "XYZ"},
+		OriginalHealthProbes: models.HealthProbes{
+			Liveness:  &models.HealthProbe{Path: "/liveness", Port: 81, IsHTTP: true},
+			Readiness: &models.HealthProbe{Path: "/readiness", Port: 82, IsHTTP: true},
+			Startup:   &models.HealthProbe{Path: "/startup", Port: 83, IsHTTP: true},
+		},
 	}
 
-	bootstrapConfig, err := BuildFromConfig(config)
+	bootstrapConfig, err := b.Build()
 	assert.Nil(err)
 	assert.NotNil(bootstrapConfig)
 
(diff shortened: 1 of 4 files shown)


pkg/envoy/bootstrap/types.go
```diff
@@ -8,20 +8,11 @@ import (
...
-// Config is the type used to represent the information needed to build the Envoy bootstrap config
-type Config struct {
-	// Admin port is the Envoy admin port
-	AdminPort uint32
-
-	// XDSClusterName is the name of the XDS cluster to connect to
-	XDSClusterName string
-
+// Builder is the type used to build the Envoy bootstrap config.
+type Builder struct {
```
draychev, line 12: I think it would increase readability if we renamed `Builder` to `EnvoyBootstrapConfigBuilder`? This to me provides instant context and I understand what this struct and functions will do.

pkg/envoy/bootstrap/config_test.go
```diff
@@ -115,128 +102,199 @@ static_resources:
         '@type': type.googleapis.com/envoy.extensions.upstreams.http.v3.HttpProtocolOptions
         explicit_http_config:
           http2_protocol_options: {}
+  - load_assignment:
```
draychev, line 105: Thanks for adding tests for the health probes. At some point these YAMLs get unwieldy (I'm guilty of starting this pattern). So we may need to rethink this or chunk it into smaller YAMLs perhaps...

pkg/injector/bootstrap.go
```diff
@@ -53,7 +50,7 @@ func (wh *mutatingWebhook) createEnvoyBootstrapConfig(name, namespace, osmNamesp
 		CipherSuites:          wh.configurator.GetMeshConfig().Spec.Sidecar.CipherSuites,
 		ECDHCurves:            wh.configurator.GetMeshConfig().Spec.Sidecar.ECDHCurves,
 	}
-	config, err := bootstrap.GenerateEnvoyConfig(configMeta, wh.configurator)
+	config, err := builder.Build()
```
draychev, line 53: The way this line is rewritten - it carries much less semantic meaning.
Looking at it (without seeing what was replaced) - I see that a builder builds config. But all 3 of these words are so generic - it carries little meaning.
I'd propose we rename `config` to `envoyBootstrapConfig` and `builder.Build()` to `bootstrapConfig.Build()` or something of that nature.

draychev (commented):
I left some minor comments around naming. I really care about readability. I think readability makes the project more approachable by new folks wanting to contribute. (And the rest of us reading this new code)

This looks great. Thanks for spearheading the effort. The effort you describe in [the document here](https://docs.google.com/document/d/1L_u-XlVU8Fog-uFS3Uf0O2nz8IArQy82cd6Npxm038Y/edit#) is ambitions and much needed. @eduser25 had started similar efforts and I have a few docs as well.  I'd love to join the refactoring party.

I'd love to see more details in[ the doc](https://docs.google.com/document/d/1L_u-XlVU8Fog-uFS3Uf0O2nz8IArQy82cd6Npxm038Y/edit#) around what specific changes you want to make.  We could then divide and conquer. I'd love to help!

pkg/envoy/bootstrap/types.go
trstringer: My vote would be to keep `Builder`. The import "path" includes "envoy" and "bootstrap", so that context is contained there and duplicating it doesn't seem to provide any benefit.

pkg/injector/bootstrap.go
steeling: changed to bootstrapConfig. 

[I'll point out that go explicitly prefers short variable names when it is used in a limited scope](https://github.com/golang/go/wiki/CodeReviewComments#variable-names)

In this case, the method config, and builder both appear within a few lines of the function name `createEnvoyBootstrapConfig`, and the instantion of the builder: `builder := bootstrap.Builder{...`

I don't think it's too important for a single line to have that much context, but too quickly skim a handful of lines and understand what is happening.

Anyways, just wanted to add that as a counter point, I've changed to `bootstrapConfig` which should be a happy medium

steeling:
> I left some minor comments around naming. I really care about readability. I think readability makes the project more approachable by new folks wanting to contribute. (And the rest of us reading this new code)
> 
> This looks great. Thanks for spearheading the effort. The effort you describe in [the document here](https://docs.google.com/document/d/1L_u-XlVU8Fog-uFS3Uf0O2nz8IArQy82cd6Npxm038Y/edit#) is ambitions and much needed. @eduser25 had started similar efforts and I have a few docs as well. I'd love to join the refactoring party.
> 
> I'd love to see more details in[ the doc](https://docs.google.com/document/d/1L_u-XlVU8Fog-uFS3Uf0O2nz8IArQy82cd6Npxm038Y/edit#) around what specific changes you want to make. We could then divide and conquer. I'd love to help!

Sure, I'll be creating some specific bugs soon which will have more detail!

pkg/injector/bootstrap.go
```diff
@@ -53,12 +50,12 @@ func (wh *mutatingWebhook) createEnvoyBootstrapConfig(name, namespace, osmNamesp
 		CipherSuites:          wh.configurator.GetMeshConfig().Spec.Sidecar.CipherSuites,
 		ECDHCurves:            wh.configurator.GetMeshConfig().Spec.Sidecar.ECDHCurves,
 	}
-	config, err := bootstrap.GenerateEnvoyConfig(configMeta, wh.configurator)
+	bootstrapConfig, err := builder.Build()
```
draychev, line 53: @steeling Is it fair to say that `builder` in this context is specif to Envoy Bootstrap config?
It is not a generic builder - it always only builds config for Envoy?
If that's the case - I'd love to see `builder` renamed to something that is a bit higher specificity. It could be `envoyBootstrapBuilder` or `bootstrapBuilder`.  I can imagine having multiple builders in the same file/context and this would disambiguate.

draychev (changes requested):
It'd be good to be more specific on `builder` and name it so it indicates what it's expected to build.

pkg/injector/bootstrap.go
steeling: it's a few lines away from the instantiation,`builder := bootstrap.Builder`.

Did you see my prior comment above? Copying here:

> [I'll point out that go explicitly prefers short variable names when it is used in a limited scope](https://github.com/golang/go/wiki/CodeReviewComments#variable-names)
> 
> In this case, the method config, and builder both appear within a few lines of the function name createEnvoyBootstrapConfig, and the instantion of the builder: builder := bootstrap.Builder{...
> 
> I don't think it's too important for a single line to have that much context, but too quickly skim a handful of lines and understand what is happening.
> 
> Anyways, just wanted to add that as a counter point, I've changed to bootstrapConfig which should be a happy medium

draychev (approved):
Thanks for starting this refactor @steeling

jaellio:
@draychev thank you for the review! Just a reminder to clean up/consolidate the commit message before merging

Issue #4839, opened 2022-06-21
Support updating TLS config without restarting server


jaellio:
**Please describe the Improvement and/or Feature Request**

On each request, get the tls config using [getConfigForClient](https://pkg.go.dev/go.step.sm/crypto/tlsutil#Renewer.GetConfigForClient). The tls config returned will contain a certificate issued using the up-to-date manager struct issuer settings. This change will allow OSM to update certificates used in the TLS without having to restart the servers. This helps facilitate root certificate rotation with minimal downtime.

The change will impact the webhook servers and the ads server. 

To do:
Webhooks
- [x] Abstract webhook server creation logic
- [x] Subscribe to webhook certificate updates on the cert rotation chan in the certificate manager. Update the server certificate and the webhook configuration

ADS
- [ ] Refactor ADS server creation. Update the tls config to use a dynamic tls config. Subscribe to ADS server certificate updates on the cert rotation chan in the certificate manager. Update the tls config with the new certificate for client and service to utilize mTLS.

**Scope (please mark with X where applicable)**
- New Functionality          [x]
- Certificate Management     [x]

Issue #4847, opened 2022-06-23
Fix head of line blocking in pubsub


steeling:
With the pubsub used in the message broker, we allow for many publishers and many subscribers. In the current implementation, if a subscriber fails to read from its subscribed channel (or is slow to do so), it blocks publishers from being able to publish. In the worst case, a backed up subscriber can prevent certificates from being rotated, or block updates to proxies.

This can essentially take down a single osm-controller. We can degrade (more) gracefully by instead leveraging an unbounded queue when publishing, prior to reading publishing onto the 

While we're at it, we should hide the internals of the pubsub mechanism from the broker's API calls, and simplify using channels. An API may look like:

```go

func (b* Broker) PublishCertEvent(e Event, topics ...string) {
   ...
} 

// Named return params just to show what the unsub func is used for.
func (b* Broker SubscribeCertEvents(topics ...string) (stream ch <- Event, unsub func()) {
   ...
}
```

This still prevents updates from being published, but at least frees up routines that are doing the publishing. We should look into other mechanisms to further degrade gracefully. Some may include:

1. Removing/reducing usage of pubsub
2. Having a policy of no blocking calls on any routine that subscribes to a pubsub topic.
3. Use an unbounded queue on the subscriber end. If we do this, we may want to further investigate message coalescing.

keithmattix:
Still valid

PR #4868, opened 2022-06-30
Consolidate cert rotation chan into cert manager

steeling:
The biggest change is moving the cert rotation pubsub to the cert manager itself, and changing the semantics to subscribe to single certs. This will make future changes that listen for cert rotations much simpler. 

The motivation for this is that every time we subscribe to cert rotation events we filter against the cert CommonName we are interested in.

In addition, with upcoming changes to cert rotation there are several places we need to add a messaging.Broker just to get access to the rotations. This, to me, was a code smell, and meant that the rotations access should be on the manager itself.

In addition the following bugs were fixed

This also fixes the following bugs:
1. Use TrustedCAs instead of IssuingCA where appropriate
2. Prevent missed rotations due to subscribing after IssueCertificate has been called.

Part of https://github.com/openservicemesh/osm/issues/4839

I've added some comments to this PR on specific code lines to provide context of the current change vs the prior context

### DIFF
--- a/pkg/certificate/manager.go
+++ b/pkg/certificate/manager.go
@@ -7,12 +7,11 @@ import (
 	"sync"
 	"time"
 
-	"github.com/openservicemesh/osm/pkg/announcements"
+	"github.com/cskr/pubsub"
+
 	"github.com/openservicemesh/osm/pkg/constants"
 	"github.com/openservicemesh/osm/pkg/errcode"
-	"github.com/openservicemesh/osm/pkg/k8s/events"
 	"github.com/openservicemesh/osm/pkg/logger"
-	"github.com/openservicemesh/osm/pkg/messaging"
 )
 
 var (
@@ -20,11 +19,11 @@ var (
 )
 
 // NewManager creates a new CertificateManager with the passed MRCClient and options
-func NewManager(ctx context.Context, mrcClient MRCClient, getServiceCertValidityPeriod func() time.Duration, getIngressCertValidityDuration func() time.Duration, msgBroker *messaging.Broker, checkInterval time.Duration) (*Manager, error) {
+func NewManager(ctx context.Context, mrcClient MRCClient, getServiceCertValidityPeriod func() time.Duration, getIngressCertValidityDuration func() time.Duration, checkInterval time.Duration) (*Manager, error) {
 	m := &Manager{
 		serviceCertValidityDuration: getServiceCertValidityPeriod,
 		ingressCertValidityDuration: getIngressCertValidityDuration,
-		msgBroker:                   msgBroker,
+		pubsub:                      pubsub.New(1),
 	}
 
 	err := m.start(ctx, mrcClient)
@@ -312,11 +311,7 @@ func (m *Manager) issueCertificate(prefix string, ct CertType, opts ...IssueOpti
 
 	if rotate {
 		// Certificate was rotated
-		m.msgBroker.GetCertPubSub().Pub(events.PubSubMessage{
-			Kind:   announcements.CertificateRotated,
-			NewObj: newCert,
-			OldObj: cert,
-		}, announcements.CertificateRotated.String())
+		m.pubsub.Pub(newCert, prefix)
 
 		log.Debug().Msgf("Rotated certificate (old SerialNumber=%s) with new SerialNumber=%s", cert.SerialNumber, newCert.SerialNumber)
 	}
@@ -338,4 +333,18 @@ func (m *Manager) ListIssuedCertificates() []*Certificate {
 		return true // continue the iteration
 	})
 	return certs
+}
+
+// SubscribeRotations returns a channel that outputs every certificate that is rotated by the manager.
+// The caller must call the returned method to close the channel.
+// WARNING: you cannot call wait on the returned channel on the same go routine you are issuing a certificate on.
+func (m *Manager) SubscribeRotations(key string) (chan interface{}, func()) {
+	ch := m.pubsub.Sub(key)
+	return ch, func() {
+		go m.pubsub.Unsub(ch)
+		// must empty the channel to prevent deadlock
+		// https://github.com/openservicemesh/osm/issues/4847
+		for range ch {
+		}
+	}
 }
--- a/pkg/certificate/manager_test.go
+++ b/pkg/certificate/manager_test.go
@@ -378,6 +378,11 @@ func TestSubscribeRotations(t *testing.T) {
 		pubsub:           pubsub.New(0),
 	}
 
+	ch1, unsub1 := cm.SubscribeRotations(cnPrefix1)
+	ch2, unsub2 := cm.SubscribeRotations(cnPrefix2)
+	defer unsub1()
+	defer unsub2()
+
 	cert, err := cm.IssueCertificate(cnPrefix1, Service)
 	assert.NoError(err)
 	assert.Equal("fake-cert-cn1.fake1.domain.com", cert.GetCommonName().String())
@@ -385,11 +390,6 @@ func TestSubscribeRotations(t *testing.T) {
 	cert, err = cm.IssueCertificate(cnPrefix2, Service)
 	assert.NoError(err)
 	assert.Equal("fake-cert-cn2.fake1.domain.com", cert.GetCommonName().String())
-
-	ch1, unsub1 := cm.SubscribeRotations(cnPrefix1)
-	ch2, unsub2 := cm.SubscribeRotations(cnPrefix2)
-	defer unsub1()
-	defer unsub2()
 
 	var wg sync.WaitGroup
 	wg.Add(1)
--- a/pkg/certificate/manager_test.go
+++ b/pkg/certificate/manager_test.go
@@ -2,18 +2,18 @@ package certificate
 
 import (
 	"context"
+	"sync"
 	"testing"
 	time "time"
 
+	"github.com/cskr/pubsub"
 	tassert "github.com/stretchr/testify/assert"
 	trequire "github.com/stretchr/testify/require"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
-	"github.com/openservicemesh/osm/pkg/announcements"
 	"github.com/openservicemesh/osm/pkg/apis/config/v1alpha2"
 	"github.com/openservicemesh/osm/pkg/certificate/pem"
 	"github.com/openservicemesh/osm/pkg/constants"
-	"github.com/openservicemesh/osm/pkg/messaging"
 )
 
 func TestShouldRotate(t *testing.T) {
@@ -85,13 +85,13 @@ func TestRotor(t *testing.T) {
 
 	stop := make(chan struct{})
 	defer close(stop)
-	msgBroker := messaging.NewBroker(stop)
-	certManager, err := NewManager(context.Background(), &fakeMRCClient{}, getServiceCertValidityPeriod, getIngressGatewayCertValidityPeriod, msgBroker, 5*time.Second)
+	certManager, err := NewManager(context.Background(), &fakeMRCClient{}, getServiceCertValidityPeriod, getIngressGatewayCertValidityPeriod, 5*time.Second)
 	require.NoError(err)
 
 	certA, err := certManager.IssueCertificate(cnPrefix, Service)
 	require.NoError(err)
-	certRotateChan := msgBroker.GetCertPubSub().Sub(announcements.CertificateRotated.String())
+	certRotateChan, unsub := certManager.SubscribeRotations(cnPrefix)
+	defer unsub()
 
 	// Wait for two certificate rotations to be announced and terminate
 	<-certRotateChan
@@ -182,7 +182,6 @@ func TestIssueCertificate(t *testing.T) {
 
 	stop := make(chan struct{})
 	defer close(stop)
-	msgBroker := messaging.NewBroker(stop)
 
 	t.Run("single key issuer", func(t *testing.T) {
 		cm := &Manager{
@@ -190,7 +189,7 @@ func TestIssueCertificate(t *testing.T) {
 			// The root certificate signing all newly issued certificates
 			signingIssuer:    &issuer{ID: "id1", Issuer: &fakeIssuer{id: "id1"}, CertificateAuthority: pem.RootCertificate("id1"), TrustDomain: "fake1.domain.com"},
 			validatingIssuer: &issuer{ID: "id1", Issuer: &fakeIssuer{id: "id1"}, CertificateAuthority: pem.RootCertificate("id1"), TrustDomain: "fake2.domain.com"},
-			msgBroker:        msgBroker,
+			pubsub:           pubsub.New(0),
 		}
 		// single signingIssuer, not cached
 		cert1, err := cm.IssueCertificate(cnPrefix, Service)
@@ -228,7 +227,7 @@ func TestIssueCertificate(t *testing.T) {
 			// The root certificate signing all newly issued certificates
 			signingIssuer:    &issuer{ID: "id1", Issuer: &fakeIssuer{id: "id1"}, CertificateAuthority: pem.RootCertificate("id1"), TrustDomain: "fake1.domain.com"},
 			validatingIssuer: &issuer{ID: "id2", Issuer: &fakeIssuer{id: "id2"}, CertificateAuthority: pem.RootCertificate("id2"), TrustDomain: "fake2.domain.com"},
-			msgBroker:        msgBroker,
+			pubsub:           pubsub.New(0),
 		}
 
 		// Not cached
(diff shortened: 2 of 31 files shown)


pkg/ingress/gateway.go
```diff
@@ -22,61 +21,14 @@ import (
...
 //    updates/removes the certificate and secret as necessary.
-func (c *client) provisionIngressGatewayCert(stop <-chan struct{}) error {
-	defaultCertSpec := c.cfg.GetMeshConfig().Spec.Certificate.IngressGateway
-	if defaultCertSpec != nil {
-		// Issue a certificate for the default certificate spec
-		if err := c.createAndStoreGatewayCert(*defaultCertSpec); err != nil {
-			return errors.Wrap(err, "Error provisioning default ingress gateway cert")
-		}
-	}
+func (c *client) provisionIngressGatewayCert(stop <-chan struct{}) {
```
steeling, line 24: There was a bug here before, where a rotation could occur prior to the subscription starting, which would be missed.

pkg/ingress/gateway.go
```diff
@@ -17,66 +16,10 @@ import (
...
-// provisionIngressGatewayCert does the following:
-// 1. If an ingress gateway certificate spec is specified in the MeshConfig resource, issues a certificate
-//    for it and stores it in the referenced secret.
-// 2. Starts a goroutine to watch for changes to the MeshConfig resource and certificate rotation, and
-//    updates/removes the certificate and secret as necessary.
-func (c *client) provisionIngressGatewayCert(stop <-chan struct{}) error {
-	defaultCertSpec := c.cfg.GetMeshConfig().Spec.Certificate.IngressGateway
-	if defaultCertSpec != nil {
-		// Issue a certificate for the default certificate spec
-		if err := c.createAndStoreGatewayCert(*defaultCertSpec); err != nil {
```
steeling, line 29: By creating the certificate before subscribing to updates, we were able to miss cert rotations

steeling:
I'll get code coverage back up

keithmattix:
Is this ready for review, or are there more iterations on the way?

steeling:
> Is this ready for review, or are there more iterations on the way?

it's ready!

pkg/certificate/manager.go
```diff
@@ -336,3 +334,15 @@ func (m *Manager) ListIssuedCertificates() []*Certificate {
...
 }
+
+// SubscribeRotations returns a channel that outputs every certificate that is rotated by the manager.
+// The caller must call the returned method to close the channel.
+// WARNING: you cannot call wait on the returned channel on the same go routine you are issuing a certificate on.
+func (m *Manager) SubscribeRotations(key string) (chan interface{}, func()) {
+	ch := m.pubsub.Sub(key)
+	return ch, func() {
+		go m.pubsub.Unsub(ch)
+		for range ch {
```
keithmattix, line 345: Why drain the channel here?

pkg/ingress/gateway.go
```diff
@@ -17,66 +16,10 @@ import (
...
-	}
-
-	return nil
-}
-
 // storeCertInSecret stores the certificate in the specified k8s TLS secret
 func (c *client) storeCertInSecret(cert *certificate.Certificate, secret corev1.SecretReference) error {
 	secretData := map[string][]byte{
-		"ca.crt":  cert.GetIssuingCA(),
+		"ca.crt":  cert.GetTrustedCAs(),
```
keithmattix, line 22: This is specifically for the cert the ingress gateway will present to the upstream service; why does the ingress gateway need the whole cert chain?

pkg/ingress/gateway.go
```diff
@@ -134,58 +77,80 @@ func (c *client) handleCertificateChange(currentCertSpec *configv1alpha2.Ingress
...
-			// Only update the secret if the cert rotated matches the cert spec
-			if cert.GetCommonName() != certificate.CommonName(cnInCertSpec) {
-				continue
-			}
+// watchCertificateRotations creates and stores a certificate with the given common name into the given secret ref.
+// It then watches for all cert rotations and updates the secret on changes.
+func (c *client) watchCertificateRotations(cn string, secret corev1.SecretReference) func() {
+	// This is some fanciness to provide the following guarantees:
+	// 1. The subscription is called before a goroutine starts, allowing the subscription creation to be deterministic.
+	// 2. By signalling to stop (instead of just closing it), we guarantee the close func doesn't return before the goroutine stops.
```
keithmattix, line 115: Why does this matter? I'm not saying it doesn't, it's just not clear to me from the comment + code

pkg/ingress/gateway_test.go
```diff
@@ -17,191 +17,17 @@ import (
...
 	tresorFake "github.com/openservicemesh/osm/pkg/certificate/providers/tresor/fake"
-	"github.com/openservicemesh/osm/pkg/configurator"
 	"github.com/openservicemesh/osm/pkg/k8s/events"
 	"github.com/openservicemesh/osm/pkg/messaging"
 )
 
 const (
 	maxTimeToSubscribe           = 500 * time.Millisecond
-	maxTimeForEventToBePublished = 2 * time.Second
+	maxTimeForEventToBePublished = 500 * time.Millisecond
```
keithmattix, line 26: What's the rationale for these decreases?

pkg/ingress/gateway.go
```diff
@@ -134,58 +77,80 @@ func (c *client) handleCertificateChange(currentCertSpec *configv1alpha2.Ingress
...
+		log.Err(err).Msg("Error updating ingress gateway cert secret after cert rotation")
+	}
+
+	go func() {
+		for {
+			select {
+			// A certificate was rotated
+			case msg := <-certRotateChan:
+				cert := msg.(*certificate.Certificate)
+				log.Info().Msg("Ingress gateway certificate was rotated, updating corresponding secret")
```
draychev, line 137: You have the `corresponding secret` - you should state exactly which secret is being updated.
Are you sure this should be an `Info`? I'd put it at Debug.

pkg/ingress/gateway.go
```diff
@@ -134,58 +77,80 @@ func (c *client) handleCertificateChange(currentCertSpec *configv1alpha2.Ingress
...
+
+			case <-stop:
+				return
+			}
 		}
+	}()
+	return func() {
+		// Allow close to be called multiple times by wrapping it in a once.
+		once.Do(func() {
+			stop <- struct{}{}
```
draychev, line 150: Shouldn't `close(stop)` be sufficient to call it done?

pkg/ingress/gateway.go
```diff
@@ -134,58 +77,80 @@ func (c *client) handleCertificateChange(currentCertSpec *configv1alpha2.Ingress
...
+	return oldSpec.SubjectAltNames[0] != newSpec.SubjectAltNames[0]
+}
 
-			// Only update the secret if the cert rotated matches the cert spec
-			if cert.GetCommonName() != certificate.CommonName(cnInCertSpec) {
-				continue
-			}
+// watchCertificateRotations creates and stores a certificate with the given common name into the given secret ref.
+// It then watches for all cert rotations and updates the secret on changes.
+func (c *client) watchCertificateRotations(cn string, secret corev1.SecretReference) func() {
```
draychev, line 112: Why do we need `watchCertificateRotations`? Doesn't the existing pub/sub infrastructure do the same? 
Whoever needs to watch for cert rotation could subscribe to the `cert-rotated` topic, no?

pkg/certificate/manager.go
```diff
@@ -7,21 +7,23 @@ import (
...
 )
 
 // NewManager creates a new CertificateManager with the passed MRCClient and options
-func NewManager(ctx context.Context, mrcClient MRCClient, getServiceCertValidityPeriod func() time.Duration, getIngressCertValidityDuration func() time.Duration, msgBroker *messaging.Broker, checkInterval time.Duration) (*Manager, error) {
+func NewManager(ctx context.Context, mrcClient MRCClient, getServiceCertValidityPeriod func() time.Duration, getIngressCertValidityDuration func() time.Duration, checkInterval time.Duration) (*Manager, error) {
 	m := &Manager{
 		serviceCertValidityDuration: getServiceCertValidityPeriod,
 		ingressCertValidityDuration: getIngressCertValidityDuration,
-		msgBroker:                   msgBroker,
+		pubsub:                      pubsub.New(1),
```
draychev, line 26: Why not reuse the existing message broker? It uses the same library (github.com/cskr/pubsub)

draychev (commented):
@steeling this PR is rather large and includes multiple areas. These could be split into separate PRs.

Splitting this into multiple PRs would make it more palatable to review these PRs - especially if they were small w/ atomic, isolated changes.

I'd make one PR for the renames: tresorfake -> tresorFake
Another PR could be the change from GetIssuingCA() to GetTrustedCAs()
Another is the fix for github.com/rs/zerolog/log --> 	github.com/openservicemesh/osm/pkg/logger

pkg/ingress/gateway.go
steeling: If the ingress gateway is doing mTLS, wouldn't it need both CA's during a rotation in case it is misaligned from the upstream svc?

pkg/ingress/gateway.go
```diff
@@ -134,58 +77,80 @@ func (c *client) handleCertificateChange(currentCertSpec *configv1alpha2.Ingress
...
-			if !ok {
-				log.Error().Msgf("Received unexpected message %T on channel, expected PubSubMessage", event)
-				continue
-			}
-			cert, ok := event.NewObj.(*certificate.Certificate)
-			if !ok {
-				log.Error().Msgf("Received unexpected message %T on cert rotation channel, expected Certificate", cert)
-				continue
-			}
+func shouldRelisten(oldSpec *configv1alpha2.IngressGatewayCertSpec, newSpec *configv1alpha2.IngressGatewayCertSpec) bool {
```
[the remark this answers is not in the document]
steeling: the semantics are a little more specific than is newSpec, for example if it is new but nil, or has 0 SAN's.

shouldRelisten feels more correct in that case, but happy to hear out alternatives that may be more descriptiv

pkg/ingress/gateway.go
steeling: I'll add more comments, but both 1 & 2 it prevent race conditions. Comments describing how/why coming soon

pkg/ingress/gateway.go
keithmattix: Ah you're right; the ingress gateway only sends, but it does need to validate the certificate of the receiver 👍🏾

pkg/ingress/gateway.go
steeling: this was a copy/paste of the prior comment. i can change it, although I'd argue a write to an external service warrants an INFO log

pkg/ingress/gateway.go
steeling: close is non-blocking, which could allow race conditions. I'll add a comment explaining more

pkg/ingress/gateway_test.go
steeling: pretty much everything below is covered in an eventually, and the tests were taking a long time to run for me. 

The originals felt aggressive

steeling:
> @steeling this PR is rather large and includes multiple areas. These could be split into separate PRs.
> 
> Splitting this into multiple PRs would make it more palatable to review these PRs - especially if they were small w/ atomic, isolated changes.
> 
> I'd make one PR for the renames: tresorfake -> tresorFake Another PR could be the change from GetIssuingCA() to GetTrustedCAs() Another is the fix for github.com/rs/zerolog/log --> github.com/openservicemesh/osm/pkg/logger

I've carved out the recommended componenets into a single PR; creating 3 separate ones felt like overkill 
https://github.com/openservicemesh/osm/pull/4870 It's still really small though.

What's your opinion on an appropriate size of a PR? I've typically worked under the assumption of ~400 lines (although I've been guilty of going over that quite a bit). (Not including tests unless tests are complicated, and not including removing large swaths of code).

If that PR gets merged prior to this one, I can rebase those changes, however I'd prefer not to block this PR's review

pkg/certificate/manager.go
steeling: this is about 15 lines of code, and requires 1 of either 2 things:

1. The message broker gets passed around in tandem with the certificate manager everywhere. This is a code smell IMO.
2. The cert manager exposes the message broker itself. 

With the current singleton message broker users still need to DI the broker into the cert manager. We could create another instance, but at that point its an implementation detail, and probably more code to do it that ways anyways.

I'd say why not do it this way, decoupling from another implementation, and [allowing us to move off of pubsub later](https://github.com/openservicemesh/osm/issues/4847)

pkg/certificate/manager.go
steeling: nope, i was wrong. This is required due to the head of line blocking issue https://github.com/openservicemesh/osm/issues/4847.

If the channel is not empty, then the unsub command will never even occur

pkg/certificate/manager.go
```diff
@@ -7,24 +7,19 @@ import (
...
+
 	"github.com/openservicemesh/osm/pkg/constants"
 	"github.com/openservicemesh/osm/pkg/errcode"
-	"github.com/openservicemesh/osm/pkg/k8s/events"
-	"github.com/openservicemesh/osm/pkg/logger"
-	"github.com/openservicemesh/osm/pkg/messaging"
-)
-
-var (
-	log = logger.New("certificate")
```
keithmattix, line 19: Can we keep this so we have more context in the log output?

pkg/certificate/manager.go
```diff
@@ -339,3 +330,17 @@ func (m *Manager) ListIssuedCertificates() []*Certificate {
...
+
+// SubscribeRotations returns a channel that outputs every certificate that is rotated by the manager.
+// The caller must call the returned method to close the channel.
+// WARNING: you cannot call wait on the returned channel on the same go routine you are issuing a certificate on.
+func (m *Manager) SubscribeRotations(key string) (chan interface{}, func()) {
+	ch := m.pubsub.Sub(key)
+	return ch, func() {
+		m.pubsub.Unsub(ch)
+		// must empty the channel to prevent deadlock
+		// https://github.com/openservicemesh/osm/issues/
```
keithmattix, line 342: Missing issue number in URL

keithmattix (approved):
Looks good; just a couple of small changes

pkg/certificate/manager.go
steeling: ya that was a mistake from the rebase, fixing

pkg/certificate/manager.go
steeling: good catch!

steeling:
Friendly ping for review

shashankram:
@steeling could you confirm the following 2 demos work after the changes in this PR:
https://main--osm-docs.netlify.app/docs/demos/ingress_contour/
https://main--osm-docs.netlify.app/docs/demos/ingress_k8s_nginx/

Since the HTTPs ingress scenarios aren't covered in the existing e2e tests, would be good to confirm these are impacted.

steeling:
> @steeling could you confirm the following 2 demos work after the changes in this PR: https://main--osm-docs.netlify.app/docs/demos/ingress_contour/ https://main--osm-docs.netlify.app/docs/demos/ingress_k8s_nginx/
> 
> Since the HTTPs ingress scenarios aren't covered in the existing e2e tests, would be good to confirm these are impacted.

done!

pkg/certificate/manager_test.go
```diff
@@ -365,3 +364,47 @@ func TestHandleMRCEvent(t *testing.T) {
...
+	cert, err := cm.IssueCertificate(cnPrefix1, Service)
+	assert.NoError(err)
+	assert.Equal("fake-cert-cn1.fake1.domain.com", cert.GetCommonName().String())
+
+	cert, err = cm.IssueCertificate(cnPrefix2, Service)
+	assert.NoError(err)
+	assert.Equal("fake-cert-cn2.fake1.domain.com", cert.GetCommonName().String())
+
+	ch1, unsub1 := cm.SubscribeRotations(cnPrefix1)
+	ch2, unsub2 := cm.SubscribeRotations(cnPrefix2)
```
jaellio, line 390: Should we subscribe here before issuing? Even though in this scenario we won't miss the rotation, would it be beneficial to be consistent about calling subscribe before issuing?

pkg/ingress/gateway_test.go
```diff
@@ -341,20 +156,17 @@ func TestHandleCertificateChange(t *testing.T) {
 				msgBroker:    msgBroker,
 			}
 
-			go c.handleCertificateChange(tc.previousCertSpec, tc.stopChan)
-			defer close(tc.stopChan)
+			go c.provisionIngressGatewayCert(tc.previousCertSpec, stop)
```
jaellio, line 159: Could you explain why the `defer close(tc.stop)` is unnecessary here?

steeling: i took off stop from the test case (it was never really a variable, just always the same), and reuse the stop chan from line 135, which already has a deferred close on it.

pkg/certificate/manager_test.go
steeling: ya i could see that, done now

pkg/ingress/gateway_test.go
jaellio: Got it! Missed the the earlier defer close(stop). Thanks!

jaellio (approved):

PR #4869, opened 2022-07-01
fips: add workflow for building and publishing FIPS-compliant control plane

### LINKED ISSUE #4888: Build and publish a FIPS compliant OSM control plane

Our users may need to run OSM in FIPS compliant environments. Therefore, we need to publish a set of control plane images that are FIPS compliant.

- CI System                  [X]

**Possible use cases**

keithmattix:
**Description**:
Creates a workflow for building and publishing FIPS-compliant versions of the OSM control plane. The modifications to the release process are in docs/release_guide.md. 
Fixes #4888

**Testing done**:

**Affected area**:

3. Has documentation corresponding to this change been updated in the [osm-docs](https://github.com/openservicemesh/osm-docs) repo (if applicable)? Not yet


keithmattix:
This version of OSM needs a custom version of Go (the Microsoft fork), so our CI/CD pipeline will need to be reworked to deploy this

keithmattix:
@nshankar13 I'd love your POV to make sure that nothing needs to change for the nightly or openshift builds

@nojnhuh @shashankram  I would really love your input around the changes to the CI and release processes.

dockerfiles/Dockerfile.demo.fips
```diff
@@ -0,0 +1,20 @@
+ARG GO_VERSION
```
nojnhuh, line 1: I would need to play with it a bit to see if this would work, but I'd prefer if we could add this logic to the existing Dockerfile instead of keeping this separate. The only hint I can offer is the end of this Dockerfile where we pick the base image for the right OS, but there might be a better way.

scripts/go-mod-tidy.sh
```diff
@@ -1,7 +1,12 @@
 #!/bin/bash
 
 go mod tidy
-if ! git diff --exit-code go.mod go.sum ; then
+if ! git diff --exit-code go.mod; then
+    echo -e "\nPlease commit the changes made by 'go mod tidy'"
+    exit 1
+fi
```
nojnhuh, line 7: Is this necessary?

.github/workflows/fips.yml
```diff
@@ -0,0 +1,332 @@
+name: Go FIPS
```
nojnhuh, line 1: If we could expose a single `FIPS=0/1` environment variable or something like that then it seems like we could use the existing workflow files and use a `strategy.matrix` to do everything for FIPS and non-FIPS.

Makefile
```diff
@@ -186,12 +205,46 @@ docker-build-osm-preinstall:
 docker-build-osm-healthcheck:
 	docker buildx build --builder osm --platform=$(DOCKER_BUILDX_PLATFORM) -o $(DOCKER_BUILDX_OUTPUT) -t $(CTR_REGISTRY)/osm-healthcheck:$(CTR_TAG) -f dockerfiles/Dockerfile.osm-healthcheck --build-arg GO_VERSION=$(DOCKER_GO_VERSION) --build-arg LDFLAGS=$(LDFLAGS) .
 
+.PHONY: docker-build-init-fips
+docker-build-init-fips:
```
nojnhuh, line 209: In general, I would much prefer to parameterize what we need to in the existing builds to make it possible to build FIPS-compliant images without needing to add new completely separate codepaths for them. It seems like the main things that are different are a couple things in the source and then most of how the images are built, but I would like to keep from having those kinds of changes from bubbling all the way up into places like the release guide.

docs/release_guide.md
```diff
@@ -13,13 +13,16 @@ This guide describes the process to create a GitHub Release for this project.
 Once an RC has been found to be stable, cut a release tagged `vX.Y.Z` using the following steps.
 
 - [Release Guide](#release-guide)
-  - [Release Candidates](#release-candidates)
-  - [Create a release branch](#create-a-release-branch)
+  - [Create release branches](#create-release-branches)
```
nojnhuh, line 16: If we can parameterize the builds enough to build FIPS and non-FIPS from the same source and utilize that in CI and use something like a `strategy.matrix` to do both, I can see where we shouldn't need to change the release guide at all.

Makefile
```diff
@@ -218,6 +276,7 @@ docker-build-cross-osm: docker-build-osm
 docker-build-cross-demo: DOCKER_BUILDX_PLATFORM=linux/amd64,windows/amd64
 docker-build-cross-demo: docker-build-demo
 docker-build-cross: docker-build-cross-osm docker-build-cross-demo
+docker-build-fips: docker-build-osm-fips docker-build-demo-fips
```
nshankar13, line 279: If we want to run / test FIPS complaint images on ARM64, we may also want to consider a cross-build option in the makefile here.

.github/workflows/fips.yml
```diff
@@ -0,0 +1,332 @@
...
+    env:
+      CTR_TAG: ${{ github.sha }}
+      CTR_REGISTRY: "localhost:5000"
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Build docker images
+        env:
+          DOCKER_BUILDX_OUTPUT: type=docker
+        run: make docker-build-osm
```
nshankar13, line 218: Should this be docker-build-osm-fips?

keithmattix: Yes, good catch!

nshankar13:
@keithmattix I don't think we would need to change anything for the OpenShift job or the nightly-noinstall job, unless we want to run the FIPS tests nightly as well. In that case we can open a follow up PR to edit the nightly job and run the fips tests there.

codegen/fips.go.txt
```diff
@@ -0,0 +1,5 @@
+package main
+
+import _ "crypto/tls/fipsonly"
```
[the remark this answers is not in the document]
keithmattix: I didn't consider build tags; I've honestly not come across a use-case where I needed them. Good idea though; I'll look into it!

.devcontainer/devcontainer.json
```diff
@@ -2,13 +2,16 @@
 // https://github.com/microsoft/vscode-dev-containers/tree/master/containers/docker-existing-dockerfile
 {
 	"name": "Dev Docker Environment",
-	"image": "osm:dev",
+	// "image": "osm:dev",
```
shashankram, line 5: I thought this file isn't even used anymore. Is this needed here or can we remove the file from the repo?

.github/workflows/fips.yml
shashankram: +1. 

@keithmattix is that feasible?

.github/workflows/pre-release-fips.yml
```diff
@@ -0,0 +1,63 @@
+name: FIPS Pre-release
```
shashankram, line 1: Can we use the existing pre-release workflow and add the FIPS job to a strategy matrix as suggested for the build job?

Makefile
shashankram: +1

docs/release_guide.md
shashankram: +1

.devcontainer/devcontainer.json
keithmattix: This file is needed for local FIPS development because vanilla Go doesn't have the `fipsonly` package. Without using devcontainers, compilation errors will abound

nojnhuh: I imagine if the file importing that package was conditionally enabled with a build tag then building locally would still work fine when the tag isn't enabled.

scripts/go-mod-tidy.sh
keithmattix: There was an issue with the version of `git` installed in the Mariner image. I _think_ I fixed this elsewhere in CI, so I'll try reverting this change

docs/release_guide.md
keithmattix: We would still need to publish different tags/images, but in general I agree if we can parameterize this to that degree.

.github/workflows/fips.yml
keithmattix: I don't think GitHub actions has the capability to conditionally set the image and other config that's necessary for the FIPS version of the build. That's the only limitation I can think of for the build matrix plan

dockerfiles/Dockerfile.demo.fips
keithmattix: I think the big blocker is going to be the fact that the mariner images aren't multi-platform AND there are different package install commands necessary (e.g. `yum install -y make ca-certificates`). I'm not a Docker build expert by any means, but unless there's conditional logic in Dockerfile syntax that allows switching between final images (e.g. `mcr.microsoft.com/cbl-mariner/distroless/base:2.0` and `gcr.io/distroless/static`) as well as selecting only a single builder image based on args, then these may have to stay separate

nojnhuh:
I played with parameterizing things and am fairly confident it will work. I have the changes limited to just osm-controller on a branch in my fork: https://github.com/openservicemesh/osm/compare/main...nojnhuh:osm:fips-jon

- Build FIPS images with `make docker-build-osm-controller FIPS=1`
  - FIPS=1 is just a convenience wrapper around setting CGO_ENABLED, the go build tag, and the base images for docker
- The only hack is I've set up the dockerfile to run the `yum install` based on whether `yum` exists or not, but I don't see that causing issues unless people try to use other base images than the defaults.
- I haven't tried making the CI changes yet, but I don't see any major roadblocks there. It just might be a bit messy to express "if $FIPS=1, then add '-fips' to the image tag" where we need to but that should be doable.

keithmattix:
@nojnhuh Thanks for that PoC; I haven't had to parameterize Dockerfiles like that before, so it was really helpful! The only blockers as far as CI workflows is [this one](https://github.com/openservicemesh/osm/pull/4869#discussion_r931491764). I don't _think_ GitHub allows for conditional images. We could potentially create FIPS specific jobs within a workflow, and that may actually be the right direction to go if we can keep the release guide unchanged. What are your thoughts?

nojnhuh:
I wonder if a reasonable first step for CI would be to setup builds for FIPS images for releases and run e2e tests nightly, but skip all of the other stuff we do normally. I'd like to think the FIPS build of Go is close enough to the normal one that the risk of any difference breaking us is low. If we find a need to run more things more frequently, we can address the problem of how to configure that in CI then.

> I don't _think_ GitHub allows for conditional images.

I think the way we'd express that is by defining the image as part of the matrix.

keithmattix:
@nojnhuh that's a good point; I'm in favor of that direction. I'll get working on that

keithmattix:
CodeQL is complaining about InsecureTLSVerify, but the meat of the PR should be fixed. Let me know your thoughts @nojnhuh

Makefile
```diff
@@ -163,88 +164,46 @@ DOCKER_DEMO_TARGETS = $(addprefix docker-build-, $(DEMO_TARGETS))
 .PHONY: $(DOCKER_DEMO_TARGETS)
 $(DOCKER_DEMO_TARGETS): NAME=$(@:docker-build-%=%)
 $(DOCKER_DEMO_TARGETS):
-	docker buildx build --builder osm --platform=$(DOCKER_BUILDX_PLATFORM) -o $(DOCKER_BUILDX_OUTPUT) -t $(CTR_REGISTRY)/$(NAME):$(CTR_TAG) -f dockerfiles/Dockerfile.demo --build-arg GO_VERSION=$(DOCKER_GO_VERSION) --build-arg BINARY=$(NAME) .
+	docker buildx build --builder osm --platform=$(DOCKER_BUILDX_PLATFORM) -o $(DOCKER_BUILDX_OUTPUT) -t $(CTR_REGISTRY)/$(NAME):$(CTR_TAG) -f dockerfiles/Dockerfile.demo --build-arg GO_BASE_IMAGE=$(DOCKER_GO_BASE_IMAGE) --build-arg BINARY=$(NAME) --build-arg FINAL_BASE_IMAGE=$(DOCKER_FINAL_BASE_IMAGE) --build-arg CGO_ENABLED=$(CGO_ENABLED) --build-arg GO_BUILD_FLAGS="$(DOCKER_GO_BUILD_FLAGS)" .
```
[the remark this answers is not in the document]
keithmattix: Oh good call; mass copypasta bit me here. We shouldn't be making these demo images FIPS compliant; removing.

.github/workflows/nightly-fips.yml
```diff
@@ -0,0 +1,144 @@
...
+          CERT_MANAGER: "tresor"
+          BOOKSTORE_SVC: "bookstore"
+          BOOKTHIEF_EXPECTED_RESPONSE_CODE: "0"
+          ENABLE_EGRESS: "false"
+          ENABLE_RECONCILER: "false"
+          PERMISSIVE_MODE: "false"
+          DEPLOY_TRAFFIC_SPLIT: "true"
+          CTR_TAG: ${{ github.sha }}-fips
+          USE_PRIVATE_REGISTRY: "false"
+          DOCKER_BUILD_CMD: "make docker-build"
```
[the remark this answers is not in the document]
keithmattix: Nope, removing

.github/workflows/pre-release.yml
```diff
@@ -22,38 +22,45 @@ jobs:
...
+      FIPS: ${{ contains(matrix.tags, 'fips') && 1 || 0 }}
     steps:
       - name: Checkout
         uses: actions/checkout@v2
       - name: Docker Login
         run: docker login --username "$DOCKER_USER" --password-stdin <<< "$DOCKER_PASS"
       - name: Push images with version tag
         env:
-          CTR_TAG: ${{ needs.version.outputs.version }}
+          CTR_TAG: ${{ contains(matrix.tags, 'fips') && format('{0}-fips', needs.version.outputs.version) || needs.version.outputs.version }}
```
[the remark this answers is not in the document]
keithmattix: Here's a working nightly run: https://github.com/keithmattix/osm/actions/runs/2762068316

Working on the pre-release and release workflows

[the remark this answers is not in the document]
keithmattix: @nojnhuh Turns out, my pre-release and image based runs are failing because I don't have a docker login and password in my fork: https://github.com/keithmattix/osm/actions/runs/2763317578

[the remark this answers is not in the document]
keithmattix: Passing Pre-release workflow: https://github.com/openservicemesh/osm/blob/b1682d62f4828264328f8b7bc96a241e16aac749/pkg/identity/types.go#L16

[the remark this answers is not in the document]
keithmattix: Passing release workflow: https://github.com/keithmattix/osm/actions/runs/2799846163

.github/workflows/nightly-fips.yml
```diff
@@ -0,0 +1,135 @@
...
+env:
+  CI_WAIT_FOR_OK_SECONDS: 60
+  CI_MAX_ITERATIONS_THRESHOLD: 60
+  CI_CLIENT_CONCURRENT_CONNECTIONS: 1
+  CI_MAX_WAIT_FOR_POD_TIME_SECONDS: 60
+  CI_MIN_SUCCESS_THRESHOLD: 1
+  OSM_HUMAN_DEBUG_LOG: true
+
+jobs:
+  build:
```
nojnhuh, line 18: I'm not sure this job really adds anything if we already do it for every change, so I'd be in favor of removing it here. This was originally defined as a dependency of these jobs to populate the build cache, but that's irrelevant now that the builds are all done in containers and I'm not really sure it sped things up at all in the first place.

.devcontainer/devcontainer.json
```diff
@@ -2,13 +2,16 @@
...
+	"image": "osm-fips:dev",
+	
 
 	// Sets the run context to one level up instead of the .devcontainer folder.
 	"context": "..",
 
 	// Update the 'dockerFile' property if you aren't using the standard 'Dockerfile' filename.
-	"dockerFile": "..\\Dockerfile",
+	// "dockerFile": "../Dockerfile",
+	"dockerFile": "../Dockerfile.fips",
```
nojnhuh, line 14: Should we leave the non-FIPS one as the default?

docs/release_guide.md
nojnhuh: Looks like this file needs another pass.

pkg/cli/verifier/pod.go
```diff
@@ -198,7 +198,7 @@ func (p *podProber) Probe(pod types.NamespacedName) error {
 			transport := &http.Transport{
 				TLSClientConfig: &tls.Config{
 					InsecureSkipVerify: true,
-					MinVersion:         tls.VersionTLS13,
+					MinVersion:         tls.VersionTLS12,
```
nojnhuh, line 201: Should this be conditional based on the `fips` build tag?

docs/release_guide.md
keithmattix: Ah yeah; I don't think there needs to be any changes in here after all right? Except maybe a note that ARM64 images will only be available via MCR

.github/workflows/pre-release.yml
```diff
@@ -22,38 +22,46 @@ jobs:
...
+    strategy:
+      matrix:
+        tags: ['', 'fips']
     env:
       DOCKER_USER: ${{ secrets.RELEASE_DOCKER_USER }}
       DOCKER_PASS: ${{ secrets.RELEASE_DOCKER_PASS }}
       VERSION: ${{ needs.version.outputs.version }}
       BUILD_DATE: "$$(date +%Y-%m-%d-%H:%M)"
+      FIPS: ${{ contains(matrix.tags, 'fips') && 1 || 0 }}
+      DOCKER_BUILDX_PLATFORM_OSM_CROSS: ${{ contains(matrix.tags, 'fips') && 'linux/amd64' || 'linux/amd64,linux/arm64' }}
```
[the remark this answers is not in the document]
keithmattix: No because the make target overrides it

.devcontainer/devcontainer.json
keithmattix: Yeah that's probably fair. I don't think anyone's using it, but defaulting to non-FIPS is probably clearer

pkg/cli/verifier/pod.go
keithmattix: Good call

nojnhuh (approved):

.devcontainer/devcontainer.json
```diff
@@ -2,13 +2,16 @@
...
+	"image": "osm-fips:dev",
+	
 
 	// Sets the run context to one level up instead of the .devcontainer folder.
 	"context": "..",
 
 	// Update the 'dockerFile' property if you aren't using the standard 'Dockerfile' filename.
-	"dockerFile": "..\\Dockerfile",
+	"dockerFile": "../Dockerfile",
+	// "dockerFile": "../Dockerfile.fips",
```
steeling, line 14: remove comments?

.github/workflows/main.yml
```diff
@@ -112,6 +112,33 @@ jobs:
       - name: Go Build
         run: make build-ci
 
+  build-fips:
+    name: Go build (FIPS)
+    runs-on: ubuntu-latest
+    container:
+      image: mcr.microsoft.com/oss/go/microsoft/golang:1.17-fips-cbl-mariner2.0
```
steeling, line 119: bump to 1.19? assuming https://github.com/openservicemesh/osm/pull/4972

.github/workflows/main.yml
```diff
@@ -134,6 +161,39 @@ jobs:
         with:
           flags: unittests
 
+  unittest-fips:
+    name: Go test (FIPS)
+    runs-on: ubuntu-latest
+    container:
+      image: mcr.microsoft.com/oss/go/microsoft/golang:1.17-fips-cbl-mariner2.0
```
steeling, line 168: same

.github/workflows/nightly-fips.yml
```diff
@@ -0,0 +1,107 @@
+name: Nightly FIPS
```
steeling, line 1: Would be nice to "templatize" some of these files. Not sure would the best option would be, but something to consider.

Pre-generation hooks, jsonnet, etc may all be good options. Not blocking this PR though

steeling (approved):

.github/workflows/nightly-fips.yml
keithmattix: Hmm, does GitHub Actions natively support those options? If not, we'd have to use something like a pre-commit hook. `matrix` strategies in GitHub actions get use decently far

PR #4871, opened 2022-07-01
ref: use binary flag to enable use of MeshRootCertificate

jaellio:
**Description**:

- Uses a binary flag to enable the use of MeshRootCertificate. 
- Updates the MRC id to be the MRC name
- Passes vault token secret reference options to the controller, injector, and bootstrap 

**Testing done**:
- CI
- Demo
  - verified the ability to use a secret to store the vault token and set the osm.vault.secret.name and osm.vault.secret.key on install

**Affected area**:
| Certificate Management     | [ x ] |

3. Has documentation corresponding to this change been updated in the [osm-docs](https://github.com/openservicemesh/osm-docs) repo (if applicable)? https://github.com/openservicemesh/osm-docs/pull/407


pkg/certificate/providers/config.go
```diff
@@ -44,14 +44,16 @@ var getCA func(certificate.Issuer) (pem.RootCertificate, error) = func(i certifi
 // NewCertificateManager returns a new certificate manager, with an MRC compat client.
 // TODO(4713): Use an informer behind a feature flag.
 func NewCertificateManager(ctx context.Context, kubeClient kubernetes.Interface, kubeConfig *rest.Config, cfg configurator.Configurator,
-	providerNamespace string, options Options, msgBroker *messaging.Broker, ic *informers.InformerCollection, checkInterval time.Duration) (*certificate.Manager, error) {
-	if err := options.Validate(); err != nil {
+	providerNamespace string, certOptions *CertProviderOptions, msgBroker *messaging.Broker, ic *informers.InformerCollection, checkInterval time.Duration) (*certificate.Manager, error) {
```
steeling, line 47: nit: instead of fiddling with the options maybe have 2 functions:

`NewCertificateManagerFromMRC` (takes options as a param, along with informers)

NewLegacyCertificateManager`. (doesn't take options, no ctx, etc)

pkg/certificate/providers/config_test.go
```diff
@@ -165,51 +165,14 @@ func TestGetCertificateManager(t *testing.T) {
 			cfg:         mockConfigurator,
 			expectError: true,
 		},
-		{
```
steeling, line 168: why remove this test?

jaellio: Added it back

charts/osm/README.md
```diff
@@ -265,6 +265,8 @@ The following table lists the configurable parameters of the osm chart and their
 | osm.tracing.port | int | `9411` | Port of the tracing collector service |
 | osm.tracing.tolerations | list | `[]` | Node tolerations applied to control plane pods. The specified tolerations allow pods to schedule onto nodes with matching taints. |
 | osm.trustDomain | string | `"cluster.local"` | The trust domain to use as part of the common name when requesting new certificates. |
+| osm.unsafe | object | `{"enableMeshRootCertificate":false}` | Unsafe values. Behavior is not supported. |
```
keithmattix, line 268: Perhaps `experimental` is a better name here? Outside of certain contexts (i.e. Rust), shipping "unsafe" features may not be a great look

cmd/cli/install_test.go
```diff
@@ -290,14 +290,6 @@ var _ = Describe("Running the install command", func() {
 			err := installCmd.run(config)
 			Expect(err.Error()).To(ContainSubstring("osm.vault.host is required"))
 		})
-
-		It("should error when token isn't set", func() {
```
keithmattix, line 294: Could we replace this with a test that makes sure either `token` or the secret ref is set?

charts/osm/values.yaml
```diff
@@ -167,6 +167,11 @@ osm:
     # The specified tolerations allow pods to schedule onto nodes with matching taints.
     tolerations: []
 
+  # -- Experimental values. Behavior is not supported.
+  experimental:
+    # -- Enable the MeshRootCertificate to configure the OSM certificate provider.
+    enableMeshRootCertificate: false
```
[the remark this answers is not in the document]
jaellio: That makes sense to me. Should the flag be added to the MeshConfig?

pkg/certificate/providers/config.go
```diff
@@ -208,11 +196,8 @@ func (c *MRCProviderGenerator) getHashiVaultOSMCertificateManager(mrc *v1alpha2.
 	if err != nil {
 		return nil, "", fmt.Errorf("error instantiating Hashicorp Vault as a Certificate Manager: %w", err)
 	}
-	id, err := getMRCID(mrc)
-	if err != nil {
-		return nil, "", err
-	}
-	return vaultClient, id, nil
+
+	return vaultClient, mrc.Name, nil
```
steeling, line 200: At this point, i would change the func signature to (issuer, error), and let the manager set the ID on it's own

steeling (approved):

charts/osm/values.yaml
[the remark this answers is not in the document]
jaellio: I originally hesitated to add it as a part of the featureFlag spec since I thought all values in that spec supported dynamic configuration. I've added `enableMeshRootCertificate` to the featureFlag spec and did not include it in the MeshConfig

pkg/certificate/providers/config.go
jaellio: Updated the signature to no longer return the issuer id.

shashankram (approved):

nojnhuh (approved):

PR #4883, opened 2022-07-08
add e2e test for ingress gateway cert

steeling:


tests/e2e/e2e_ingress_gateway_test.go
```diff
@@ -0,0 +1,187 @@
...
+					SubjectAltNames:  []string{"myhost.test.com"},
+					ValidityDuration: "24h",
+					Secret: corev1.SecretReference{
+						Name:      secretName,
+						Namespace: ingressNS,
+					},
+				}
+
+				_, err = Td.UpdateOSMConfig(meshConfig)
+				Expect(err).NotTo(HaveOccurred())
```
jaellio, line 61: Should we also check if the secret was successfully created?

tests/e2e/e2e_ingressbackend_test.go
```diff
@@ -33,7 +37,10 @@ var _ = OSMDescribe("Ingress using IngressBackend API",
 	})
 
 func testIngressBackend() {
-	var destNs = framework.RandomNameWithPrefix("server")
+	var (
+		destNs     = framework.RandomNameWithPrefix("server")
+		secretName = "ingress-gateway-cert"
+	)
 
 	It("allows ingress traffic", func() {
```
jaellio, line 45: Could the changes made to the IngressBackend test be added to a separate "It" block? We could have "allows ingress traffic without TLS" and "allows ingress traffic with mTLS"? This would allow us to test our support for both documented scenarios https://main--osm-docs.netlify.app/docs/demos/ingress_k8s_nginx/

keithmattix:
Could I request adding a test or two that actually installs Contour? We could've caught #4898 earlier if we had one

steeling:
> Could I request adding a test or two that actually installs Contour? We could've caught #4898 earlier if we had one

Let's get the fix in for that first, as I'd rather not block this PR on that

keithmattix:
Is the expectation that the user will just add the trust domain when specifying the SAN?

PR #4885, opened 2022-07-11
update demo to allow multi-arch/multi-os 

whitneygriffith:
Signed-off-by: Whitney Griffith whitney.griffith16@gmail.com

**Description**: The demo has been explicitly set to run on nodes with Linux OS and and AMD64 Architecture. These changes unblock running the demo on clusters with exclusively linux/arm64 nodes. The node selectors were removed for the demo apps and [node affinity](ttps://kubernetes.io/docs/concepts/configuration/assign-pod-node/) added where necessary. 

**Testing done**: The demo was run locally with these changes. 

**Affected area**:
| Demo                       |  [x] |

No

2. Is this a breaking change?

No

### DIFF
--- a/demo/deploy-vault.sh
+++ b/demo/deploy-vault.sh
@@ -27,9 +27,6 @@ spec:
       labels:
         app: vault
     spec:
-      nodeSelector:
-        kubernetes.io/arch: amd64
-        kubernetes.io/os: linux
       terminationGracePeriodSeconds: 10
       containers:
       - name: vault
--- a/demo/deploy-vault.sh
+++ b/demo/deploy-vault.sh
@@ -27,6 +27,24 @@ spec:
       labels:
         app: vault
     spec:
+      ## Affinity settings for pod assignment
+      ## Ref: https://kubernetes.io/docs/concepts/configuration/assign-pod-node/
+      affinity:
+        nodeAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+            nodeSelectorTerms:
+            - matchExpressions:
+              - key: kubernetes.io/os
+                operator: In
+                values:
+                - linux
+              - key: kubernetes.io/arch
+                operator: In
+                values:
+                - amd64
+                - arm64
+                - arm 
+                - 386
       terminationGracePeriodSeconds: 10
       containers:
       - name: vault
--- a/Makefile
+++ b/Makefile
@@ -22,7 +22,7 @@ BUILD_DATE_VAR := github.com/openservicemesh/osm/pkg/version.BuildDate
 BUILD_VERSION_VAR := github.com/openservicemesh/osm/pkg/version.Version
 BUILD_GITCOMMIT_VAR := github.com/openservicemesh/osm/pkg/version.GitCommit
 DOCKER_GO_VERSION = 1.17
-DOCKER_BUILDX_PLATFORM ?= linux/amd64
+DOCKER_BUILDX_PLATFORM ?= linux/$(shell go env GOARCH)
 # Value for the --output flag on docker buildx build.
 # https://docs.docker.com/engine/reference/commandline/buildx_build/#output
 DOCKER_BUILDX_OUTPUT ?= type=registry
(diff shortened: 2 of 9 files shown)


shashankram (changes requested):
@whitneygriffith Since these are Linux apps, the node selectors are in place so that they don't get scheduled on Windows nodes in a cluster that has both Linux and Windows nodes.

What's the reason to remove these?

demo/deploy-vault.sh
```diff
@@ -27,9 +27,6 @@ spec:
       labels:
         app: vault
     spec:
-      nodeSelector:
-        kubernetes.io/arch: amd64
-        kubernetes.io/os: linux
```
nojnhuh, line 32: Since this isn't an image we build, let's specify `affinity` here so we can match exactly what the image supports, which `docker manifest inspect vault:1.4.0` shows should be:
- linux/amd64
- linux/arm
- linux/arm64
- linux/386

We do something similar here:
https://github.com/openservicemesh/osm/blob/3a7c924c9ebdedf9513220fc5d1c527b933f71b9/charts/osm/values.yaml#L230-L248

nojnhuh (commented):
> @whitneygriffith Since these are Linux apps, the node selectors are in place so that they don't get scheduled on Windows nodes in a cluster that has both Linux and Windows nodes.
> 
> What's the reason to remove these?

The demo apps are currently built and published for Windows, so I'm fairly confident they worked at some point but I haven't tried recently.

I think it's reasonable to remove the node selector for the demo apps so developers are as free as possible to run the demo. If tweaking is necessary for some setups, then the deploy scripts can always be edited manually. These changes unblock running the demo on clusters with exclusively linux/arm64 nodes which I think is at least a step in the right direction.

shashankram:
> > These changes unblock running the demo on clusters with exclusively linux/arm64 nodes which I think is at least a step in the right direction.
> 
> That's fair. @whitneygriffith could you update the commit && PR description to reflect the motivation for this change?

@nojnhuh the images in the demo are not deploying the `-windows` version of the images. So we should still ensure these images don't get scheduled onto Windows nodes. This can be done by specifying a [nodeAffinity](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#node-affinity) that allows the pod to be scheduled on either amd64 or arm64 nodes.

nojnhuh:
> @nojnhuh the images in the demo are not deploying the `-windows` version of the images. So we should still ensure these images don't get scheduled onto Windows nodes. This can be done by specifying a [nodeAffinity](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#node-affinity) that allows the pod to be scheduled on either amd64 or arm64 nodes.

We don't build separate`-windows` versions of  the demo images anymore. We publish multi-os/arch manifests so the same reference works for both Windows and linux.

shashankram:
> > @nojnhuh the images in the demo are not deploying the `-windows` version of the images. So we should still ensure these images don't get scheduled onto Windows nodes. This can be done by specifying a [nodeAffinity](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#node-affinity) that allows the pod to be scheduled on either amd64 or arm64 nodes.
> 
> We don't build separate`-windows` versions of the demo images anymore. We publish multi-os/arch manifests so the same reference works for both Windows and linux.

I see, I forgot about that. Have we tested the multi-arch images to work on Windows? Given that Windows support is still highly experimental, we could potentially not worry about whether the images work on Windows, but it would be good to know rather than hoping it doesn't.

nojnhuh:
I don't remember whether or not those were tested on Windows for that change, and I don't have a Windows setup handy to test them myself.

shashankram (approved):
Not blocking this on Windows testing.

@whitneygriffith, please describe the motivation for this change in the commit and PR descriptions.

whitneygriffith:
Addressed the comments in last commit

nojnhuh (approved):

PR #4909, opened 2022-07-18
Pin the alpine image to a minor version (3.16)

trstringer:
Prior to this change the alpine image was using the sliding tag of `3`.
This PR introduces are stricter pin to the minor version to keep the
version drift to a minimum, so that we pull in patch updates but not
minor updates.

**Affected area**:

3. Has documentation corresponding to this change been updated in the [osm-docs](https://github.com/openservicemesh/osm-docs) repo (if applicable)? N/A.

### DIFF
--- a/dockerfiles/Dockerfile.init
+++ b/dockerfiles/Dockerfile.init
@@ -1,2 +1,2 @@
-FROM alpine:3
+FROM alpine:3.16
 RUN apk add --no-cache iptables


nojnhuh (changes requested):
This is what we were doing originally. I changed the tag a while ago to just `3` so we don't have to keep updating this every time this minor version goes out of support and a vulnerability pops up. Considering all we're using the base image for is to install iptables, I don't see what the risk is in letting this update itself.

keithmattix:
Closing since conversation has stalled

Issue #4783, opened 2022-06-02
[Optimization] Leverage standard map + mutex in ProxyRegistry


steeling:
The ProxyRegistry uses a sync.Map, which has the downfall of not being able to get the count. The current solution is to iterate the map (while holding its lock) and count the number of items. 

Using a RW mutex + a normal map would yield improvements

PR #4912, opened 2022-07-18
ref: optimize ProxyRegistry counting

allenlsy:
**Description**:

* Use a counter with mutex to get the size of ProxyRegistry
instead of looping the whole registry
* Use a normal map instead of sync.Map to store proxies

Resolves https://github.com/openservicemesh/osm/issues/4783

**Testing done**:

* Unit test

**Affected area**:

No

2. Is this a breaking change?

No

No


pkg/envoy/registry/types.go
```diff
@@ -14,9 +15,13 @@ var log = logger.New("proxy-registry")
 type ProxyRegistry struct {
 	ProxyServiceMapper
 
-	connectedProxies sync.Map
+	connectedProxies map[string]*envoy.Proxy
```
keithmattix, line 18: A raw map is not threadsafe. A consequence of this change is that the locks protecting this data structure are held longer. How confident are we that this change results in higher performance?

keithmattix (changes requested):
The rationale for this change isn't clear to me. Is looping over the proxyregistry causing some bottleneck? What other approaches might be taken to achieve a similar performance result?

pkg/envoy/registry/types.go
allenlsy: I have similar doubt actually. I suggest doing a benchmark to compare these two approaches.

steeling: there's no need to hold the lock longer, but the locks are locking on a single int, instead of on the (non-threadsafe) map itself.

It would be faster (no benchmark needed), simply grab the lock whenever reading/writing to the map

pkg/envoy/registry/registry.go
```diff
@@ -10,43 +10,64 @@ func NewProxyRegistry(mapper ProxyServiceMapper, msgBroker *messaging.Broker) *P
...
+
+	pr.connectedProxies[uuid] = proxy
+	pr.proxiesCount++
 	log.Debug().Str("proxy", proxy.String()).Msg("Registered new proxy")
 }
 
 // GetConnectedProxy loads a connected proxy from the registry.
 func (pr *ProxyRegistry) GetConnectedProxy(uuid string) *envoy.Proxy {
-	p, ok := pr.connectedProxies.Load(uuid)
+	p, ok := pr.connectedProxies[uuid]
```
steeling, line 35: these accesses are not thread safe and will result in a panic

pkg/envoy/registry/registry.go
```diff
@@ -10,43 +10,64 @@ func NewProxyRegistry(mapper ProxyServiceMapper, msgBroker *messaging.Broker) *P
...
 // UnregisterProxy unregisters the given proxy from the catalog.
-func (pr *ProxyRegistry) UnregisterProxy(p *envoy.Proxy) {
-	pr.connectedProxies.Delete(p.UUID.String())
-	log.Debug().Msgf("Unregistered proxy %s", p.String())
+func (pr *ProxyRegistry) UnregisterProxy(proxy *envoy.Proxy) {
+	pr.countLock.Lock()
+	defer pr.countLock.Unlock()
+
+	uuid := proxy.UUID.String()
+	if pr.GetConnectedProxy(uuid) == nil {
```
steeling, line 48: calling this while the lock is held will result in a deadlock

pkg/envoy/registry/types.go
```diff
@@ -14,9 +15,13 @@ var log = logger.New("proxy-registry")
...
 
 	msgBroker *messaging.Broker
+
+	proxiesCount int
```
steeling, line 22: standard practice is to call it `mu sync.Mutex`

and simply place the protected fields underneath it, in this case the map itself. No need to store the length, you can just len whenever you need it. that's a fast op

pkg/envoy/registry/types.go
keithmattix: My comment about holding the lock longer was referring to the PR's implementation since the lock was being acquired and then not released until the count was updated.

Are we sure this is a hot path though? How often are we checking the length of this map?

allenlsy: I did a test comparing this the normal map v.s. original sync.Map. The test consists of two parts:

* Add 10000 proxies to the registry with 10000 goroutines
* When registry contains 10000 proxies, get the proxies count.

Both tests ensured the correctness of proxies count.

|  | sync.Map | normal map with mutex |
| --- | --- | --- |
| Add 10000 proxies | 9.5s | 0.26s |
| GetCount | 957642 ns | 13ns  |

keithmattix: Great, thanks @allenlsy. That's what I was looking for

pkg/envoy/registry/types.go
keithmattix: +1

pkg/envoy/registry/registry.go
```diff
@@ -10,43 +10,64 @@ func NewProxyRegistry(mapper ProxyServiceMapper, msgBroker *messaging.Broker) *P
...
+	delete(pr.connectedProxies, uuid)
+	pr.proxiesCount--
+	log.Debug().Msgf("Unregistered proxy %s", proxy.String())
 }
 
 // GetConnectedProxyCount counts the number of connected proxies
-// TODO(steeling): switch to a regular map with mutex so we can get the count without iterating the entire list.
 func (pr *ProxyRegistry) GetConnectedProxyCount() int {
-	return len(pr.ListConnectedProxies())
+	pr.countLock.RLock()
```
steeling, line 60: there's actually 2 locks associated with a read lock. Read locks are great for when you do a lot of reads, and your ops (read/write), might have some longer blocking function. In this case, since it's just setting/deleting/reading values it should be faster to use a standard lock

pkg/envoy/registry/types.go
```diff
@@ -14,9 +15,13 @@ var log = logger.New("proxy-registry")
...
+
+	countLock sync.RWMutex
```
steeling, line 24: see not above about standard mutex vs rw

steeling:
hey @allenlsy friendly reminder not to squash commits to make reviews easier

pkg/envoy/registry/announcement_handlers.go
```diff
@@ -35,8 +35,7 @@ func (pr *ProxyRegistry) ReleaseCertificateHandler(certManager certificateReleas
 			}
 
 			proxyUUID := deletedPodObj.Labels[constants.EnvoyUniqueIDLabelName]
-			if proxyIface, ok := pr.connectedProxies.Load(proxyUUID); ok {
-				proxy := proxyIface.(*envoy.Proxy)
+			if proxy, ok := pr.connectedProxies[proxyUUID]; ok {
```
steeling, line 38: leverage GetCOnnected proxy here, since this needs to grab the lock

pkg/envoy/registry/registry_benchmark.go
```diff
@@ -0,0 +1,66 @@
+package registry
```
steeling, line 1: benchmark tests still belong int `*_test.go` files, so as not to be included in non test files

pkg/envoy/registry/registry_test.go
```diff
@@ -40,5 +43,29 @@ var _ = Describe("Test catalog proxy register/unregister", func() {
 			connectedProxies := proxyRegistry.ListConnectedProxies()
 			Expect(len(connectedProxies)).To(Equal(0))
 		})
+
+		It("ensures the correctness of proxy count", func() {
```
steeling, line 47: Not going to block on this, but would be nice to move away from ginko in unit tests

pkg/envoy/registry/types.go
```diff
@@ -14,9 +15,11 @@ var log = logger.New("proxy-registry")
 type ProxyRegistry struct {
 	ProxyServiceMapper
 
-	connectedProxies sync.Map
+	connectedProxies map[string]*envoy.Proxy
```
steeling, line 18: nit: place the connected proxies field under the mutex field, which signifies what the mutex locks.

pkg/envoy/registry/registry_test.go
keithmattix: +1 table driven tests or at least testify assert is more idiomatic

trstringer: Also a +1 on this! We are moving away from ginkgo for testing, so we shouldn't be introducing more at this point.

pkg/envoy/registry/registry.go
```diff
@@ -10,43 +10,66 @@ func NewProxyRegistry(mapper ProxyServiceMapper, msgBroker *messaging.Broker) *P
...
-	log.Debug().Msgf("Unregistered proxy %s", p.String())
+func (pr *ProxyRegistry) UnregisterProxy(proxy *envoy.Proxy) {
+	uuid := proxy.UUID.String()
+	if pr.GetConnectedProxy(uuid) == nil {
+		log.Debug().Str("proxy", proxy.String()).Msgf("Proxy %s does not exist", proxy.String())
+		return
+	}
+
+	pr.mu.Lock()
+	delete(pr.connectedProxies, uuid)
```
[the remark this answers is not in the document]
allenlsy: In our test, after removing lock, concurrent map read and write is found. I think the program may add and delete the same key simultaneously.

allenlsy:
Ping here for the review @steeling @keithmattix

pkg/envoy/ads/response_benchmark_test.go
```diff
@@ -187,7 +187,6 @@ func BenchmarkSendXDSResponse(b *testing.B) {
 					b.Fatalf("Failed to send response: %s", err)
 				}
 			}
-			b.StopTimer()
```
steeling, line 190: you intend to remove this?

pkg/envoy/registry/registry.go
```diff
@@ -10,43 +10,66 @@ func NewProxyRegistry(mapper ProxyServiceMapper, msgBroker *messaging.Broker) *P
...
 		msgBroker:          msgBroker,
+		connectedProxies:   make(map[string]*envoy.Proxy),
 	}
 }
 
 // RegisterProxy registers a newly connected proxy.
 func (pr *ProxyRegistry) RegisterProxy(proxy *envoy.Proxy) {
-	pr.connectedProxies.Store(proxy.UUID.String(), proxy)
+	uuid := proxy.UUID.String()
+	if pr.GetConnectedProxy(uuid) != nil {
```
steeling, line 20: The way in which a proxy can disconnect and reconnect in rapid succession would actually allow a proxy to still be "registered", prior to it being deregistered.

This is a pre-existing bug, and isn't quite in the critical path, so likely not a big concern, although we should make a note of it and consider addressing it later.

One solution would require a nonce, or a version number to check when unregistering, and only unregistering if those match.

pkg/envoy/registry/registry.go
```diff
@@ -10,43 +10,66 @@ func NewProxyRegistry(mapper ProxyServiceMapper, msgBroker *messaging.Broker) *P
...
 }
 
 // GetConnectedProxy loads a connected proxy from the registry.
 func (pr *ProxyRegistry) GetConnectedProxy(uuid string) *envoy.Proxy {
-	p, ok := pr.connectedProxies.Load(uuid)
+	pr.mu.Lock()
+	p, ok := pr.connectedProxies[uuid]
+	pr.mu.Unlock()
+
 	if !ok {
```
steeling, line 37: nit: no need for nil check, a bit simpler:

```go
pr.mu.Lock()
defer pr.mu.Unlock()
return pr.connectedProxies[uuid]
```

pkg/envoy/registry/registry.go
```diff
@@ -10,43 +10,66 @@ func NewProxyRegistry(mapper ProxyServiceMapper, msgBroker *messaging.Broker) *P
...
+	return p
 }
 
 // UnregisterProxy unregisters the given proxy from the catalog.
-func (pr *ProxyRegistry) UnregisterProxy(p *envoy.Proxy) {
-	pr.connectedProxies.Delete(p.UUID.String())
-	log.Debug().Msgf("Unregistered proxy %s", p.String())
+func (pr *ProxyRegistry) UnregisterProxy(proxy *envoy.Proxy) {
+	uuid := proxy.UUID.String()
+	if pr.GetConnectedProxy(uuid) == nil {
```
steeling, line 46: nit: this log doesn't add much value IMO, can just simply call delete within the lock

pkg/envoy/ads/response_benchmark_test.go
allenlsy: Yes. It's not necessary in this case since it's the end of the test.

pkg/envoy/registry/registry.go
allenlsy: Let's create a new issue for this bug. Nonce might not be a good solution. Sequential version number should maintain the request sequence. 

E.g. a proxy sends unregister and register requests. Due to network delay, register request arrives earlier. Without sequence number, it will eventually be deregistered.

keithmattix (approved):

pkg/envoy/registry/registry.go
jaellio: Can you create the issue and add a TODO here referencing the issue number?

jaellio (approved):
@allenlsy What is the force rerun file included in this PR? Waiting on @steeling's review to merge.

allenlsy:
@jaellio That's an empty file commit to force re-run a flaky e2e test. I can squash all commits here into a single one pre-merge

jaellio:
> @jaellio That's an empty file commit to force re-run a flaky e2e test. I can squash all commits here into a single one pre-merge

Thanks for clarifying! Is this something we want to add in this PR or in a separate PR? Should we review the flaky test?

allenlsy:
> > @jaellio That's an empty file commit to force re-run a flaky e2e test. I can squash all commits here into a single one pre-merge
> 
> Thanks for clarifying! Is this something we want to add in this PR or in a separate PR? Should we review the flaky test?

Based on the failed test [here](https://github.com/openservicemesh/osm/runs/7526003652?check_suite_focus=true#step:5:263), I don't think it's related to the changes in this PR. We should address it separately.

steeling (approved):
approved after comments are addressed, thanks Allen!

pkg/envoy/registry/registry.go
allenlsy: New issue opened to track the bug: https://github.com/openservicemesh/osm/issues/4950

jaellio: Awesome! Did you add the TODO(#4950) comment?

allenlsy: Just added. Let's wait for the CI pass.

PR #4913, opened 2022-07-18
fix: add missing nosec

jaellio:
**Description**:

Adds missing nosec causing lint errors on http server
initialization.

Thanks @shalier for finding this!

**Testing done**:
- CI

**Affected area**:
| Functional Area            |     |
| -------------------------- | --- |
| New Functionality          | [ ] |
| CI System                  | [ ] |
| CLI Tool                   | [ ] |
| Certificate Management     | [ ] |
| Control Plane              | [ ] |
| Demo                       | [ ] |
| Documentation              | [ ] |
| Egress                     | [ ] |
| Ingress                    | [ ] |
| Install                    | [ ] |
| Networking                 | [ ] |
| Observability              | [ ] |
| Performance                | [ ] |
| SMI Policy                 | [ ] |
| Security                   | [ ] |
| Sidecar Injection          | [ ] |
| Tests                      | [x] |
| Upgrade                    | [ ] |
| Other                      | [ ] |

Please answer the following questions with yes/no.

1. Does this change contain code from or inspired by another project? No
    -   Did you notify the maintainers and provide attribution?

2. Is this a breaking change? No

3. Has documentation corresponding to this change been updated in the [osm-docs](https://github.com/openservicemesh/osm-docs) repo (if applicable)?

PR #4915, opened 2022-07-19
fix golints and security scan issues

steeling:
1. Bump prom client
2. Add ReadHeaderTimeout of 10s

### DIFF
--- a/pkg/httpserver/httpserver.go
+++ b/pkg/httpserver/httpserver.go
@@ -101,6 +101,8 @@ func (s *HTTPServer) Stop() error {
 	s.server = &http.Server{
 		Addr:    fmt.Sprintf(":%d", s.port),
 		Handler: s.httpServeMux,
+		// Needs a default for gosec. This can probably be brought down to a lower value.
+		ReadHeaderTimeout: time.Second * 10,
 	}
 
 	return nil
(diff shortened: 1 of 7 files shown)


pkg/httpserver/httpserver.go
```diff
@@ -101,6 +101,8 @@ func (s *HTTPServer) Stop() error {
 	s.server = &http.Server{
 		Addr:    fmt.Sprintf(":%d", s.port),
 		Handler: s.httpServeMux,
+		// Needs a default for gosec. This can probably be brought down to a lower value.
+		ReadHeaderTimeout: time.Second * 10,
```
nojnhuh, line 105: #4913 from @jaellio is also open to address this by ignoring the error. In general I'd prefer to fix it like this does rather than ignore it, but I admit I don't have a good sense for exactly what we should set this to.

nojnhuh (approved):

pkg/httpserver/httpserver.go
jaellio: There are multiple other places where we ignore G112. If we decide to make the fix, I think we should fix it in the other locations as well.

steeling: done!

PR #4919, opened 2022-07-19
rate-limiting: add global rate limiting schema

shashankram:
**Description**:
Adds the global rate limiting schema to the existing
rate limiting spec for per VirtualHost and Route
level rate limiting. The schema is based on the
Go types for [UpstreamTrafficSetting](https://github.com/openservicemesh/osm/blob/main/pkg/apis/policy/v1alpha1/upstreamtrafficsetting.go).

Part of #2018

**Testing done**:
Applied sample configs such as:
```
apiVersion: policy.openservicemesh.io/v1alpha1
kind: UpstreamTrafficSetting
metadata:
  name: http-rate-limit
  namespace: demo
spec:
  host: test.demo.svc.cluster.local
  rateLimit:
    global:
      http:
        rateLimitService:
          host: rls.default.svc.cluster.local
          port: 80
        domain: osm
        failOpen: true
        timeout: 20ms
        enableXRateLimitHeaders: true
        descriptors:
          - entries:
            - remoteAddress: {}
          - entries:
            - genericKey:
                key: foo
                value: bar
            - requestHeader:
                name: x-service
                key: service
            - headerValueMatch:
                key: header_match
                value: get
                expectMatch: true
                headers:
                  - name: ":method"
                    prefix: /get
  httpRoutes:
    - path: /foo
      rateLimit:
        global:
          descriptors:
          - entries:
            - remoteAddress: {}
```

**Affected area**:

3. Has documentation corresponding to this change been updated in the [osm-docs](https://github.com/openservicemesh/osm-docs) repo (if applicable)? `no`

### DIFF
--- a/cmd/osm-bootstrap/crds/policy_upstream_traffic_setting.yaml
+++ b/cmd/osm-bootstrap/crds/policy_upstream_traffic_setting.yaml
@@ -91,7 +91,7 @@ spec:
                   type: object
                   properties:
                     local:
-                      description: Policy responsible for rate limiting traffic at the upstream service.
+                      description: Local rate limit policy responsible for rate limiting traffic at the upstream service.
                       type: object
                       properties:
                         tcp:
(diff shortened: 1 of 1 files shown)


cmd/osm-bootstrap/crds/policy_upstream_traffic_setting.yaml
```diff
@@ -162,6 +162,242 @@ spec:
...
+                                          description: Key of the descriptor entry.
+                                          type: string
+                                        value:
+                                          description: Value of the descriptor entry.
+                                          type: string
+                            timeout:
+                              description: Timeout (optional) interval for calls to the rate limit service. Defaults to 20ms.
+                              type: string
+                            failOpen:
+                              description: FailOpen (optional) defines whether to allow traffic in case of communication failure between
```
jaellio, line 230: For my own understanding could you explain why the default failOpen value is true?

shashankram: That's the default behavior in Envoy. So we tend to use Envoy defaults for optional values (within the controller) unless there is a reason not to.

cmd/osm-bootstrap/crds/policy_upstream_traffic_setting.yaml
```diff
@@ -225,6 +461,138 @@ spec:
...
+                                description: List of rate limit descriptors to use in the rate limit service request.
+                                type: array
+                                minItems: 1
+                                items:
+                                  description: Rate limit descriptor to use in the rate limit service request.
+                                  type: object
+                                  required:
+                                  - entries
+                                  properties:
+                                    entries:
```
jaellio, line 480: For each "entries" you can only set one of the following: genericKey, remoteAddress, requestHeader, headerValueMatch. Would it make sense for "entries" to be "entry"?

shashankram: No, multiple entries are allowed per descriptor. Entries and each entry needs to exactly map to the Go type schema: https://github.com/openservicemesh/osm/blob/main/pkg/apis/policy/v1alpha1/upstreamtrafficsetting.go#L320.

What's your concern here?

jaellio: I understand that multiple entries can be specified per descriptor. My concern is that the plural "entries" is misleading given only one of the following fields can be set per entry: enericKey, remoteAddress, requestHeader, headerValueMatch.

```
        descriptors:
          # generates descriptor entry [ remote_address=<client-ip from x-forwarded-for header> ]
          - entries:
            - remoteAddress: {}
          - entries:
            # generates descriptor entry [ foo=bar ]
            - genericKey:
                key: foo # optional, default: generic_key
                value: bar
          - entries:
            # generates descriptor entries [ header_match=get, service=<x-service header value> ]
            - headerValueMatch:
              key: header_match # optional, default: header_match
              value: get
              expectMatch: true # optional, default: true
              headers:
                - name: ":method"
                  prefix: /get
```

shashankram: I think there may be a misunderstanding here. The plural `entries` makes sense because there can be multiple descriptor entries as follows:
```
          - entries:
            - genericKey:
                key: foo
                value: bar
            - requestHeader:
                name: x-service
                key: service
            - headerValueMatch:
                key: header_match
                value: get
                expectMatch: true
                headers:
                  - name: ":method"
                    prefix: /get
```

Think of the descriptor entry being polymorphic while being one of `genericKey`, `requestHeader`, `headerValueMatch`, `remoteAddress`.

jaellio: Got it. I misunderstood what the `oneOf` applied to. Thanks!

cmd/osm-bootstrap/crds/policy_upstream_traffic_setting.yaml
jaellio: Got it. I see that Envoy's [`failure_mode_deny` is set to false ](https://www.envoyproxy.io/docs/envoy/v1.23.0/api-v3/extensions/filters/network/thrift_proxy/filters/ratelimit/v3/rate_limit.proto#extensions-filters-network-thrift-proxy-filters-ratelimit-v3-ratelimit) which makes sense. I agree with the default `failOpen` value being true. I was looking around at other implementations and I saw that the default `failOpen` value for [contour's rate limiting service api](https://projectcontour.io/docs/main/config/api/#projectcontour.io/v1alpha1.RateLimitServiceConfig) is false which is interesting. Thanks!

keithmattix (approved):

PR #4937, opened 2022-07-25
rate-limiting: add global L4 rate limiting capability

shashankram:
**Description**:
Adds the capability to enable global L4 rate limiting
for traffic to upstream services. The rate limiting
config is applied via the L4 network global rate
limit filter on the upstream listener.

Part of #2018

**Testing done**:
Tested rate limiting using an external rate limit service. 

**Affected area**:

3. Has documentation corresponding to this change been updated in the [osm-docs](https://github.com/openservicemesh/osm-docs) repo (if applicable)? https://github.com/openservicemesh/osm-docs/pull/425

### DIFF
--- a/pkg/catalog/inbound_traffic_policies.go
+++ b/pkg/catalog/inbound_traffic_policies.go
@@ -310,7 +310,7 @@ func (mc *MeshCatalog) getUpstreamServicesIncludeApex(upstreamServices []service
 	return allServices
 }
 
-// getRateLimitServiceClusters returns a list MeshClusterConfig objects corresponding to the global
+// getRateLimitServiceClusters returns a list of MeshClusterConfig objects corresponding to the global
 // rate limit service instance. It ensures only a single cluster config if the same rate limit service
 // is used for both TCP and HTTP rate limiting.
 func getRateLimitServiceClusters(upstreamTrafficSetting *policyv1alpha1.UpstreamTrafficSetting, clusterSet mapset.Set) []*trafficpolicy.MeshClusterConfig {
--- a/pkg/catalog/inbound_traffic_policies.go
+++ b/pkg/catalog/inbound_traffic_policies.go
@@ -46,6 +46,10 @@ func (mc *MeshCatalog) GetInboundMeshTrafficPolicy(upstreamIdentity identity.Ser
 		upstreamSvcSet.Add(svc)
 	}
 
+	// Used to avoid duplicate clusters that can arise when multiple
+	// upstream services reference the same global rate limit service
+	rlsClusterSet := mapset.NewSet()
+
 	// A policy (traffic match, route, cluster) must be built for each upstream service. This
 	// includes apex/root services associated with the given upstream service.
 	allUpstreamServices := mc.getUpstreamServicesIncludeApex(upstreamServices)
@@ -66,6 +70,7 @@ func (mc *MeshCatalog) GetInboundMeshTrafficPolicy(upstreamIdentity identity.Ser
 
 		upstreamTrafficSetting := mc.policyController.GetUpstreamTrafficSetting(
 			policy.UpstreamTrafficSettingGetOpt{MeshService: &upstreamSvc})
+		clusterConfigs = append(clusterConfigs, getRateLimitServiceClusters(upstreamTrafficSetting, rlsClusterSet)...)
 
 		// ---
 		// Create a TrafficMatch for this upstream servic.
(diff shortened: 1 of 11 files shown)


trstringer (approved):

pkg/trafficpolicy/types.go
```diff
@@ -154,7 +154,12 @@ type MeshClusterConfig struct {
 	EnableEnvoyActiveHealthChecks bool
 
 	// UpstreamTrafficSetting is the traffic setting for the upstream cluster
+	// +optional
 	UpstreamTrafficSetting *policyv1alpha1.UpstreamTrafficSetting
+
+	// Protocol to use for the cluster
```
steeling, line 160: can you add the set of possible options to the comment?

pkg/catalog/inbound_traffic_policies.go
```diff
@@ -304,3 +309,45 @@ func (mc *MeshCatalog) getUpstreamServicesIncludeApex(upstreamServices []service
 
 	return allServices
 }
+
+// getRateLimitServiceClusters returns a list MeshClusterConfig objects corresponding to the global
+// rate limit service instance. It ensures only a single cluster config if the same rate limit service
+// is used for both TCP and HTTP rate limiting.
+func getRateLimitServiceClusters(upstreamTrafficSetting *policyv1alpha1.UpstreamTrafficSetting, clusterSet mapset.Set) []*trafficpolicy.MeshClusterConfig {
+	if upstreamTrafficSetting == nil || upstreamTrafficSetting.Spec.RateLimit == nil || upstreamTrafficSetting.Spec.RateLimit.Global == nil {
+		return nil
```
steeling, line 318: just double checking, it's ok to have a nil cluster?

Tracing through the code looks like there's a few places we can return a nil cluster in `cds.NewResponse`

pkg/envoy/cds/cluster.go
```diff
@@ -314,13 +345,7 @@ func localClustersFromClusterConfigs(configs []*trafficpolicy.MeshClusterConfig)
 }
 
 func getDefaultHTTPProtocolOptions() *extensions_upstream_http.HttpProtocolOptions {
```
steeling, line 347: nit: can you replace calls to `getDefaultHTTPProtocolOptions`, with `getHTTPProtocolOptions("")` (for the purpose of readability)

steeling (commented):
partial review, will review the remainder tomorrow.

non-blocking nits on what i've reviewed so far

pkg/catalog/inbound_traffic_policies.go
```diff
@@ -66,6 +70,7 @@ func (mc *MeshCatalog) GetInboundMeshTrafficPolicy(upstreamIdentity identity.Ser
 
 		upstreamTrafficSetting := mc.policyController.GetUpstreamTrafficSetting(
 			policy.UpstreamTrafficSettingGetOpt{MeshService: &upstreamSvc})
+		clusterConfigs = append(clusterConfigs, getRateLimitServiceClusters(upstreamTrafficSetting, rlsClusterSet)...)
```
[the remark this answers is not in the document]
shashankram: That's implicitly happens in getRateLimitServiceClusters which already happens a comment pertaining to handling duplicates.

pkg/catalog/inbound_traffic_policies.go
shashankram: This returns a nil slice, not a nil cluster. So when we do `append(slice, nil_slice...)`, the result is we don't append anything to the original slice.

pkg/envoy/cds/cluster.go
shashankram: Sure, done.

pkg/trafficpolicy/types.go
shashankram: Done.

pkg/envoy/lds/inmesh.go
```diff
@@ -310,6 +330,59 @@ func buildTCPLocalRateLimitFilter(config *policyv1alpha1.TCPLocalRateLimitSpec,
...
+		descriptors = append(descriptors, &xds_common_ratelimit.RateLimitDescriptor{Entries: entries})
+	}
+	rateLimit.Descriptors = descriptors
+
+	if config.Timeout != nil {
+		rateLimit.Timeout = durationpb.New(config.Timeout.Duration)
+		rateLimit.RateLimitService.GrpcService.Timeout = durationpb.New(config.Timeout.Duration)
+	}
+
+	if config.FailOpen != nil {
```
steeling, line 369: nit: we could make this not a pointer, and use a default in the yaml definition for the CRD

pkg/envoy/lds/inmesh.go
```diff
@@ -88,6 +92,14 @@ func (lb *listenerBuilder) getInboundHTTPFilters(trafficMatch *trafficpolicy.Tra
 		}
 		filters = append(filters, rateLimitFilter)
 	}
+	// Apply the network level global rate limit filter if configured for the TrafficMatch
+	if trafficMatch.RateLimit != nil && trafficMatch.RateLimit.Global != nil && trafficMatch.RateLimit.Global.TCP != nil {
+		rateLimitFilter, err := buildTCPGlobalRateLimitFilter(trafficMatch.RateLimit.Global.TCP, trafficMatch.Name)
```
steeling, line 97: this is all in the func `getInboundHTTPFilters`, should this be using an HTTP Global rate limit? Or will you eventually be checking both TCP and HTTP Rate limit existance here?

If the latter, is it possible to specify both on a single traffic match?

pkg/envoy/lds/inmesh.go
shashankram: That's fine as well. I used this approach to be consistent with other optional fields for which we rely on Envoy defaults.

pkg/envoy/lds/inmesh.go
shashankram: Yes, the TCP rate limiting is applicable to HTTP traffic, so much like local TCP rate limiting filter, the global TCP rate limiting filter can be applied on the HTTP filter_chain as well. Think of it as performing connection level rate limiting at the TCP level, for HTTP traffic.

steeling (approved):

jaellio (approved):

keithmattix (approved):

Issue #4928, opened 2022-07-20
Move init*Monitor to the informer collection


steeling:
Part of #4863

In pkg/k8s/client.go, the init*Monitor funcs should be on the informer collection itself.

PR #4938, opened 2022-07-25
Collapse informer event handlers into the informer collection.

steeling:
This collapses all of the event handler logic into the single informer collection as part of #4928.

This is accomplished by taking the `events.PubSubMessage`, and breaking up the `Kind` into 2 fields

1. The field `Kind` now represents the K8s `Kind`
2. The field `Type` is one of `updated, added, deleted`

By breaking this up we can dynamically determine the message type, and lose a lot of the structure around the `EventFuncs`.

Finally this allows us to remove the message.Broker from the rest of the clients, as it was only used for this purpose initially.


keithmattix:
I'll wait for the PR to be marked as ready before doing a full review, but it's interesting seeing msgBroker be added into the informer collection. I thought we were moving away from msgBroker. Also, I'm not totally sure about the dependency inversion there.

steeling:
> I'll wait for the PR to be marked as ready before doing a full review, but it's interesting seeing msgBroker be added into the informer collection. I thought we were moving away from msgBroker. Also, I'm not totally sure about the dependency inversion there.

Informer Collection and the K8s client will likely fold into each other as well. The message broker itself should probably be revisited after #2683 IMO. 

So this first of PR's is more about moving things around than changing things if that makes sense.

PR #4951, opened 2022-07-28
feat(certs): update certificate Manager using MRC Status changes

### LINKED ISSUE #4815: Configure certificate manager in response to MRC and MRC status changes

To do:
- [ ] Update the issuers in the Manager based on the MRC status changes 

In response to MRC and MRC rotation stage status changes update the `Manager` struct. 

All MRCs events should be considered valid. The validating webhook is responsible for performing validation on adds, updates, and deletions of MRCs.

Outstanding Questions: 
- For the phase 1 of the auto root certificate rotation implementation should the user be responsible for setting the initial status of validatingRollout? Or should OSM handle setting this status? With manual certificate rotation, the user will be responsible for setting the remaining status to progress through rotation. 
  - **Proposed solution**: create a mutating webhook for MRCs. On add, set the `providerState` to `validatingIssuing`. The admission request will be denied by the validating webhook if it shouldn't be created. 
- How to implement controller, injector, and bootstrap specific rotation logic?
  - **Proposed solution**: We could have specific watches for each control plane component, but we'd either have to coordinate manager updates or support updating the manager in multiple places. One alternative is to publish manager updates to msg broker. The OSM component specific logic doesn't need to have any knowledge of the actual MRC updates, it only needs to know that the manager's issuers have been updated and it can reissue certificates.
- Who and when should the error status on the MRC be set?
 
events to handle | response
-- | --
MRC added | **See outstanding question above** - If using the proposed solution above, the ProviderState is `validatingRollout`. Create the Issuer and set it as the `validatingIssuer` in the `Manager` struct 
ProviderState status is validatingRollout | **See above**
ProviderState status is issuingRollout | Verify the ID of the `validatingIssuer` is the same as the ID on the updated MRC.  Set `signingIssuer` to the MRC being rotated in and set the `validatingIssuer` to the MRC being rotated in. <pre> tempIssuer := signingIssuer <br> signingIssuer = validatingIssuer <br> validatingIssuer = tempIssuer </pre> Since the `signingIssuer` and the `validatingIssuer` are not equal the newly issued certs will still contain both CAs for validation. <br><br>Set the `providerState` status of the MRC being rotated out to `issuingRollback`.
ProviderState status is active | Verify the ID of the `signingIssuer` is the same as the ID on the updated MRC. Set the `validatingIssuer` to the `signingIssuer`. At this point, the `signingIssuer` and `validatingIssuer` should be the same. <br><br>Set the `providerState` status of the MRC being rotated out to `validatingRollback`.
ProviderState status is issuingRollback | Noop. Handled by status change to `issuingRollout` on the MRC being rolled out
ProviderState status is validatingRollback | Noop. Handled by status change to `active` on the MRC being rolled out
ProviderState status is inactive | Noop. After the `providerState` status of the MeshRootCertificate has been set to `inactive` it can be deleted
ProviderState status is error | **See outstanding questions above**

jaellio:
**Description**:

Updates the Manager in response to MRC state status changes.

Note: This change is not sufficient for OSM to support OSM root
certificate rotation. More updates are required to support root
certificate rotation in non-Envoy service certificates.

Resolves #4815

**Testing done**:

**Affected area**:

### DIFF
--- a/pkg/certificate/manager.go
+++ b/pkg/certificate/manager.go
@@ -3,6 +3,7 @@ package certificate
 import (
 	"context"
 	"errors"
+	"fmt"
 	"math/rand"
 	"sync"
 	"time"
@@ -117,9 +118,16 @@ func (m *Manager) start(ctx context.Context, mrcClient MRCClient) error {
 func (m *Manager) handleMRCEvent(mrcClient MRCClient, event MRCEvent) error {
 	switch event.Type {
 	case MRCEventAdded:
-		mrc := event.MRC
+		mrc := event.NewMRC
+
+		// ignore add event if MRC has no state
+		if mrc.Status.State == "" {
+			log.Debug().Msgf("received MRC add event for MRC %s. Ignoring because MRC has no state", mrc.GetName())
+			return nil
+		}
+		// ignore add event if MRC state is error
 		if mrc.Status.State == constants.MRCStateError {
-			log.Debug().Msgf("skipping MRC with error state %s", mrc.GetName())
+			log.Debug().Msgf("received MRC add event for MRC %s. Skipping MRC with error state", mrc.GetName())
 			return nil
 		}
 
@@ -129,20 +137,22 @@ func (m *Manager) handleMRCEvent(mrcClient MRCClient, event MRCEvent) error {
 		}
 
 		c := &issuer{Issuer: client, ID: mrc.Name, CertificateAuthority: ca, TrustDomain: mrc.Spec.TrustDomain}
-		switch {
-		case mrc.Status.State == constants.MRCStateActive:
+		switch mrc.Status.State {
+		case constants.MRCStateActive:
 			m.mu.Lock()
 			m.signingIssuer = c
 			m.validatingIssuer = c
 			m.mu.Unlock()
-		case mrc.Status.State == constants.MRCStateIssuingRollback || mrc.Status.State == constants.MRCStateIssuingRollout:
+		case constants.MRCStateIssuingRollout:
 			m.mu.Lock()
 			m.signingIssuer = c
 			m.mu.Unlock()
-		case mrc.Status.State == constants.MRCStateValidatingRollback || mrc.Status.State == constants.MRCStateValidatingRollout:
+		case constants.MRCStateIssuingRollback, constants.MRCStateValidatingRollout:
 			m.mu.Lock()
 			m.validatingIssuer = c
 			m.mu.Unlock()
+		case constants.MRCStateValidatingRollback:
+			// do nothing, the issuer of active MRC should be used
 		default:
 			m.mu.Lock()
 			m.signingIssuer = c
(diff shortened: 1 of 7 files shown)


pkg/certificate/manager.go
```diff
@@ -117,9 +118,16 @@ func (m *Manager) start(ctx context.Context, mrcClient MRCClient) error {
 func (m *Manager) handleMRCEvent(mrcClient MRCClient, event MRCEvent) error {
 	switch event.Type {
 	case MRCEventAdded:
-		mrc := event.MRC
+		mrc := event.NewMRC
+
+		// ignore add event if MRC has no state
+		if mrc.Status.State == "" {
+			log.Debug().Msgf("received MRC add event for MRC %s. Ignoring because MRC has no state", mrc.GetName())
```
jsturtevant, line 125: the logging framework allows for keyvalue pairs with Structured logging.  Do we have any guidance on what we publish?  It might be helpful here to have something like mrc=mrcname when debugging down the line

pkg/certificate/manager.go
```diff
@@ -117,9 +118,16 @@ func (m *Manager) start(ctx context.Context, mrcClient MRCClient) error {
 func (m *Manager) handleMRCEvent(mrcClient MRCClient, event MRCEvent) error {
 	switch event.Type {
 	case MRCEventAdded:
-		mrc := event.MRC
+		mrc := event.NewMRC
+
+		// ignore add event if MRC has no state
```
jsturtevant, line 123: What are the use cases for the MRC having no state?

pkg/certificate/manager.go
```diff
@@ -117,9 +118,16 @@ func (m *Manager) start(ctx context.Context, mrcClient MRCClient) error {
...
+		if mrc.Status.State == "" {
+			log.Debug().Msgf("received MRC add event for MRC %s. Ignoring because MRC has no state", mrc.GetName())
+			return nil
+		}
+		// ignore add event if MRC state is error
 		if mrc.Status.State == constants.MRCStateError {
-			log.Debug().Msgf("skipping MRC with error state %s", mrc.GetName())
+			log.Debug().Msgf("received MRC add event for MRC %s. Skipping MRC with error state", mrc.GetName())
```
jsturtevant, line 130: would we want to raise an event here?

pkg/certificate/manager.go
```diff
@@ -129,34 +137,113 @@ func (m *Manager) handleMRCEvent(mrcClient MRCClient, event MRCEvent) error {
...
 			m.signingIssuer = c
 			m.validatingIssuer = c
 			m.mu.Unlock()
 		}
 	case MRCEventUpdated:
-		// TODO
+		newMRC := event.NewMRC
+		oldMRC := event.OldMRC
+		if newMRC.Status.State != oldMRC.Status.State {
+			// TODO(jaellio): Do I need to validate status updates here to determine if they are allowed?
```
jsturtevant, line 166: are there state transitions that are not allowed?  How does it get to that?

pkg/certificate/manager_test.go
```diff
@@ -318,21 +318,19 @@ func TestIssueCertificate(t *testing.T) {
 
 func TestHandleMRCEvent(t *testing.T) {
```
jsturtevant, line 319:
```suggestion
func TestHandleMRCEventAdded(t *testing.T) {
```

I think all of these tests cases cover the type `MCREventAdded`?

pkg/certificate/manager_test.go
```diff
@@ -342,25 +340,244 @@ func TestHandleMRCEvent(t *testing.T) {
...
+					Spec: v1alpha2.MeshRootCertificateSpec{
+						TrustDomain: "foo.bar.com",
+					},
+					Status: v1alpha2.MeshRootCertificateStatus{
+						State: constants.MRCStateValidatingRollout,
+					},
+				},
+				OldMRC: &v1alpha2.MeshRootCertificate{
+					ObjectMeta: v1.ObjectMeta{
+						Name:      "newmrc",
```
jsturtevant, line 482: should this be `oldmrc` or is this just then status of the MRC?

pkg/certificate/types.go
```diff
@@ -134,15 +134,19 @@ type MRCEventType string
 // MRCEvent describes a change event on a given MRC
 type MRCEvent struct {
 	Type MRCEventType
+
+	// The previous observed version of the MRC as of the time of this event
+	OldMRC *v1alpha2.MeshRootCertificate
+
 	// The last observed version of the MRC as of the time of this event
-	MRC *v1alpha2.MeshRootCertificate
+	NewMRC *v1alpha2.MeshRootCertificate
```
jsturtevant, line 142: Is there any concerns about renaming this? is this exposed outside of the system here?

pkg/certificate/types.go
```diff
@@ -134,15 +134,19 @@ type MRCEventType string
...
 }
 
 var (
 	// MRCEventAdded is the type of announcement emitted when we observe an addition of a Kubernetes MeshRootCertificate
 	MRCEventAdded MRCEventType = "meshrootcertificate-added"
 
-	// MRCEventUpdated is the type of announcement emitted when we observe an update to a Kubernetes MeshRootCertificate
+	// MRCEventUpdated is the type of announcement emitted when we observe a non status update to a Kubernetes MeshRootCertificate
```
jsturtevant, line 149: The tests above are testing changes to the status? under the new test `TestHandleMRCStatusUpdate`

jsturtevant (commented):
new to some of the concepts here so if I asked questions that don't make sense just let me know.

pkg/certificate/manager.go
jaellio: That's a great idea! I'm not sure if we have explicit guidelines. We do this for osm error codes already (for example https://github.com/openservicemesh/osm/blob/2fb95a2159bfc75ebc75891d8170ecdad7d74d66/pkg/envoy/lds/ingress.go#L25)

pkg/certificate/manager.go
jaellio: MRCs with no state are not in use. This allows the user to create an MRC without it automatically triggering a rotation.

pkg/certificate/manager.go
jaellio: I am still working through how to respond an MRC with an error state

pkg/certificate/manager.go
jaellio: The MRC status must be updated in the following order (note that at any time the state can be set to error): "", validatingRollout, issuingRollout, active, issuingRollback, validatingRollback, inactive. The validating webhook should enforce this behavior since the status at this point can only be updated manually and not by OSM. We should never have to handle MRC event here for invalid status changes but it is possible the validator has failed or a MRC has been updated before the validating server has started. The TODO here is to question whether or not we should perform redundant validation

pkg/certificate/manager_test.go
jaellio: Good point :)

pkg/certificate/manager_test.go
jaellio: In this test I think I should just be using "mrc" for the MRC's name. Good point!

pkg/certificate/types.go
jaellio: Good catch! I had explored creating an event type just for status changes and this must be left over from that. Will remove it!

pkg/certificate/manager.go
jsturtevant: cool, I did it in #4952 but not sure if there was any consistency and also saw it in a few other areas of the code base. It would help with operating/debugging if we did define some general guidelines.  for example kubernetes has this: https://github.com/kubernetes/community/blob/a7cb5e7b61ad7e85b60d6beb2170c1edd50a4cb3/contributors/devel/sig-instrumentation/migration-to-structured-logging.md#name-arguments

pkg/certificate/manager.go
jsturtevant: Even though validation *should* catch it in the webhook, I would lead towards doing final validation to ensure we are not moving into a bad workflow state  (i.e moving from `validatingrollout` directly to `active` with out first doing `issuingrollout` if that is the statemachine we are enforcing.

jaellio: Okay, I agree. I'll make those changes one the MRC validation PR is merged.

jsturtevant: this https://github.com/openservicemesh/osm/pull/4811?

Issue #4817, opened 2022-06-14
Update conversion, validating, and mutating webhook configurations based on certificate manager updates


jaellio:
**Please describe the Improvement and/or Feature Request**

Update conversion, validating, and mutating webhook configurations based on certificate manager updates

To do:
- [ ] Subscribe to updates to the conversion, validating, and mutating webhook certificates on the cert rotation chan in the certificate Manager.  On updates, update the respective webhook configuration

**Scope (please mark with X where applicable)**
- New Functionality          [x]
- Certificate Management     [x]

jaellio:
Part of #4839

jsturtevant:
Just added a few notes from looking into this issue:

Webhooks that need updates:

[Mutating Webhook](https://github.com/openservicemesh/osm/blob/8030047c6dd9d237522b63e28fe175ec9a08fb74/pkg/injector/webhook.go#L39)
[CRD conversion webhook](https://github.com/openservicemesh/osm/blob/fa17242a34b39d87b6555774795563aede46efaa/pkg/crdconversion/crdconversion.go#L41)
[Validating webhook](https://github.com/openservicemesh/osm/blob/8030047c6dd9d237522b63e28fe175ec9a08fb74/pkg/validator/server.go#L35)

Rotation subscriptions: https://github.com/openservicemesh/osm/blob/7abf10d787d37cbc79c207888e70248c3a9eef66/pkg/certificate/manager.go#L341

Cert management details: https://release-v1-2.docs.openservicemesh.io/docs/guides/certificates/

rotation steps: https://release-v1-2.docs.openservicemesh.io/docs/guides/certificates/#root-certificate-rotation

jsturtevant:
after looking into I believe this might already be partially implemented via:

https://github.com/openservicemesh/osm/blob/8030047c6dd9d237522b63e28fe175ec9a08fb74/pkg/validator/server.go#L52-L54


and 


https://github.com/openservicemesh/osm/blob/8030047c6dd9d237522b63e28fe175ec9a08fb74/pkg/injector/webhook.go#L352-L361

It looks like the other two web hooks do similar things:

https://github.com/openservicemesh/osm/blob/fa17242a34b39d87b6555774795563aede46efaa/pkg/crdconversion/crdconversion.go#L74
https://github.com/openservicemesh/osm/blob/fa17242a34b39d87b6555774795563aede46efaa/pkg/crdconversion/crdconversion.go#L95-L98

[...]

PR #4952, opened 2022-07-28
feat(certs): update webhooks to handle root certificate rotations

jsturtevant:

**Description**:
Fixes: https://github.com/openservicemesh/osm/issues/4817

This PR subscripts to updates to certificate changes for the webhooks and uses existing code that will update the Webhook CRD configuration.

Builds on functionality added in: https://github.com/openservicemesh/osm/pull/4833 which added ability to patch the configurations but wasn't listing to the cert changes produced.
 

**Testing done**:

- [x] unit testing
- [x] manual testing
- [ ] e2e - not sure if all the parts are in place to do e2e test?
-

**Affected area**:

2. Is this a breaking change?

### DIFF
--- a/pkg/certificate/providers/tresor/fake/fake.go
+++ b/pkg/certificate/providers/tresor/fake/fake.go
@@ -17,14 +17,58 @@ import (
 
 const (
 	rootCertOrganization = "Open Service Mesh Tresor"
+	initialRootName      = "osm-mesh-root-certificate"
 )
 
-type fakeMRCClient struct{}
+type fakeMRCClient struct {
+	mrcChannel chan certificate.MRCEvent
+}
 
+// NewFakeMRC allows for publishing events on to the watch channel to generate MRC events
+func NewFakeMRC() *fakeMRCClient {
+	ch := make(chan certificate.MRCEvent)
+
+	return &fakeMRCClient{
+		mrcChannel: ch,
+	}
+}
+
+// NewCertEvent allows pushing MRC events which can trigger cert changes
+func (c *fakeMRCClient) NewCertEvent(name, state string) {
+	c.mrcChannel <- certificate.MRCEvent{
+		Type: certificate.MRCEventAdded,
+		MRC: &v1alpha2.MeshRootCertificate{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      name,
+				Namespace: "osm-system",
+			},
+			Spec: v1alpha2.MeshRootCertificateSpec{
+				TrustDomain: "cluster.local",
+				Provider: v1alpha2.ProviderSpec{
+					Tresor: &v1alpha2.TresorProviderSpec{
+						CA: v1alpha2.TresorCASpec{
+							SecretRef: v1.SecretReference{
+								Name:      "osm-ca-bundle",
+								Namespace: "osm-system",
+							},
+						},
+					},
+				},
+			},
+			Status: v1alpha2.MeshRootCertificateStatus{
+				State: state,
+			},
+		},
+	}
+}
+
+// GetCertIssuerForMRC will return a root cert for testing.
 func (c *fakeMRCClient) GetCertIssuerForMRC(mrc *v1alpha2.MeshRootCertificate) (certificate.Issuer, pem.RootCertificate, error) {
 	rootCertCountry := "US"
 	rootCertLocality := "CA"
-	ca, err := tresor.NewCA("Fake Tresor CN", 1*time.Hour, rootCertCountry, rootCertLocality, rootCertOrganization)
+	cn := certificate.CommonName(mrc.Name)
+
+	ca, err := tresor.NewCA(cn, 1*time.Hour, rootCertCountry, rootCertLocality, rootCertOrganization)
 	if err != nil {
 		return nil, nil, err
 	}
@@ -39,37 +83,12 @@ func (c *fakeMRCClient) List() ([]*v1alpha2.MeshRootCertificate, error) {
 }
 
 func (c *fakeMRCClient) Watch(ctx context.Context) (<-chan certificate.MRCEvent, error) {
-	ch := make(chan certificate.MRCEvent)
+	// send event for first CA created
 	go func() {
-		ch <- certificate.MRCEvent{
-			Type: certificate.MRCEventAdded,
-			MRC: &v1alpha2.MeshRootCertificate{
-				ObjectMeta: metav1.ObjectMeta{
-					Name:      "osm-mesh-root-certificate",
-					Namespace: "osm-system",
-				},
-				Spec: v1alpha2.MeshRootCertificateSpec{
-					TrustDomain: "cluster.local",
-					Provider: v1alpha2.ProviderSpec{
-						Tresor: &v1alpha2.TresorProviderSpec{
-							CA: v1alpha2.TresorCASpec{
-								SecretRef: v1.SecretReference{
-									Name:      "osm-ca-bundle",
-									Namespace: "osm-system",
-								},
-							},
-						},
-					},
-				},
-				Status: v1alpha2.MeshRootCertificateStatus{
-					State: constants.MRCStateActive,
-				},
-			},
-		}
-		close(ch)
+		c.NewCertEvent(initialRootName, constants.MRCStateActive)
 	}()
 
-	return ch, nil
+	return c.mrcChannel, nil
 }
 
 // NewFake constructs a fake certificate client using a certificate
@@ -80,7 +99,18 @@ func NewFake(checkInterval time.Duration) *certificate.Manager {
 
 // NewFakeWithValidityDuration constructs a fake certificate manager with specified cert validity duration
 func NewFakeWithValidityDuration(getCertValidityDuration func() time.Duration, checkInterval time.Duration) *certificate.Manager {
-	tresorCertManager, err := certificate.NewManager(context.Background(), &fakeMRCClient{}, getCertValidityDuration, getCertValidityDuration, checkInterval)
+	tresorCertManager, err := certificate.NewManager(context.Background(), NewFakeMRC(), getCertValidityDuration, getCertValidityDuration, checkInterval)
+	if err != nil {
+		log.Error().Err(err).Msg("error encountered creating fake cert manager")
+		return nil
+	}
+	return tresorCertManager
+}
+
+// NewFakeWithMRC constructs a fake certificate manager with specified cert validity duration and fake MRC client
+func NewFakeWithMRC(fakeMRCClient *fakeMRCClient, checkInterval time.Duration) *certificate.Manager {
+	getValidityDuration := func() time.Duration { return 1 * time.Hour }
+	tresorCertManager, err := certificate.NewManager(context.Background(), fakeMRCClient, getValidityDuration, getValidityDuration, checkInterval)
 	if err != nil {
 		log.Error().Err(err).Msg("error encountered creating fake cert manager")
 		return nil
--- a/pkg/webhook/server.go
+++ b/pkg/webhook/server.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/openservicemesh/osm/pkg/certificate"
 	"github.com/openservicemesh/osm/pkg/constants"
+	"github.com/openservicemesh/osm/pkg/k8s/events"
 	"github.com/openservicemesh/osm/pkg/metricsstore"
 )
 
@@ -19,6 +20,7 @@ type CertRotatedFunc func(cert *certificate.Certificate) error
 // Server is a construct to run generic HTTPS webhook servers.
 type Server struct {
 	name         string
+	namespace    string
 	cm           *certificate.Manager
 	server       *http.Server
 	onCertChange CertRotatedFunc
@@ -28,7 +30,7 @@ type Server struct {
 }
 
 // NewServer returns a new server based on the input. Run() must be called to start the server.
-func NewServer(name, namespace string, port int, cm *certificate.Manager, handlers map[string]http.HandlerFunc, onCertChange CertRotatedFunc) (*Server, error) {
+func NewServer(name, namespace string, port int, cm *certificate.Manager, handlers map[string]http.HandlerFunc, onCertChange CertRotatedFunc) *Server {
 	mux := http.NewServeMux()
 
 	for path, h := range handlers {
@@ -37,6 +39,7 @@ func NewServer(name, namespace string, port int, cm *certificate.Manager, handle
 
 	s := &Server{
 		name:         name,
+		namespace:    namespace,
 		cm:           cm,
 		onCertChange: onCertChange,
 	}
@@ -54,38 +57,41 @@ func NewServer(name, namespace string, port int, cm *certificate.Manager, handle
 			MinVersion: constants.MinTLSVersion,
 		},
 	}
-	// set the certificate once, which will also call onCertChange.
-	if err := s.setCert(name, namespace); err != nil {
-		return nil, err
-	}
-	return s, nil
+	return s
 }
 
-// Run actually starts the server. It blocks until the passed in context is done.
-func (s *Server) Run(ctx context.Context) {
+// Run actually starts the server.
+func (s *Server) Run(ctx context.Context) error {
+	if err := s.configureCertificateRotation(ctx); err != nil {
+		return err
+	}
+
 	log.Info().Msgf("Starting %s webhook server on: %s", s.name, s.server.Addr)
 	go func() {
 		err := s.server.ListenAndServeTLS("", "") // err is always non-nil
 		log.Error().Err(err).Msgf("%s webhook HTTP server shutdown", s.name)
 	}()
 
-	// Wait on exit signals
-	<-ctx.Done()
+	go func() {
+		// Wait on exit signals
+		<-ctx.Done()
 
-	// Stop the servers
-	if err := s.server.Shutdown(context.TODO()); err != nil {
-		log.Error().Err(err).Msgf("Error shutting down %s webhook HTTP server", s.name)
-	} else {
-		log.Info().Msgf("Done shutting down %s webhook HTTP server", s.name)
-	}
+		// Stop the servers
+		if err := s.server.Shutdown(context.TODO()); err != nil {
+			log.Error().Err(err).Msgf("Error shutting down %s webhook HTTP server", s.name)
+		} else {
+			log.Info().Msgf("Done shutting down %s webhook HTTP server", s.name)
+		}
+	}()
+	return nil
 }
 
-func (s *Server) setCert(name, namespace string) error {
+func (s *Server) setCert() error {
 	// This is a certificate issued for the webhook handler
 	// This cert does not have to be related to the Envoy certs, but it does have to match
 	// the cert provisioned.
 	webhookCert, err := s.cm.IssueCertificate(
-		fmt.Sprintf("%s.%s.svc", name, namespace),
+		s.certCommonName(),
 		certificate.Internal,
 		certificate.FullCNProvided())
 	if err != nil {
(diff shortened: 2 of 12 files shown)


pkg/certificate/providers/tresor/fake/fake.go
```diff
@@ -80,7 +105,18 @@ func NewFake(checkInterval time.Duration) *certificate.Manager {
...
+	tresorCertManager, err := certificate.NewManager(context.Background(), NewFakeMRC(), getCertValidityDuration, getCertValidityDuration, checkInterval)
+	if err != nil {
+		log.Error().Err(err).Msg("error encountered creating fake cert manager")
+		return nil
+	}
+	return tresorCertManager
+}
+
+// NewFakeWithMrc constructs a fake certificate manager with specified cert validity duration and fake MRC client
+func NewFakeWithMrc(fakeMRCClient *fakeMRCClient, checkInterval time.Duration) *certificate.Manager {
```
jaellio, line 117: nit:
```suggestion
// NewFakeWithMRC constructs a fake certificate manager with specified cert validity duration and fake MRC client
func NewFakeWithMRC(fakeMRCClient *fakeMRCClient, checkInterval time.Duration) *certificate.Manager {
```

jsturtevant:
note: need to deploy with ` --set=osm.featureFlags.enableMeshRootCertificate=true`

Create a new MRC by using `kubectl deploy -f` with

```
apiVersion: config.openservicemesh.io/v1alpha2
kind: MeshRootCertificate
metadata:
  name: osm-mesh-root-certificate-v2
  namespace: osm-system
spec:
  provider:
    tresor:
      ca:
        secretRef:
          name: osm-ca-bundle-v2
          namespace: osm-system
  trustDomain: cluster.local
```

Did some validation with the demo deployment.  

The validation webhook:

```
Defaulted container "osm-controller" out of: osm-controller, init-osm-controller (init)
{"level":"info","component":"webhook","time":"2022-07-29T02:25:28Z","file":"server.go:69","message":"Starting osm-validator webhook server on test: :9093"}
{"level":"info","component":"webhook","webhook":"osm-validator","cn":"osm-validator.osm-system.svc","time":"2022-07-29T02:25:28Z","file":"server.go:116","message":"Listing for certificate rotations"}
{"level":"debug","component":"webhook","webhook":"osm-validator","cn":"osm-validator.osm-system.svc","time":"2022-07-29T02:25:32Z","file":"server.go:123","message":"Certificate rotation was initiated for webhook"}
{"level":"info","component":"webhook","webhook":"osm-validator","cn":"osm-validator.osm-system.svc","time":"2022-07-29T02:25:32Z","file":"server.go:128","message":"Certificate rotated for webhook"}
{"level":"trace","component":"osm-validator","time":"2022-07-29T15:59:41Z","file":"server.go:67","message":"Received validating webhook request: Method=POST, URL=/validate?timeout=10s"}
{"level":"trace","component":"osm-validator","time":"2022-07-29T15:59:41Z","file":"server.go:67","message":"Received validating webhook request: Method=POST, URL=/validate?timeout=10s"}
{"level":"trace","component":"osm-validator","time":"2022-07-29T15:59:41Z","file":"server.go:67","message":"Received validating webhook request: Method=POST, URL=/validate?timeout=10s"}
```

And we can see that the webhook is showing the v2 cert:

```
❯ kubectl get secret -n osm-system osm-ca-bundle -o jsonpath='{.data.ca\.crt}' | base64 -d | openssl x509 -noout -modulus
Modulus=A67D65484D049C76A0006E328E2D902FF260094237E649E2342F386B027EC1CC890FE5C88D8C3BECB61D4EC267FF15C5DCB66D6C3A9FCCA80F424339CF631C036BBDE779F862F3755FFC1053BA747000CFD3BCB7B88BF8626055EC4F14A54BBE70579DB2C17FC0809FB420E006DD157966C770CFDB406E62FED49A0370935A53EC9047AFFE5CA417DBBC2BF30BE9EEC4578DD6312914ECC0D7CE9B3A77140B863B9BA047C093B0204427F6EA79CB1F7EBCD0CADA59DDC553FCE9152086CE1E96AA31148871C65F0D79173393B32CB7C9C22F1EB9F11CC841F200E036C943FB9D0119F82D077E9FF49D16314980D8CA4921B94F524E4E39536E8316E5FBDAC899

osm on  feat-certs-webhook-rotation 
❯ k get validatingwebhookconfigurations.admissionregistration.k8s.io osm-validator-mesh-osm -ojson | jq -r '.webhooks[] | .clientConfig.caBundle' | base64 --decode | openssl x509 -noout -modulus
Modulus=97B8D8A089D67137F5B56B6D591412599B3700FD2D26A0C41E285DB607A045CD92A50ED5AF9B47E5E7CA4A983962989818FD110A53C1A94BB815B3F8EE2398AB580A933179FD626CBD782824253E8DB1358907C8F2E14EE6A4162F37FAFD07B01B0FFF371EBA2CFD8F1F88DB15E68B86B24C64F3D559B9AC575D07E66F69D8D0F39C39448C64CC319E8D9D995457CDF88F2E9945958DB803A018EA60FFC528040E3D60B0F7F710BDA44E7ED871E921B183AC1974A9BC9C3B7499EB1F36A07964E44A03D1B30F789B42E84E26AABEF47311EF60823C005F9300B6B44B1010BC98484A6599E32D4A599D571813998044763EC712E449CD5C135668CE1CCF2A0A7D

osm on  feat-certs-webhook-rotation
❯ kubectl get secret -n osm-system osm-ca-bundle-v2 -o jsonpath='{.data.ca\.crt}' | base64 -d | openssl x509 -noout -modulus
Modulus=97B8D8A089D67137F5B56B6D591412599B3700FD2D26A0C41E285DB607A045CD92A50ED5AF9B47E5E7CA4A983962989818FD110A53C1A94BB815B3F8EE2398AB580A933179FD626CBD782824253E8DB1358907C8F2E14EE6A4162F37FAFD07B01B0FFF371EBA2CFD8F1F88DB15E68B86B24C64F3D559B9AC575D07E66F69D8D0F39C39448C64CC319E8D9D995457CDF88F2E9945958DB803A018EA60FFC528040E3D60B0F7F710BDA44E7ED871E921B183AC1974A9BC9C3B7499EB1F36A07964E44A03D1B30F789B42E84E26AABEF47311EF60823C005F9300B6B44B1010BC98484A6599E32D4A599D571813998044763EC712E449CD5C135668CE1CCF2A0A7D
```

jsturtevant:
same for the other crd and mutating hooks:

```
{"level":"debug","component":"webhook","webhook":"osm-bootstrap","cn":"osm-bootstrap.osm-system.svc","time":"2022-07-29T17:25:01Z","file":"server.go:123","message":"Certificate rotation was initiated for webhook"}                                                                                                                                                                                                           
{"level":"info","component":"crd-conversion","time":"2022-07-29T17:25:01Z","file":"crdconversion.go:100","message":"successfully updated conversion webhook configuration for crd : meshconfigs.config.openservicemesh.io"}                                                                                                                                                                                                     
{"level":"info","component":"webhook","webhook":"osm-bootstrap","cn":"osm-bootstrap.osm-system.svc","time":"2022-07-29T17:25:01Z","file":"server.go:128","message":"Certificate rotated for webhook"}  
```

```
{"level":"info","component":"webhook","webhook":"osm-injector","cn":"osm-injector.osm-system.svc","time":"2022-07-29T17:21:39Z","file":"server.go:116","message":"Listing for certificate rotations"}
{"level":"debug","component":"webhook","webhook":"osm-injector","cn":"osm-injector.osm-system.svc","time":"2022-07-29T17:21:43Z","file":"server.go:123","message":"Certificate rotation was initiated for webhook"}                                                                                                                                                                                                             
{"level":"info","component":"sidecar-injector","time":"2022-07-29T17:21:43Z","file":"webhook.go:370","message":"Finished creating MutatingWebhookConfiguration osm-webhook-osm"}
{"level":"info","component":"webhook","webhook":"osm-injector","cn":"osm-injector.osm-system.svc","time":"2022-07-29T17:21:43Z","file":"server.go:128","message":"Certificate rotated for webhook"} 
```

pkg/webhook/server.go
```diff
@@ -54,14 +56,16 @@ func NewServer(name, namespace string, port int, cm *certificate.Manager, handle
 		},
 	}
 	// set the certificate once, which will also call onCertChange.
-	if err := s.setCert(name, namespace); err != nil {
+	if err := s.setCert(); err != nil {
```
jaellio, line 59: Should we subscribe to the cert rotation pub sub here before ever issuing the webhook cert so we don't miss any rotations?

pkg/certificate/providers/tresor/fake/fake.go
```diff
@@ -17,14 +18,63 @@ import (
...
+
+// GetCertIssuerForMRC will return a root cert for testing. If called with MRCStateIssuingRollout it creates a
+// new cert with a different issuer name based on mrc name.
 func (c *fakeMRCClient) GetCertIssuerForMRC(mrc *v1alpha2.MeshRootCertificate) (certificate.Issuer, pem.RootCertificate, error) {
 	rootCertCountry := "US"
 	rootCertLocality := "CA"
-	ca, err := tresor.NewCA("Fake Tresor CN", 1*time.Hour, rootCertCountry, rootCertLocality, rootCertOrganization)
+	cn := certificate.CommonName("Fake Tresor CN")
+	if mrc.Status.State == constants.MRCStateIssuingRollout {
+		// this will create a new issuer ID which causes a
```
shalier, line 73: Incomplete comment

pkg/certificate/providers/tresor/fake/fake.go
```diff
@@ -17,14 +18,63 @@ import (
...
+			},
+			Status: v1alpha2.MeshRootCertificateStatus{
+				State: state,
+			},
+		},
+	}
+}
+
+// GetCertIssuerForMRC will return a root cert for testing. If called with MRCStateIssuingRollout it creates a
+// new cert with a different issuer name based on mrc name.
```
shalier, line 67: Nit:
```suggestion
// new cert with a different issuer name based on the MRC name.
```

pkg/webhook/server.go
```diff
@@ -103,3 +107,27 @@ func (s *Server) setCert(name, namespace string) error {
...
+	log.Info().Str("webhook", s.name).Str("cn", s.certCommonName()).Msg("Listing for certificate rotations")
+	certRotationChan, unsub := s.cm.SubscribeRotations(s.certCommonName())
+	defer unsub()
+
+	for {
+		select {
+		case <-certRotationChan:
+			log.Debug().Str("webhook", s.name).Str("cn", s.certCommonName()).Msg("Certificate rotation was initiated for webhook")
+			if err := s.setCert(); err != nil {
+				// raise event?
```
shalier, line 125: Could create an issue to discuss this and add a TODO comment

pkg/webhook/server_test.go
```diff
@@ -0,0 +1,108 @@
...
+			checkInterval:     10 * time.Millisecond,
+			waitForRotation:   true,
+		},
+		{name: "should not rotate certificate if checkInterval is longer than rotation",
+			rotations:         1,
+			rotationsExpected: 0,
+			checkInterval:     1 * time.Hour,
+			waitForRotation:   false,
+		},
+		{name: "should not rotate certificate if checkInterval is longer than rotation",
```
shalier, line 48: Could this be named to be different from the previous test at L42?

pkg/webhook/server.go
steeling: Ya this is a broader problem we have throughout the code base.

Would you mind filing an issue along the lines of "Handle/record metric for errors better", and add a TODO here with the issue #

pkg/certificate/providers/tresor/fake/fake.go
```diff
@@ -17,14 +18,63 @@ import (
...
+func NewFakeMRC() *fakeMRCClient {
+	ch := make(chan certificate.MRCEvent)
 
+	return &fakeMRCClient{
+		mrcChannel: ch,
+	}
+}
+
+// NewCertEvent allows pushing MRC events which can trigger cert changes
+func (c *fakeMRCClient) NewCertEvent(name, state string) {
```
steeling, line 38: This works, alternatively you could leverage a real informer with a fake client, and create new MRC's which would trigger the event.

Don't really have an opinion just wanted to present the alternative :)

pkg/webhook/server.go
```diff
@@ -103,3 +107,27 @@ func (s *Server) setCert(name, namespace string) error {
...
+				log.Error().Err(err).Str("webhook", s.name).Msgf("Error rotating certificate for webhook")
+			}
+			log.Info().Str("webhook", s.name).Str("cn", s.certCommonName()).Msg("Certificate rotated for webhook")
+		case <-ctx.Done():
+			return
```
steeling, line 130: maybe log that we're exiting?

pkg/webhook/server_test.go
```diff
@@ -0,0 +1,108 @@
...
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			assert := tassert.New(t)
+
+			mrc := tresorFake.NewFakeMRC()
+			cm := tresorFake.NewFakeWithMRC(mrc, tc.checkInterval)
+
+			count := 0
+			wait := make(chan struct{})
```
steeling, line 64: This could be simplified a tad, but not blocking on the review if you prefer to keep it.


Due to semantics of certmanager, you can "pre-issue" the certificate, by calling `firstCert, err := cm.IssueCertificate(...)`

That saves the use of needing to wait for the first onRotation to be called.

Now you can switch the chan stuff to a sync.WaitGroup. Makes things a bit more readable, but does depend on knowing the internals of certmanager. Just a thought!

steeling (approved):

pkg/webhook/server.go
jsturtevant: https://github.com/openservicemesh/osm/issues/4956

pkg/webhook/server.go
jsturtevant: good catch, I added a test case that demonstrates this issue and working through a solution.

jsturtevant: thoughts?

pkg/webhook/server_test.go
jsturtevant: I think `onRotation` isn't named right here, it should be `onCertChange` (as the parameter is called in `NewServer`). 

Since this gets called for every instance, this if I pre-issue the cert it will this function is still called, but I tried out the waitgroup version and it is a bit simpler I think. Thanks!

pkg/webhook/server.go
jaellio: Could we add a health probe for the validating webhook server to indicate if we failed to issue the certificate and therefore failed to run the server?

pkg/certificate/providers/tresor/fake/fake.go
jsturtevant: makes sense, I didn't see the informer usage so didn't introduce it.  As we add more scenarios we may want to go that route

pkg/webhook/server.go
```diff
@@ -103,3 +111,40 @@ func (s *Server) setCert(name, namespace string) error {
 
 	return s.onCertChange(webhookCert)
 }
+
+func (s *Server) certCommonName() string {
+	return fmt.Sprintf("%s.%s.svc", s.name, s.namespace)
+}
+
+// configureCertificateRotation gets the certificate from the certificate manager and spawns a goroutine to watch for certificate rotation.
+func (s *Server) configureCertificateRotation() func() {
```
jsturtevant, line 120: This is using the same pattern in https://github.com/openservicemesh/osm/blob/7abf10d787d37cbc79c207888e70248c3a9eef66/pkg/ingress/gateway.go#L116. 

There might be opportunity to refactor and re-use.

pkg/webhook/server.go
jsturtevant: good idea, adding health probe would be helpful in both scenarios

jaellio: I'm not sure if there is more context around why we haven't chosen to do this in the past. @steeling do you know?

steeling: there's a handful of places we silently ignore failures. This one is seems more critical in that we should panic on an error

pkg/webhook/server.go
```diff
@@ -103,3 +111,40 @@ func (s *Server) setCert(name, namespace string) error {
...
+		for {
+			select {
+			case <-certRotationChan:
+				log.Debug().Str("webhook", s.name).Str("cn", s.certCommonName()).Msg("Certificate rotation was initiated for webhook")
+				if err := s.setCert(); err != nil {
+					// TODO(#4956): raise event for failures
+					log.Error().Err(err).Str("webhook", s.name).Msgf("Error rotating certificate for webhook")
+				}
+				log.Info().Str("webhook", s.name).Str("cn", s.certCommonName()).Msg("Certificate rotated for webhook")
+			case <-stop:
```
steeling, line 137: We'll need to pass in a new stop as well, since we're no longer returning a cancel func. Using the context from Run could work

jsturtevant: I don't think we can use the context from the Run (this is what I initially did) because it will miss the cert rotations since the cert is set during the call to `NewWebhook`. One option is to move the initial cert creation to run but this changes the current behavior (https://github.com/openservicemesh/osm/pull/4952/files#r935739443).  

Should we change that behavior and panic if we can call [Tresor](https://github.com/openservicemesh/osm/pull/4952/files#r937169543)? In the case of certmanager and other external system this could happen pretty regularly and retries might be a better option with a healthprobe.

steeling: could remove run, and call go run() from within NewServer. 

Just a thought

jaellio (approved):

jsturtevant:
/hold
Going to see what this looks like if I merge run with and NewServer

pkg/certificate/providers/tresor/fake/fake.go
```diff
@@ -17,14 +18,63 @@ import (
...
+}
+
+// GetCertIssuerForMRC will return a root cert for testing. If called with MRCStateIssuingRollout it creates a
+// new cert with a different issuer name based on the MRC name.
 func (c *fakeMRCClient) GetCertIssuerForMRC(mrc *v1alpha2.MeshRootCertificate) (certificate.Issuer, pem.RootCertificate, error) {
 	rootCertCountry := "US"
 	rootCertLocality := "CA"
-	ca, err := tresor.NewCA("Fake Tresor CN", 1*time.Hour, rootCertCountry, rootCertLocality, rootCertOrganization)
+	cn := certificate.CommonName("Fake Tresor CN")
+	if mrc.Status.State == constants.MRCStateIssuingRollout {
```
jaellio, line 72: Could you elaborate on why the issueingRollout state needs to be handled uniquely? Could the id for every issuer be the mrc name?

steeling: i believe it's to indicate that the issuer trust domain has changed, thereby changing the CN of the CA.

Although it doesn't compare the CA so there's actually nothing to do here.. the actual manager will handle that logic properly

(If i'm understanding the reasoning here)

jsturtevant: yes, I didn't want to put too much logic in here since it's a test function and this was enough to cause the rotation

jsturtevant: spoke with @jaellio who helped me understand the rotation logic a bit better and found this check isn't needed to fire the rotation so removed it.

pkg/webhook/server.go
jsturtevant: I tested this out some more and found I was wrong. the `events.GenericEventRecorder().FatalEvent` forces a webhook shut down so it goes into a crashloop.  I've updated everything to continue to work that way.

jaellio:
LGTM!

Issue #4929, opened 2022-07-20
Remove the configurator object


steeling:
The configurator object essentially provides 2 useful methods:

1. GetMeshConfig
2. GetOSMNamespace

These 2 methods should be moved to the pkg/providers/kube client object.

The rest of the methods are convenience field accessors on the MeshConfig object itself. These can be moved to the meshconfig object, such that new calls will be:

provider.GetMeshConfig().GetServiceValidityDuration().

This is part of #4863

PR #4953, opened 2022-07-29
Migrate MeshConfig Getters out of the configurator

steeling:
Move the MeshConfig getters onto the meshconfig itself, and out of the configurator class. This is part of [#4929](https://github.com/openservicemesh/osm/issues/4929), and will make it simpler to move the remaining methods into the k8s controller.

It makes the interface significantly smaller


shashankram (dismissed):
I believe we discussed the idea of moving the Configurator implementation into the autogenerated API code part in the past, and I was against this for several reasons:
1. It results in extra maintenance, as every time a new API version is added/updated/removed, we have to manually copy over these non-autogenerated code over. I think it's best to keep the autogenerated code separate and keep custom wrapper implementations used by the controller within the controller code, and not `pkg/apis`.
2. This couples controller defaults and fallback behavior (on nil, empty strings, etc.) into the API code. I would argue this is confusing from an API perspective, as such defaults should be correctly encoded in the API. For e.g., when the init container image is unset, it returns `OSM_DEFAULT_INIT_CONTAINER_IMAGE`. This behavior is not intuitive from an API perspective. Such instances are better handled within the controller code.
3. I don't believe we need to expose multiple sub-fields via methods on a large type such as `MeshConfig`. Clients can directly reference the fields regardless of the nesting level. We should aim to reduce the number of accessors exposed. For e.g., the k8s `corev1.Pod` doesn't have unnecessary wrappers that client implementations may require. I would suggest following the k8s API style for the Go types as much as possible.

Can we first make a pass on reducing the number of getters on the Configurator type, and then see if having specific accessors makes sense?

steeling:
> I believe we discussed the idea of moving the Configurator implementation into the autogenerated API code part in the past, and I was against this for several reasons:
> 
> 1. It results in extra maintenance, as every time a new API version is added/updated/removed, we have to manually copy over these non-autogenerated code over. I think it's best to keep the autogenerated code separate and keep custom wrapper implementations used by the controller within the controller code, and not `pkg/apis`.
> 2. This couples controller defaults and fallback behavior (on nil, empty strings, etc.) into the API code. I would argue this is confusing from an API perspective, as such defaults should be correctly encoded in the API. For e.g., when the init container image is unset, it returns `OSM_DEFAULT_INIT_CONTAINER_IMAGE`. This behavior is not intuitive from an API perspective. Such instances are better handled within the controller code.
> 3. I don't believe we need to expose multiple sub-fields via methods on a large type such as `MeshConfig`. Clients can directly reference the fields regardless of the nesting level. We should aim to reduce the number of accessors exposed. For e.g., the k8s `corev1.Pod` doesn't have unnecessary wrappers that client implementations may require. I would suggest following the k8s API style for the Go types as much as possible.
> 
> Can we first make a pass on reducing the number of getters on the Configurator type, and then see if having specific accessors makes sense?

done, moved remaining accessors to the utils package

pkg/k8s/events/types.go
```diff
@@ -30,7 +39,145 @@ const (
 
 // PubSubMessage represents a common messages abstraction to pass through the PubSub interface
 type PubSubMessage struct {
-	Kind   announcements.Kind
+	Kind   Kind
+	Type   EventType
 	OldObj interface{}
 	NewObj interface{}
 }
+
```
shashankram, line 47: Are these event related changes accidentally included in this PR?

shashankram (commented):
Could you drop the event related changes in this PR as I believe that is unrelated to the MeshConfig changes.

pkg/k8s/events/types.go
steeling: Those got caught in here somehow. I'll  fix once https://github.com/openservicemesh/osm/pull/4939 is submitted

steeling:
> Could you drop the event related changes in this PR as I believe that is unrelated to the MeshConfig changes.

rebased and ready to go!

keithmattix (approved):
Looks like this is functionally a no-op, so I'm approving

pkg/utils/mesh_config.go
```diff
@@ -0,0 +1,152 @@
...
+
+const (
+	// defaultServiceCertValidityDuration is the default validity duration for service certificates
+	defaultServiceCertValidityDuration = 24 * time.Hour
+
+	// defaultIngressGatewayCertValidityDuration is the default validity duration for ingress gateway certificates
+	defaultIngressGatewayCertValidityDuration = 24 * time.Hour
+
+	// DefaultCertKeyBitSize is the default certificate key bit size
+	DefaultCertKeyBitSize = 2048
```
allenlsy, line 22: nit: no need to export this?

allenlsy (approved):

pkg/utils/mesh_config.go
steeling: good call, fixed

shashankram (approved):

jaellio (approved):

Issue #4923, opened 2022-07-20
Leverage Builder pattern for EDS


steeling:
Subtask of #4862, this is specifically for the EDS configs.

We should create a builder object that takes as inputs all data, and calls build. It should not make any network calls, or call out to other methods to get *more data*.

It *can* use nested builders if it makes sense to do so. See the parent issue and associated design doc for more info.

allenlsy:
https://github.com/openservicemesh/osm/pull/4957

PR #4957, opened 2022-08-01
ref: use builder pattern in EDS resources generation

allenlsy:
**Description**:

Apply builder pattern on EDS resource generation. 

The purpose of the refactoring is to decouple the external resource fetching, core business logic and intermediate result generation.

Resolves issue #4923 

**Testing done**:

No feature changes. 

- [x] Unit test

**Affected area**:

No

2. Is this a breaking change?

No

No

### DIFF
--- a/pkg/envoy/eds/builder.go
+++ b/pkg/envoy/eds/builder.go
@@ -22,6 +22,17 @@ type endpointsBuilder struct {
 	upstreamSvcEndpoints map[service.MeshService][]endpoint.Endpoint
 }
 
+func newEndpointsBuilder() *endpointsBuilder {
+	return &endpointsBuilder{
+		upstreamSvcEndpoints: make(map[service.MeshService][]endpoint.Endpoint),
+	}
+}
+
+func (b *endpointsBuilder) AddEndpoints(svc service.MeshService, endpoints []endpoint.Endpoint) {
+	b.upstreamSvcEndpoints[svc] = endpoints
+}
+
+// Build generate Envoy endpoint resources based on stored endpoints
 func (b *endpointsBuilder) Build() []types.Resource {
 	var edsResources []types.Resource
 
--- a/pkg/envoy/eds/builder.go
+++ b/pkg/envoy/eds/builder.go
@@ -3,6 +3,7 @@ package eds
 import (
 	xds_core "github.com/envoyproxy/go-control-plane/envoy/config/core/v3"
 	xds_endpoint "github.com/envoyproxy/go-control-plane/envoy/config/endpoint/v3"
+	"github.com/envoyproxy/go-control-plane/pkg/cache/types"
 
 	"github.com/golang/protobuf/ptypes/wrappers"
 
@@ -16,6 +17,19 @@ const (
 	localClusterPriority  = uint32(0)
 	remoteClusterPriority = uint32(1)
 )
+
+type endpointsBuilder struct {
+	upstreamSvcEndpoints map[service.MeshService][]endpoint.Endpoint
+}
+
+func (b *endpointsBuilder) Build() []types.Resource {
+	var edsResources []types.Resource
+
+	for svc, endpoints := range b.upstreamSvcEndpoints {
+		edsResources = append(edsResources, newClusterLoadAssignment(svc, endpoints))
+	}
+	return edsResources
+}
 
 // newClusterLoadAssignment returns the cluster load assignments for the given service and its endpoints
 func newClusterLoadAssignment(svc service.MeshService, serviceEndpoints []endpoint.Endpoint) *xds_endpoint.ClusterLoadAssignment {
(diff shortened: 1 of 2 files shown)


shashankram (changes requested):
It doesn't seem like this change simplifies the existing code, rather it wraps the existing code with the builder type, which seems unnecessary in its current form. The builder pattern is most useful when generic complex objects, but it seems like the existing code here is simple enough to not even require the additional wrapping.

Let's update the code to something as follows:
```
b := EndpointsBuilder()
for something : range requestedResources {
    endpoints := getEndpoints(something)
    b.AddEndpoint(meshSvc, endpoints)
}

return b.Build()
```

pkg/envoy/eds/response.go
```diff
@@ -35,7 +35,8 @@ func fulfillEDSRequest(meshCatalog catalog.MeshCataloger, proxy *envoy.Proxy, re
 		return nil, fmt.Errorf("Endpoint discovery request for proxy %s cannot be nil", proxy.Identity)
```
steeling, line 35: Let's remove fulfillEDSRequest and generateEDSConfig, and merge them directly into NewResponse

steeling: similarly i think we can also collapse `getUpstreamEndpointsForProxyIdentity` into NewResponse

pkg/envoy/eds/builder.go
```diff
@@ -17,6 +18,19 @@ const (
...
 
+type endpointsBuilder struct {
+	upstreamSvcEndpoints map[service.MeshService][]endpoint.Endpoint
+}
+
+func (b *endpointsBuilder) Build() []types.Resource {
+	var edsResources []types.Resource
+
+	for svc, endpoints := range b.upstreamSvcEndpoints {
+		edsResources = append(edsResources, newClusterLoadAssignment(svc, endpoints))
```
steeling, line 29: it's preferrable (but not required), to not pass fields to methods that further operate on those fields. The reason being that it is much easier to see where the fields are used otherwise, in that with an IDE that leverages gopls, I can simply clickthrough that field to view all of its uses. 

When it gets passed around however, I then need to continue to manually drill into each function call.

This isn't so bad in this scenario, but there's instances in our code base where that cognitive overhead does add up.

One option here would be to remove unused parts of the code, and inline that function call here. You could get rid of lines 54-56 and 75-89 which are currently unused

allenlsy: I prefer not removing these lines in this PR. Although they are not used now, but the features were previously completed. Removing them will break these features. We can remove or change them in separate PR.

steeling:
> It doesn't seem like this change simplifies the existing code, rather it wraps the existing code with the builder type, which seems unnecessary in its current form. The builder pattern is most useful when generic complex objects, but it seems like the existing code here is simple enough to not even require the additional wrapping.
> 
> Let's update the code to something as follows:
> 
> ```
> b := EndpointsBuilder()
> for something : range requestedResources {
>     endpoints := getEndpoints(something)
>     b.AddEndpoint(meshSvc, endpoints)
> }
> 
> return b.Build()
> ```

I agree, I hadn't taken a deep look at EDS, but the builder pattern provides little to no benefit here.

With the above recommendation I'm still not sure that helps much. One thing with the builder pattern is that I don't think we should be doing partial builds until Build is called. Meaning `AddEndpoint` should not generate any envoy config, and the resulting Build call would be identical to what it is now.

I'm ok with moving forward with the builder pattern for consistency, but if we decide not to since it doesn't add anything here that's ok too.

shashankram:
> > It doesn't seem like this change simplifies the existing code, rather it wraps the existing code with the builder type, which seems unnecessary in its current form. The builder pattern is most useful when generic complex objects, but it seems like the existing code here is simple enough to not even require the additional wrapping.
> > Let's update the code to something as follows:
> > ```
> > b := EndpointsBuilder()
> > for something : range requestedResources {
> >     endpoints := getEndpoints(something)
> >     b.AddEndpoint(meshSvc, endpoints)
> > }
> > 
> > return b.Build()
> > ```
> 
> I agree, I hadn't taken a deep look at EDS, but the builder pattern provides little to no benefit here.
> 
> With the above recommendation I'm still not sure that helps much. One thing with the builder pattern is that I don't think we should be doing partial builds until Build is called. Meaning `AddEndpoint` should not generate any envoy config, and the resulting Build call would be identical to what it is now.
> 
> I'm ok with moving forward with the builder pattern for consistency, but if we decide not to since it doesn't add anything here that's ok too.

I agree. I was suggesting using `AddEndpoint` to let the builder keep track of the internal object state, vs passing it a final state in the form of a map, in which case it's best to just drop the builder for this change.

allenlsy:
@shashankram @steeling I agree with using `AddEndpoint` and not partially generate config before `build()`. Since in this PR, we did merge the logic of `fulfillEDSRequest()` and `generateEDSConfig()` a bit. I think that's a simpler implementation. I'm going to keep it and address the feedbacks

steeling:
I'm going to approve this PR, but agree with @shashankram that the builder pattern provides little benefit here, so we may want to scrap it altogether. I'll let others have final say

steeling (approved):

allenlsy:
ping @shashankram

keithmattix (approved):
I'm approving because I think having all of the xDS response types look the same is a benefit. Having one xDS handler that functions differently than the rest can cause extra cognitive overhead IMO

shashankram (approved):

PR #4958, opened 2022-08-02
Remove the proxy service mapper in favor of the kube provider

steeling:
Functionality between proxy mapper and kube provider was deuplicated. This removes that duplicated functionality, and helps decouple from k8s.


pkg/envoy/cds/response.go
```diff
@@ -26,12 +26,7 @@ func NewResponse(meshCatalog catalog.MeshCataloger, proxy *envoy.Proxy, _ *xds_d
 	}
 
 	// Build local clusters based on allowed inbound traffic policies
-	proxyServices, err := proxyRegistry.ListProxyServices(proxy)
-	if err != nil {
-		log.Error().Err(err).Str(errcode.Kind, errcode.GetErrCodeWithMetric(errcode.ErrFetchingServiceList)).
```
nojnhuh, line 31: Looks like `errcode.ErrFetchingServiceList` is no longer used and can be removed.

nojnhuh (approved):

PR #4961, opened 2022-08-02
envoy/lds: use builders for envoy listener configs

### LINKED ISSUE #4920: Leverage Builder pattern for LDS
Subtask of #4862, this is specifically for the LDS configs.

We should create a builder object that takes as inputs all data, and calls build. It should not make any network calls, or call out to other methods to get *more data*.

It *can* use nested builders if it makes sense to do so. See the parent issue and associated design doc for more info.

shashankram:
**Description**:
- Uses builders to build LDS Envoy configs -
  listeners, HTTP and TCP filters. Config is only
  built when `build()` is called on a builder, with
  required info set on the builder. 3 types of 
  builders are introduced: `ListenerBuilder` for
  building the listener, `FilterBuilder` for building
  various filter, and `HttpConnManagerBuilder`
  for HTTP filters.

- Updates EgressPolicy to be applicable only
   when permissive egress (global egress) is
   disabled.

- Renames existing config builder utitlity funcs to
   buildXyz instead of getXyz, for consistency.

- Removes the need for mocks in all tests other
  than the one calling lds.NewResponse().

- Increases code coverage of pkg/lds to >80%.

- Removes unused code after refactoring.

- Removes ginkgo dependency for the pkg

Resolves #4920 

**Testing done**:
- CI
- Improved unit test coverage

**Affected area**:

2. Is this a breaking change? Egress policy behavior is a breaking change.


pkg/catalog/ingress.go
```diff
@@ -20,6 +20,24 @@ const (
...
+func (mc *MeshCatalog) GetIngressTrafficPolicies(meshServices []service.MeshService) []*trafficpolicy.IngressTrafficPolicy {
+	var policies []*trafficpolicy.IngressTrafficPolicy
+
+	for _, meshSvc := range meshServices {
+		policy, err := mc.GetIngressTrafficPolicy(meshSvc)
+		if err != nil {
+			log.Error().Err(err).Msgf("Error getting ingress traffic policy for service %s, skippint it", meshSvc)
+			continue
+		}
+		if policy != nil {
```
steeling, line 33: nit: 

would it make sense to change GetIngressTrafficPolicy to return an error if one doesn't exist? Even an error like 

```go
policy, err := GetIngressTrafficPolicy()
if err == errNotFound {
   continue // ignore
} else if err != nil {
 log(err)
continue
}

policies = append(...)
```

Small thing, but as a user, I'd expect if no err then not nil.

pkg/envoy/lds/builder_test.go
```diff
@@ -0,0 +1,336 @@
...
+			if f.Name == filterName {
+				return true
+			}
+		}
+		return false
+	}
+
+	testCases := []struct {
+		name       string
+		buildFunc  func(b *httpConnManagerBuilder)
```
steeling, line 31: Alternatively, you could simply pass in the builder object with the build data supplied, although this doesn't test any potential logic in the builder's Setter methods

ie:

```
testCases := []struct {
		name       string
		buildFunc  b *httpConnManagerBuilder
		...
}
```

pkg/envoy/lds/builder_test.go
```diff
@@ -0,0 +1,336 @@
...
+		assertFunc func(*assert.Assertions, *xds_hcm.HttpConnectionManager)
```
steeling, line 32: and then the assert could be on the actual bytes/yaml definition of the output from the builder. That said it may hurt readability to do so

pkg/envoy/lds/egress.go
```diff
@@ -32,15 +32,15 @@ func (lb *listenerBuilder) getEgressFilterChainsForMatches(matches []*trafficpol
 		switch match.DestinationProtocol {
 		case constants.ProtocolHTTP:
 			// HTTP protocol --> HTTPConnectionManager filter
-			if filterChain, err := lb.getEgressHTTPFilterChain(*match); err != nil {
+			if filterChain, err := lb.buildEgressHTTPFilterChain(*match); err != nil {
```
steeling, line 35: *This comment is less of a code review, and more meant to elicit conversation.*

I think it's ever-so-slightly preferable to not pass fields that are on a struct to methods on that same struct, as it becomes easier in code navigation to see instances of that field in one place vs needing to click through to the next func.

That's mostly a pre-amble for the following comment:

I'm wondering, if while you've done this refactoring, if you've noticed instances where it might make sense to keep the data on the builder in a different structure. ie, maybe for match we have on the builder a `matchesByName map[string]*trafficpolicy.TrafficMatch`

It may also leave us inclined to refactor some of the meshcatalog methods, or split up some of the trafficpolicy objects.

steeling: For instance in `buildEgressHTTPFilterChain`, it looks like only the Name and DestinationPort are used.

Maybe that means we can add another meshcatalog query `GetPortsByMatchForService(...)` or something of the sort. Maybe not, and definitely doesn't need to be done in this PR, but might be worth leaving some comments throughout the code if you've noticed places

steeling (commented):
Awesome! impressive PR, a few comments but mostly thoughts. I didn't go through at a low level though, and may take a slightly bigger look at which tests are removed

pkg/envoy/lds/builder_test.go
shashankram: Yes I did it this way to test the builder setters, some of which such as `addFilter` have additional logic. I'll keep this as is out of convenience.

pkg/envoy/lds/builder_test.go
shashankram: I feel asserting on the fields is easier as we may not want to compare  the whole object

pkg/envoy/lds/egress.go
shashankram: Some of the existing utility builders operate on TrafficMatch objects that are not a part of the builder struct directly but that may be generated by the caller within the `build()` function. I think in this case, it makes sense to pass the param to the function as we don't have it readily available on the method receiver.

pkg/catalog/ingress.go
shashankram: Currently, it's okay for a policy to not exist, in which case we return `nil`. I left a TODO to remove GetIngressTrafficPolicy as the function even in its current form never returns an error. I'll consider handling the return semantics when I address that TODO to merge the 2 GetIngressPolic.* APIs. We simply log an error and ignore the erring policy, as we try to program as many configs on the proxy as possible.

pkg/envoy/lds/builder.go
```diff
@@ -0,0 +1,594 @@
...
+	for i, f := range hb.filters {
+		if f.Name == envoy.HTTPRouterFilterName {
+			routerFilterIndex = i
+			break
+		}
+	}
+
+	// Guard against the case where multiple router filters are accidentally programmed.
+	// This panics to ensure this never happens in code.
+	if routerFilterIndex == -1 && filter.Name == envoy.HTTPRouterFilterName {
```
keithmattix, line 467: How would `routerFilterIndex` be equal to -1 after matching 2 filters? There's also a break? Seems like this guards against no router filters as opposed to multiple

pkg/envoy/lds/builder_test.go
keithmattix: Agree with Shashank here; this feels like the more readable option

keithmattix (approved):
Just some small questions/nits, but LGTM overall

pkg/catalog/types.go
```diff
@@ -57,7 +57,11 @@ type MeshCataloger interface {
 	// is allowed access the upstream service
 	ListAllowedUpstreamEndpointsForService(identity.ServiceIdentity, service.MeshService) []endpoint.Endpoint
 
+	// GetIngressTrafficPolicies returns a list of IngressTrafficPolicy objects for the given MeshService list
+	GetIngressTrafficPolicies([]service.MeshService) []*trafficpolicy.IngressTrafficPolicy
+
 	// GetIngressTrafficPolicy returns the ingress traffic policy for the given mesh service
+	// TODO: deprecate in favor of GetIngressTrafficPolicies
```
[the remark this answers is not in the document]
shashankram: GetIngressTrafficPolicy always returns a nil error. It makes sense to use the newer API in pkg/envoy/rds, but it needs more work which I didn't want to club with this change.

pkg/envoy/lds/builder.go
```diff
@@ -0,0 +1,594 @@
...
+	"github.com/openservicemesh/osm/pkg/envoy"
+	"github.com/openservicemesh/osm/pkg/protobuf"
+	"github.com/openservicemesh/osm/pkg/trafficpolicy"
+)
+
+const (
+	websocketUpgradeType = "websocket"
+)
+
+func ListenerBuilder() *listenerBuilder { //nolint: revive // unexported-return
```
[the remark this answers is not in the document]
shashankram: I am envisioning the builder to be instantiated from outside the pkg at some point, most likely after we introduce the snapshot cache mechanism. So the function is exported. the type isn't exported because the setters in the builder may perform additional logic than just setting the field (e.g. AddFilters in HttpConnManagerBuilder). It's irrelevant at this moment because both the caller of the builder and the builder live in the same pkg.

pkg/envoy/lds/builder.go
shashankram: Good catch. Should be `!=`

pkg/envoy/lds/builder.go
```diff
@@ -0,0 +1,594 @@
...
+				TypedConfig: &any.Any{
+					TypeUrl: envoy.OriginalDstFilterTypeURL,
+				},
+			},
+		},
+	)
+	return lb
+}
+
+func (lb *listenerBuilder) Tracing(endpoint string) *listenerBuilder {
```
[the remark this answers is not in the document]
shashankram: Done.

pkg/envoy/lds/builder.go
keithmattix: Should there be a test added to catch this?

pkg/catalog/types.go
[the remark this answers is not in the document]
keithmattix: Ah, gotcha.

pkg/envoy/lds/builder.go
shashankram: Added.

shashankram:
There are some lint failures unrelated to this change. Going to ignore those.

pkg/envoy/lds/builder.go
```diff
@@ -0,0 +1,599 @@
...
+}
+
+func (tb *tcpProxyBuilder) WeightedClusters(wc []service.WeightedCluster) *tcpProxyBuilder {
+	tb.weightedClusters = wc
+	return tb
+}
+
+// Build builds the TCP proxy filter
+func (tb *tcpProxyBuilder) Build() (*xds_listener.Filter, error) {
+	if tb.cluster != "" && len(tb.weightedClusters) > 0 {
```
steeling, line 559: i would consider changing the builder to have only 1 way to accept clusters

`AddCluster(cluster string, weight int)`

pkg/envoy/lds/ingress.go
```diff
@@ -9,66 +9,59 @@ import (
...
-		log.Error().Err(err).Str(errcode.Kind, errcode.GetErrCodeWithMetric(errcode.ErrIngressFilterChain)).
-			Msgf("Error getting ingress filter chain for proxy with identity %s and service %s", lb.serviceIdentity, svc)
-		return nil
-	}
-
-	if ingressPolicy == nil {
-		log.Trace().Msgf("No ingress policy confiugred for proxy with identity %s and service %s", lb.serviceIdentity, svc)
+func (lb *listenerBuilder) buildIngressFilterChains() []*xds_listener.FilterChain {
+	if lb.ingressTrafficPolicies == nil {
 		return nil
```
steeling, line 21: this looks like it gets appended to the filter chains, that's ok?

Would it be better to wrap the call around `buildIngressFilterChains` within the nil check?

pkg/envoy/lds/response.go
```diff
@@ -27,47 +30,88 @@ func NewResponse(meshCatalog catalog.MeshCataloger, proxy *envoy.Proxy, _ *xds_d
...
+		egressPolicy, err := meshCatalog.GetEgressTrafficPolicy(proxy.Identity)
+		if err != nil {
+			return nil, fmt.Errorf("error building LDS response: %w", err)
 		}
+		outboundLis.EgressTrafficPolicy(egressPolicy)
+	}
+	if cfg.GetMeshConfig().Spec.Observability.Tracing.Enable {
+		outboundLis.TracingEndpoint(cfg.GetTracingEndpoint())
+	}
+	if cfg.GetMeshConfig().Spec.FeatureFlags.EnableWASMStats {
```
steeling, line 64: Would this be a good opportunity to remove other knobs from featureflags, or do we want to keep it here?

pkg/envoy/lds/builder.go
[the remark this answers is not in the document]
keithmattix: Right, I'm saying that that reasoning may be useful in a comment

pkg/envoy/lds/ingress.go
```diff
@@ -9,66 +9,59 @@ import (
...
-	for _, trafficMatch := range ingressPolicy.TrafficMatches {
-		if filterChain, err := lb.getIngressFilterChainFromTrafficMatch(trafficMatch, lb.cfg.GetMeshConfig().Spec.Sidecar); err != nil {
-			log.Error().Err(err).Msgf("Error building ingress filter chain for proxy with identity %s service %s", lb.serviceIdentity, svc)
-		} else {
+
+	for _, policy := range lb.ingressTrafficPolicies {
+		for _, trafficMatch := range policy.TrafficMatches {
+			filterChain, err := lb.buildIngressFilterChainFromTrafficMatch(trafficMatch)
+			if err != nil {
+				log.Error().Err(err).Msgf("Error building ingress filter chain for traffic match %s for proxy with identity %s", trafficMatch.Name, lb.proxyIdentity)
```
jsturtevant, line 30: This doesn't look to have changed here. But out of curiosity, Is this an expected error in some cases? what happens if this filter chain isn't there? traffic doesn't flow?

pkg/envoy/lds/ingress_test.go
```diff
@@ -163,23 +149,12 @@ func TestGetIngressFilterChainFromTrafficMatch(t *testing.T) {
...
-
-			mockConfigurator := configurator.NewMockConfigurator(mockCtrl)
 
 			lb := &listenerBuilder{
-				serviceIdentity: tests.BookstoreServiceIdentity,
-				cfg:             mockConfigurator,
+				proxyIdentity: tests.BookstoreServiceIdentity,
 			}
 
-			mockConfigurator.EXPECT().IsTracingEnabled().Return(false)
```
jsturtevant, line 176: dropping this seems like a nice improvement, making the tests easier to debug/comprehend 💯

pkg/envoy/lds/listener_test.go
```diff
@@ -3,62 +3,21 @@ package lds
...
-			listener, _ := buildPrometheusListener(connManager)
-			Expect(listener.Address).To(Equal(envoy.GetAddress(constants.WildcardIPAddr, constants.EnvoyPrometheusInboundListenerPort)))
-			Expect(len(listener.ListenerFilters)).To(Equal(0)) //  no listener filters
-			Expect(listener.TrafficDirection).To(Equal(xds_core.TrafficDirection_INBOUND))
-		})
-	})
-})
+	connManager := getPrometheusConnectionManager()
+	listener, err := buildPrometheusListener(connManager)
+	a.NotNil(listener)
```
jsturtevant, line 18: should there be more checks here?  seems like the previous test had more checks on the listener such as: 

```
Expect(listener.Address).To(Equal(envoy.GetAddress(constants.WildcardIPAddr, constants.EnvoyPrometheusInboundListenerPort)))
```

pkg/envoy/lds/builder.go
```diff
@@ -0,0 +1,599 @@
...
+	lb.trafficDirection = dir
+	return lb
+}
+
+func (lb *listenerBuilder) OutboundMeshTrafficPolicy(t *trafficpolicy.OutboundMeshTrafficPolicy) *listenerBuilder {
+	lb.outboundMeshTrafficPolicy = t
+	return lb
+}
+
+func (lb *listenerBuilder) InboundMeshTrafficPolicy(t *trafficpolicy.InboundMeshTrafficPolicy) *listenerBuilder {
```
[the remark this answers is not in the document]
shashankram: I prefer leaving this as is and letting `build()` do the magic of currently building the listener based on the fields set on it.

pkg/envoy/lds/builder.go
```diff
@@ -0,0 +1,599 @@
...
+	lb.permissiveMesh = enable
+	return lb
+}
+
+func (lb *listenerBuilder) PermissiveEgress(enable bool) *listenerBuilder {
+	lb.permissiveEgress = enable
+	return lb
+}
+
+func (lb *listenerBuilder) FilterBuilder() *filterBuilder {
```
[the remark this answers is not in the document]
shashankram: See https://github.com/openservicemesh/osm/pull/4961#discussion_r938236877 on why I exported this.

pkg/envoy/lds/builder.go
```diff
@@ -0,0 +1,599 @@
...
+	if lb.filterBuilder == nil {
+		lb.filterBuilder = FilterBuilder()
+	}
+	return lb.filterBuilder
+}
+
+func (lb *listenerBuilder) DefaultOutboundListenerFilters() *listenerBuilder {
```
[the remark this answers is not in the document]
shashankram: Good idea, will do.

pkg/envoy/lds/builder.go
```diff
@@ -0,0 +1,599 @@
...
+	for i, f := range hb.filters {
+		if f.Name == envoy.HTTPRouterFilterName {
+			routerFilterIndex = i
+			break
+		}
+	}
+
+	// Guard against the case where multiple router filters are accidentally programmed.
+	// This panics to ensure this never happens in code.
+	if routerFilterIndex != -1 && filter.Name == envoy.HTTPRouterFilterName {
```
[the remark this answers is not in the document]
shashankram: I don't see anything wrong with a panic here as it's meant to catch developer bugs (similar to the use of asserts in other languages)

pkg/envoy/lds/builder.go
```diff
@@ -0,0 +1,599 @@
...
+	case xds_core.TrafficDirection_INBOUND:
+		l.FilterChains = append(l.FilterChains, lb.buildInboundFilterChains()...)
+
+	default:
+		return nil, fmt.Errorf("listener %s: unsupported traffic direction %s", l.Name, l.TrafficDirection)
+	}
+
+	if lb.permissiveEgress {
+		// Enable permissive (global) egress to unknown destinations
+		egressFilterChain, err := getDefaultPassthroughFilterChain()
```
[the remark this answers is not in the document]
shashankram: That's a good suggestion. Some of the protos that are dynamically generated need to be verified if the marshalling is correct, but agree about the static configs. I'll leave this as is for now since unit tests would detect static config marshalling issues at present.

pkg/envoy/lds/builder.go
```diff
@@ -0,0 +1,599 @@
...
+}
+
+func (fb *filterBuilder) TCPProxy() *tcpProxyBuilder {
+	if fb.tcpProxyBuilder == nil {
+		fb.tcpProxyBuilder = TCPProxyBuilder()
+	}
+	return fb.tcpProxyBuilder
+}
+
+func (fb *filterBuilder) Build() ([]*xds_listener.Filter, error) {
```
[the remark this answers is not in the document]
shashankram: I prefer letting the FilterBuilder handle this.

pkg/envoy/lds/builder.go
```diff
@@ -0,0 +1,599 @@
...
+}
+
+// RouteConfigName sets the route config name on the builder
+func (hb *httpConnManagerBuilder) RouteConfigName(name string) *httpConnManagerBuilder {
+	hb.routeConfigName = name
+	return hb
+}
+
+// DefaultFilters sets the default HTTP filters on the builder
+func (hb *httpConnManagerBuilder) DefaultFilters() *httpConnManagerBuilder {
```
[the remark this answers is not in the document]
shashankram: Nope, the order of filters matters and this should be called in the current order.

pkg/envoy/lds/builder.go
```diff
@@ -0,0 +1,599 @@
...
+				TypedConfig: &any.Any{
+					TypeUrl: envoy.HTTPRBACFilterTypeURL,
+				},
+			},
+		},
+		// HTTP local rate limit filter - required to perform local rate limiting
+		&xds_hcm.HttpFilter{
+			Name: envoy.HTTPLocalRateLimitFilterName,
+			ConfigType: &xds_hcm.HttpFilter_TypedConfig{
+				TypedConfig: protobuf.MustMarshalAny(
```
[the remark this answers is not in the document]
shashankram: For static configs, it makes sense to use this. Dynamic configs can still fail proto marshalling.

pkg/envoy/lds/builder.go
shashankram: I prefer to keep this as is since I want to ensure you can set both a single cluster as well as weighted clusters through the builder.

pkg/envoy/lds/ingress.go
shashankram: No, `filterChains = append(filterChains, lb.buildIngressFilterChains()...)` does not append nil values.

pkg/envoy/lds/response.go
shashankram: This change is definitely not relevant to making API changes. Feel free to open an issue to discuss this further.

pkg/envoy/lds/ingress.go
shashankram: Yes. It's an expected error case that must be logged.

pkg/envoy/lds/listener_test.go
shashankram: I considered it but it's unnecessary testing the expected config by populating the fields the same way as the actual config. We already have e2e tests for this, and this isn't going to increase code coverage anyway.

pkg/envoy/lds/builder.go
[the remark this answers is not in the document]
shashankram: I separated out the building of inbound and outbound listeners to their own functions within build().

pkg/envoy/lds/builder.go
```diff
@@ -0,0 +1,594 @@
...
+	var filterChains []*xds_listener.FilterChain
+
+	filterChains = append(filterChains, lb.buildInboundMeshFilterChains()...)
+	filterChains = append(filterChains, lb.buildIngressFilterChains()...)
+
+	return filterChains
+}
+
+// FilterBuilder returns an instance used to build filters
+func FilterBuilder() *filterBuilder { //nolint: revive // unexported-return
```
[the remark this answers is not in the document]
shashankram: Will leave this as is. Feel free to break it up into multiple files in a subsequent PR.

pkg/envoy/lds/builder.go
```diff
@@ -0,0 +1,599 @@
...
+
+	var outboundTrafficMatches []*trafficpolicy.TrafficMatch // used to configure FilterDisabled match predicate
+	switch l.TrafficDirection {
+	case xds_core.TrafficDirection_OUTBOUND:
+		l.FilterChains = append(l.FilterChains, lb.buildOutboundFilterChains()...)
+		if lb.outboundMeshTrafficPolicy != nil {
+			outboundTrafficMatches = append(outboundTrafficMatches, lb.outboundMeshTrafficPolicy.TrafficMatches...)
+		}
+
+	case xds_core.TrafficDirection_INBOUND:
```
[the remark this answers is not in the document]
shashankram: Done.

pkg/envoy/lds/builder.go
```diff
@@ -0,0 +1,599 @@
...
+		// and fallback to the default filter chain.
+		l.ContinueOnListenerFiltersTimeout = true
+	}
+
+	if len(l.FilterChains) == 0 && l.DefaultFilterChain == nil {
+		// Programming a listener with no filter chains is an error.
+		// It is possible for the outbound listener to have no filter chains if
+		// there are no allowed upstreams for this proxy and egress is disabled.
+		// In this case, return a nil filter chain so that it doesn't get programmed.
+		return nil, nil
```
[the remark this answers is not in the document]
shashankram: Nope, it's common to both. I fixed the comment.

pkg/envoy/lds/builder.go
```diff
@@ -0,0 +1,599 @@
...
+	}
+
+	if hb.tracing != nil {
+		connManager.GenerateRequestId = &wrappers.BoolValue{
+			Value: true,
+		}
+		connManager.Tracing = hb.tracing
+	}
+
+	marshalled, err := anypb.New(connManager)
```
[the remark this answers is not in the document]
shashankram: Since connManager is built using dynamic configs, we shouldn't use MustMarshalAny, which will panic. We need proto validation to happen for dynamic configs.

pkg/envoy/lds/builder.go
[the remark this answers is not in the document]
steeling: Can you add a TODO at least for my comment.. i see it as pretty objective that removing an error case is an improvement. 

It also removes a decent amount of logic, with better guarantees around correctness

pkg/envoy/lds/builder.go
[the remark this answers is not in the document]
steeling: but since this is always called first, doesn't that leave potential for ommitting it and therefore incorrect behavior? Why not bring it into the build method and remove that possibility? Looks like it's an implementation detail to me and shouldn't be exported

pkg/envoy/lds/builder.go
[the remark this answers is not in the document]
steeling: even if that's the case, `FilterBuilder` is called within build methods, so it wouldn't be called from outside the package

pkg/envoy/lds/builder.go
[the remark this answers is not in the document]
shashankram: ok will modify this to keep track of the routerfilter dirctly on the hcmBuilder struct

pkg/envoy/lds/builder.go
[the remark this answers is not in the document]
shashankram: Updated to `defaultFilters()` to handle the ordering logic within `build()`.

pkg/envoy/lds/builder.go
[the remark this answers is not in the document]
shashankram: Done.

steeling (approved):

PR #4963, opened 2022-08-03
begin snapshot cache implementation.

### LINKED ISSUE #4950: Potential incorrectness when proxy rapidly connect and disconnect to the control plane
**Bug description**:

When a proxy sends connect and disconnect request to the registry almost simultaneously, the end result is indeterministic. It will depends on sequence that requests are processed. This may lead to incorrect result.

Possible solution is to associate a nonce or version number to the requests.

[Discussion in previous PR here](https://github.com/openservicemesh/osm/pull/4912#discussion_r930209026). 

Related issue: https://github.com/openservicemesh/osm/issues/4783

- Envoy Control Plane        [x]

**Expected behavior**:

When the proxy sends connect and disconnect requests simultaneously, the end result should be consistent according to the sequence of requests from the proxy

steeling:
Leverage the callbacks to register and unregister the proxy from the proxyregistry.

The snapshot cache leverages a Stream ID, and that's all that's provided on the stream closed callback. Because of this, we need to maintain a mapping of the stream ID to the envoy.Proxy. The most logical place to put this is in the proxyRegistry, so we change the mapping from UUID to stream ID. 

`OnStreamOpen` is mostly copy/pasted form the beginning of `StreamAggregatedResources`

I also now maintain a stream ID on the server itself, for the legacy `StreamAggregatedResources`, although that will be removed when we migrate to snapshot cache.

Finally, I'm able to delete some unused code after these refactors, including `GetProxyFromPod`, and the proxy registry's announcement handler, which was only used in the osm-controller on bootstrap certs (which are never issued in the osm-controller, so therefore did nothing)

Part of #2683 
Fixes #4950


allenlsy:
> so we change the mapping from UUID to stream ID.

So this is a breaking change when users upgrade OSM, right?

steeling:
> > so we change the mapping from UUID to stream ID.
> 
> So this is a breaking change when users upgrade OSM, right?

No the mappings are strictly held in memory so it’s recomputed on every restart anyways

pkg/envoy/ads/stream.go
```diff
@@ -48,7 +49,8 @@ func (s *Server) StreamAggregatedResources(server xds_discovery.AggregatedDiscov
 	// This is the Envoy proxy that just connected to the control plane.
 	// NOTE: This is step 1 of the registration. At this point we do not yet have context on the Pod.
 	//       Details on which Pod this Envoy is fronting will arrive via xDS in the NODE_ID string.
-	proxy := envoy.NewProxy(kind, uuid, si, utils.GetIPFromContext(server.Context()))
+	streamID := atomic.AddInt64(&s.streamCount, 1)
+	proxy := envoy.NewProxy(kind, uuid, si, utils.GetIPFromContext(server.Context()), streamID)
```
allenlsy, line 53: Looks like uuid is replaced by connectionID? We can get rid of it in future PR if that's easier.

pkg/envoy/proxy.go
```diff
@@ -169,6 +173,14 @@ func (p *Proxy) GetConnectedAt() time.Time {
 	return p.connectedAt
 }
 
+// GetConnectionID returns the connection ID of the proxy.
```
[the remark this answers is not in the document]
steeling: We actually do want a different ID for each connection. It's what lets us map an envoy.Proxy to a specific stream.

This also solves that proxyRegistry issue where a proxy could reconnect and connect quickly. If you drill into the snapshot cache implementation, this is pretty much exactly how they generate the ID

pkg/envoy/ads/stream.go
steeling: A good way to think about it is that a UUID maps to a proxy, while a ConnectionID maps to the connection (grpc Stream). If the same proxy connects and reconnects we need to know which stream, so we can remove the old one, and which UUID (to identify which proxy it is)

pkg/envoy/registry/registry.go
```diff
@@ -10,38 +12,32 @@ func NewProxyRegistry(mapper ProxyServiceMapper, msgBroker *messaging.Broker) *P
...
-	if pr.GetConnectedProxy(uuid) != nil {
-		log.Debug().Str("proxy", proxy.String()).Msgf("Proxy %s already registered", proxy.String())
-		return
-	}
-
 	pr.mu.Lock()
 	defer pr.mu.Unlock()
-	pr.connectedProxies[uuid] = proxy
+	pr.connectedProxies[proxy.GetConnectionID()] = proxy
 	log.Debug().Str("proxy", proxy.String()).Msg("Registered new proxy")
```
jsturtevant, line 26: nit: this might be slightly misleading comment in since the proxy potentially already registered?

pkg/envoy/ads/stream.go
```diff
@@ -48,7 +49,8 @@ func (s *Server) StreamAggregatedResources(server xds_discovery.AggregatedDiscov
 	// This is the Envoy proxy that just connected to the control plane.
 	// NOTE: This is step 1 of the registration. At this point we do not yet have context on the Pod.
 	//       Details on which Pod this Envoy is fronting will arrive via xDS in the NODE_ID string.
-	proxy := envoy.NewProxy(kind, uuid, si, utils.GetIPFromContext(server.Context()))
+	streamID := atomic.AddInt64(&s.streamCount, 1)
```
jsturtevant, line 52: maybe rename `streamCount`? It took me a moment to realize where there was no decrement on this value.  The function `GetConnectedProxyCount` doing a len(on connected proxies) to get the number of streams not using `streamCount`.

steeling: Good call! Renaming to streamID, although if you have an alternative happy to use it

pkg/envoy/ads/server.go
```diff
@@ -79,7 +79,7 @@ func (s *Server) Start(ctx context.Context, cancel context.CancelFunc, port int,
 
 	if s.cacheEnabled {
 		s.ch = cachev3.NewSnapshotCache(false, cachev3.IDHash{}, &scLogger{})
-		s.srv = serverv3.NewServer(ctx, s.ch, &Callbacks{})
+		s.srv = serverv3.NewServer(ctx, s.ch, s)
```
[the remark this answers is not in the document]
steeling: yup, it actually doesn't need a name at all!

pkg/envoy/registry/registry.go
```diff
@@ -10,38 +12,32 @@ func NewProxyRegistry(mapper ProxyServiceMapper, msgBroker *messaging.Broker) *P
 	return &ProxyRegistry{
 		ProxyServiceMapper: mapper,
 		msgBroker:          msgBroker,
-		connectedProxies:   make(map[string]*envoy.Proxy),
+		connectedProxies:   make(map[int64]*envoy.Proxy),
+		pubsub:             pubsub.New(0),
+		// ch 							: make(chan *envoy.Proxy, 10),
```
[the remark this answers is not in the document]
steeling: good catch

pkg/envoy/registry/registry.go
steeling: fixed!

pkg/envoy/registry/registry.go
```diff
@@ -10,38 +12,32 @@ func NewProxyRegistry(mapper ProxyServiceMapper, msgBroker *messaging.Broker) *P
...
+	return pr.connectedProxies[connectionID]
 }
 
 // UnregisterProxy unregisters the given proxy from the catalog.
-func (pr *ProxyRegistry) UnregisterProxy(proxy *envoy.Proxy) {
+func (pr *ProxyRegistry) UnregisterProxy(connectionID int64) {
 	pr.mu.Lock()
 	defer pr.mu.Unlock()
-	delete(pr.connectedProxies, proxy.UUID.String())
-	log.Debug().Msgf("Unregistered proxy %s", proxy.String())
```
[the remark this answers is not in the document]
steeling: re-added

allenlsy (approved):

keithmattix:
Forgive me if I've missed it, but is there a TL;DR on the benefits of using Envoy's snapshot cache and even what that cache is?

steeling:
> Forgive me if I've missed it, but is there a TL;DR on the benefits of using Envoy's snapshot cache and even what that cache is?

I've updated the comment here https://github.com/openservicemesh/osm/issues/2683, let me know if i can provide more context :)

pkg/envoy/ads/response_benchmark_test.go
```diff
@@ -153,7 +153,7 @@ func setupTestServer(b *testing.B) {
 		b.Fatalf("Failed to create service: %v", err)
 	}
 
-	proxy = envoy.NewProxy(envoy.KindSidecar, proxyUUID, proxySvcAccount.ToServiceIdentity(), nil)
+	proxy = envoy.NewProxy(envoy.KindSidecar, proxyUUID, proxySvcAccount.ToServiceIdentity(), nil, 1)
```
keithmattix, line 156: It might be nice to surface any performance improvement we see in tests based on this change

keithmattix (approved):
LGTM

pkg/envoy/ads/response_benchmark_test.go
steeling: this test would actually have no differences compared to the last, since it's not doing any extra work here

pkg/debugger/proxy.go
```diff
@@ -49,18 +58,18 @@ func (ds DebugConfig) printProxies(w http.ResponseWriter) {
 	_, _ = fmt.Fprint(w, "<tr><td>#</td><td>Envoy's Service Identity</td><td>Envoy's UUID</td><td>Connected At</td><td>How long ago</td><td>tools</td></tr>")
 	for idx, proxy := range proxies {
 		ts := proxy.GetConnectedAt()
-		proxyURL := fmt.Sprintf("/debug/proxy?%s=%s", uuidQueryKey, proxy.UUID)
+		proxyURL := fmt.Sprintf("/debug/proxy?%s=%d", streamIDQueryKey, proxy.GetConnectionID())
```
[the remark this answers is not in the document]
steeling: Nope, the mapping changes but the proxy still has the UUID on it. These links are placed on buttons, and still labeled by the UUID

allenlsy (approved):

pkg/envoy/ads/stream.go
```diff
@@ -47,7 +48,8 @@ func (s *Server) StreamAggregatedResources(server xds_discovery.AggregatedDiscov
 	// This is the Envoy proxy that just connected to the control plane.
 	// NOTE: This is step 1 of the registration. At this point we do not yet have context on the Pod.
 	//       Details on which Pod this Envoy is fronting will arrive via xDS in the NODE_ID string.
-	proxy := envoy.NewProxy(kind, uuid, si, utils.GetIPFromContext(server.Context()))
+	streamID := atomic.AddInt64(&s.nextStreamID, 1)
+	proxy := envoy.NewProxy(kind, uuid, si, utils.GetIPFromContext(server.Context()), streamID)
```
shashankram, line 52: Are we creating a proxy instance per gRPC stream? I am wondering about the impact on the control plane proxy bookkeeping workflow if that's the case.

steeling: it is, but under the assumption that the prior stream is about to disconnect (but there are race conditions here).

I think more work will be pushed to the proxy registry (as currently we have some bookeeping happening in each per proxy stream).

With the proxy registry there are controls around querying for/listing by proxy uuid to return the proxy with the highest stream ID.

steeling: also it should be on the order of milliseconds that there are multiple proxies connected with the same UUID

shashankram: But multiple gRPC streams are allowed per connection. Wouldn't it make sense to keep track of proxies (envoy.Proxy instance) per UUID? Creating a proxy instance per stream would mean we are tracking individual gRPC streams and not unique proxies. 

How are we certain that multiple streams can only momentarily exist per proxy?

steeling: you're right there's no guarantee, other than knowing the envoy sidecar won't open multiple connections.

But we don't have that guarantee currently anyways. In fact, there's currently a bug where it's possible to delete the proxy from the registry after it connects.

Maintaing a registry for each connection makes sense IMO, sense it keeps track of what's connected. However the API of the registry will only return the newest connection

shashankram: I agree maintaining a proxy per connection makes sense, but connection != stream. So I am still wondering why to keep track of a proxy instance per stream, which could result in incorrect bookkeeping of unique proxies.

steeling: maybe we're getting terminologies crossed? `Stream` refers to the grpc Stream, `StreamAggregatedResources`. We use the callback `OnStreamOpen`, which is triggered when the sidecar makes a new request to `StreamAggregatedResources`

Are you saying that a single proxy may have multiple calls to StreamAggregatedResources ongoing simultaneously?

shashankram: Though that should not happen based on observation, I don't see it being explicitly stated anywhere in the Envoy proto. Hence, I am wondering if we should be careful about bookkeeping per stream vs per proxy connection. I haven't seen this happen with the current XDS state machine, but not sure if the snapshot cache behaves differently in this regard.

steeling: gotcha. So this implementation is "more correct" than the current implementation, in that we avoid a race condition of deleting recently joined proxies.

How we store the proxy on the registry cache is really an implementation detail IMO, and since we filter by the most recent StreamID, to an observer there's no knowledge of multiple streams.

shashankram (approved):

PR #4971, opened 2022-08-05
move several methods between the compute client and the k8s client.

steeling:
With our current refactoring the k8s client will become a simple accessor with little/no business logic,
whereas the compute.Client will have the business logic for translating osm objects to/from k8s objects.
This also brings us closer to removing the kubecontroller from the meshcatalog


steeling:
@keithmattix  can you take a look at this with respect to me removing the `GetServiceFromHostname` and `GetSubdomainFromHostname` which I believe was an eroeneous check due to splits not working outside of the namespace, nor with statefulsets

keithmattix:
@steeling Will give a more thorough review later, but you're correct; good catch! Statefulsets aren't supported with TrafficSplits, so there's no need to add that extra logic

pkg/k8s/osm_injector_test.go
```diff
@@ -1,102 +0,0 @@
...
-	tassert "github.com/stretchr/testify/assert"
-	corev1 "k8s.io/api/core/v1"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/runtime"
-	"k8s.io/client-go/kubernetes/fake"
-
-	"github.com/openservicemesh/osm/pkg/constants"
-)
-
-func TestGetOSMInjectorPods(t *testing.T) {
```
keithmattix, line 15: This test doesn't appear to be replaced

keithmattix (commented):
Looks good generally, but I see several deleted test files that don't appear to be replaced or accounted for in added tests. Holding off on an approval pending an an explanation

pkg/k8s/osm_injector_test.go
steeling: The method in the corresponding file was deleted

steeling:
> Looks good generally, but I see several deleted test files that don't appear to be replaced or accounted for in added tests. Holding off on an approval pending an an explanation

I believe those are all due to removing the original method

keithmattix (approved):

shashankram (commented):
Needs a rebase after my recent lds changes.

steeling:
> Needs a rebase after my recent lds changes.

done

pkg/cli/verifier/envoy_config.go
```diff
@@ -461,7 +460,7 @@ func (v *EnvoyConfigVerifier) verifyDestination() Result {
 	return result
 }
 
-func (v *EnvoyConfigVerifier) getDstMeshServicesForSvcPod(svc corev1.Service, podName string) ([]service.MeshService, error) {
+func (v *EnvoyConfigVerifier) getDstMeshServicesForSvcPod(svc corev1.Service) ([]service.MeshService, error) {
```
shashankram, line 463: nit: rename function to `getDstMeshServicesForK8sSvc`

pkg/compute/kube/fake/fake.go
```diff
@@ -95,3 +96,32 @@ func (f fakeClient) GetResolvableEndpointsForService(svc service.MeshService) []
...
 	return endpoints
 }
+
+func (f fakeClient) IsMetricsEnabled(*envoy.Proxy) (bool, error) {
+	return true, nil
+}
+
+// GetHostnamesForService returns the hostnames over which the service is accessible
+func (f fakeClient) GetHostnamesForService(svc service.MeshService, localNamespace bool) []string {
+	var hostnames []string
```
shashankram, line 106: seems like the fake has the same implementation as the actual one. Can this be consolidated?

shashankram (approved):
Could you confirm the test coverage hasn't been significantly reduced after this refactor. Would be good to have codecov bot auto-detect this, which currently isn't configured to flag code coverage drops.

keithmattix:
@shashankram CodeCov code coverage computation has been VERY unreliable in my experience. It's just anecdotal, but I'm hesitant to experience that again 😅

shashankram:
> @shashankram CodeCov code coverage computation has been VERY unreliable in my experience. It's just anecdotal, but I'm hesitant to experience that again 😅

Which computation are you referring to? There's a GH action that is meant to reliably detect drop in coverage. The codecov bot that comments on PRs is known to be ambiguous.

keithmattix:
Ah, I was referring to the codecov bot that comments on PRs. I'm open to a more consistent GitHub action 👍🏾

pkg/compute/kube/fake/fake.go
steeling: done

pkg/cli/verifier/envoy_config.go
steeling: done

steeling:
> Could you confirm the test coverage hasn't been significantly reduced after this refactor. Would be good to have codecov bot auto-detect this, which currently isn't configured to flag code coverage drops.

Test coverage percent has dropped in package k8s, but mostly due to deleting a few large methods that had 100% coverage (no longer needed), and slightly increasing the mocke kube client which has a very low percentage.

The overall lines of code tested looks to be more or less unchanged

Issue #4924, opened 2022-07-20
Leverage builder pattern for RDS


steeling:
Subtask of #4862, this is specifically for the RDS configs.

We should create a builder object that takes as inputs all data, and calls build. It should not make any network calls, or call out to other methods to get *more data*.

It *can* use nested builders if it makes sense to do so. See the parent issue and associated design doc for more info.

shashankram:
@nshankar13, this will be impacted by the work I am doing for HTTP rate limiting as a part of https://github.com/openservicemesh/osm/issues/2018. I might wrap that in a week or so. Can this wider refactor of the RDS code wait till then?

shashankram:
@nshankar13 the rate limiting work is complete.

PR #4973, opened 2022-08-08
Route builder for RDS

nshankar13:

**Description**: Resolves https://github.com/openservicemesh/osm/issues/4924

**Testing done**:

**Affected area**:

2. Is this a breaking change?


nshankar13:
@steeling @shashankram few thoughts: 
1) Wasn't sure if "getTrafficPolicy" methods should be in builder or remain in response.go. 
2) Bulk of config building was happening already in route_config.go. Should we merge this into the builder class, or continue to keep them separate? Advantage of having them separate is that builder.go code looks a lot cleaner and more intuitive.

pkg/envoy/rds/builder.go
```diff
@@ -0,0 +1,97 @@
...
+	"github.com/openservicemesh/osm/pkg/trafficpolicy"
+)
+
+type routesBuilder struct {
+	routes []*xds_route.RouteConfiguration
+}
+
+func newRoutesBuilder() *routesBuilder {
+	return &routesBuilder {
+		routes: make([]*xds_route.RouteConfiguration, 0),
```
steeling, line 22: instead of building here, we should be setting all inputs upfront, then call build at the end.

ie: add the TrustDomain as a field, trafficmatches, etc.

pkg/envoy/rds/builder.go
```diff
@@ -0,0 +1,97 @@
...
+	if discoveryReq != nil {
+		rdsResources = ensureRDSRequestCompletion(discoveryReq, rdsResources)
+	}
+
+	return rdsResources, nil
+}
+
+// ensureRDSRequestCompletion computes delta between requested resources and response resources.
+// If any resources requested were not responded to, this function will fill those in with empty RouteConfig stubs
+func ensureRDSRequestCompletion(discoveryReq *xds_discovery.DiscoveryRequest, rdsResources []types.Resource) []types.Resource {
```
steeling, line 75: let's bring this back to the response.go file

pkg/envoy/rds/response.go
```diff
@@ -35,10 +32,7 @@ func NewResponse(cataloger catalog.MeshCataloger, proxy *envoy.Proxy, discoveryR
 	// clients on allowed routes.
 	inboundMeshTrafficPolicy := cataloger.GetInboundMeshTrafficPolicy(proxy.Identity, proxyServices)
 	if inboundMeshTrafficPolicy != nil {
-		inboundMeshRouteConfig := route.BuildInboundMeshRouteConfiguration(inboundMeshTrafficPolicy.HTTPRouteConfigsPerPort, proxy, cfg, trustDomain)
-		for _, config := range inboundMeshRouteConfig {
-			rdsResources = append(rdsResources, config)
-		}
+		builder.AddInboundMeshRouteConfig(inboundMeshTrafficPolicy.HTTPRouteConfigsPerPort, proxy, cfg, trustDomain)
```
steeling, line 35: so for instance here, we may have separate calls:

```go
builder.SetHTTPRouteCOnfigsPerPort(configsPerPort)
.SetProxy(proxy)
.SetTrustDomain(trustDomain)
```

Note that we should not set the cfg, or any other client objects. Instead we should call the client objects here, and set the data on the builder

pkg/envoy/rds/response.go
```diff
@@ -72,8 +63,7 @@ func NewResponse(cataloger catalog.MeshCataloger, proxy *envoy.Proxy, discoveryR
 		ingressTrafficPolicies = trafficpolicy.MergeInboundPolicies(ingressTrafficPolicies, ingressPolicy.HTTPRoutePolicies...)
```
steeling, line 63: Notice that on a lot of these we call GetIngressTrafficPolicy (or inboundTrafficPolicy, outboundTrafficPolicy), but then we only access one field. That means we should probably add a method to just get that one field, so we don't perform the extra work.

No need to tackle in this PR but probably worth a TODO

pkg/envoy/rds/builder.go
```diff
@@ -0,0 +1,80 @@
...
+func (b *routesBuilder) Build() ([]types.Resource, error) {
+
+	var rdsResources []types.Resource
+
+	// ---
+	// Build inbound mesh route configurations. These route configurations allow
+	// the services associated with this proxy to accept traffic from downstream
+	// clients on allowed routes.
+	if b.inboundMeshTrafficPolicy != nil {
+		inboundMeshRouteConfig := route.BuildInboundMeshRouteConfiguration(b.inboundMeshTrafficPolicy.HTTPRouteConfigsPerPort, b.proxy, b.cfg, b.trustDomain)
```
steeling, line 43: instead of passing these to the route package's `BuildInboundMeshRouteConfiguration`, let's make BuildInboundMeshRouteConfiguration an unexported method on the builder object.

From the [design doc](https://docs.google.com/document/d/1L_u-XlVU8Fog-uFS3Uf0O2nz8IArQy82cd6Npxm038Y/edit#heading=h.86n6n6h1rqmo), we call out that it's easier when leveraging code navigation to highlight ie: `b.proxy`, and see all of its uses. Whereas when we pass it to another function, we lose that ability.

steeling (changes requested):

pkg/envoy/rds/builder.go
nshankar13: Makes sense, this is sort of what I was alluding to in my top comment (https://github.com/openservicemesh/osm/pull/4973#issuecomment-1208494343) about merging / borrowing some of the methods in route_config.go and implementing that in builder.go instead. 

Let's discuss more over a call, I think that would be easier.

nshankar13:
Putting this on hold temporarily until globalRateLimiting has been merged, but plan on updating this with by moving some of the methods in route_config (BuildInboundMeshConfiguration) etc, to builder.go, and editing the tests accordingly.

pkg/envoy/rds/builder.go
```diff
@@ -0,0 +1,210 @@
...
+	outboundVirtualHost = "outbound_virtual-host"
+
+	// egressVirtualHost is the prefix for the virtual host's name in the egress route configuration
+	egressVirtualHost = "egress_virtual-host"
+
+	// ingressVirtualHost is the prefix for the virtual host's name in the ingress route configuration
+	ingressVirtualHost = "ingress_virtual-host"
+)
+
+type routesBuilder struct {
```
steeling, line 27: i think the builder and the route package should be joined together. It would make sense to me to move this file into pkg/envoy/rds/route

steeling (approved):

steeling:
approved! please make sure tests pass before submitting :)

jsturtevant (approved):
lgtm, pending linter updates

pkg/envoy/rds/response.go
```diff
@@ -69,23 +57,22 @@ func NewResponse(cataloger catalog.MeshCataloger, proxy *envoy.Proxy, discoveryR
...
-		for _, egressConfig := range egressRouteConfigs {
-			rdsResources = append(rdsResources, egressConfig)
-		}
+		routesBuilder.EgressPortSpecificRouteConfigs(egressTrafficPolicy.HTTPRouteConfigsPerPort)
+	}
+
+	rdsResources, err := routesBuilder.Build()
+
+	if err != nil {
+		log.Error().Err(err).Msgf("Error building route config.")
```
shashankram, line 75: `return nil, err` here.

shashankram (commented):
Could you collapse pkg/envoy/rds/route/* under pkg/envoy/rds for consistency with other verticals (cds, lds, etc.).

pkg/envoy/rds/builder.go
nshankar13: As per Shashank's comment I will be removing `/route` and moving everything to `/rds`

pkg/envoy/rds/route_config_test.go
```diff
@@ -0,0 +1,908 @@
+package rds
```
shashankram, line 1: Was this file moved using `git mv`? Curious because this file diff doesn't appear the same as `route_config.go`:
image

nshankar13: I moved it and removed the original `route` folder through VS code directly, but I also had to make changes to `route_config.go` because I moved a few of the methods there to `builder.go` and also needed to make some of the methods in `route_config.go` exportable to call in builder.

shashankram: Sure, but could you try to confirm if file history can be preserved using `git mv` versus removing the file directly on your IDE?

nshankar13: Still not working with `git mv` not sure why it considers route_config_test a new file altogether but preserves version history for the others. If I recreate the /route folder and add route_config_test back there, git lists it as a modification, but if I use `git mv` and move it to /rds and delete the /route folder, then it thinks it's a new file and doesn't preserve the file history.

shashankram: Based on discussing this offline, it seems the file was moved after being edited, instead of being moved first. This could cause the file history to be unexpected. @nshankar13 will try to fix this.

nshankar13: Tried resetting, moving the file with `git mv`, and then applying edits on top of the moved files. However, git still thinks `rds/builder_test.go` was previously `rds/route/route_config_test.go`, and that `rds/route_config_test.go` is now a new file.

nshankar13: `pkg/envoy/rds/route/route_config_test.go → pkg/envoy/rds/builder_test.go`

nshankar13: This might have to do with how `git` detects renames: https://chelseatroy.com/2020/05/09/question-how-does-git-detect-renames/

pkg/envoy/rds/builder.go
```diff
@@ -0,0 +1,205 @@
...
+	xds_route "github.com/envoyproxy/go-control-plane/envoy/config/route/v3"
+	"github.com/envoyproxy/go-control-plane/pkg/cache/types"
+
+	"github.com/openservicemesh/osm/pkg/envoy"
+	"github.com/openservicemesh/osm/pkg/trafficpolicy"
+)
+
+const (
+	// inboundVirtualHost is prefix for the virtual host's name in the inbound route configuration
+	inboundVirtualHost = "inbound_virtual-host"
```
keithmattix, line 14: nit: if these are prefixes, that should be reflected in the variable name

keithmattix (approved):
One naming nit, but it's not blocking. Great work!

PR #4978, opened 2022-08-09
egress: deprecate EgressPolicy feature flag

shashankram:
**Description**:
After 87fd049d, EgressPolicy API is implicitly
enabled when global mesh-wide egress is disabled
by setting `traffic.enableEgress=false` in MeshConfig.

This change updates the remaining changes pertaining
to the now deprecated EgressPolicy feature flag.

**Testing done**:
CI

**Affected area**:

2. Is this a breaking change? `yes, release noted for v1.3`

3. Has documentation corresponding to this change been updated in the [osm-docs](https://github.com/openservicemesh/osm-docs) repo (if applicable)? `pending`

### DIFF
--- a/charts/osm/values.yaml
+++ b/charts/osm/values.yaml
@@ -470,7 +470,7 @@ osm:
     enableWASMStats: true
     # -- Enable OSM's Egress policy API.
     # When enabled, fine grained control over Egress (external) traffic is enforced
-    # DEPRECATED. To use EgressPolicy, disable global mesh-wide egress using '--set osm.enableEgress=false'.
+    # DEPRECATED, do not use. To use EgressPolicy API, disable global mesh-wide egress using '--set osm.enableEgress=false'.
     enableEgressPolicy: true
     # -- Enable async proxy-service mapping
     enableAsyncProxyServiceMapping: false
--- a/pkg/catalog/egress.go
+++ b/pkg/catalog/egress.go
@@ -27,7 +27,8 @@ const (
 
 // GetEgressTrafficPolicy returns the Egress traffic policy associated with the given service identity
 func (mc *MeshCatalog) GetEgressTrafficPolicy(serviceIdentity identity.ServiceIdentity) (*trafficpolicy.EgressTrafficPolicy, error) {
-	if !mc.configurator.GetMeshConfig().Spec.FeatureFlags.EnableEgressPolicy {
+	if mc.configurator.GetMeshConfig().Spec.Traffic.EnableEgress {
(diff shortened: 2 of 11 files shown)


charts/osm/values.yaml
```diff
@@ -470,7 +470,7 @@ osm:
     enableWASMStats: true
     # -- Enable OSM's Egress policy API.
     # When enabled, fine grained control over Egress (external) traffic is enforced
-    # DEPRECATED. To use EgressPolicy, disable global mesh-wide egress using '--set osm.enableEgress=false'.
+    # DEPRECATED, do not use. To use EgressPolicy API, disable global mesh-wide egress using '--set osm.enableEgress=false'.
```
steeling, line 473: do we need to maintain cli flags as part of our deprecation policy?

pkg/catalog/egress.go
```diff
@@ -27,7 +27,8 @@ const (
 
 // GetEgressTrafficPolicy returns the Egress traffic policy associated with the given service identity
 func (mc *MeshCatalog) GetEgressTrafficPolicy(serviceIdentity identity.ServiceIdentity) (*trafficpolicy.EgressTrafficPolicy, error) {
-	if !mc.configurator.GetFeatureFlags().EnableEgressPolicy {
+	if mc.configurator.GetMeshConfig().Spec.Traffic.EnableEgress {
```
steeling, line 30: super nit: Slightly of the opinion we should just wrap the call of `GetEgressTrafficPolicy` within that check instead of placing it here. Removes the awkward `return nil, nil`

steeling (approved):

charts/osm/values.yaml
shashankram: This is a Helm chart value that can be set using an opaque `--set` CLI flag. We need to keep this property at least for 1 more release after announcing deprecation, after which we can drop it completely. It's the safer thing to do.

pkg/catalog/egress.go
shashankram: I prefer this approach, instead of having a bunch of conditionals on the caller side wherever this function is invoked. A `nil` policy is a no-op for the XDS code generation path.

keithmattix (approved):

keithmattix:
@shashankram is there a follow up PR to update the release notes?

shashankram:
> @shashankram is there a follow up PR to update the release notes?

No, it has already been updated: https://github.com/openservicemesh/osm/blob/main/docs/release_notes.md#breaking-changes

PR #4980, opened 2022-08-09
ref: leverage builder pattern for sds

whitneygriffith:
Description: 

Leverage builder pattern in SDS resources generation. 

Resolves issue #4922 

Signed-off-by: Whitney Griffith <whitney.griffith16@gmail.com>

- [ ]  Tests 

**Description**:

**Testing done**:

**Affected area**:

2. Is this a breaking change?

### DIFF
--- a/pkg/envoy/sds/builder.go
+++ b/pkg/envoy/sds/builder.go
@@ -1,0 +1,169 @@
+package sds
+
+import (
+	xds_core "github.com/envoyproxy/go-control-plane/envoy/config/core/v3"
+	xds_auth "github.com/envoyproxy/go-control-plane/envoy/extensions/transport_sockets/tls/v3"
+
+	"github.com/openservicemesh/osm/pkg/certificate"
+	"github.com/openservicemesh/osm/pkg/envoy"
+	"github.com/openservicemesh/osm/pkg/envoy/secrets"
+	"github.com/openservicemesh/osm/pkg/errcode"
+	"github.com/openservicemesh/osm/pkg/identity"
+	"github.com/openservicemesh/osm/pkg/service"
+)
+
+type sdsBuilder struct {
+	proxy *envoy.Proxy
+
+	// Service certificate for this proxy
+	serviceCert *certificate.Certificate
+
+	// The DiscoveryRequest contains the certs to generate secrets for
+	// requestedCerts is expected to be a list of either "service-cert:namespace/service" or "root-cert:namespace/service"
+	requestedCerts []string
+
+	trustDomain string
+
+	serviceIdentitiesForOutboundServices map[service.MeshService][]identity.ServiceIdentity
+}
+
+// NewBuilder returns a new sdsBuilder
+func NewBuilder() *sdsBuilder { //nolint: revive // unexported-return
+	return &sdsBuilder{}
+}
+
+func (b *sdsBuilder) SetProxy(proxy *envoy.Proxy) *sdsBuilder {
+	b.proxy = proxy
+	return b
+}
+
+func (b *sdsBuilder) SetProxyCert(cert *certificate.Certificate) *sdsBuilder {
+	b.serviceCert = cert
+	return b
+}
+
+func (b *sdsBuilder) SetRequestedCerts(requestedCerts []string) *sdsBuilder {
+	b.requestedCerts = requestedCerts
+	return b
+}
+
+func (b *sdsBuilder) SetTrustDomain(trustDomain string) *sdsBuilder {
+	b.trustDomain = trustDomain
+	return b
+}
+
+func (b *sdsBuilder) SetServiceIdentitiesForService(serviceIdentitiesForServices map[service.MeshService][]identity.ServiceIdentity) *sdsBuilder {
+	b.serviceIdentitiesForOutboundServices = serviceIdentitiesForServices
+	return b
+}
+
+// Build generates SDS Secret Resources based on requested certs in the DiscoveryRequest
+func (b *sdsBuilder) Build() []*xds_auth.Secret {
+	log.Info().Str("proxy", b.proxy.String()).Msgf("Creating SDS response for request for resources %v", b.requestedCerts)
+	var sdsResources = make([]*xds_auth.Secret, 0, len(b.requestedCerts))
+	// requestedCerts is expected to be a list of either of the following:
+	// - "service-cert:namespace/service-account"
+	// - "root-cert-for-mtls-outbound:namespace/service"
+	// - "root-cert-for-mtls-inbound:namespace/service-account"
+	// The Envoy makes a request for a list of resources (aka certificates), which we will send as a response to the SDS request.
+	for _, requestedCertificate := range b.requestedCerts {
+		sdsCert, err := secrets.UnmarshalSDSCert(requestedCertificate)
+		if err != nil {
+			log.Error().Err(err).Str(errcode.Kind, errcode.GetErrCodeWithMetric(errcode.ErrUnmarshallingSDSCert)).
+				Msgf("Invalid resource kind requested: %q", requestedCertificate)
+			continue
+		}
+
+		log.Debug().Str("proxy", b.proxy.String()).Msgf("Proxy requested cert %s", requestedCertificate)
+
+		switch sdsCert.CertType {
+		// A service certificate is requested
+		case secrets.ServiceCertType:
+			envoySecret, err := b.buildServiceCertSecret(requestedCertificate)
+			if err != nil {
+				log.Error().Err(err).Str(errcode.Kind, errcode.GetErrCodeWithMetric(errcode.ErrGettingServiceCertSecret)).
+					Str("proxy", b.proxy.String()).Msgf("Error getting service cert %s for proxy", requestedCertificate)
+				continue
+			}
+			sdsResources = append(sdsResources, envoySecret)
+
+		// A root certificate used to validate a service certificate is requested
+		case secrets.RootCertTypeForMTLSInbound, secrets.RootCertTypeForMTLSOutbound:
+			envoySecret, err := b.buildRootCertSecret(*sdsCert)
+			if err != nil {
+				log.Error().Err(err).Str("proxy", b.proxy.String()).Msgf("Error getting root cert %s for proxy", requestedCertificate)
+				continue
+			}
+			sdsResources = append(sdsResources, envoySecret)
+
+		default:
+			log.Error().Str("proxy", b.proxy.String()).Msgf("Unexpected certificate type %s requested by proxy", requestedCertificate)
+		}
+	}
+	return sdsResources
+}
+
+// buildServiceCertSecret creates the struct with certificates for the service, which the
+// connected Envoy proxy belongs to.
+func (b *sdsBuilder) buildServiceCertSecret(name string) (*xds_auth.Secret, error) {
+	secret := &xds_auth.Secret{
+		// The Name field must match the tls_context.common_tls_context.tls_certificate_sds_secret_configs.name in the Envoy yaml config
+		Name: name,
+		Type: &xds_auth.Secret_TlsCertificate{
+			TlsCertificate: &xds_auth.TlsCertificate{
+				CertificateChain: &xds_core.DataSource{
+					Specifier: &xds_core.DataSource_InlineBytes{
+						InlineBytes: b.serviceCert.GetCertificateChain(),
+					},
+				},
+				PrivateKey: &xds_core.DataSource{
+					Specifier: &xds_core.DataSource_InlineBytes{
+						InlineBytes: b.serviceCert.GetPrivateKey(),
+					},
+				},
+			},
+		},
+	}
+	return secret, nil
+}
+
+func (b *sdsBuilder) buildRootCertSecret(sdscert secrets.SDSCert) (*xds_auth.Secret, error) {
+	secret := &xds_auth.Secret{
+		// The Name field must match the tls_context.common_tls_context.tls_certificate_sds_secret_configs.name
+		Name: sdscert.String(),
+		Type: &xds_auth.Secret_ValidationContext{
+			ValidationContext: &xds_auth.CertificateValidationContext{
+				TrustedCa: &xds_core.DataSource{
+					Specifier: &xds_core.DataSource_InlineBytes{
+						InlineBytes: b.serviceCert.GetTrustedCAs(),
+					},
+				},
+			},
+		},
+	}
+
+	// SAN validation should not be performed by the root validation certificate used by the upstream server
+	// to validate a downstream client. This is because of the following:
+	// 1. SAN validation is already performed by the RBAC filter on the inbound listener's filter chain (using
+	//    network RBAC filter) and each HTTP route in the inbound route ocnfiguration (using HTTP RBAC per route).
+	// 2. The same root validation certificate is used to validate both in-mesh and ingress downstreams.
+	// For these reasons, we only perform SAN validation of peer certificates on downstream clients (ie. outbound SAN
+	// validation).
+	if sdscert.CertType == secrets.RootCertTypeForMTLSInbound {
+		return secret, nil
+	}
+
+	// For the outbound certificate validation context, the SANs needs to match the list of service identities
+	// corresponding to the upstream service. This means, if the sdscert.Name points to service 'X',
+	// the SANs for this certificate should correspond to the service identities of 'X'.
+	meshSvc, err := sdscert.GetMeshService()
+	if err != nil {
+		log.Error().Err(err).Str(errcode.Kind, errcode.GetErrCodeWithMetric(errcode.ErrGettingMeshService)).
+			Msgf("Error unmarshalling upstream service for outbound cert %s", sdscert)
+		return nil, err
+	}
+	svcIdentitiesInCertRequest := b.serviceIdentitiesForOutboundServices[*meshSvc]
+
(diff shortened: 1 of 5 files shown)


pkg/envoy/sds/response.go
```diff
@@ -34,14 +39,130 @@ func NewResponse(meshCatalog catalog.MeshCataloger, proxy *envoy.Proxy, request
...
-	builder.SetProxy(proxy.String())
 
-	// TODO: decouple meshcatalog
-	return builder.Build(requestedCerts)
+	// Set trust domain
+	builder.SetTrustDomain(certManager.GetTrustDomain())
+
+	// getSDSSecrets (cert, requestedCerts, proxy)
+	// Create envoy secret resources based on the requested certs in the DiscoveryRequest
+	envoyCerts := getSDSSecrets(cert, requestedCerts, proxy, meshCatalog, builder.TrustDomain)
```
steeling, line 50: In this format the responsibility of building the SDS config is mixed between `NewResponse`, and the builder.

We should maintain single responsbility, meaning that we should set the `requestedCerts`, and the `cert` object on the builder, then the builder object deals with the `getSDSSecrets` in the build method

whitneygriffith: I have been attempting to convert `getRootCert` into a builder method and focused on decoupling `meshCatalog` but due to the difficulty of untangling it I realized that `getSDSSecrets` can remain as a composer. 

Calling `getSDSSecrets` from the Builder fits the builder pattern?

pkg/envoy/sds/response.go
```diff
@@ -78,23 +69,23 @@ func (s *sdsImpl) getSDSSecrets(cert *certificate.Certificate, requestedCerts []
...
 		// A root certificate used to validate a service certificate is requested
 		case secrets.RootCertTypeForMTLSInbound, secrets.RootCertTypeForMTLSOutbound:
-			envoySecret, err := s.getRootCert(cert, *sdsCert)
+			envoySecret, err := getRootCert(cert, *sdsCert, meshCatalog, trustDomain)
 			if err != nil {
 				log.Error().Err(err).Str("proxy", proxy.String()).Msgf("Error getting root cert %s for proxy", requestedCertificate)
 				continue
 			}
-			certs = append(certs, envoySecret)
+			sdsSecrets = append(sdsSecrets, envoySecret)
```
nshankar13, line 81: I think this logic of building the secrets should be builder.go, otherwise with the current implementation technically response.go is still handling some of the building logic for the secrets config. 

So getSDSSecrets and getRootCert would be methods in builder.go. The builder struct would have cert, requestedCerts, proxy, meshcatalog.ListServiceIdentitiesForService (we don't want to pass in the entire meshcatalog object, just what's necessary), and trustDomain as fields, and we set those when we create the builder object. 

Then, builder.Build() will call the getSDSSecrets and getRootCert methods and return the sdsResources.

pkg/envoy/sds/response.go
nshankar13: I think the way it would work (based on the feedback I've received for RDS) is that getSDSSecrets and getRootCert would be methods in builder.go. The builder struct would have cert, requestedCerts, proxy, meshcatalog.ListServiceIdentitiesForService (we don't want to pass in the entire meshcatalog object, just what's necessary), and trustDomain as fields, and we set those when we create the builder object. 

Then, builder.Build() will call the getSDSSecrets and getRootCert methods and return the sdsResources.

whitneygriffith: @nshankar13 that helped a lot! I did a spike attempting to get the results from `meshCatalog` before passing it to the builder but that turnt out messy.

pkg/envoy/sds/builder.go
```diff
@@ -0,0 +1,172 @@
...
+
+func (b *sdsBuilder) SetListServiceIdentitiesForService(listServiceIdentitiesForService func(service.MeshService) []identity.ServiceIdentity) {
+	b.ListServiceIdentitiesForService = listServiceIdentitiesForService
+}
+
+// Build generates SDS Secret Resources based on requested certs in the DiscoveryRequest
+func (b *sdsBuilder) Build() ([]types.Resource, error) {
+	var sdsResources []types.Resource
+
+	for _, envoyProto := range b.getSDSSecrets(b.proxyCert, b.requestedCerts, b.proxy) {
```
steeling, line 57: no need to pass in fields of the builder to a method on the builder. `getSDSSecrets` can simply access b.cert, b.requestedCerts, b.proxy`

steeling: same if passing the field to other methods

steeling, line 57: I might suggest refactoring this a bit, since we have 2 loops:

```go
for _, request := range b.requestedCerts {
   envoyProto, err := b.getSDSSecret(request)
   if err != nil {
      // log the error
      continue
   }
   .....
}
```

This would allow simplifying the getSDSSecrets func and also reduce to just one iteration

pkg/envoy/sds/builder.go
```diff
@@ -0,0 +1,172 @@
...
+			envoySecret, err := b.getServiceCertSecret(cert, requestedCertificate)
+			if err != nil {
+				log.Error().Err(err).Str(errcode.Kind, errcode.GetErrCodeWithMetric(errcode.ErrGettingServiceCertSecret)).
+					Str("proxy", proxy.String()).Msgf("Error getting service cert %s for proxy", requestedCertificate)
+				continue
+			}
+			certs = append(certs, envoySecret)
+
+		// A root certificate used to validate a service certificate is requested
+		case secrets.RootCertTypeForMTLSInbound, secrets.RootCertTypeForMTLSOutbound:
```
steeling, line 92: as a side note: we can notice that every single rootcert is identical.

These get picked based on the name that is set in `envoy.GetUpstreamTLSContext` in each cluster config.

With this refactoring we could also switch these names to be a static name: `osm-root-cert`, and have just 1 single root cert for validation, instead of one identical root cert for each individual service.

Not necessary in this PR, but worth exploring to understand ramifications :)

pkg/envoy/sds/builder.go
whitneygriffith: Good catch!

pkg/envoy/sds/builder.go
```diff
@@ -0,0 +1,171 @@
...
+	b.trustDomain = trustDomain
+}
+
+func (b *sdsBuilder) SetListServiceIdentitiesForService(listServiceIdentitiesForService func(service.MeshService) []identity.ServiceIdentity) {
+	b.ListServiceIdentitiesForService = listServiceIdentitiesForService
+}
+
+// Build generates SDS Secret Resources based on requested certs in the DiscoveryRequest
+func (b *sdsBuilder) Build() ([]types.Resource, error) {
+	b.getSDSSecrets()
```
steeling, line 58: nit: no need to have a separate function at this point

pkg/envoy/sds/builder.go
```diff
@@ -0,0 +1,171 @@
...
+	return b.sdsResources, nil
+}
+
+func (b *sdsBuilder) getSDSSecrets() {
+	// requestedCerts is expected to be a list of either of the following:
+	// - "service-cert:namespace/service-account"
+	// - "root-cert-for-mtls-outbound:namespace/service"
+	// - "root-cert-for-mtls-inbound:namespace/service-account"
+
+	b.sdsResources = nil
```
steeling, line 68: nit: no need to set the output on the builder, can simply just return this (from the build method as well)

pkg/envoy/sds/response.go
```diff
@@ -12,155 +10,39 @@ import (
...
-	// the SANs for this certificate should correspond to the service identities of 'X'.
-	meshSvc, err := sdscert.GetMeshService()
-	if err != nil {
-		log.Error().Err(err).Str(errcode.Kind, errcode.GetErrCodeWithMetric(errcode.ErrGettingMeshService)).
-			Msgf("Error unmarshalling upstream service for outbound cert %s", sdscert)
-		return nil, err
-	}
-	svcIdentitiesInCertRequest := s.meshCatalog.ListServiceIdentitiesForService(*meshSvc)
+	// Lists service identities for certs in request
+	builder.SetListServiceIdentitiesForService(meshCatalog.ListServiceIdentitiesForService)
```
steeling, line 43: instead of having a setter that takes in a method, we want to remove all API calls from the builder, and only set data.

What we can do instead is maintain a `map[service.MeshService][]identity.ServiceIdentity` on the builder.

And we can set the data from the response.go's `NewResponse` function.

whitneygriffith:
Next up: 

- [ ] Test Build function which replaces `TestGetSDSSecrets`
- [ ] Test Get Service Identities

pkg/envoy/sds/builder.go
```diff
@@ -0,0 +1,163 @@
...
+	serviceIdentitiesForServices map[service.MeshService][]identity.ServiceIdentity
+}
+
+func newSdsBuilder() *sdsBuilder {
+	return &sdsBuilder{}
+}
+
+func (b *sdsBuilder) SetProxy(proxy *envoy.Proxy) {
+	b.proxy = proxy
+}
```
shashankram, line 35: Why is `newSdsBuilder()` unexported while the methods on the builder type are exported?

pkg/envoy/sds/builder.go
```diff
@@ -0,0 +1,163 @@
...
+	b.trustDomain = trustDomain
+}
+
+func (b *sdsBuilder) SetServiceIdentitiesForService(serviceIdentitiesForServices map[service.MeshService][]identity.ServiceIdentity) {
+	b.serviceIdentitiesForServices = serviceIdentitiesForServices
+}
+
+// Build generates SDS Secret Resources based on requested certs in the DiscoveryRequest
+func (b *sdsBuilder) Build() (sdsResources []types.Resource) {
+
```
shashankram, line 55: nit: remove unnecessary newline

pkg/envoy/sds/builder.go
```diff
@@ -0,0 +1,163 @@
...
+func (b *sdsBuilder) SetTrustDomain(trustDomain string) {
+	b.trustDomain = trustDomain
+}
+
+func (b *sdsBuilder) SetServiceIdentitiesForService(serviceIdentitiesForServices map[service.MeshService][]identity.ServiceIdentity) {
+	b.serviceIdentitiesForServices = serviceIdentitiesForServices
+}
+
+// Build generates SDS Secret Resources based on requested certs in the DiscoveryRequest
+func (b *sdsBuilder) Build() (sdsResources []types.Resource) {
```
shashankram, line 54: Let's drop the named return type since it doesn't improve readability here.
```suggestion
func (b *sdsBuilder) Build() []types.Resource {
```

pkg/envoy/sds/builder.go
```diff
@@ -0,0 +1,163 @@
...
+	"github.com/openservicemesh/osm/pkg/errcode"
+	"github.com/openservicemesh/osm/pkg/identity"
+	"github.com/openservicemesh/osm/pkg/service"
+)
+
+type sdsBuilder struct {
+	proxy *envoy.Proxy
+
+	// Service certificate for this proxy
+	proxyCert *certificate.Certificate
```
shashankram, line 19:
```suggestion
	serviceCert *certificate.Certificate
```

pkg/envoy/sds/builder.go
```diff
@@ -0,0 +1,163 @@
...
+		default:
+			log.Error().Str("proxy", b.proxy.String()).Msgf("Unexpected certificate type %s requested by proxy", requestedCertificate)
+		}
+	}
+	return sdsResources
+}
+
+// getServiceCertSecret creates the struct with certificates for the service, which the
+// connected Envoy proxy belongs to.
+func (b *sdsBuilder) getServiceCertSecret(name string) (*xds_auth.Secret, error) {
```
shashankram, line 101: nit: s/get/build 
```suggestion
func (b *sdsBuilder) buildServiceCertSecret(name string) (*xds_auth.Secret, error) {
```

pkg/envoy/sds/builder.go
```diff
@@ -0,0 +1,163 @@
...
+	b.requestedCerts = requestedCerts
+}
+
+func (b *sdsBuilder) SetTrustDomain(trustDomain string) {
+	b.trustDomain = trustDomain
+}
+
+func (b *sdsBuilder) SetServiceIdentitiesForService(serviceIdentitiesForServices map[service.MeshService][]identity.ServiceIdentity) {
+	b.serviceIdentitiesForServices = serviceIdentitiesForServices
+}
```
shashankram, line 51: Let's allow chaining the calls to set functions by returning the builder type in these methods. See https://github.com/openservicemesh/osm/blob/main/pkg/envoy/lds/builder.go#L31 and https://github.com/openservicemesh/osm/blob/main/pkg/envoy/lds/response.go#L42 for examples on how the calls can be chained by the caller.

pkg/envoy/sds/response.go
```diff
@@ -14,50 +12,45 @@ import (
...
-	// 2. Create SDS secret resources based on the requested certs in the DiscoveryRequest
-	// request.ResourceNames is expected to be a list of either "service-cert:namespace/service" or "root-cert:namespace/service"
-	for _, envoyProto := range s.getSDSSecrets(cert, requestedCerts, proxy) {
-		sdsResources = append(sdsResources, envoyProto)
-	}
+	builder.SetTrustDomain(certManager.GetTrustDomain())
 
-	return sdsResources, nil
+	// Set service identities for services in requests
+	log.Info().Msgf("Getting Service Identities for services in request for resources %v", builder.requestedCerts)
```
shashankram, line 44:
```suggestion
	log.Trace().Msgf("Getting Service Identities for services in request for resources %v", builder.requestedCerts)
```

pkg/envoy/sds/response_test.go
```diff
@@ -337,43 +308,28 @@ func TestGetSDSSecrets(t *testing.T) {
...
-				actualSANs := subjectAltNamesToStr(sdsSecret.GetValidationContext().GetMatchSubjectAltNames())
-				assert.ElementsMatch(actualSANs, tc.expectedSANs)
-
-				// Check trusted CA
-				assert.NotNil(sdsSecret.GetValidationContext().GetTrustedCa().GetInlineBytes())
-
-			case secrets.ServiceCertType:
-				assert.NotNil(sdsSecret.GetTlsCertificate().GetCertificateChain().GetInlineBytes())
-				assert.NotNil(sdsSecret.GetTlsCertificate().GetPrivateKey().GetInlineBytes())
-			}
```
shashankram, line 376: Why are these being removed? I don't see them being tested with the new changes.

shashankram (changes requested):

pkg/envoy/sds/response.go
```diff
@@ -66,100 +59,30 @@ func (s *sdsImpl) getSDSSecrets(cert *certificate.Certificate, requestedCerts []
 			continue
 		}
 
-		log.Debug().Str("proxy", proxy.String()).Msgf("Proxy requested cert %s", requestedCertificate)
-
 		switch sdsCert.CertType {
```
steeling, line 62: nit: a switch case is overkill here. just doing ` if sdsCert.CertType == secrets.RootCertTypeForMTLSOutbound {...}`

pkg/errcode/errcode.go
```diff
@@ -162,6 +162,9 @@ const (
 
 	// ErrMeshConfigFetchFromCache indicates failed to fetch MeshConfig from cache with specific key
 	ErrMeshConfigFetchFromCache
+
+	// ErrGettingServiceIdentitiesForService indicates failed to fetch Service Identities for a Mesh Service
+	ErrGettingServiceIdentitiesForService
```
shashankram, line 167: You also want to add this error code and its description to the `ErrCodeMap` so that the error code tool `osm support error-info` can display the error info.

pkg/envoy/sds/builder.go
```diff
@@ -0,0 +1,167 @@
+package sds
+
+import (
+	xds_core "github.com/envoyproxy/go-control-plane/envoy/config/core/v3"
+	xds_auth "github.com/envoyproxy/go-control-plane/envoy/extensions/transport_sockets/tls/v3"
+	"github.com/openservicemesh/osm/pkg/certificate"
```
shashankram, line 6: To address this lint error, add an empty line between import groups:
```
Go stdlib imports

external imports

OSM API imports

OSM internal pkg imports
```

pkg/envoy/sds/builder.go
```diff
@@ -0,0 +1,167 @@
...
+	// The DiscoveryRequest contains the certs to generate secrets for
+	// requestedCerts is expected to be a list of either "service-cert:namespace/service" or "root-cert:namespace/service"
+	requestedCerts []string
+
+	trustDomain string
+
+	serviceIdentitiesForOutboundServices map[service.MeshService][]identity.ServiceIdentity
+}
+
+func NewBuilder() *sdsBuilder {
```
shashankram, line 29: The linter expects exported functions and types to have a comment, so you need to add those for the linter to pass.

shashankram (commented):
The changes look good. There are a few lint failures to address.

steeling (approved):

pkg/errcode/errcode.go
```diff
@@ -581,6 +584,10 @@ Failed to cast object to MeshConfig.
 `,
 	ErrMeshConfigFetchFromCache: `
 Failed to fetch MeshConfig from cache with specific key.
+`,
+
+	ErrGettingServiceIdentitiesForService: `
+Failed to fetch Service Identities for Mesh Service
```
shashankram, line 590: nit: missing period at the end of the description.

shashankram (approved):

PR #4982, opened 2022-08-10
remove smi resource specific healthcheck

steeling:
This PR is meant to facilitate discussion, I can close if we decide we want to keep it.

The SMI healthcheck is the only healthcheck that is constantly probing to see if the SMI resources exist. In a proper installation of OSM this should be unnecessary, and is inconsistent with our other CRD's, and the CRD reconciler.

I believe we should use validating webhooks/reconcilers instead.


shashankram (approved):

keithmattix (approved):
+1 from me

Issue #1684, opened 2020-09-10
Better API to list services based on filters


eduser25:
We are missing a way to apply filtering on the `ListServices` API on the kubernetes package, similar to what the `selectors` that can be filled in on an API call to K8s API endpoints. Some services are not interested in seeing all services, and might just be interested ie. on services on a particular namespace. They currently are required to implement apply the filtering themselves.

Has to be researched.

- New Functionality          [X]
- Envoy Control Plane        [X]

**Possible use cases**
`getServicesByLabels` for example, requires to apply selectors only on services on a specific namespace.

Issue #4863, opened 2022-06-30
Push all k8s clients into a layer of clients


steeling:
We currently have ~5 flavors of kubernetes clients:

* k8s.Controller
* generated "controller" clients for our CRD's
* A large informerCollection type
* a `kube.Client`
* kubernetes.Interface

Instead, we should leverage "layers" of clients, such that our business logic is only aware of 1 layer. This might looks like:

```go
// K8sAPIClient is an abstraction of the K8s API. It has inputs of OSM objects and returns OSM objects.
// It performs the translations, and makes calls to the underlying K8sClient.
type K8sInfraClient struct {
   client K8sClient
}

// K8sClient is aware of k8s types, and has inputs/outputs of k8s objects. It may leverage the informer cache
type K8sClient struct {
   informerCollection
}
```

As we do this, no functions should accept a client object just to make a call to that client. It should only be accepted for dependency injection. An example of this includes `GetKubernetesServerVersionNumber`. Instead, this should be called `GetServerVersionNumber`, and be a method on a k8s client.

This improves our ability to decouple from k8s, and leverage interfaces. Please take a look at the [guiding principles section](https://docs.google.com/document/d/1L_u-XlVU8Fog-uFS3Uf0O2nz8IArQy82cd6Npxm038Y/edit#) for how we can better leverage interfaces.

The new layer for access to these clients will be:

1. MeshCatalog
    a) Contains business logic for constructing the trafficpolicy types
    b) Has no concept of anything related to K8s
    c) Has an embedded compute.Interface
2. The compute.Interface
   a) Only accepts/returns non-k8s type objects. (The exception to this is custom objects where we do not have an analogous OSM type
   b) Performs translations to/from k8s type objects
   c) There may be multiple instances of this, which is how we can interface outside of k8s.
3. The k8s.Client
   a) Basic Get/Set/Update to kubernetes API
   c) Maintains the informer cache
   d) Has a client for each k8s kind
   e) Is only aware of k8s types, is not aware of OSM types

Some nice results:
1. Eventually the MeshCatalog will only contain business logic, so its interface can go away, and any mocks can supply a mock compute.Interface to a real MeshCatalog instance. This will improve our testing of business logic in the MeshCatalog.
2. With the combination of the Builder pattern, which reduces the need for the mesh catalog, and the potential deconstruction of trafficpolicy types, the mesh catalog may be able to be removed completely. This direction should become more clear in the future.

> 

> contains business logic

steeling:
[...]

PR #4984, opened 2022-08-10
Move K8s calls out of ProxyRegistry

steeling:
Move the GetProxyServices call from the proxy registry into the compute provider

part of #4863


shashankram (approved):

jaellio:
Is there a related issue for this change?

steeling:
> Is there a related issue for this change?

Added to description

pkg/compute/kube/client.go
```diff
@@ -141,6 +139,27 @@ func (c *client) GetServicesForServiceIdentity(svcIdentity identity.ServiceIdent
 	return meshServices
 }
 
+// GetServicesForProxy maps an Envoy instance to a number of Kubernetes services.
+func (c *client) GetServicesForProxy(p *envoy.Proxy) ([]service.MeshService, error) {
```
keithmattix, line 143: Not something to fix in this PR, but we should try and avoid using this function as much as possible or finally fix #1684. A loop through k8s service in our cache is a bottleneck for large clusters

pkg/envoy/cds/response.go
```diff
@@ -14,7 +14,7 @@ import (
 )
 
 // NewResponse creates a new Cluster Discovery Response.
-func NewResponse(meshCatalog catalog.MeshCataloger, proxy *envoy.Proxy, _ *xds_discovery.DiscoveryRequest, _ *certificate.Manager, proxyRegistry *registry.ProxyRegistry) ([]types.Resource, error) {
+func NewResponse(meshCatalog catalog.MeshCataloger, proxy *envoy.Proxy, _ *xds_discovery.DiscoveryRequest, _ *certificate.Manager, _ *registry.ProxyRegistry) ([]types.Resource, error) {
```
keithmattix, line 17: Are these unused args necessary to fulfill an interface?

pkg/compute/kube/client.go
```diff
@@ -141,6 +139,27 @@ func (c *client) GetServicesForServiceIdentity(svcIdentity identity.ServiceIdent
 	return meshServices
 }
 
+// GetServicesForProxy maps an Envoy instance to a number of Kubernetes services.
```
keithmattix, line 142: Not something to fix in this PR, but we should try and avoid using this function as much as possible or finally fix https://github.com/openservicemesh/osm/issues/1684. A loop through k8s service in our cache is a bottleneck for large clusters

pkg/envoy/lds/response.go
```diff
@@ -31,7 +31,7 @@ func NewResponse(meshCatalog catalog.MeshCataloger, proxy *envoy.Proxy, _ *xds_d
 		statsHeaders = proxy.StatsHeaders()
 	}
 
-	svcList, err := proxyRegistry.ListProxyServices(proxy)
+	svcList, err := meshCatalog.GetServicesForProxy(proxy)
```
keithmattix, line 34: nit: by convention, our code seems to use `Get` for singular queries and `List` for plural queries. Is there a reason for the switch here?

cmd/osm-controller/osm-controller.go
```diff
@@ -240,8 +240,7 @@ func main() {
 		msgBroker,
 	)
 
-	proxyMapper := &registry.KubeProxyServiceMapper{KubeController: k8sClient}
-	proxyRegistry := registry.NewProxyRegistry(proxyMapper, msgBroker)
+	proxyRegistry := registry.NewProxyRegistry()
```
keithmattix, line 243: So it seems like the gist of this change is that the proxyRegistry loses a dependency on a `proxyMapper` and instead relies on the new `GetServicesForProxy` method declared on the client (accessible to CDS via the meshcatalog's k8s client instead of the proxy registry). Is that correct? If so, it seems like this changeset further decouples our Envoy config from k8s primitives, increasing the use of our domain objects. Nice!

keithmattix (approved):
Approving pending some clarifications

cmd/osm-controller/osm-controller.go
steeling: that's correct!

pkg/envoy/lds/response.go
steeling: will make the change back to List

pkg/compute/kube/client.go
steeling: ya we have a lot of methods that iterate over the iterate over the entire informer cache. Unfortunately the informer cache doesn't have a great way to implement that, but agree we can revisit in a subsequent PR (maybe maintain our own internal mappings)

pkg/envoy/cds/response.go
steeling: yes they are, there will be an opportunity to remove that though with the rest of the snapshotcache + refactoring that is currently ongoing

pkg/compute/kube/client.go
keithmattix: One option is to implement our own [SharedIndexInformers](https://pkg.go.dev/k8s.io/client-go/tools/cache#SharedIndexInformer) for certain types and auto-index based on arbitrary attributes. In this case, for v1.Service, we could index based on a set of selectors.

steeling: would be really cool if we could create an indexer by proxy. Especially if we could then have the EventHandlers be more proxy aware, so we could simply say "get proxies affected by this update", and schedule the resulting update.

keithmattix: +1 - let's see if there's performance impact after these updates; if so, we may need to take another look at this

PR #4985, opened 2022-08-10
demo: script to set up vanilla multicluster env

### LINKED ISSUE #4974: Create a workflow to generate 2 clusters provisioned with ServiceImports/Exports
@allenlsy has been experimenting with Submariner. Submariner is a unique MCS implementation that creates ServiceImports/Exports and populates with the IP address of the Submariner gateway, which handles network connectivity across clusters.

As a service mesh, we will need to create our own network connectivity solution, agnostic to submariner. This bug is explicitly to create a workflow to "join" to clusters, such that they are aware of a set of ServiceImports and ServiceExports for appropriate services in each cluster.

This may be as simple as a script that wraps kubectl commands to:

1. Create ServiceImport and ServiceExport CRD's
2. Create a loadbalancer in each cluster
3. create ServiceExports in each cluster
4. Create the appropriate ServiceImports in the remote clusters, corresponding to the ServiceImport.
5. The ServiceImport clusterIP should also work with CoreDNS, so we will likely need to add a CoreDNS entry

allenlsy:
**Description**:

Add a script to set up a demo multicluster environment for OSM development.

It deploys the sample bookstore application on cluster c1 and c2. C2 exports its bookstore service via `ServiceExport` and c1 imports it via `ServiceImport`. A new `EndpointSlice` resource on c1 is created to store accessible endpoint from c2.

The steps listed in the script are:

1. Setup CRDs
2. Deploy applications
3. Create `ServiceExport` in c2
4. Create `ServiceImport` in c1. It will have a ClusterSetIP
5. Expose bookstore on c2 and get service public IP
6. On c1, create `EndpointSlice` containing the remote service IP 
7. (optional) update coreDNS entries of domain `bookstore.bookstore.svc.clusterset.local`

Partially resolves #4974 

**Testing done**:

Script is tested by executing with 2 clusters.

The end result of running the script:

* [c1] Created ServiceImport with MultiClusterService IP
* [c1] Created EndpointSlice with public service IP from c2
* [c2] Created ServiceExport

**Affected area**:

| Demo                       | [x] |

No.

The demo is inspired by [mcs-api](https://github.com/kubernetes-sigs/mcs-api/tree/master/demo)

8. Is this a breaking change?

No

9. Has documentation corresponding to this change been updated in the [osm-docs](https://github.com/openservicemesh/osm-docs) repo (if applicable)?

No

### DIFF
--- a/scripts/multicluster/rbac/role.yaml
+++ b/scripts/multicluster/rbac/role.yaml
@@ -1,0 +1,50 @@
+
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  creationTimestamp: null
+  name: mcs-derived-service-manager
+rules:
+- apiGroups:
+  - ""
+  resources:
+  - services
+  verbs:
+  - create
+  - get
+  - list
+  - patch
+  - update
+  - watch
+- apiGroups:
+  - ""
+  resources:
+  - services/status
+  verbs:
+  - create
+  - get
+  - list
+  - patch
+  - update
+  - watch
+- apiGroups:
+  - discovery.k8s.io
+  resources:
+  - endpointslices
+  verbs:
+  - get
+  - list
+  - patch
+  - update
+  - watch
+- apiGroups:
+  - multicluster.x-k8s.io
+  resources:
+  - serviceimports
+  verbs:
+  - get
+  - list
+  - patch
+  - update
+  - watch
--- a/scripts/multicluster/setup.sh
+++ b/scripts/multicluster/setup.sh
@@ -1,0 +1,176 @@
+#!/bin/bash
+
+# This script sets up a demo multicluster environment for OSM development.
+# It deploys the sample bookstore application on cluster c1 and c2. C2 exports its
+# bookstore service via ServiceExport and c1 imports it via ServiceImport. A new EndpointSlice
+# resource on c1 is created to store accessible endpoint from c2.
+
+# shellcheck disable=SC2128
+# shellcheck disable=SC2086
+cd "$(dirname ${BASH_SOURCE})"
+set -e
+
+c1=${CLUSTER1:-c1}
+c2=${CLUSTER2:-c2}
+k1="kubectl --context ${c1}"
+k2="kubectl --context ${c2}"
+app_port=14001
+
+# ==== 1. Setup CRDs
+
+${k1} apply -f ./crd -f ./rbac
+${k2} apply -f ./crd -f ./rbac
+
+# ==== 2. Deploy applications
+for ctx in ${c1} ${c2}; do
+    kubectl config use-context "${ctx}"
+
+    echo "[${ctx}] Installing osm"
+    osm install
+
+    echo "[${ctx}] Deploying demo application"
+    kubectl create namespace bookstore
+    kubectl create namespace bookbuyer
+    kubectl create namespace bookthief
+    kubectl create namespace bookwarehouse
+    osm namespace add bookstore bookbuyer bookthief bookwarehouse
+    kubectl apply -f https://raw.githubusercontent.com/openservicemesh/osm-docs/release-v1.2/manifests/apps/bookbuyer.yaml
+    kubectl apply -f https://raw.githubusercontent.com/openservicemesh/osm-docs/release-v1.2/manifests/apps/bookstore.yaml
+    kubectl apply -f https://raw.githubusercontent.com/openservicemesh/osm-docs/release-v1.2/manifests/apps/bookwarehouse.yaml
+    kubectl apply -f https://raw.githubusercontent.com/openservicemesh/osm-docs/release-v1.2/manifests/apps/mysql.yaml
+done
+
+# ==== Deploy a curl client for testing
+${k1} apply -f - <<EOF
+apiVersion: v1
+kind: Deployment
+metadata:
+  name: curl
+  namespace: bookbuyer
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app: curl
+  template:
+    metadata:
+      labels:
+        app: curl
+    spec:
+      containers:
+      - name: curl
+        image: curlimages/curl
+        command: ["sleep", "1000000"]
+EOF
+
+# ==== 3. Create ServiceExport
+${k2} apply -f - <<EOF
+apiVersion: multicluster.x-k8s.io/v1alpha1
+kind: ServiceExport
+metadata:
+  name: bookstore
+  namespace: bookstore
+EOF
+
+# ==== 4. Create ServiceImport
+
+# create a meta-service to get a valid local service IP
+${k1} apply -f - <<EOF
+apiVersion: v1
+kind: Service
+metadata:
+  name: mcs-bookstore
+  namespace: bookstore
+spec:
+  type: ClusterIP
+  ports:
+  - port: ${app_port}
+    protocol: TCP
+EOF
+sleep 5
+mcs_ip=$(${k1} get service mcs-bookstore -n bookstore -o jsonpath='{.spec.clusterIP}')
+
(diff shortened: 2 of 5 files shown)


scripts/multicluster/rbac/role.yaml
```diff
@@ -0,0 +1,50 @@
+
+---
```
keithmattix, line 2: If there's no ClusterRoleBinding, I don't think this does anything

scripts/multicluster/setup.sh
```diff
@@ -0,0 +1,176 @@
...
+  ports:
+  - name: http
+    protocol: TCP
+    port: ${app_port}
+EOF
+
+# ==== 5. Expose service on c2 and get the IP addresses.
+${k2} patch service bookstore -n bookstore -p '{"spec":{"type":"LoadBalancer"}}'
+for _ in {1..10}; do
+    remote_ip=$(${k2} get service bookstore -n bookstore -o jsonpath='{.status.loadBalancer.ingress[0].ip}')
```
keithmattix, line 112: I believe the MCS API doesn't apply to LoadBalancer IPs; only in-cluster IPs

scripts/multicluster/setup.sh
```diff
@@ -0,0 +1,176 @@
...
+    name: bookstore
+    uid: ${uid}
+addressType: IPv4
+ports:
+  - name: http
+    protocol: TCP
+    port: ${app_port}
+endpoints:
+  - addresses:
+    - "${remote_ip}"
```
keithmattix, line 151: This should be a list of actual pod IPs

keithmattix (changes requested):
The shape of this looks generally good; just some things to fix to make sure we're compliant with the spec

scripts/multicluster/setup.sh
allenlsy: The actual pod IPs needs the assumption that two clusters are in the flat network (this applies to the last comment also). If they are not in a flat network, AFAIK an extra layer (like Submariner which provides cluster gateway) is required. In our simple setup, we don't have these.

scripts/multicluster/setup.sh
```diff
@@ -0,0 +1,176 @@
...
+
+# ==== 4. Create ServiceImport
+
+# create a meta-service to get a valid local service IP
+${k1} apply -f - <<EOF
+apiVersion: v1
+kind: Service
+metadata:
+  name: mcs-bookstore
+  namespace: bookstore
```
steeling, line 83: let's create this in a non-meshed namespace so it doesn't conflict with anything else in OSM with permissive mode

scripts/multicluster/setup.sh
```diff
@@ -0,0 +1,176 @@
...
+  name: bookstore
+  namespace: bookstore
+spec:
+  ips:
+  - ${mcs_ip}
+  type: ClusterSetIP
+  ports:
+  - name: http
+    protocol: TCP
+    port: ${app_port}
```
steeling, line 106: Let's add the status that has the list of other clusters, as described in the kep https://github.com/kubernetes/enhancements/tree/master/keps/sig-multicluster/1645-multi-cluster-services-api#importing-services

scripts/multicluster/setup.sh
steeling: agreed with @allenlsy  in a non-flat network, we will want this field here.

We may add an OSM object that will enable this, by putting the remote IP in a separate location, and we use that instead, TBD.

For now I think this is acceptable, although maybe we should add a comment about that

scripts/multicluster/setup.sh
```diff
@@ -0,0 +1,176 @@
...
+  name: coredns-custom
+  namespace: kube-system
+data:
+  mcs-bookstore.server: |
+    bookstore.bookstore.svc.clusterset.local:53 {
+      errors
+      cache 30
+      hosts {
+        ${mcs_ip} bookstore.bookstore.svc.clusterset.local
+        ${local_ip} bookstore.bookstore.svc.clusterset.local
```
steeling, line 168: This should not resolve to the local ip, only the above mcs_ip is necessary.

Remember it is kube proxy that knows how to resolve from the ClusterSetIP to the appropriate set of EndpointSlices

scripts/multicluster/setup.sh
keithmattix: Yeah this is likely fine for our setup...but there's some merging with duplicate serviceexport names in the original spec that I want to make sure we're aware of

scripts/multicluster/setup.sh
keithmattix: Kube-proxy doesn't natively know how to resolve to clusterset.local  without an update right?

steeling: Correct, but our traffic also doesn't go through kube-proxy, so we'll need to bring the functionality to the sidecar either way

scripts/multicluster/setup.sh
steeling: can you elaborate?

keithmattix: From [the KEP](https://github.com/kubernetes/enhancements/tree/master/keps/sig-multicluster/1645-multi-cluster-services-api)

> If multiple clusters export a Service with the same namespaced name, they will be recognized as a single combined service. For example, if 5 clusters export my-svc.my-ns, each importing cluster will have one ServiceImport named my-svc in the my-ns namespace and it will be associated with endpoints from all exporting clusters. Properties of the ServiceImport (e.g. ports, topology) will be derived from a merger of component Service properties.

I don't see anything that requires a merge between a commonly named local Service and ServiceImports, but if the owning cluster creates a ServiceExport for that internal service, then this seems to imply that the endpoints should be merged. The non-flat network question may need some clarification in the upstream KEP

scripts/multicluster/setup.sh
keithmattix: Ah yes, I was just addressing this:
> Remember it is kube proxy that knows how to resolve from the ClusterSetIP to the appropriate set of EndpointSlices

Did you perhaps mean that kube-proxy knows how to resolve the local ClusterIP rather than the ClustersetIP?

scripts/multicluster/setup.sh
allenlsy: I think when importing external services, the importing cluster will copy the EndpointSlices from the exporting cluster, but not merging the endpoints to existing EndpointSlice. Hope this clarifies the process.

Either way (using a single remote service IP or a list of remote pod IPs) are reasonable here. It depends on how we create the multiple clusters.

scripts/multicluster/setup.sh
keithmattix: I see that we're using the LB IP to simulate a cluster gateway. Resolving this comment

allenlsy:
FYI I executed the script with c1 and c2 deployed on same subnet without OSM installed, c1 bookbuyer cannot access c2 bookstore by ClusterIP. But c1 bookbuyer can access c2 bookstore by:

* c2 bookstore endpoint's pod IP
* OR c2 bookstore external IP (if exposed)

With OSM installed, c1 cannot access c2 services due to failed mTLS authn.

steeling (approved):

keithmattix (approved):

scripts/multicluster/rbac/role.yaml
allenlsy: These manifests are referred from https://github.com/kubernetes-sigs/mcs-api/tree/master/config. ClusterRole is not used in this demo. We can keep it for future use.

jsturtevant:
> With OSM installed, c1 cannot access c2 services due to failed mTLS authn.

Is it possible to make them talk with OSM installed?

Maybe a readme could be helpful for things like: what the configuration of the two clusters was (kind/aks) and was there additional components installed (Submariner)?

PR #4996, opened 2022-08-12
feat(certs): update ADS server tls config on cert rotation

### LINKED ISSUE #4839: Support updating TLS config without restarting server

jaellio:
**Description**:

Updates the gRPC server's tls config to support dynamic reloading.
Subscribes to the server's certificate and watches for rotations.
On a rotation, updates the client validation and server certs in
the tls config.

Resolves #4839

**Testing done**:
- [x] manually verify the tls config is updated on ads server cert rotation on expiration 
- [ ] write unit test or e2e to verify the tls config is updated on cert rotation and connections can still be established with the server

**Affected area**:


pkg/utils/grpc.go
```diff
@@ -27,44 +46,139 @@ func NewGrpc(serverType string, port int, certPem, keyPem, rootCertPem []byte) (
...
 	grpcOptions = append(grpcOptions, mutualTLS)
 
-	return grpc.NewServer(grpcOptions...), lis, nil
+	s.server = grpc.NewServer(grpcOptions...)
+	return s, lis, nil
+}
+
+// GetServer returns the gRPC server
+func (s *Server) GetServer() *grpc.Server {
+	s.mu.Lock()
```
jaellio, line 82: Is a lock necessary for the grpc server?

steeling: looks like you're never changing it so nope

pkg/utils/grpc.go
```diff
@@ -27,44 +46,139 @@ func NewGrpc(serverType string, port int, certPem, keyPem, rootCertPem []byte) (
...
-	log.Info().Msgf("[grpc][%s] stopping %s server", serverType, serverType)
+		log.Info().Msgf("[grpc][%s] stopping %s server", s.name, s.name)
 
-	if grpcServer != nil {
-		log.Info().Msgf("[grpc][%s] Gracefully stopping %s gRPC server", serverType, serverType)
-		grpcServer.GracefulStop()
-		log.Info().Msgf("[grpc][%s] gRPC Server stopped", serverType)
-	}
-	log.Info().Msgf("[grpc][%s] exiting %s gRPC server", serverType, serverType)
+		if s.server != nil {
```
steeling, line 176: no nil check needed here either

pkg/utils/grpc.go
```diff
@@ -27,44 +46,139 @@ func NewGrpc(serverType string, port int, certPem, keyPem, rootCertPem []byte) (
...
+		Certificates:       []tls.Certificate{certif},
+		ClientCAs:          certPool,
+		MinVersion:         constants.MinTLSVersion,
+	}
+	s.mu.Unlock()
+
+	return nil
+}
+
+func (s *Server) configureConfigUpdates(ctx context.Context) error {
```
steeling, line 122: can we try a different name?

watchCertRotations maybe?

steeling (approved):

pkg/envoy/ads/server.go
```diff
@@ -32,6 +32,9 @@ const (
 
 	// workerPoolSize is the default number of workerpool workers (0 is GOMAXPROCS)
 	workerPoolSize = 0
+
+	// xdsServerCertificateCommonName is the name of the certificate for the ADS server
+	xdsServerCertificateCommonName = "ads"
```
shashankram, line 37: The variable name indicates this being a CN, while the comment refers to this as the certificate name. Could you make the terminology consistent.

pkg/envoy/ads/server.go
```diff
@@ -70,23 +73,28 @@ func (s *Server) withXdsLogMutex(f func()) {
 
 // Start starts the ADS server
 func (s *Server) Start(ctx context.Context, cancel context.CancelFunc, port int, adsCert *certificate.Certificate) error {
-	grpcServer, lis, err := utils.NewGrpc(ServerType, port, adsCert.GetCertificateChain(), adsCert.GetPrivateKey(), adsCert.GetTrustedCAs())
+	grpcServer, lis, err := utils.NewGrpc(ServerType, port, xdsServerCertificateCommonName, s.certManager)
 	if err != nil {
 		log.Error().Err(err).Str(errcode.Kind, errcode.GetErrCodeWithMetric(errcode.ErrStartingADSServer)).
-			Msg("Error starting ADS server")
+			Msg("Error creating ADS server")
```
shashankram, line 79: The error code refers to an error starting the ADS server, while the log refers to an error creating the server. Let's make this consistent.

Further, since `utils.NewGrpc` already logs an error, let's skip logging an error here and return by wrapping the original error:
`return fmt.Errorf("error starting ADS server: %w, err)`

pkg/utils/grpc.go
```diff
@@ -2,21 +2,40 @@ package utils
...
+	server   *grpc.Server
+	certName string
+
+	mu     sync.Mutex
+	config tls.Config
+}
+
 // NewGrpc creates a new gRPC server
-func NewGrpc(serverType string, port int, certPem, keyPem, rootCertPem []byte) (*grpc.Server, net.Listener, error) {
+func NewGrpc(serverType string, port int, certName string, cm *certificate.Manager) (*Server, net.Listener, error) {
```
shashankram, line 38: This file is meant for a generic `gRPC` server, and not specifically ADS. Passing ADS specific  getters such as `certificate.Manager` indicates it's best to move this code under `pkg/envoy/ads` instead.

In other words, this no longer seems to be a utility function of some kind, like before.

pkg/utils/grpc.go
```diff
@@ -27,44 +46,139 @@ func NewGrpc(serverType string, port int, certPem, keyPem, rootCertPem []byte) (
...
+}
+
+// GetServer returns the gRPC server
+func (s *Server) GetServer() *grpc.Server {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	return s.server
+}
+
+func (s *Server) setConfig() error {
```
shashankram, line 87: A setter method without any arguments seems ambiguous. Would rename this to `initConfig` or something similar.

pkg/utils/grpc.go
```diff
@@ -27,44 +46,139 @@ func NewGrpc(serverType string, port int, certPem, keyPem, rootCertPem []byte) (
...
+		s.certName,
+		certificate.Internal,
+		certificate.FullCNProvided())
+	if err != nil {
+		return err
+	}
+
+	certif, err := tls.X509KeyPair(grpcCert.GetCertificateChain(), grpcCert.GetPrivateKey())
+	if err != nil {
+		return fmt.Errorf("[grpc][mTLS][%s] Failed loading Certificate (%+v) and Key (%+v) PEM files", s.name, grpcCert.GetCertificateChain(), grpcCert.GetPrivateKey())
```
shashankram, line 98: No need for `[grpc][mTLS]` prefixes in log lines. This predates the use of zerolog in the project, which already logs filenames and components. This was only necessary when we were using glog when filenames/components weren't logged.

pkg/utils/grpc.go
```diff
@@ -27,44 +46,139 @@ func NewGrpc(serverType string, port int, certPem, keyPem, rootCertPem []byte) (
...
+	}
+
+	certPool := x509.NewCertPool()
+
+	// Load the set of Root CAs
+	if ok := certPool.AppendCertsFromPEM(grpcCert.GetTrustedCAs()); !ok {
+		return fmt.Errorf("[grpc][mTLS][%s] Failed to append client certs", s.name)
+	}
+
+	s.mu.Lock()
```
shashankram, line 108: Could you describe via a comment why a lock is necessary here?

pkg/utils/grpc.go
```diff
@@ -27,44 +46,139 @@ func NewGrpc(serverType string, port int, certPem, keyPem, rootCertPem []byte) (
...
+	}
+	s.mu.Unlock()
+
+	return nil
+}
+
+func (s *Server) configureConfigUpdates(ctx context.Context) error {
+	// listen for certificate rotation first, so we don't miss any events
+	certRotationChan, unsubscribeRotation := s.cm.SubscribeRotations(s.certName)
+
```
shashankram, line 125: Add `defer unsubscribeRotation()` and drop other unsubscriptions below.

pkg/utils/grpc.go
```diff
@@ -27,44 +46,139 @@ func NewGrpc(serverType string, port int, certPem, keyPem, rootCertPem []byte) (
...
-		grpcServer.GracefulStop()
-		log.Info().Msgf("[grpc][%s] gRPC Server stopped", serverType)
-	}
-	log.Info().Msgf("[grpc][%s] exiting %s gRPC server", serverType, serverType)
+		if s.server != nil {
+			log.Info().Msgf("[grpc][%s] Gracefully stopping %s gRPC server", s.name, s.name)
+			s.server.GracefulStop()
+			log.Info().Msgf("[grpc][%s] gRPC Server stopped", s.name)
+		}
+		log.Info().Msgf("[grpc][%s] exiting %s gRPC server", s.name, s.name)
```
shashankram, line 181: No need for `[grpc][]` in log lines.

shashankram (changes requested):
Thanks for the change.
Have suggested a few changes inline to improve readability and update legacy code. Moreover, it seems that the `grpc.go` file is tightly coupled to ADS usage, so it no longer seems to be a utility file similar to other functions in `pkg/utils`, so it might make more sense to move the `gRPC` related code to `pkg/envoy/ads/grpc.go`.

pkg/utils/grpc.go
jaellio: I think for any gRPC server we create we would want to support updating the tls config on certificate rotation without restart. Supporting updates on rotation requires access to the `certificate.Manager`. However, I believe at this time a gRPC server is only created for ADS.

pkg/utils/grpc.go
jaellio: Does `unsubscribeRotation()` have to be called in the go routine? If it is called outside of the go routine then we'll unsubscribe from the pubsub while we're still listening for rotation events.

shashankram: In that case, would it make sense to only subscribe within the goroutine? Alternatively,  It might be more readable to let the caller invoke this as `go watchForConfigUpdates()` and drop the wrapping of the goroutine in this function, but it seems like you may be relying on some sort of error propagation down the call stack to terminate the process when this fails to happen. I think a log.Fatal() should suffice in such cases, or by passing a signal handler from main to such functions, vs structuring the code such that a lot of function calls in the call stack need to propagate the error all the way back to `main()`.

jaellio: We try to subscribe to certificate rotations prior to issuing the certificate to ensure that rotations are not missed between issuing and subscribing. However, this is an edge case

shashankram: Got it. Would be nice to update the comment in L124 to indicate that the issuance happens via the call to `initConfig()`

pkg/utils/grpc.go
```diff
@@ -27,44 +46,136 @@ func NewGrpc(serverType string, port int, certPem, keyPem, rootCertPem []byte) (
...
+	}
+
+	// Handle the rotations until the context is cancelled
+	go func() {
+		log.Info().Str("grpc", s.name).Str("cn", s.certName).Msg("Listening for certificate rotations")
+		for {
+			select {
+			case <-certRotationChan:
+				log.Debug().Str("grpc", s.name).Str("cn", s.certName).Msg("Certificate rotation was initiated for grpc server")
+				if err := s.initConfig(); err != nil {
```
jsturtevant, line 142: minor but maybe `initConfig` is the right term here since its not only happening during initialization?

pkg/utils/grpc.go
```diff
@@ -27,44 +46,136 @@ func NewGrpc(serverType string, port int, certPem, keyPem, rootCertPem []byte) (
...
+	}
+	s.mu.Unlock()
+
+	return nil
+}
+
+func (s *Server) watchCertRotations(ctx context.Context) error {
+	// listen for certificate rotation first, so we don't miss any events
+	certRotationChan, unsubscribeRotation := s.cm.SubscribeRotations(s.certName)
+	defer unsubscribeRotation()
```
jsturtevant, line 127: won't this unsubscribe when this func returns?

pkg/utils/grpc.go
```diff
@@ -27,44 +46,136 @@ func NewGrpc(serverType string, port int, certPem, keyPem, rootCertPem []byte) (
...
+
+	// Handle the rotations until the context is cancelled
+	go func() {
+		log.Info().Str("grpc", s.name).Str("cn", s.certName).Msg("Listening for certificate rotations")
+		for {
+			select {
+			case <-certRotationChan:
+				log.Debug().Str("grpc", s.name).Str("cn", s.certName).Msg("Certificate rotation was initiated for grpc server")
+				if err := s.initConfig(); err != nil {
+					events.GenericEventRecorder().ErrorEvent(err, events.CertificateIssuanceFailure, "Error rotating the certificate for grpc server")
```
jsturtevant, line 143: Nice to see the consistency here with the webserver rotations.  Don't need to do anything here, but keep an eye on if we can re-use this in someway.  Probably way too early to do this now.

pkg/utils/grpc.go
```diff
@@ -27,44 +46,136 @@ func NewGrpc(serverType string, port int, certPem, keyPem, rootCertPem []byte) (
 
 	log.Debug().Msgf("Parameters for %s gRPC server: MaxConcurrentStreams=%d;  KeepAlive=%+v", serverType, maxStreams, streamKeepAliveDuration)
 
+	s := &Server{
+		name:     serverType,
+		cm:       cm,
+		certName: certName,
```
jsturtevant, line 52:
```suggestion
		certCommonName: certCommonName,
```

pkg/utils/grpc.go
```diff
@@ -27,44 +46,136 @@ func NewGrpc(serverType string, port int, certPem, keyPem, rootCertPem []byte) (
 
 	log.Debug().Msgf("Parameters for %s gRPC server: MaxConcurrentStreams=%d;  KeepAlive=%+v", serverType, maxStreams, streamKeepAliveDuration)
 
+	s := &Server{
+		name:     serverType,
```
jsturtevant, line 50:
```suggestion
		name:     serverName,
```

pkg/utils/grpc.go
```diff
@@ -27,44 +46,136 @@ func NewGrpc(serverType string, port int, certPem, keyPem, rootCertPem []byte) (
...
 	}
 
-	mutualTLS, err := setupMutualTLS(false, serverType, certPem, keyPem, rootCertPem)
-	if err != nil {
-		log.Error().Err(err).Msg("Error setting up mutual tls for GRPC server")
-		return nil, nil, err
+	// #nosec G402: TLS MinVersion too low
+	tlsConfig := tls.Config{
+		InsecureSkipVerify: false,
+		ServerName:         serverType,
```
jsturtevant, line 65:
```suggestion
		ServerName:         s.name,
```

jsturtevant (commented):
a few minor things and question about the defer unsubscribing from the rotations.

pkg/utils/grpc.go
shashankram: What I meant to say is that `grpc.go` no longer belongs under `pkg/utils`, seems coupled to ADS requirements, so is better suited to be under `pkg/envoy/ads`.

pkg/utils/grpc.go
jaellio: Yes, that was accidentally included here. I highlighted this issue in [this conversation], but included it anyway :)(https://github.com/openservicemesh/osm/pull/4996#discussion_r948311899).

pkg/utils/grpc.go
jaellio: I previously had `initConfig()` as `setConfig()`, but I think `updateConfig()` might make the most sense here.

pkg/envoy/ads/server.go
```diff
@@ -32,6 +32,9 @@ const (
 
 	// workerPoolSize is the default number of workerpool workers (0 is GOMAXPROCS)
 	workerPoolSize = 0
+
+	// xdsServerCertificateCommonName is the common name of the certificate for the ADS server
+	xdsServerCertificateCommonName = "ads"
```
jaellio, line 37: Should the ads cert cn also contain the namespace of the control plane?

pkg/utils/grpc.go
jaellio: I can move it. If we need another grpc server for any reason we can move it back to `pkg/utils`. Like I stated in my above comment, I think `grpc.go` could live in either location.

jaellio: Is this blocking?

shashankram (approved):

pkg/utils/grpc.go
shashankram: I'd prefer to have this addressed with this change, given it's just a `git mv` command with little overhead. But feel free to take a call based on your preference.

pkg/envoy/ads/server.go
keithmattix: I would think so

pkg/utils/grpc.go
```diff
@@ -34,37 +59,124 @@ func NewGrpc(serverType string, port int, certPem, keyPem, rootCertPem []byte) (
...
+	tlsConfig := tls.Config{
+		InsecureSkipVerify: false,
+		ServerName:         s.name,
+		ClientAuth:         tls.RequireAndVerifyClientCert,
+		MinVersion:         constants.MinTLSVersion,
+		GetConfigForClient: func(*tls.ClientHelloInfo) (*tls.Config, error) {
+			// use lock to prevent concurrent updates and reads to the tls config
+			s.mu.Lock()
+			defer s.mu.Unlock()
+			return &s.config, nil
```
keithmattix, line 72: `s.config` is initially nil here; is that intentional?

pkg/utils/grpc.go
```diff
@@ -34,37 +59,124 @@ func NewGrpc(serverType string, port int, certPem, keyPem, rootCertPem []byte) (
...
+	certPool := x509.NewCertPool()
+
+	// Load the set of Root CAs
+	if ok := certPool.AppendCertsFromPEM(grpcCert.GetTrustedCAs()); !ok {
+		return fmt.Errorf("failed to append client certs")
+	}
+
+	// use lock to prevent concurrent updates and reads to the tls config
+	s.mu.Lock()
+	// #nosec G402: TLS MinVersion too low
```
keithmattix, line 110: You should theoretically be able to remove this since non-FIPS builds have a MinTLSVersion of 1.3

jaellio: I was getting a lint error without this. I'll try removing it again

pkg/utils/grpc.go
jaellio: It was intentional. I want to delay issuing the ads server certificate until after we subscribe to rotations of that certificate. Therefore, we're holding off on issuing the ads cert until the caller starts the server (calls `GrpcServe`). By holding off on the certificate creation, I think it also makes sense to hold off on creating the tls config.

pkg/utils/grpc.go
jaellio: Updated. Merged with existing ads/grpc.go

pkg/envoy/ads/server.go
jaellio: Will address this in a separate issue/PR

pkg/utils/grpc.go
keithmattix: Ah, I see. Just be cautious (perhaps comment) about accessing `s.config` around the caller so as to avoid nil pointer errros

keithmattix (approved):

Issue #4967, opened 2022-08-04
Error when performing upgrades on v1.0.0


steeling:
We have a bug in our upgrade path when upgrading from v1.0.0 forwards:

1. The v1.0.0 osm-bootstrap attempts to write a mesh config on startup @ version v1alpha1. It does this prior to starting it's servers, which include the crd conversion webhook.
2. In an upgrade, the new one comes up, which now writes the new version v1alpha2. Similarly, it does not start it's servers until this is successfully written.
3. This triggers a call from K8s to the conversion webhook. However if the old pod is down, the call fails, because the new pod has not yet started its server. ie: there are no servers to respond to the conversion webhook.
4. The old pod can attempt to come back up, but it also fails, due to trying to path the mesh config CRD, but the patch omits v1alpha2, which fails k8s validation.

Therefore no bootstrap pods can come up, which the remainder of the control plane depends upon.

steeling:
It looks like if `enableReconciler` is set to true it can lead to issues as well

steeling:
expanding on my last comment: the reconciler attempts to keep things at the current version. If 2 OSM versions are up for a brief period of time, one which has written the new CRD's/webhook/etc, the old version will try to rewrite them.

@jaellio this also breaks cert rotation for those webhooks

keithmattix:
Doesn't the reconciler also re-create the mutatingwebhookconfiguration if it's deleted?

nshankar13:
The issue seems to be due to the way that the CRD patching is coupled with the cert rotation handler code -> we pass in the `patchCrds` method in `crdconversion.go` as `onCertChange` - thus, the first time the certificate gets issued to the conversionwebhook, patchCrds is called before the server is started. 

One potential solution is to not immediately call `patchCrds` the first time the certificate is issued - the actual rotation handler logic will be the same, since that is configured to handle future cert changes. But the first time `setCert`  is called before the server starts, we don't call `patchCrds` - instead we call it directly in `crdconversion.go` once the server starts. 

This way, if the old osm-bootstrap pod is down, the conversionwebhook could respond to the server on the new bootstrap pod. If both osm-bootstrap pods are still up, k8s can use either the old or new osm-bootstrap service to update the apiversion on the meshconfig.

nshankar13:
[...]

Issue #4845, opened 2022-06-23
Endless envoy-bootstrap-config secrets


thorker:
**Bug description**:

Missing clean up of unreferenced envoy-bootstrap-config secrets, so we got 4k of it.
All secrets got no OwnerReference.

**Affected area (please mark with X where applicable)**:
- Install                    [ ]
- SMI Traffic Access Policy  [ ]
- SMI Traffic Specs Policy   [ ]
- SMI Traffic Split Policy   [ ]
- Permissive Traffic Policy  [ ]
- Ingress                    [ ]
- Egress                     [ ]
- Envoy Control Plane        [ ]
- CLI Tool                   [ ]
- Metrics                    [ ]
- Certificate Management     [ ]
- Sidecar Injection          [ x ]
- Logging                    [ ]
- Debugging                  [ ]
- Tests                      [ ]
- Demo                       [ ]
- CI System                  [ ]

**Expected behavior**:
No unreferenced secrets.
**Steps to reproduce the bug (as precisely as possible)**:

**How was OSM installed?**:
AKS AddOn

**Anything else we need to know?**:

**Bug report archive**:

**Environment**:
- OSM version (use `osm version`): v1.0
- Kubernetes version (use `kubectl version`): 1.23
- Size of cluster (number of worker nodes in the cluster):
- Others:

shashankram:
This seems like a regression as the [code](https://github.com/openservicemesh/osm/blob/main/pkg/k8s/announcement_handlers.go) should handle the cleanup of secrets via k8s garbage collection through OwnerReferences.

/cc @jaellio

jaellio:
@steeling has also been seeing this issue. I'll sync with him.

patst:
Would be interesting to have a trace log for the cases at https://github.com/openservicemesh/osm/blob/8fd236e8e104279b4d951a32720e06f4257fd80a/pkg/k8s/announcement_handlers.go#L71 when a conflict error is raised. 

Don't know if something else is updating Secret objects in parallel but that may interfere with the handler.

keithmattix:
We're looking to address this in our upcoming release. There should be an error log in the OSM controller when this occurs: https://github.com/openservicemesh/osm/blob/e6304c1/pkg/k8s/announcement_handlers.go#L70

Could anyone running into this error please confirm that you see this log line in the osm-controller?

trstringer:
[...]

PR #5005, opened 2022-08-16
feat(certs): update bootstrap secret on certificate manager update

### LINKED ISSUE #4818: Update OSM bootstrap secret on certificate manager updates

shalier:
**Description**:
Rotate bootstrap secret when corresponding cert is updated,
added unit tests, changed issuer struct -> IssuerMetadata struct
and resolves #4818

**Testing done**:
unit tests written for added functions

**Affected area**:

2. Is this a breaking change?


pkg/injector/bootstrap.go
```diff
@@ -100,3 +105,103 @@ func (wh *mutatingWebhook) marshalAndSaveBootstrap(name, namespace string, confi
...
+				log.Error().Err(err).Str(errcode.Kind, errcode.GetErrCodeWithMetric(errcode.ErrRotatingCert)).
+					Msgf("Error rotating cert SerialNumber=%s", cert.GetSerialNumber())
+			}
+			// if the secret and issued cert are the same no need to update the secret
+			if reflect.DeepEqual(c.GetIssuingCA, secret.Data["ca.crt"]) && reflect.DeepEqual(c.PrivateKey, secret.Data["tls.key"]) && reflect.DeepEqual(c.CertChain, secret.Data["tls.crt"]) {
+				continue
+			}
+			// TODO: check mapping
+			secretData := map[string][]byte{
+				"ca.crt":  cert.GetIssuingCA(),
```
jaellio, line 170:
```suggestion
				"ca.crt":  cert.GetTrustedCA(),
```
The issuing ca is the ca that was used to sign the issued certificate. In this case, we want to set the ca.crt to the trusted cas of the certificate. The trusted ca will contain the issuing ca and may additionally contain the validating ca during root certificate rotation.

During some stages of rotation, all certificates must have both cas available for validation. For example, there may be a set of certs issued by one ca and a set of certs issued by another. In order for those services (or in this case the ADS server and envoy) to communicate with mTLS both services must trust both cas.

pkg/certificate/types.go
```diff
@@ -103,16 +104,17 @@ type issuer struct {
...
 	// TODO(#4711): define serviceCertValidityDuration in the MRC
 	serviceCertValidityDuration func() time.Duration
 
-	mu            sync.Mutex // mu syncrhonizes acces to the below resources.
-	signingIssuer *issuer
-	// equal to signingIssuer if there is no additional public cert issuer.
-	validatingIssuer *issuer
+	mu sync.Mutex // mu syncrhonizes acces to the below resources.
+	// SigningIssuer represents the signing key IssuerMetadata
+	SigningIssuer *IssuerMetadata
```
jaellio, line 115: Could you explain why the signing and validating issuers need to be exported?

jaellio (commented):
Took a quick first look! Will review in more detail tomorrow

pkg/injector/bootstrap_test.go
```diff
@@ -318,3 +331,159 @@ var _ = Describe("Test functions creating Envoy bootstrap configuration", func()
...
+					},
+				},
+			},
+		},
+	}
+
+	for i, tc := range testCases {
+		t.Run(fmt.Sprintf("Running test case %d: %s", i, tc.name), func(t *testing.T) {
+			getCertValidityPeriod := func() time.Duration { return 5 * time.Second }
+			certManager, err := certificate.NewManager(context.Background(), &certificate.FakeMRCClient{}, getCertValidityPeriod, getCertValidityPeriod, 5*time.Second)
```
steeling, line 445: you can also just call certificate.FakeCertManager() and keep fakeMRCClient unexposed

pkg/certificate/types.go
```diff
@@ -82,7 +81,8 @@ type Certificate struct {
 	signingIssuerID    string
 	validatingIssuerID string
 
-	certType CertType
+	// CertType is the type of certificate. This is only used by OSM.
+	CertType CertType
```
steeling, line 85: no need to export certType

pkg/injector/bootstrap.go
```diff
@@ -100,3 +105,103 @@ func (wh *mutatingWebhook) marshalAndSaveBootstrap(name, namespace string, confi
...
+	for _, secretPtr := range secretPtrs {
+		if secretPtr == nil {
+			continue
+		}
+		secret, ok := secretPtr.(*corev1.Secret)
+		if !ok {
+			continue
+		}
+		// finds bootstrap secrets
+		if strings.Contains(secret.Name, bootstrapSecretPrefix) {
```
steeling, line 136: let's check the annotations here as well, there should be an annotation specific to OSM. 

app.kubernetes.io/instance=<meshName>
app.kubernetes.io/name=openservicemesh.io

pkg/injector/bootstrap.go
```diff
@@ -100,3 +105,103 @@ func (wh *mutatingWebhook) marshalAndSaveBootstrap(name, namespace string, confi
...
+
+	return secrets
+}
+
+// rotateBootstrapSecrets finds the bootstrap secret of the given certificate
+// from the list of secrets stored in the informerCollection's store and updates the secret.
+func (b *BootstrapSecretRotator) rotateBootstrapSecrets() {
+	secrets := b.listBootstrapSecrets()
+	for _, secret := range secrets {
+		secretProxyUUID := strings.ReplaceAll(secret.Name, bootstrapSecretPrefix, "")
```
steeling, line 149: nit: `strings.TrimPrefix`

pkg/injector/bootstrap.go
```diff
@@ -100,3 +105,103 @@ func (wh *mutatingWebhook) marshalAndSaveBootstrap(name, namespace string, confi
...
+		certs := b.certManager.ListIssuedCertificates()
+		for _, cert := range certs {
```
steeling, line 151: alternatively, we can query the proxyRegistry, by UUID, and reissue the certificate. Remember that the ListIssuedCertificates only has a cache, and may not actually have the certificate here

steeling: this is also currently an O(MXN) operation (roughly N^2 if everything is held in memory), which we can reduce to O(N) (assuming constant lookup on proxy registry)

pkg/injector/bootstrap.go
```diff
@@ -100,3 +105,103 @@ func (wh *mutatingWebhook) marshalAndSaveBootstrap(name, namespace string, confi
...
+			certProxyUUID := strings.Split(cert.CommonName.String(), ".")[0]
+			// checks if this is the corresponding cert for the secret
+			if secretProxyUUID != certProxyUUID {
+				continue
+			}
+			opts := []certificate.IssueOption{certificate.FullCNProvided()}
+			c, err := b.certManager.IssueCertificate(cert.CommonName.String(), cert.CertType, opts...)
+			if err != nil {
+				log.Error().Err(err).Str(errcode.Kind, errcode.GetErrCodeWithMetric(errcode.ErrRotatingCert)).
+					Msgf("Error rotating cert SerialNumber=%s", cert.GetSerialNumber())
```
steeling, line 162: nit: we recently added a `String` method to Certificate, so we can just do `Msgf("Error rotating cert: %s", cert)`

pkg/injector/bootstrap.go
```diff
@@ -100,3 +105,103 @@ func (wh *mutatingWebhook) marshalAndSaveBootstrap(name, namespace string, confi
...
+			}
+			// if the secret and issued cert are the same no need to update the secret
+			if reflect.DeepEqual(c.GetTrustedCAs, secret.Data["ca.crt"]) && reflect.DeepEqual(c.PrivateKey, secret.Data["tls.key"]) && reflect.DeepEqual(c.CertChain, secret.Data["tls.crt"]) {
```
steeling, line 165: It would be good to also check expiration times. This would allow us reduce the expiration on bootstrap certs from 10 years. Not required as of now however.

@jaellio curious on your thoughts

pkg/injector/bootstrap.go
```diff
@@ -100,3 +105,103 @@ func (wh *mutatingWebhook) marshalAndSaveBootstrap(name, namespace string, confi
...
+				continue
+			}
+			// TODO: check mapping
```
steeling, line 168: what's this TODO? is it for your reference or intended to live past this PR?

pkg/injector/bootstrap.go
```diff
@@ -100,3 +105,103 @@ func (wh *mutatingWebhook) marshalAndSaveBootstrap(name, namespace string, confi
 	log.Debug().Msgf("Creating bootstrap config for Envoy: name=%s, namespace=%s", name, namespace)
 	return wh.kubeClient.CoreV1().Secrets(namespace).Create(context.Background(), secret, metav1.CreateOptions{})
 }
+
+// NewBootstrapSecretRotator returns a new bootstrap secret rotator.
+func NewBootstrapSecretRotator(ctx context.Context, kubeClient kubernetes.Interface, informerCollection *informers.InformerCollection, certManager *certificate.Manager, checkInterval time.Duration) *BootstrapSecretRotator {
+	return &BootstrapSecretRotator{
+		context:            ctx,
```
steeling, line 112: from go: https://go.dev/blog/context-and-structs, you should avoid setting contexts on structs. Instead, pass it into the function that requires it

pkg/injector/bootstrap.go
```diff
@@ -100,3 +105,103 @@ func (wh *mutatingWebhook) marshalAndSaveBootstrap(name, namespace string, confi
...
+			}
+			newSecret := &corev1.Secret{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      secret.Name,
+					Namespace: secret.Namespace,
+				},
+				Type: corev1.SecretTypeTLS,
+				Data: secretData,
+			}
+			_, err = b.kubeClient.CoreV1().Secrets(secret.GetNamespace()).Update(b.context, newSecret, metav1.UpdateOptions{})
```
steeling, line 182: Can you add an UpdateSecret method to the k8s.client?

Can you also add a ListSecret method?

WIth those 2, we can keep a single k8s.Client, and avoid having both a kubeClient, and an informer

pkg/certificate/fake_manager.go
```diff
@@ -34,7 +36,8 @@ func (c *fakeMRCClient) List() ([]*v1alpha2.MeshRootCertificate, error) {
 	}}, nil
 }
 
-func (c *fakeMRCClient) Watch(ctx context.Context) (<-chan MRCEvent, error) {
+// Watch returns a channel that has one MRCEventAdded. It is intended to implement the certificate.MRCClient interface.
```
jaellio, line 39: Thanks for adding these comments here! Super helpful!

jaellio:
@shalier thanks for this change! Could you add some context to this PR describing the design choices made in order to ensure all bootstrap certificates are updated due to expiration or root certificate rotation? For example, could you explain why it isn't sufficient to subscribe to bootrstrap cert rotations? And why a ticker is being used?

pkg/injector/bootstrap.go
jaellio: If we use `ListIssuedCertificate()` and the cert cache is empty or doesn't contain the bootstrap cert then it will never be rotated even though it is expiring. 

I think this is one potential solution:
1. Iterate over the bootstrap secrets
2. Check if the bootstrap cert is going to expire. We could call `IssueCertificate` here first but there is a high probability that the cert is not in the local cache and therefore a new cert would be returned. In this case, we'd be updating the secret a lot more frequently. Perhaps there is some other check that we could perform to prevent this and simply call `IssueCertificate` first.
3. If the cert is going to expire, call `IssueCertificate()`
4. Update the secret with the new cert
5. If the update fails, log the error and continue

jaellio: > alternatively, we can query the proxyRegistry, by UUID, and reissue the certificate. Remember that the ListIssuedCertificates only has a cache, and may not actually have the certificate here

This check could then be moved to the osm-controller from the osm-injector

pkg/injector/bootstrap.go
jaellio: I believe this label filtering is already done in `informer.go`.

pkg/certificate/fake_manager.go
```diff
@@ -18,14 +18,16 @@ var (
 	validity = time.Hour
 )
 
-type fakeMRCClient struct{}
+// FakeMRCClient implements the MRCClient interface
+type FakeMRCClient struct{}
```
jsturtevant, line 22: does this need be exported?

pkg/injector/bootstrap.go
jsturtevant: >We could call IssueCertificate here first but there is a high probability that the cert is not in the local cache

Why is it a high probability?

pkg/injector/bootstrap.go
```diff
@@ -100,3 +105,103 @@ func (wh *mutatingWebhook) marshalAndSaveBootstrap(name, namespace string, confi
...
+					Msgf("Error updating bootstrap secret %s/%s with issued cert %s", secret.Namespace, secret.Name, cert.CommonName.String())
+				continue
+			}
+		}
+	}
+}
+
+// StartBootstrapSecretRotationTicker will start a ticker to check if the bootstrap secrets should be
+// updated every BootstrapSecretRotator check interval
+func (b *BootstrapSecretRotator) StartBootstrapSecretRotationTicker() {
```
jsturtevant, line 194: can pass the context from the calling function here.

pkg/injector/bootstrap.go
jsturtevant: > Check if the bootstrap cert is going to expire. 

This makes sense to me.  If I am following correctly, this currently is only rotating when a cert is in the local cache. If its not in the local cache we would potentially miss it.

How are we responding to Root Certificate rotation?  Right now, If the root rotates in but the cert isn't expiring then we wouldn't update it.

Seems like there are two times we want to rotate the bootstrap:  

- if we are close to expiration
- if new root rotates in

jaellio: > > We could call IssueCertificate here first but there is a high probability that the cert is not in the local cache
> 
> Why is it a high probability?

Good point, we can't assume this is a high probability. I guess what we can say is it is a possibility so we can't rely on the cache alone.

jaellio: > > Check if the bootstrap cert is going to expire.
> 
> This makes sense to me. If I am following correctly, this currently is only rotating when a cert is in the local cache. If its not in the local cache we would potentially miss it.
> 
> How are we responding to Root Certificate rotation? Right now, If the root rotates in but the cert isn't expiring then we wouldn't update it.
> 
> Seems like there are two times we want to rotate the bootstrap:
> 
> * if we are close to expiration
> * if new root rotates in

Calling `IssueCertificate()` will check if the cert (if present in the cache) needs to be rotated due to root certificate rotation by comparing the issuing and validating IDs to the issuers in the `certificate.Manager`. You're correct that if we pre-check for expiration we also need to pre-check for mismatched IDs.

jaellio: If the likelihood of injector churn is low enough, then I think calling `IssueCertificate()` without a pre check would be acceptable.

jsturtevant: I think you are right that there needs to be some type of check. I think `listBootstrapSecrets()` iterates over every secret in the cluster? If there are multiple instances of the injector, we would be potential calling this for every secret, sometime multiple times across the instances.

jaellio: Yeah, that's correct. I am not sure how we could divide secret/certificate updates across the injectors so there is no overlap.

jaellio: Maybe this process would make more sense in the osm-bootstrap since there are likely to be less replicas. @steeling what are your thoughts on this?

pkg/injector/bootstrap.go
jaellio: Are you suggesting making the bootstrap certificate expiration configurable?

pkg/injector/bootstrap.go
jaellio: I guess this problem might be good justification for using the ProxyRegistry. We wouldn't see the same overlap using the proxyRegistry and moving this process to the osm-controller If we construct the certificate names from the proxies in the registry we would still need to know the namespace to look up the Kubernetes secret.

Additionally, if we use the proxyRegistry to lookup bootstrap secrets/certificates and a proxy is failing to connect to the control plane for whatever reason its bootstrap certificate may never be updated it isn't in the proxyRegistry and therefore may never be able to connect to the control plane.

shalier: if we implement using the ProxyRegistry then I don't think the code should be in the pkg/injector folder, but not sure where would be a better place, any ideas? @jaellio @jsturtevant @steeling

steeling: @jaellio it's quite possible we're going to remove the bootstrap pod with https://github.com/openservicemesh/osm/issues/4967, so let's hhold off on that.

pkg/injector/bootstrap.go
```diff
@@ -100,3 +106,76 @@ func (wh *mutatingWebhook) marshalAndSaveBootstrap(name, namespace string, confi
...
+		checkInterval:  checkInterval,
+	}
+}
+
+// rotateBootstrapSecrets updates the bootstrap secret from the connectedProxy by
+// getting the current or issuing a new certificate.
+func (b *BootstrapSecretRotator) rotateBootstrapSecrets(ctx context.Context) {
+	proxies := b.proxyRegistry.ListConnectedProxies()
+	for _, proxy := range proxies {
+		envoyBootstrapConfigName := bootstrapSecretPrefix + proxy.UUID.String()
```
jaellio, line 125:
```suggestion
		envoyBootstrapConfigSecretName := bootstrapSecretPrefix + proxy.UUID.String()
```

pkg/injector/bootstrap.go
jaellio: @steeling If we continue iterating over the secrets directly (without the proxyRegistry) how could we limit the high churn of multiple injectors trying to update all bootstrap secrets at the same time? Unless they all check the same certs in exactly the same order and at the same time over and over again I think they will eventually be updated. I just don't want to run into a similar issue that we are seeing with #4845

steeling: an update wouldn't change the owner reference nor create a secret that was previously deleted, so won't result in https://github.com/openservicemesh/osm/issues/4845.

I'd say that secrets should be rotated infrequently enough that attempting to solve this issue now would be a premature optimization

pkg/injector/bootstrap.go
steeling: we can make it configurable later if we choose, but yes if we're adding rotation functionality we may as well check if the certs are expiring

pkg/injector/bootstrap_test.go
```diff
@@ -312,3 +325,201 @@ var _ = Describe("Test functions creating Envoy bootstrap configuration", func()
...
+	proxyUUID := uuid.New()
+	secretName := bootstrapSecretPrefix + proxyUUID.String()
+
+	testCases := []struct {
+		name     string
+		certName string
+		secrets  []*corev1.Secret
+	}{
+		{
+			name:     "update bootstrap secret with new cert",
```
jsturtevant, line 404: should we have a test case to verify it doesn't rotate?

jsturtevant: and where there are more than one secret, etc?

pkg/k8s/client.go
```diff
@@ -28,12 +29,13 @@ import (
 )
 
 // NewClient returns a new kubernetes.Controller which means to provide access to locally-cached k8s resources
-func NewClient(osmNamespace, meshConfigName string, informerCollection *osminformers.InformerCollection, policyClient policyv1alpha1Client.Interface, msgBroker *messaging.Broker, selectInformers ...InformerKey) *Client {
+func NewClient(osmNamespace, meshConfigName string, informerCollection *osminformers.InformerCollection, kubeClient kubernetes.Interface, policyClient policyv1alpha1Client.Interface, msgBroker *messaging.Broker, selectInformers ...InformerKey) *Client {
```
jsturtevant, line 32: Is this the right object/place to be referencing the Kubeclient in? Still getting use to the various interfaces/classes used to interact with k8s.

jsturtevant (commented):
like the improvements based on the feedback.  minor comments on add some more tests.

pkg/k8s/client.go
jsturtevant: based on https://github.com/openservicemesh/osm/pull/5018 it looks like k8s client is for CRUD, so this would be the place to add ability to update secrets.

jsturtevant: I was surprised since we didn't have it in this pkg/k8s/client but looks like it was primarily informers in this package and pass the k8s client around when needed but in process of simplifying it 👍

pkg/injector/bootstrap_test.go
shalier: @jsturtevant not sure how we'd go about testing if it doesn't rotate, since that's determined by ShouldRotate(), any ideas?

pkg/injector/bootstrap.go
jaellio: @steeling is the goal to no longer have a kubeClient and rely entirely on the informers?

pkg/injector/bootstrap.go
```diff
@@ -100,3 +104,116 @@ func (wh *mutatingWebhook) marshalAndSaveBootstrap(name, namespace string, confi
...
+				Msgf("Error updating bootstrap secret %s/%s with issued cert %s", secret.Namespace, secret.Name, issuedCert)
+			continue
+		}
+	}
+}
+
+func getCert(secret *corev1.Secret) (*certificate.Certificate, error) {
+	pemCert, ok := secret.Data[bootstrap.EnvoyXDSCertFile]
+	if !ok {
+		// TODO(#3962): metric might not be scraped before process restart resulting from this error
```
jaellio, line 172: nit: I don't think these TODOs are applicable here

pkg/injector/bootstrap.go
```diff
@@ -100,3 +104,116 @@ func (wh *mutatingWebhook) marshalAndSaveBootstrap(name, namespace string, confi
...
+		log.Error().Err(err).Str(errcode.Kind, errcode.GetErrCodeWithMetric(errcode.ErrDecodingPEMCert)).
+			Msg("Error converting PEM cert to x509 to obtain serial number")
+		return nil, err
+	}
+
+	return &certificate.Certificate{
+		CommonName:   certificate.CommonName(x509Cert.Subject.CommonName),
+		SerialNumber: certificate.SerialNumber(x509Cert.SerialNumber.String()),
+		CertChain:    pemCert,
+		TrustedCAs:   secret.Data[bootstrap.EnvoyXDSCACertFile],
```
jaellio, line 198: nit: We are not verifying that the secret has the ca cert file field as we do for the key file and pem cert fields above. Not sure if this check is necessary.

pkg/injector/bootstrap.go
```diff
@@ -100,3 +104,116 @@ func (wh *mutatingWebhook) marshalAndSaveBootstrap(name, namespace string, confi
...
+		err = b.kubeController.UpdateSecret(ctx, secret, secret.Data)
+		if err != nil {
+			if apierrors.IsConflict(err) {
+				log.Error().Err(err).Str(errcode.Kind, errcode.GetErrCodeWithMetric(errcode.ErrUpdatingBootstrapSecret)).
+					Msgf("There was an update conflict while trying to update the envoy bootstrap config secret %s/%s with issued cert %s", secret.Namespace, secret.Name, issuedCert)
+				continue
+			}
+			log.Error().Err(err).Str(errcode.Kind, errcode.GetErrCodeWithMetric(errcode.ErrUpdatingBootstrapSecret)).
+				Msgf("Error updating bootstrap secret %s/%s with issued cert %s", secret.Namespace, secret.Name, issuedCert)
+			continue
```
jaellio, line 164: nit: I don't think `continue` is necessary here.

pkg/k8s/client.go
```diff
@@ -134,6 +136,32 @@ func (c *Client) GetService(svc service.MeshService) *corev1.Service {
...
+		}
+
+		secrets = append(secrets, secret)
+	}
+
+	return secrets
+}
+
+// UpdateSecret updates the secret with the provided data
+func (c *Client) UpdateSecret(ctx context.Context, secret *corev1.Secret, secretData map[string][]byte) error {
```
jaellio, line 159: nit: Does it make sense for this to be `UpdateSecretData`?

pkg/injector/bootstrap_test.go
```diff
@@ -312,3 +325,233 @@ var _ = Describe("Test functions creating Envoy bootstrap configuration", func()
...
+			},
+		},
+	}
+
+	for i, tc := range testCases {
+		t.Run(fmt.Sprintf("Running test case %d: %s", i, tc.name), func(t *testing.T) {
+			certManager, err := certificate.FakeCertManager()
+			assert.Nil(err)
+
+			mockController := k8s.NewMockController(gomock.NewController(t))
```
jaellio, line 457: Is it possible to create a fake `NewSimpleClientSet{[]runtime.Object{...}...)` for the kubeClient instead of mocking the controller? This would allow us to verify the secrets were updated as expected.

pkg/k8s/client_test.go
```diff
@@ -69,7 +69,8 @@ func TestIsMonitoredNamespace(t *testing.T) {
 
 			ic, err := informers.NewInformerCollection(testMeshName, nil, informers.WithKubeClient(testclient.NewSimpleClientset()))
 			a.Nil(err)
-			c := NewClient("osm", tests.OsmMeshConfigName, ic, nil, nil)
+
+			c := NewClient("osm", tests.OsmMeshConfigName, ic, nil, nil, nil)
```
jaellio, line 73: Were tests added to `client_test.go` for `ListSecrets()` and `UpdateSecrets()`?

pkg/injector/bootstrap_test.go
keithmattix: +1 a SimpleClientset is the more robust approach

pkg/injector/bootstrap.go
```diff
@@ -100,3 +104,116 @@ func (wh *mutatingWebhook) marshalAndSaveBootstrap(name, namespace string, confi
...
+	return &BootstrapSecretRotator{
+		kubeController: kubeController,
+		certManager:    certManager,
+		checkInterval:  checkInterval,
+	}
+}
+
+// listBootstrapSecrets returns the bootstrap secrets stored in the informerCollection's store.
+// this function is used to get the namespace of the secrets.
+func (b *BootstrapSecretRotator) listBootstrapSecrets() []*corev1.Secret {
```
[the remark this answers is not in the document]
shalier: Both ways described end up being O(N) complexity since we drop constants in big-O notation. I think instead of calling it `listBootstrapSecrets` I'm leaning towards `getBootstrapSecrets`.

pkg/injector/bootstrap_test.go
shalier: That would create a cert that's not about to expire, but `shouldRotate `also checks if the `signingIssuer `or `validatingIssuer` are different [here ](https://github.com/openservicemesh/osm/blob/main/pkg/certificate/manager.go#L200) and I don't think there's a way to set those variables unless I exported them

shalier: @jsturtevant that does create a cert that's not going to expire, but `shouldRotate` also [checks if the `signingIssuer` or `validatingIssuer` are different](https://github.com/openservicemesh/osm/blob/main/pkg/certificate/manager.go#L200), and I don't think I can set those variables unless I export them 🤔

shalier: I ended up setting the `signingIssuer` and `validatingIssuer` IDs to empty strings in `pkg/certificate/fake_manager.go`

pkg/certificate/certificate.go
```diff
@@ -91,3 +96,47 @@ func NewFromPEM(pemCert pem.Certificate, pemKey pem.PrivateKey) (*Certificate, e
 		Expiration:   x509Cert.NotAfter,
 	}, nil
 }
+
+// CreateValidCertAndKey creates a non-expiring PEM certificate and private key
+func CreateValidCertAndKey(notBefore, notAfter time.Time) (pem.Certificate, pem.PrivateKey, error) {
```
jsturtevant, line 101: Since this is a test helper function, can we move it to the test file where it is used?

pkg/certificate/certificate_test.go
```diff
@@ -16,40 +12,11 @@ import (
 
 func TestNewFromPEM(t *testing.T) {
 	assert := tassert.New(t)
-	cn := CommonName("Test CA")
-	rootCertCountry := "US"
-	rootCertLocality := "CA"
-	rootCertOrganization := "Root Cert Organization"
 
+	cn := CommonName("Test CA")
```
jsturtevant, line 16: would pass this into the `CreateValidCertAndKey`  to make it explicit that is the CN for that cert

pkg/certificate/certificate.go
jaellio, line 101: Is this method only used by tests? Maybe it would make sense in a test helper file or something like that?

pkg/injector/bootstrap.go
```diff
@@ -101,3 +105,118 @@ func (wh *mutatingWebhook) marshalAndSaveBootstrap(name, namespace string, confi
...
+		}
+
+		secret.Data[bootstrap.EnvoyXDSCACertFile] = issuedCert.GetTrustedCAs()
+		secret.Data[bootstrap.EnvoyXDSCertFile] = issuedCert.GetCertificateChain()
+		secret.Data[bootstrap.EnvoyXDSKeyFile] = issuedCert.GetPrivateKey()
+
+		err = b.kubeController.UpdateSecretData(ctx, secret, secret.Data)
+		if err != nil {
+			if apierrors.IsConflict(err) {
+				log.Error().Err(err).Str(errcode.Kind, errcode.GetErrCodeWithMetric(errcode.ErrUpdatingBootstrapSecret)).
```
jsturtevant, line 158: is this more of a warning?  Is there enough wiggle room in expiration in `ShouldRotate()` to absorb this error? I.e.  If we have a conflict and then postpone updating for another round will we get to rotation before we cause a disruption?

 It looks like `30s` for ShouldRotate (https://github.com/openservicemesh/osm/blob/194856e2fbe909cc9d2830951aed97e9e7ddb033/pkg/certificate/certificate.go#L13)  and we run every 5 seconds.  The thing to be wary of would be large cluster or other delays that might cause this to tick up to high.

There is a helper function in k8s libraries that will do retry on conflict that could eliminate any potential for this being a  sneaky bug long term

jsturtevant (changes requested):
we should move the test helper function to a test file but other than that looking good

pkg/injector/bootstrap.go
jsturtevant: https://github.com/kubernetes/client-go/blob/v0.25.0/util/retry/util.go#L103

jaellio: Docs for `RetryOnConflict`: https://pkg.go.dev/k8s.io/client-go/util/retry#RetryOnConflict

keithmattix (approved):
Approving pending other feedback. Good work!

pkg/injector/bootstrap.go
```diff
@@ -101,3 +105,118 @@ func (wh *mutatingWebhook) marshalAndSaveBootstrap(name, namespace string, confi
...
+	}
+
+	x509Cert, err := certificate.DecodePEMCertificate(pemCert)
+	if err != nil {
+		log.Error().Err(err).Str(errcode.Kind, errcode.GetErrCodeWithMetric(errcode.ErrDecodingPEMCert)).
+			Msg("Error converting PEM cert to x509 to obtain serial number")
+		return nil, err
+	}
+
+	return &certificate.Certificate{
```
jaellio, line 197: Currently, the `siginingIssuerID` and the `validatingIssuerID` will always be empty strings. When we call `ShouldRotate()` the certificate's `siginingIssuerID` and `validatingIssuerID`  will never match the Managers `signingIssuer.ID` and `validatingIssuer.ID` respectively since they will be set to the name of the appropriate MRC resource. Therefore, `ShouldRotate()` will always return true and the certificates will constantly be rotated. 

To resolve this, we could add additional data fields to the osm-bootrstrap secret for the `siginingIssuerID` and `validatingIssuerID`. `GetCertFromSecret` could access those values to create a `Certificate`. Since the `siginingIssuerID` and `validatingIssuerID` are not exported, we could add a `Certificate` builder method like `NewCertificate()` in pkg/certificate.

jaellio (changes requested):

pkg/injector/bootstrap.go
```diff
@@ -95,9 +104,122 @@ func (wh *mutatingWebhook) marshalAndSaveBootstrap(name, namespace string, confi
...
+		if !b.certManager.ShouldRotate(bootstrapCert) {
+			continue
+		}
+
+		issuedCert, err := b.certManager.IssueCertificate(bootstrapCert.CommonName.String(), certificate.Internal, certificate.FullCNProvided())
+		if err != nil {
+			log.Error().Err(err).Str(errcode.Kind, errcode.GetErrCodeWithMetric(errcode.ErrRotatingCert)).Msgf("Error rotating cert %s", issuedCert)
+		}
+
+		err = retry.RetryOnConflict(retry.DefaultRetry, func() error {
```
jaellio, line 158: I think `RetryOnConflict` needs to include the "get" operation. Without a call to get the updated version of the secret, the conflict error will occur on every retry. 

Here is a good example of `RetryOnConflict` from the [docs](https://pkg.go.dev/k8s.io/client-go/util/retry#RetryOnConflict):

```go
err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
    // Fetch the resource here; you need to refetch it on every try, since
    // if you got a conflict on the last update attempt then you need to get
    // the current version before making your own changes.
    pod, err := c.Pods("mynamespace").Get(name, metav1.GetOptions{})
    if err != nil {
        return err
    }

    // Make whatever updates to the resource are needed
    pod.Status.Phase = v1.PodFailed

    // Try to update
    _, err = c.Pods("mynamespace").UpdateStatus(pod)
    // You have to return err itself here (not wrapped inside another error)
    // so that RetryOnConflict can identify it correctly.
    return err
})
if err != nil {
    // May be conflict if max retries were hit, or may be something unrelated
    // like permissions or a network error
    return err
}
```

To implement a similar pattern, we will need to add a `GetSecret` method in `k8s/client.go`. After getting the secret and the certificate we will also need to check if the certificate still needs to be updated since it might have already been updated by another injector.

pkg/injector/bootstrap.go
```diff
@@ -95,9 +104,127 @@ func (wh *mutatingWebhook) marshalAndSaveBootstrap(name, namespace string, confi
...
+		}
+	}
+	return bootstrapSecrets
+}
+
+// rotateBootstrapSecrets updates the bootstrap secret by getting the current or issuing a new certificate.
+func (b *BootstrapSecretRotator) rotateBootstrapSecrets(ctx context.Context) {
+	bootstrapSecrets := b.getBootstrapSecrets()
+	for _, secret := range bootstrapSecrets {
+		bootstrapCert, err := getCertFromSecret(secret)
```
jaellio, line 143: I think we might need to get the cert from the secret each time we retry. If the secret was updated by another injector, we need the cert from the updated secret to pass to `ShouldRotate`.

pkg/injector/bootstrap.go
```diff
@@ -95,9 +104,127 @@ func (wh *mutatingWebhook) marshalAndSaveBootstrap(name, namespace string, confi
...
+			corev1Secret, err := b.kubeController.GetSecret(ctx, secret.Namespace, secret.Name)
+			if err != nil {
+				return err
+			}
+
+			if !b.certManager.ShouldRotate(bootstrapCert) {
+				return nil
+			}
+
+			issuedCert, err := b.certManager.IssueCertificate(certificate.ForCommonName(bootstrapCert.CommonName.String()))
```
jaellio, line 159: We shouldn't update the secret if issuing the certificate fails.

pkg/injector/bootstrap.go
```diff
@@ -140,25 +140,28 @@ func (b *BootstrapSecretRotator) getBootstrapSecrets() []*corev1.Secret {
...
-
-		err = retry.RetryOnConflict(retry.DefaultRetry, func() error {
+		cert := &certificate.Certificate{}
+		err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
 			corev1Secret, err := b.kubeController.GetSecret(ctx, secret.Namespace, secret.Name)
 			if err != nil {
 				return err
 			}
 
+			bootstrapCert, err := getCertFromSecret(corev1Secret)
```
jaellio, line 150: I think this could be set to `cert` directly.

pkg/injector/bootstrap.go
```diff
@@ -140,25 +140,28 @@ func (b *BootstrapSecretRotator) getBootstrapSecrets() []*corev1.Secret {
...
+			cert = bootstrapCert
+			if err != nil {
+				log.Error().Err(err).Str(errcode.Kind, errcode.GetErrCodeWithMetric(errcode.ErrObtainingCertFromSecret)).
+					Msgf("Error getting cert %s from bootstrap secret %s/%s", bootstrapCert, corev1Secret.Namespace, corev1Secret.Name)
```
jaellio, line 154: Let's only log the CN or serial number of the certificate. In `bootstrapCert` could be nil here, so I think it is sufficient to log "Error getting cert from bootstrap secret %s/%s".

pkg/injector/bootstrap.go
```diff
@@ -140,25 +140,28 @@ func (b *BootstrapSecretRotator) getBootstrapSecrets() []*corev1.Secret {
...
+				return err
+			}
 			if !b.certManager.ShouldRotate(bootstrapCert) {
 				return nil
 			}
 
 			issuedCert, err := b.certManager.IssueCertificate(certificate.ForCommonName(bootstrapCert.CommonName.String()))
 			if err != nil {
-				log.Error().Err(err).Str(errcode.Kind, errcode.GetErrCodeWithMetric(errcode.ErrRotatingCert)).Msgf("Error rotating cert %s", issuedCert)
+				log.Error().Err(err).Str(errcode.Kind, errcode.GetErrCodeWithMetric(errcode.ErrRotatingCert)).Msgf("Error rotating cert %s for bootstrap secret %s/%s", issuedCert, corev1Secret.Namespace, corev1Secret.Name)
```
jaellio, line 163: Same here. We shouldn't log the cert

pkg/injector/bootstrap.go
```diff
@@ -95,9 +104,129 @@ func (wh *mutatingWebhook) marshalAndSaveBootstrap(name, namespace string, confi
...
+			corev1Secret.Data[bootstrap.EnvoyXDSKeyFile] = issuedCert.GetPrivateKey()
+			corev1Secret.Data[signingIssuerID] = []byte(issuedCert.GetSigningIssuerID())
+			corev1Secret.Data[validatingIssuerID] = []byte(issuedCert.GetValidatingIssuerID())
+
+			err = b.kubeController.UpdateSecretData(ctx, corev1Secret, corev1Secret.Data)
+			return err
+		})
+		if err != nil {
+			log.Error().Err(err).Str(errcode.Kind, errcode.GetErrCodeWithMetric(errcode.ErrUpdatingBootstrapSecret)).
+				Msgf("Error updating bootstrap secret %s/%s with cert %s", secret.Namespace, secret.Name, cert)
```
jaellio, line 177: Still logging the cert here, but otherwise looks good!!

pkg/k8s/client.go
```diff
@@ -155,6 +158,41 @@ func (c *Client) GetService(name, namespace string) *corev1.Service {
...
+		}
+
+		secrets = append(secrets, secret)
+	}
+
+	return secrets
+}
+
+// GetSecret returns the secret for a given namespace and secret name, returns nil if the API errored
+func (c *Client) GetSecret(ctx context.Context, namespace, name string) (*corev1.Secret, error) {
```
jaellio, line 181: Should we get from the informer cache here?

jaellio (approved):

pkg/certificate/castorage/k8s/k8s.go
```diff
@@ -43,7 +43,7 @@ func GetCertFromKubernetes(ns string, secretName string, kubeClient kubernetes.I
 		return nil, certificate.ErrInvalidCertSecret
 	}
 
-	cert, err := certificate.NewFromPEM(pemCert, pemKey)
+	cert, err := certificate.NewRootCertificateFromPEM(pemCert, pemKey)
```
keithmattix, line 46: I'm slightly concerned about the ways that we're using the term "root certificate". By definition, a root certificate is the last node in the trust chain. This isn't at all blocking, but I wonder if we could remove some ambiguity by being stricter with our terms

pkg/injector/bootstrap.go
```diff
@@ -3,21 +3,30 @@ package injector
...
 	"github.com/openservicemesh/osm/pkg/errcode"
+	"github.com/openservicemesh/osm/pkg/k8s"
 	"github.com/openservicemesh/osm/pkg/models"
 	"github.com/openservicemesh/osm/pkg/utils"
 	"github.com/openservicemesh/osm/pkg/version"
 )
 
+const (
+	signingIssuerID    = "signing_issuer_id"
+	validatingIssuerID = "validating_issuer_id"
```
keithmattix, line 27: What's the purpose of these constants? Are they keys for the secret? If so, that should be clear based on the var name

keithmattix (approved):
Looks good; two non-blocking comments

pkg/k8s/client.go
```diff
@@ -155,6 +158,42 @@ func (c *Client) GetService(name, namespace string) *corev1.Service {
...
+	secretIf, exists, err := c.informers.GetByKey(osminformers.InformerKeySecret, key(name, namespace))
+	if exists && err == nil {
+		secret := secretIf.(*corev1.Secret)
+		return secret
+	}
+	return nil
+}
+
+// UpdateSecretData updates the secret with the provided data
+func (c *Client) UpdateSecretData(ctx context.Context, secret *corev1.Secret, secretData map[string][]byte) error {
```
steeling, line 191: can this just take in the secret object to update? ie: require the caller to do secret.Data = secretData.

Because if they've udpated anything else, then we're updating that as well, so that's a misnomer

pkg/injector/bootstrap.go
```diff
@@ -95,9 +104,125 @@ func (wh *mutatingWebhook) marshalAndSaveBootstrap(name, namespace string, confi
...
 	}
 
 	log.Debug().Msgf("Creating bootstrap config for Envoy: name=%s, namespace=%s", name, namespace)
 	return wh.kubeClient.CoreV1().Secrets(namespace).Create(context.Background(), secret, metav1.CreateOptions{})
 }
+
+// NewBootstrapSecretRotator returns a new bootstrap secret rotator.
+func NewBootstrapSecretRotator(kubeController k8s.Controller, certManager *certificate.Manager, checkInterval time.Duration) *BootstrapSecretRotator {
+	return &BootstrapSecretRotator{
+		kubeController: kubeController,
```
steeling, line 119: We've recently put a lot of effort into hiding our k8s abstractions. Instead of giving this a kubeController, which propagates our leak of the k8s abstraction, can you give this a compute.Interface.

From there, let's add a new type 

```
type Secret struct {
 data map[string][]byte
}
```

Then we'll want to add methods to the compute.Interface (and kube.Client) to translate between the abstraction and the k8s client. So:

```
func (c *client) UpdateSecret(ctx context.Context, secret models.Secret) error {
   _, err := c.kubeController.UpdateSecret(ctx, secret)
  return err
}
```

steeling (changes requested):

pkg/injector/bootstrap.go
```diff
@@ -95,9 +104,123 @@ func (wh *mutatingWebhook) marshalAndSaveBootstrap(name, namespace string, confi
...
+	}
+	return bootstrapSecrets
+}
+
+// rotateBootstrapSecrets updates the bootstrap secret by getting the current or issuing a new certificate.
+func (b *BootstrapSecretRotator) rotateBootstrapSecrets(ctx context.Context) {
+	bootstrapSecrets := b.getBootstrapSecrets()
+	for _, secret := range bootstrapSecrets {
+		err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
+			cert, err := getCertFromSecret(secret)
```
jaellio, line 144: I think we had added a `GetSecret()` method and call here. Why was this removed? Thanks!

shalier: `GetSecret()` was essentially used to convert the bootstrap secret to a corev1.Secret, but we don't need to explicitly use the corev1.Secret, the models.Secret works here since we only need the secret data to get the cert

jaellio: I thought we added `GetSecret()` to ensure we had the latest version of the secret. Currently, if there is a conflict on a secret update, retry will always fail because the same outdated secret will be used to try to make the update.

shalier: Oops you are right, I forgot that's how RetryOnConflict worked, will add it back

pkg/constants/constants.go
```diff
@@ -97,6 +97,9 @@ const (
 	// OSMCertificateValidityPeriod is the TTL of the certificates used in the OSM control plane or for Envoy to xDS communication.
 	OSMCertificateValidityPeriod = 87600 * time.Hour // a decade
 
+	// CertCheckInterval is the period of ticks for a Ticker
+	CertCheckInterval = 5 * time.Second
```
steeling, line 101: 5 seconds feels a little aggressive for this check, thoughts? I guess it is just checking in the informer cache, but still we can probably get away with reducing this.

Particularly if we change the remaining time till expiration in shouldRotate https://github.com/openservicemesh/osm/issues/5150

steeling (approved):

pkg/constants/constants.go
keithmattix: s/reducing/increasing?

PR #5006, opened 2022-08-16
Reduce the configurator into the k8s.Client object

steeling:
Part of #4863 

This separates out some of the methods into a new interface called the PassthroughInterface. The idea is that these methods would not change, even in a world where OSM was running outside of K8s, which is due to the fact that we do not have a mapping of the k8s object to an internal representation of that object.

Note: this PR has a lot of line changes, but is mostly getting the new mocks to work with the removal of the MeshConfig object.

Now that it is tied to the kubeController, which had a static implementation when used with the NewFakeMeshCatalog, I also had to remove/reduce the usage of NewFakeMeshCatalog for instances that wanted to mock the underlying kubeController


pkg/catalog/outbound_traffic_policies.go
```diff
@@ -144,13 +146,14 @@ func (mc *MeshCatalog) GetOutboundMeshTrafficPolicy(downstreamIdentity identity.
...
 func (mc *MeshCatalog) ListOutboundServicesForIdentity(serviceIdentity identity.ServiceIdentity) []service.MeshService {
-	if mc.configurator.GetMeshConfig().Spec.Traffic.EnablePermissiveTrafficPolicyMode {
+	if mc.GetMeshConfig().Spec.Traffic.EnablePermissiveTrafficPolicyMode {
 		return mc.ListServices()
 	}
 
 	svcAccount := serviceIdentity.ToK8sServiceAccount()
 	serviceSet := mapset.NewSet()
 	var allowedServices []service.MeshService
+	fmt.Println("mc.meshSpec.ListTrafficTargets() ::: ", mc.meshSpec.ListTrafficTargets())
```
allenlsy, line 156: Is this a debugging print?

pkg/compute/kube/client.go
```diff
@@ -118,7 +117,7 @@ func (c *client) GetServicesForServiceIdentity(svcIdentity identity.ServiceIdent
 	svcSet := mapset.NewSet() // mapset is used to avoid duplicate elements in the output list
 
 	svcAccount := svcIdentity.ToK8sServiceAccount()
-
+	fmt.Println("c.kubeController.ListPods() ", svcIdentity, c.kubeController.ListPods())
```
allenlsy, line 120: ditto

pkg/envoy/lds/response_test.go
```diff
@@ -11,68 +10,66 @@ import (
...
-func getProxy(kubeClient kubernetes.Interface) (*envoy.Proxy, *v1.Pod, error) {
-	podLabels := map[string]string{
-		constants.AppLabel:               tests.BookbuyerService.Name,
-		constants.EnvoyUniqueIDLabelName: tests.ProxyUUID,
-	}
+func TestNewResponse(t *testing.T) {
+	assert := tassert.New(t)
+	mockCtrl := gomock.NewController(t)
+	policyCtrl := policy.NewMockController(mockCtrl)
+	// mockK8s := k8s.NewMockController(mockCtrl)
```
allenlsy, line 38: Remove this line?

pkg/compute/kube/client.go
```diff
@@ -127,9 +126,10 @@ func (c *client) GetServicesForServiceIdentity(svcIdentity identity.ServiceIdent
 		if pod.Spec.ServiceAccountName != svcAccount.Name {
 			continue
 		}
-
+		fmt.Println("\n\npod.Spec.ServiceAccountName ", pod.Spec.ServiceAccountName)
```
shashankram, line 129: can be removed

pkg/k8s/types.go
```diff
@@ -102,14 +109,20 @@ type Controller interface {
...
 
-	GetTargetPortForServicePort(types.NamespacedName, uint16) (uint16, error)
-
 	ServiceToMeshServices(svc corev1.Service) []service.MeshService
 }
+
+// PassthroughInterface is the interface for methods that are implemented by the k8s.Client, but are not considered
+// k8s "specific", and thus do not need further abstraction, and can be used throughout the code base without fear of
+// coupling to k8s.
+type PassthroughInterface interface {
```
allenlsy, line 121: Consider renaming it to OSMInterface? Since these features are closely tied to OSM.

pkg/compute/kube/client.go
```diff
@@ -127,9 +126,10 @@ func (c *client) GetServicesForServiceIdentity(svcIdentity identity.ServiceIdent
...
 		podLabels := pod.ObjectMeta.Labels
 		meshServicesForPod := c.getServicesByLabels(podLabels, pod.Namespace)
+		fmt.Println("which had the following services: ", meshServicesForPod)
```
shashankram, line 132: can be removed

pkg/k8s/client.go
```diff
@@ -434,3 +441,21 @@ func (c *Client) GetTargetPortForServicePort(namespacedSvc types.NamespacedName,
...
 	return len(svc.Spec.ClusterIP) == 0 || svc.Spec.ClusterIP == corev1.ClusterIPNone
 }
+
+// GetMeshConfig returns the MeshConfig resource corresponding to the control plane
+func (c *Client) GetMeshConfig() configv1alpha2.MeshConfig {
+	var meshConfig configv1alpha2.MeshConfig
+	configs := c.informers.List(osminformers.InformerKeyMeshConfig)
+	if len(configs) != 1 {
+		log.Error().Msgf("Expected exactly one MeshConfig resource, got %d", len(configs))
+	}
```
shashankram, line 451: Can we keep this consistent with the previous behavior where we simply `GetByKey(mesh_config_name)` so we don't log errors on MeshConfigs not relevant to the controller if they exist on the cluster for whatever reason?

allenlsy (approved):
Overall looks good

pkg/k8s/types.go
```diff
@@ -102,14 +109,20 @@ type Controller interface {
...
+	GetMeshConfig() configv1alpha2.MeshConfig
+	GetOSMNamespace() string
+	UpdateIngressBackendStatus(obj *policyv1alpha1.IngressBackend) (*policyv1alpha1.IngressBackend, error)
+	UpdateUpstreamTrafficSettingStatus(obj *policyv1alpha1.UpstreamTrafficSetting) (*policyv1alpha1.UpstreamTrafficSetting, error)
+
+	GetTargetPortForServicePort(types.NamespacedName, uint16) (uint16, error)
```
shashankram, line 127: These 3 seem k8s specific to me, in which case they should be a part of the k8s.Controller interface because they are tightly coupled to k8s resources.

What was your reasoning to not add them within the k8s client?

pkg/k8s/client.go
steeling: the informer already has an object name field selector so it was redundant

```
istOption := configInformers.WithTweakListOptions(func(opt *metav1.ListOptions) {
	opt.FieldSelector = fields.OneTermEqualSelector(metav1.ObjectNameField, meshConfigName).String()
})
```

pkg/k8s/types.go
steeling: The Passthrough interface is part of the k8s Client as well (it's the set of Methods that we allow to "leak" out of k8s)

I'd be happy to abstract them, but not sure the best way to do that at the moment. Open to suggestions though

pkg/compute/kube/client.go
steeling: done

pkg/compute/kube/client.go
steeling: done

pkg/envoy/lds/response_test.go
steeling: done

pkg/compute/kube/client.go
steeling: done

pkg/catalog/outbound_traffic_policies.go
steeling: done

pkg/k8s/types.go
steeling: the `compute.Interface` is essentially that (which contains the PassthroughInterface), The passthrough is a subset of what would be the OSMInterface, in that it is what we allow to leak from k8s.

Not sure what the best answer is so I'll leave this comment unresolved to see if others have opinions on naming here :)

allenlsy: Ok, I think I initially misunderstood compute.Interface which I thought it should contain methods that are cluster provider related and should have nothing to do with OSM terminologies.

pkg/k8s/client.go
shashankram: In that case, just a GetByKey would suffice vs a List?

pkg/k8s/types.go
shashankram: Given these are k8s specific, the comment `but are not considered k8s "specific"` in the description of the PassthroughInterface seems a bit ambiguous. I am trying to understand how do you decide what goes into this interface. A comment describing these criteria in the interface definition would help.

shashankram (approved):
Conditionally approving this based on clarification requested in https://github.com/openservicemesh/osm/pull/5006#discussion_r948528123

pkg/k8s/client.go
steeling: sure will do

pkg/k8s/types.go
steeling: added some comments, let me know if that helps

keithmattix (approved):

PR #5011, opened 2022-08-17
cli: extend proxy cmd to set proxy config

shashankram:
**Description**:
Extends the `osm proxy` cmd by making it
able to execute `set`(POST) requests as well.

With this change, similar to `osm proxy get`,
`osm proxy set` allows updating the proxy
configuration via the Envoy admin interface.
This is extremely useful for troubleshooting,
when we may want to reset stats counters and
logging levels.

**Testing done**:
```
$ osm proxy get stats "$fortio_server" -n demo | grep over_limit
cluster.demo/fortio|8080|local.ratelimit.over_limit: 0

$ osm proxy set reset_counters fortio-c4bd7857f-8kxq5 -n demo
OK
```

**Affected area**:

### DIFF
--- a/cmd/cli/proxy.go
+++ b/cmd/cli/proxy.go
@@ -1,16 +1,36 @@
 package main
 
 import (
+	"fmt"
 	"io"
+	"os"
 
 	"github.com/spf13/cobra"
 	"helm.sh/helm/v3/pkg/action"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+
+	"github.com/openservicemesh/osm/pkg/cli"
+	"github.com/openservicemesh/osm/pkg/constants"
 )
 
 const proxyCmdDescription = `
 This command consists of subcommands related to the operations
 of the sidecar proxy on pods.
 `
+
+type proxyAdminCmd struct {
+	out        io.Writer
+	config     *rest.Config
+	clientSet  kubernetes.Interface
+	query      string
+	namespace  string
+	pod        string
+	localPort  uint16
+	outFile    string
+	sigintChan chan os.Signal
+}
 
 func newProxyCmd(config *action.Configuration, out io.Writer) *cobra.Command {
 	cmd := &cobra.Command{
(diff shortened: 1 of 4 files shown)


keithmattix (commented):
This feels potentially dangerous; by using port forwarding directly to the envoy admin port, there's no certificate auth going on. Therefore, anyone with pod access can modify the envoy config. Granted, portforward RBAC is a standard best practice, but I did want to call it out

steeling:
> This feels potentially dangerous; by using port forwarding directly to the envoy admin port, there's no certificate auth going on. Therefore, anyone with pod access can modify the envoy config. Granted, portforward RBAC is a standard best practice, but I did want to call it out

That's a good call out, although I'd vote that this is simply convenience.. if they had access before this command doesn't change anything, just wraps existing commands.

shashankram:
> This feels potentially dangerous; by using port forwarding directly to the envoy admin port, there's no certificate auth going on. Therefore, anyone with pod access can modify the envoy config. Granted, portforward RBAC is a standard best practice, but I did want to call it out

This change doesn't introduce a security hole, given it relies on the user's kubeconfig file to access the cluster, which the user would anyway be able to do without the CLI. It simply wraps a bunch of calls into 1 command, something that's easily scriptable.

Note that the Envoy admin interface binds to localhost, so only clients who already have access to the pod can use the admin interface.

cmd/cli/proxy.go
```diff
@@ -1,17 +1,37 @@
...
 of the sidecar proxy on pods.
 `
 
+type proxyAdminCmd struct {
+	out        io.Writer
+	config     *rest.Config
+	clientSet  kubernetes.Interface
+	query      string
+	namespace  string
+	pod        string
```
steeling, line 29: I'm hoping to push more stuff behind interfaces and unify behind a single client (k8s.Client for k8s, and kube.Client will abstract that)

I wonder if we can map things like pod, to proxy, and have a generic ProxyPortForwarder which abstracts the k8s one.

The rest of the client setup isn't quite there yet (a few more PR's, and need to get the k8s client to be able to issue updates, and have options for queries outside the informer cache), but if it's feasible to add some abstractions in this PR I'd push for it. 

If not, no worries.

steeling (approved):

keithmattix:
> > This feels potentially dangerous; by using port forwarding directly to the envoy admin port, there's no certificate auth going on. Therefore, anyone with pod access can modify the envoy config. Granted, portforward RBAC is a standard best practice, but I did want to call it out
> 
> 
> 
> This change doesn't introduce a security hole, given it relies on the user's kubeconfig file to access the cluster, which the user would anyway be able to do without the CLI. It simply wraps a bunch of calls into 1 command, something that's easily scriptable.
> 
> 
> 
> Note that the Envoy admin interface binds to localhost, so only clients who already have access to the pod can use the admin interface.

Yep, that makes sense, especially the note about localhost binding. I feel good having that note captured here in the PR

keithmattix (approved):

cmd/cli/proxy.go
shashankram: I am not sure I understand what that would achieve. The CLI code is heavily decoupled from internal control plane infra clients, and relies on kubeclient directly. More so, I don't believe we should refactor existing CLI behavior in this change, the purpose of which is to extend the existing `osm proxy` command.

If there isn't an issue for your proposal open already, could you open one and describe how you'd like to address the existing CLI code that uses the kubeclient?

Issue #5000, opened 2022-08-15
Certificate rotation broken in long running environments


shashankram:
**Bug description**:
I noticed that in 2 of my setups running for several days, certificate rotation is broken. This leads to expired certificates in Envoy never being rotated, leading to complete traffic disruption between apps.

Log on the client:
```
{"bytes_received":0,"response_flags":"UF","upstream_service_time":null,"response_code":503,"start_time":"2022-08-15T18:02:56.864Z","authority":"fortio.demo.svc.cluster.local:8080","duration":1,"bytes_sent":195,"protocol":"HTTP/1.1","x_forwarded_for":null,"path":"/","request_id":"26aa1f49-a09b-4b13-81bd-0acfa49852b8","user_agent":"fortio.org/fortio-1.34.1","response_code_details":"upstream_reset_before_response_started{connection_failure,TLS_error:_268435581:SSL_routines:OPENSSL_internal:CERTIFICATE_VERIFY_FAILED}","time_to_first_byte":null,"requested_server_name":null,"upstream_host":"10.244.1.5:8080","method":"GET","upstream_cluster":"demo/fortio|8080"}
```

Stat on the client:
`cluster.demo/fortio|8080.ssl.fail_verify_error: 50`

The `certs` stat `osm_bug_report_2928885602/namespaces/demo/pods/fortio-client-b9b7bbfb8-hc9wr/commands/osm_proxy_get_certs_fortio-client-b9b7bbfb8-hc9wr_-n_demo` confirms the cert not being updated, with its client cert having an expiration date of `2022-08-12T19:05:52Z`. The expiration date should be past `2022-08-15` (current date).

Both the client and server are connected to the controller as per the XDS cluster stats collected in the bug-report.

osm-controller indicates the cert `default.demo.svc.cluster.local` has expired but has not been rotated:
```
	 Common Name: "default.demo.cluster.local"
	 Valid Until: 2022-08-12 19:05:52.0391624 +0000 UTC m=+86504.429431801 (16h45m27.4651171s remaining)
	 Issuing CA (SHA256): d4bc5ec6f2ab02a7f484f5c36ee90222435250592d0942c686737ba0a77e857e
	 Trusted CAs (SHA256): d4bc5ec6f2ab02a7f484f5c36ee90222435250592d0942c686737ba0a77e857e
	 Cert Chain (SHA256): c4b68f17387579bff64ee155ae7263687c2d3fbea929ad56b0489b6ebd074dd8
[...]

PR #5012, opened 2022-08-17
cert rotation now ignores monotonic clock readings when checking expiration

steeling:
Presumed fix for https://github.com/openservicemesh/osm/issues/5000. More details in the comment https://github.com/openservicemesh/osm/issues/5000#issuecomment-1218539412

(even if this is not the issue, this would still be the correct way to check the x509 expiration)


keithmattix (approved):

shashankram (approved):

sushilsaha1111:
Any idea when this is planned to be released for implementation. We are also facing the same issue as highlighted in https://github.com/openservicemesh/osm/issues/5000

sushilsaha1111:
1.0.0 and our AKS version is 1.24

PR #5013, opened 2022-08-18
increase test coverage to look for coverage outside of current package

steeling:
Test coverage, by default, only reports the coverage within the package of the unit test. However many of our tests should be considered integration tests, with tests crossing package boundaries. By applying the coverpkg flag each test will now check coverage against all packages in the `pkg` directory. 

This will likely slow down our unit tests, but our tests suite is largely dominated by e2e tests so this should be negligible

### DIFF
--- a/scripts/test-w-coverage.sh
+++ b/scripts/test-w-coverage.sh
@@ -16,6 +16,7 @@ go test -timeout 120s \
    -failfast \
    -race \
    -v \
+   -coverpkg=./pkg/... \
    -coverprofile=coverage.txt.with_generated_code \
    -covermode atomic $modules | tee testoutput.txt || { echo "go test returned non-zero"; exit 1; }
 
--- a/scripts/test-w-coverage.sh
+++ b/scripts/test-w-coverage.sh
@@ -16,7 +16,7 @@ go test -timeout 120s \
    -failfast \
    -race \
    -v \
-   -coverpkg=./pkg/... \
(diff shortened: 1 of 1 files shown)


keithmattix (approved):

steeling:
i had missed test coverage in ./cmd/... initially, repushed with that change, so let's see what codecov bot reports now

keithmattix:
I'm confused by this [report](https://app.codecov.io/gh/openservicemesh/osm/compare/5013/overview); the unittest delta says -7.25%, but when you dig into the actual coverage changes, there's a net increase....

steeling:
> I'm confused by this [report](https://app.codecov.io/gh/openservicemesh/osm/compare/5013/overview); the unittest delta says -7.25%, but when you dig into the actual coverage changes, there's a net increase.... 

If I had to guess the positive delta's are on files that are still in report.. negative was due to dropping coverage in ./cmd/...

anybody know how to get the report to re-run? @shashankram  @nojnhuh

steeling:
oh wait no that's even worse when including cli... ya idk what that's about

steeling:
> oh wait no that's even worse when including cli... ya idk what that's about

I think it's that each package is now printing coverage of the entire code base, so pkg X coverage in pkg X is 80%, but of the whole codebase is maybe 10%.

Instead of properly tracking all the lines that are covered it's doing this? I'm just going to avoid the headache of reasoning with this and closing it.

Issue #4933, opened 2022-07-20
Collapse all k8s funcs/methods onto k8s.Client


steeling:
From #4863 collapse all of the k8s input/output methods/funcs onto the k8s.Client object.

This includes any function that accepts a controller/client object as a parameter, as well as all of

pkg/smi
pkg/configurator
pkg/policy
pkg/k8s
pkg/config

allenlsy:
`pkg/smi` cannot be directly collapsed into `k8s.Client` as there will be [circular import](https://github.com/openservicemesh/osm/blob/52bd5a7d8d190fc2e02a13f252250cd585a66619/pkg/smi/types.go#L24).  Therefore we will create a new `Client` interface in `pkg/mesh`. This MeshClient will do that job that communicate with k8s resources.

steeling:
Hey @allenlsy, there may be some misunderstanding here on how to accomplish this.

The methods on the SMI client should be moved to the k8s client, and the SMI client should go away. Any necessary fields can be added to the k8s client. Take a look at this PR which did the same for the configurator https://github.com/openservicemesh/osm/pull/5006

allenlsy:
PR for pkg/policy: https://github.com/openservicemesh/osm/pull/5014

PR #5014, opened 2022-08-18
ref: collapse pkg/policy into k8s Client

### LINKED ISSUE #4863: Push all k8s clients into a layer of clients

allenlsy:

**Description**:

Partially resolve [#4933](https://github.com/openservicemesh/osm/issues/4933)

Move functions from policy client into k8s client. The only exception is `DetectIngressBackendConflicts()` in the policy package, which is used in by CLI when checking IngressBackend conflicts.

**Testing done**:

Unit test

**Affected area**:

No

2. Is this a breaking change?

No

No


pkg/catalog/catalog.go
```diff
@@ -5,15 +5,15 @@ import (
...
 	"github.com/openservicemesh/osm/pkg/messaging"
-	"github.com/openservicemesh/osm/pkg/policy"
 	"github.com/openservicemesh/osm/pkg/smi"
 	"github.com/openservicemesh/osm/pkg/ticker"
 )
 
 // NewMeshCatalog creates a new service catalog
 func NewMeshCatalog(meshSpec smi.MeshSpec, certManager *certificate.Manager,
-	policyController policy.Controller, stop <-chan struct{},
+	policyController k8s.PassthroughInterface, stop <-chan struct{},
```
steeling, line 16: this is already present on the compute.Interface, so no need to store it twice.

pkg/catalog/ingress_test.go
```diff
@@ -390,7 +390,7 @@ func TestGetIngressTrafficPolicy(t *testing.T) {
 			defer mockCtrl.Finish()
 
 			mockProvider := compute.NewMockInterface(mockCtrl)
-			mockPolicyController := policy.NewMockController(mockCtrl)
+			mockPolicyController := k8s.NewMockController(mockCtrl)
```
steeling, line 393: let's place the mocked methods on the mockProvider

pkg/k8s/client.go
```diff
@@ -58,6 +58,18 @@ func NewClient(osmNamespace, meshConfigName string, informerCollection *osminfor
 		informerInitHandlerMap[informer]()
 	}
 
+	shouldObserve := func(obj interface{}) bool {
```
steeling, line 61: let's remove these calls here, and add the the event handlers to the informer's `With*` methods

allenlsy:
ping @openservicemesh/osm-maintainers

pkg/k8s/client.go
allenlsy: Can you clarify the idea?

pkg/catalog/egress.go
```diff
@@ -136,8 +136,8 @@ func (mc *MeshCatalog) getUpstreamTrafficSettingForEgress(egressPolicy *policyv1
 				Namespace: egressPolicy.Namespace,
 				Name:      match.Name,
 			}
-			upstreamtrafficSetting := mc.policyController.GetUpstreamTrafficSetting(
-				policy.UpstreamTrafficSettingGetOpt{NamespacedName: &namespacedName})
+			upstreamtrafficSetting := mc.Interface.GetUpstreamTrafficSetting(
+				k8s.UpstreamTrafficSettingGetOpt{NamespacedName: &namespacedName})
```
jsturtevant, line 140: It's surprising to see the k8s pkg imported into this package.  My understanding was cataglog shouldn't know about k8s?  

Is this option parameter in the wrong place?

pkg/catalog/egress.go
```diff
@@ -35,7 +35,7 @@ func (mc *MeshCatalog) GetEgressTrafficPolicy(serviceIdentity identity.ServiceId
 	var trafficMatches []*trafficpolicy.TrafficMatch
 	var clusterConfigs []*trafficpolicy.EgressClusterConfig
 	portToRouteConfigMap := make(map[int][]*trafficpolicy.EgressHTTPRouteConfig)
-	egressResources := mc.policyController.ListEgressPoliciesForSourceIdentity(serviceIdentity.ToK8sServiceAccount())
+	egressResources := mc.Interface.ListEgressPoliciesForSourceIdentity(serviceIdentity.ToK8sServiceAccount())
```
jsturtevant, line 38: may not be appropriate in this PR but it would be clear if it was `mc.ComputeInterface`

pkg/catalog/types.go
```diff
@@ -25,10 +24,6 @@ type MeshCatalog struct {
 	compute.Interface
```
keithmattix, line 24: It's not a part of this PR (and it seems like this PR largely avoids this pitfall by explicitly calling out `MeshCatalog.Interface`), but embedding this interface seems to imply a tighter coupling between MeshCatalog and a `compute.Interface` than desired IMO. Interface embedding is already discouraged/prohibited in Uber's [style guide](https://github.com/uber-go/guide/blob/master/style.md#avoid-embedding-types-in-public-structs), and it's just confusing here. In my mind, a `MeshCatalog` has a dependency on a `compute.Interface` it is not itself a `compute.Interface` like embedding may imply. /cc @steeling for some possible context

pkg/k8s/client.go
keithmattix: I actually don't think these event handlers belong in the With* methods; this code seems to be a correct usage of AddEventHandler

pkg/k8s/types.go
```diff
@@ -70,6 +70,11 @@ const (
 	MeshRootCertificate InformerKey = "MeshRootCertificate"
 )
 
+const (
+	// kindSvcAccount is the ServiceAccount kind
+	kindSvcAccount = "ServiceAccount"
```
keithmattix, line 75: Are there multiple serviceAccount kinds?

pkg/validator/server.go
```diff
@@ -32,7 +32,7 @@ type validatingWebhookServer struct {
 }
 
 // NewValidatingWebhook returns a validatingWebhookServer with the defaultValidators that were previously registered.
-func NewValidatingWebhook(ctx context.Context, webhookConfigName, osmNamespace, osmVersion, meshName string, enableReconciler, validateTrafficTarget bool, certManager *certificate.Manager, kubeClient kubernetes.Interface, policyClient policy.Controller) error {
+func NewValidatingWebhook(ctx context.Context, webhookConfigName, osmNamespace, osmVersion, meshName string, enableReconciler, validateTrafficTarget bool, certManager *certificate.Manager, kubeClient kubernetes.Interface, policyClient k8s.PassthroughInterface) error {
```
keithmattix, line 35: Just for posterity: this is a valid use of the passthrough interface because the validatingwebhook concept is Kubernetes specific.

keithmattix (commented):
LGTM!

keithmattix (approved):
Lol, I meant to approve earlier

pkg/catalog/egress.go
allenlsy: Great point. I agree that `UpstreamTrafficSettingGetOpt` should remain in the policy package

pkg/k8s/types.go
allenlsy: No. I actually don't get the question.

pkg/catalog/egress.go
allenlsy: Agree but it's better to be addressed in a separate PR.

pkg/k8s/types.go
keithmattix: Usually, with a const grouping like you have here:
```go
const (
...
) 
```
each of the consts declared represent various options of a common type/concept. Looking again, I'm guessing this const is just here to deduplicate the string "ServiceAccount"?

pkg/k8s/client.go
steeling: sorry i was mistaken ,its' actually in the k8s.client init* methods. take a look at PR https://github.com/openservicemesh/osm/pull/5006 or other instances in the k8/client.go file.

pkg/catalog/egress.go
steeling: I might suggest the following:

1. place the entire `GetUpstreamTrafficSetting` method in compute/client.
2. Add 2 new methods to k8s.Client: `GetUpstreamTrafficSetting` (takes a name and namespace, no options), and `ListUpstreamTrafficSetting`
3. Break up the `GetUpstreamTrafficSetting` into 3 methods (all in compute.Client), one for each option. The options pattern isn't very well suited here, since combining/permuting options is meaningless.

The final result of all of these changes will be:

1. k8s/client.go: Is merged with the informer package, and contains simple CRUD operations for K8s objects
2. compute.Client, abstracts over the k8s.Client, and provides some logic for retrieving OSMs domain objects
3. MeshCatalog contains higher level methods for combining multiple objects to generate OSM domain objects (may eventually get merged with compute.Client

cmd/osm-controller/osm-controller.go
```diff
@@ -254,7 +250,7 @@ func main() {
 		events.GenericEventRecorder().FatalEvent(err, events.InitializationError, "Error initializing ADS server")
 	}
 
-	if err := validator.NewValidatingWebhook(ctx, validatorWebhookConfigName, osmNamespace, osmVersion, meshName, enableReconciler, validateTrafficTarget, certManager, kubeClient, policyController); err != nil {
+	if err := validator.NewValidatingWebhook(ctx, validatorWebhookConfigName, osmNamespace, osmVersion, meshName, enableReconciler, validateTrafficTarget, certManager, kubeClient, k8sClient); err != nil {
```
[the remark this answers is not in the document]
allenlsy: I think https://github.com/openservicemesh/osm/issues/4863 fixes the issue you mentioned.

steeling:
> ping @openservicemesh/osm-maintainers

recommend taking it out of draft if its ready for review

allenlsy:
> > ping @openservicemesh/osm-maintainers
> 
> recommend taking it out of draft if its ready for review

Will mark it as ready once the tests are passed.

pkg/catalog/egress.go
steeling: the issue is the stutter of `compute.ComputeInterface`. 

Whatever we decide, this is an embedded interface, so we can simply call `mc.ListEgressPoliciesForSourceIdentity`. This is useful, because a caller outside of the meshCatalog package doesn't really need to know which one has the responsbility

pkg/catalog/types.go
steeling: Personally, I disagree with Uber's conclusion on embedding interface. Although it might be something to take a harder look at, I think it has its places.

For us here, I think there's a good chance the meshcatalog and compute.Client eventually get merged. Embedding the interface allows abstracting the methods in compute.Interface, such that the response.go files don't need to know which layer of the domain they are accessing.

In fact, I recommend we remove `meshCatalog.Interface.<MethodCall>`, and directly call the method from meshCatalog. As we deconstruct some of the meshcatalog calls it will make for less changes.

* This gets heavy into opinions, so only providing mine as a reference point :)

pkg/catalog/egress.go
allenlsy: Update:

1. `UpstreamTrafficSettingGetOpt` is removed.
2. `GetUpstreamTrafficSetting()` is split into 3 methods, one for each option. 

> compute.Client, abstracts over the k8s.Client, and provides some logic for retrieving OSMs domain objects

Do you mean compute.Client abstracts over kube.Client? Currently k8s.Client has no direct connection with compute.Client.

pkg/validator/server.go
jaellio: Could you clarify the significance of the passthrough interface? I thought the interface was for methods that were not specific to kubernetes.

Also, should `policyClient` be updated to a more generic name since the interface has methods for more than just the policy resources?

steeling: Ya I agree, it should probably either be the raw k8s.Client, or a meshCatalog, or compute.Interface

keithmattix: Maybe I'm misunderstanding the passthrough interface then; I thought it was for deliberate abstraction leaking in specific cases. Regardless, I do agree the variable should be renamed

pkg/catalog/egress.go
steeling: the compute.Client has an embedded k8s.KubeController (which is a k8s.Client object)

pkg/catalog/types.go
steeling: https://github.com/openservicemesh/osm/issues/4863 has the full description

steeling (approved):

pkg/k8s/client.go
```diff
@@ -463,3 +484,150 @@ func (c *Client) GetMeshConfig() configv1alpha2.MeshConfig {
 func (c *Client) GetOSMNamespace() string {
 	return c.osmNamespace
 }
+
+// ListEgressPoliciesForSourceIdentity lists the Egress policies for the given source identity based on service accounts
+func (c *Client) ListEgressPoliciesForSourceIdentity(source identity.K8sServiceAccount) []*policyv1alpha1.Egress {
```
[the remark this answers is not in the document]
allenlsy: Egress policy, IngressBackend etc. are OSM/SMI terminologies. I think they are not part of compute.Interface as they are not Kubernetes native.

I will update the method names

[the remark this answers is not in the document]
allenlsy: For the 3 `GetUpstreamTrafficSettingBy<Type>()` methods, I don't find a better way to name them. They are split from the original `GetUpstreamTrafficSetting()` method. Do you have any suggestion?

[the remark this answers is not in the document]
steeling: They are being queried from Kubernetes though. The interface is to the control plane of where they are housed. 

So it would looks like this:

in compute/kube/client.go, are all the methods that are defined here.

In k8s/client.go, you have: `ListEgressPolicies`, which just calls list on the informer.

Right now there's logic in k8s, which makes k8s aware of OSM primitives and mappings. The k8s package will have simple CRUD based methods (get, delete, update, create, list), and nothing more. Happy to sync if you'd like

pkg/validator/server.go
keithmattix: Oh, that wasn't my intent. I was trying to convey that I think this is a good, intended use of the passthrough interface because it's fine to let abstractions leak into validatingWebhook logic because validatingwebhook logic is explicitly a Kubernetes concept

pkg/k8s/client.go
```diff
@@ -468,3 +489,99 @@ func (c *Client) GetMeshConfig() configv1alpha2.MeshConfig {
...
+			continue
+		}
+		policies = append(policies, egressPolicy)
+	}
+
+	return policies
+}
+
+// GetIngressBackendPolicy returns the IngressBackend policy for the given backend MeshService
+func (c *Client) GetIngressBackendPolicy(namespace, svcName string, targetPort int) *policyv1alpha1.IngressBackend {
```
steeling, line 509: I would just remove this get.. technically we don't really have a get, so the get can be moved up into the compute/client, ie:


```
GetIngressBackendPolicyByService -> calls k8s.ListIngressBackendPolicy, and filters appropriately.
```

The rule of thumb is that Gets should be by just name/namespace. Any other fields are a filter on a list

pkg/k8s/client.go
```diff
@@ -468,3 +489,99 @@ func (c *Client) GetMeshConfig() configv1alpha2.MeshConfig {
...
+		}
+
+		settings = append(settings, setting)
+	}
+
+	return settings
+}
+
+// GetUpstreamTrafficSettingByNamespace returns the UpstreamTrafficSetting resources with matching namespaced name
+func (c *Client) GetUpstreamTrafficSettingByKey(key string) *policyv1alpha1.UpstreamTrafficSetting {
```
steeling, line 581: super nit: instead of key, let's pass in name, namespace and construct the key in here

steeling (approved):
2 quick comments but awesome work!

pkg/k8s/client.go
allenlsy: I just noticed that the caller of this method uses a `types.NamespacedName` obj, which is k8s native object. So it should be fine to pass the object to this method

PR #5017, opened 2022-08-23
Complete ADS snapshot cache implementation

steeling:
This completes the ADS snapshot cache implementation, but does not flip the flag to enable it. This is accomplished by:

1. Switch the AddJob method to return a writable channel, so that it can be closed. This removes a channel resource leak.
2. Move the WatchUpdate logic into the OnStreamOpen, so that there is one routine running per connected proxy. This maintains current status quo, and makes it simpler to implement with the current message broker setup.
3. Add in the missing metrics tracking

Part of #2683

### DIFF
--- a/pkg/envoy/ads/cache_callbacks.go
+++ b/pkg/envoy/ads/cache_callbacks.go
@@ -7,6 +7,7 @@ import (
 	discovery "github.com/envoyproxy/go-control-plane/envoy/service/discovery/v3"
 
 	"github.com/openservicemesh/osm/pkg/envoy"
+	"github.com/openservicemesh/osm/pkg/messaging"
 	"github.com/openservicemesh/osm/pkg/metricsstore"
 	"github.com/openservicemesh/osm/pkg/utils"
 )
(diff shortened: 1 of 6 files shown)


shashankram (approved):

pkg/envoy/ads/cache_stream.go
```diff
@@ -8,46 +8,8 @@ import (
...
-		case <-proxyUpdateChan:
-			s.allPodUpdater()
-			// TODO(2683): listen to specific pod updates and cert rotations
-		case <-ctx.Done():
-			return
-		}
-	}
-}
-
-func (s *Server) allPodUpdater() {
```
keithmattix, line 32: We really updated all the pods every time a single pod got an update event? Big save of resources here!

pkg/envoy/ads/cache_callbacks.go
```diff
@@ -44,18 +45,59 @@ func (s *Server) OnStreamOpen(ctx context.Context, streamID int64, typ string) e
 	}
 
 	s.proxyRegistry.RegisterProxy(proxy)
+	go func() {
```
keithmattix, line 48: Does this make sense to abstract into a separate function like in the previous implementation?

keithmattix (approved):
LGTM!

pkg/envoy/ads/cache_callbacks.go
steeling: sorry i prematurely merged this.. i'm still investigating ways we can abstract this stuff more. I've taken a note to revisit this :)

PR #5018, opened 2022-08-23
Move OSM domain -> k8s domain methods into the compute provider.

steeling:
This brings us closer to having the k8s.Client contain solely simple CRUD methods (Get, List, Delete, Update, Create)
While the kube.Client maintains the logic to translate between k8s objects and OSM domain objects.

A couple side effects:

1. A couple methods for querying k8s specific objects are added.
2. Some methods have the input changed. For instance, `ListServiceIdentitiesForService` had several calls where a `service.MeshService` was generated, with only the name and namespace populated. This makes it hard to tell what fields we can depend on being populated in each method. (For example ListEndpointsForService has some instances where only the name and namespace are provided, and others where the port is provided)
3. `GetTargetPortForService` was similarly changed to `GetService`. The prior implementation was another example of a partial service being generated, and then calling `GetTargetPortForService`, just to populate it's targetport. A more generic method is to simply call `GetService`

This is another large PR, but is comprised of mostly moving things around.

Part of https://github.com/orgs/openservicemesh/projects/6


pkg/compute/kube/client.go
```diff
@@ -284,3 +264,126 @@ func (c *client) GetHostnamesForService(svc service.MeshService, localNamespace
...
+
+	for svcAcc := range svcAccountsSet.Iter() {
+		identities = append(identities, svcAcc.(identity.K8sServiceAccount).ToServiceIdentity())
+	}
+	return identities, nil
+}
+
+// GetService returns the service.MeshService corresponding to the Port used by clients
+// to communicate with it.
+func (c *client) GetService(name, namespace string, port uint16) (service.MeshService, error) {
```
jsturtevant, line 303: The naming here aligns with rest of package, IMO something like GetMeshService() helps mentally keep track where I am in the code base. Not blocking just a comment as I am learning the abstractions

jsturtevant (commented):
lgtm in general but looks like unit test is related to this change: 

```
--- FAIL: TestGetService (0.10s)
    --- FAIL: TestGetService/TargetPort_found (0.10s)
FAIL
coverage: 63.9% of statements
FAIL	github.com/openservicemesh/osm/pkg/compute/kube	1.070s
```

pkg/compute/kube/client.go
steeling: Fixed!

pkg/compute/kube/client.go
```diff
@@ -406,3 +387,125 @@ func DetectIngressBackendConflicts(x policyv1alpha1.IngressBackend, y policyv1al
...
+					TargetPort: meshSvc.TargetPort,
+					Protocol:   meshSvc.Protocol,
+				})
+			}
+		}
+	}
+	return meshServices
+}
+
+func (c *client) ListNamespaces() ([]string, error) {
```
jsturtevant, line 499: this looks like it is used by the debug handler endpoint which already has a reference to the [kubecontroller](https://github.com/steeling/osm/blob/2e96617b4abe8daed6b8c963f75333f59194c4f7/pkg/debugger/server.go#L52). Is this needed by other components or could it be dropped?

jsturtevant (commented):
one minor comment, otherwise look good.  Thanks for the naming update!

pkg/compute/kube/client.go
steeling: ya i have further plans to remove the kubecontroller from the debugger.

The reason I put it here is that `ListNamespaces` in the k8s/client.go should return a k8s namespace object, while the OSM domain representation of that is simply a string, denoting that it belongs in the compute/client

pkg/compute/kube/types.go
```diff
@@ -1,22 +1,20 @@
...
-
 const (
 	// kindSvcAccount is the ServiceAccount kind
 	kindSvcAccount = "ServiceAccount"
 )
 
 var (
-	log = logger.New("kube-provider")
+	log                = logger.New("kube-provider")
+	errServiceNotFound = fmt.Errorf("Service not found")
```
keithmattix, line 17: nit: errors shouldn't be capitalized

pkg/k8s/client.go
```diff
@@ -271,101 +232,29 @@ func (c *Client) UpdateUpstreamTrafficSettingStatus(obj *policyv1alpha1.Upstream
...
-					Name:       svc.Name,
-					Subdomain:  address.Hostname,
-					Port:       meshSvc.Port,
-					TargetPort: meshSvc.TargetPort,
-					Protocol:   meshSvc.Protocol,
-				})
-			}
-		}
+// GetMeshConfig returns the current MeshConfig
+func (c *Client) GetMeshConfig() configv1alpha2.MeshConfig {
```
keithmattix, line 241: Is this function in MeshCatalog and the compute client?

keithmattix (approved):

pkg/k8s/client.go
steeling: nope it's just here. this gets embedded in the meshcatalog though so it has access to it.

pkg/compute/kube/types.go
steeling: fixed

steeling: also switched to errors.New, which has less allocations :)

keithmattix: Oh does it really? TIL

jaellio (approved):

PR #5020, opened 2022-08-23
separate out the subdomain into a separate field. 

steeling:
This is prepping us for performing queries to the k8s Controller to get service by name and namespace.

Right now there's a lot of places we construct a MeshService on the fly, simply to pass in the name and namespace as the only 2 populated fields. However there's already a handful of methods that have cases on whether certain fields are set. This makes it tough to tell whether the passed in MeshService really has all of the fields populated (and can be depended on), or if they are empty, and we just need the name/namespace.

Part of https://github.com/openservicemesh/osm/issues/4863


keithmattix (commented):
I'll leave a more in-depth review later, but I remember trying this and then moving away from it for some reason; I'll try and see if I can find the reason

steeling:
> I'll leave a more in-depth review later, but I remember trying this and then moving away from it for some reason; I'll try and see if I can find the reason

IIRC, we had previously supported SMI policies and splits based on Name, which we realized SMI doesn't support so we removed that code

pkg/service/types.go
```diff
@@ -28,14 +27,13 @@ type MeshService struct {
...
 	Namespace string
 
-	// The name of the service. May include instance (e.g. pod) information if the backing service
-	// doesn't have a single, stable ip address. For example, a MeshService created by a headless
-	// Kubernetes service named mysql-headless, will have the name "mysql.mysql-headless"
-	// A MeshService created by a normal ClusterIP service named mysql will be named "mysql"
-	// This imposes a restriction that service names cannot contain "." (which is already
-	// the case in kubernetes). Thus, MeshService.Name will be of the form: [subdomain.]providerKey
+	// The name of the service. This will not include a subdomain in the case that the service is backed by a single
+	// pod.
```
jsturtevant, line 31: This comment was a bit confusing, should this be something like clusterip service will not have subdomains?

jsturtevant (commented):
lgtm, I don't think I fully understand the scenario and was confused by the comment but logic looks like the same

keithmattix:
> > I'll leave a more in-depth review later, but I remember trying this and then moving away from it for some reason; I'll try and see if I can find the reason
> 
> IIRC, we had previously supported SMI policies and splits based on Name, which we realized SMI doesn't support so we removed that code

Hmm yeah I guess as long as CDS and RDS are aware of the subdomain, it should be fine. I'll look a bit closer though

jaellio:
I don't understand the full context of this change. Is there a related issue for this PR?

pkg/service/types.go
jaellio: Could the name always exclude the subdomain regardless of whether or not the service is backed by a single pod?

pkg/service/types.go
```diff
@@ -28,14 +27,13 @@ type MeshService struct {
...
-	// The name of the service. May include instance (e.g. pod) information if the backing service
-	// doesn't have a single, stable ip address. For example, a MeshService created by a headless
-	// Kubernetes service named mysql-headless, will have the name "mysql.mysql-headless"
-	// A MeshService created by a normal ClusterIP service named mysql will be named "mysql"
-	// This imposes a restriction that service names cannot contain "." (which is already
-	// the case in kubernetes). Thus, MeshService.Name will be of the form: [subdomain.]providerKey
+	// The name of the service
 	Name string
 
+	// Subdomain is the subdomain of the service. This is only populated if the service is a headless service backed by a
```
keithmattix, line 33: nit: we probably shouldn't couple our definition to Kubernetes concepts like pods. I won't block on it though

pkg/service/types.go
```diff
@@ -51,33 +49,17 @@ type MeshService struct {
...
 }
 
-// Subdomain is an optional subdomain for this MeshService
-// TODO: possibly memoize if performance suffers
-func (ms *MeshService) Subdomain() string {
-	nameComponents := strings.Split(ms.Name, ".")
-	if len(nameComponents) == 1 {
-		return ""
+// FullName is the name of the service including the subdomain.
+func (ms MeshService) FullName() string {
```
keithmattix, line 58: This change is the sneakiest IMO; this means that subdomain is not included by default. Is there an instance where we don't want the subdomain? It feels like there are several mistakes that could result from this change. Additionally, this change prevents Name from being a compound primary key (along with port) over the set of MeshServices. Is this intended?

pkg/service/types.go
```diff
@@ -111,11 +93,6 @@ func (ms MeshService) InboundTrafficMatchName() string {
 	return fmt.Sprintf("inbound_%s_%d_%s", ms, ms.TargetPort, ms.Protocol)
 }
 
-// IngressTrafficMatchName returns the ingress traffic match name
-func (ms MeshService) IngressTrafficMatchName() string {
```
keithmattix, line 115: What was the reason for removing this?

pkg/service/types_test.go
```diff
@@ -58,73 +58,3 @@ func TestMeshServiceCluster(t *testing.T) {
 		})
 	}
 }
-
-func TestMeshService_Subdomain(t *testing.T) {
```
keithmattix, line 62: Missing a test for FullName()

keithmattix (changes requested):
Blocking due to missing test. My main concern is the semantic change to MeshService.Name

steeling:
> I don't understand the full context of this change. Is there a related issue for this PR?

Part of https://github.com/openservicemesh/osm/issues/4863

keithmattix (approved):

jaellio (approved):

PR #4969, opened 2022-08-05
Fix linter issues and clean up unused param

jsturtevant:
Signed-off-by: James Sturtevant <jstur@microsoft.com>

**Description**:
The PRs started to fail when CI `golangci-lint` was bumped to 1.48.0: https://github.com/golangci/golangci-lint/releases

This fixes the linter issues and also cleans up an unused parameter 

**Testing done**:

**Affected area**:
| Functional Area            |     |
| -------------------------- | --- |
| New Functionality          | [ ] |
| CI System                  | [x] |
| CLI Tool                   | [ ] |
| Certificate Management     | [ ] |
| Control Plane              | [ ] |
| Demo                       | [ ] |
| Documentation              | [ ] |
| Egress                     | [ ] |
| Ingress                    | [ ] |
| Install                    | [ ] |
| Networking                 | [ ] |
| Observability              | [ ] |
| Performance                | [ ] |
| SMI Policy                 | [ ] |
| Security                   | [ ] |
| Sidecar Injection          | [ ] |
| Tests                      | [ ] |
| Upgrade                    | [ ] |
| Other                      | [x] |

Please answer the following questions with yes/no.

1. Does this change contain code from or inspired by another project?
    -   Did you notify the maintainers and provide attribution?

2. Is this a breaking change?

3. Has documentation corresponding to this change been updated in the [osm-docs](https://github.com/openservicemesh/osm-docs) repo (if applicable)?

PR #5037, opened 2022-08-24
fix golints G114 and package-comments

shalier:
**Description**:
changes:
1. Suppress for demo apps G114: Use of net/http serve function that has no support for setting timeouts
2. Ignore package-comments, known false positive https://golangci-lint.run/usage/false-positives/#exc0015
3. Change golangci-lint timeout to 20m
4. Remove pkg/mesh and moves isValidUUID() to pkg/cli/proxy_get.go
5. Remove unused pkg/k8s/fake folder

**Testing done**:

**Affected area**:

2. Is this a breaking change?

6. Has documentation corresponding to this change been updated in the [osm-docs](https://github.com/openservicemesh/osm-docs) repo (if applicable)?


demo/cmd/bookbuyer/bookbuyer.go
```diff
@@ -76,8 +76,16 @@ func debugServer() {
 		router.HandleFunc(h.path, h.fn).Methods(h.method)
 	}
 	http.HandleFunc("/favicon.ico", func(w http.ResponseWriter, r *http.Request) {})
+
+	// Initialize bookbuyer HTTP server
+	server := &http.Server{
+		Addr:              fmt.Sprintf(":%d", *port),
+		Handler:           router,
+		ReadHeaderTimeout: time.Second * 10,
+	}
```
shashankram, line 85: Is this being done just to please the linter warning? I don't see anything wrong with the previous code as it uses the default mux server from the http pkg. If that's the case, can't we simply suppress the lint warning?

jsturtevant:
I think this linter issue occurs because we are running latest linter in CI. There is lots of value to that, but the UI is pretty rough since it doesn't show you all the linter failures in one go and it takes a really long time locally.  I fixed a bunch of these last month https://github.com/openservicemesh/osm/pull/4969.  

We have the timeout set to 10m in the  https://github.com/openservicemesh/osm/blob/157a15ef7ebc99d48f60f93ec5a4a31a11a15f4e/.golangci.yml#L3 but looks like it only running for 5.  maybe add `args: --timeout=10m` to https://github.com/openservicemesh/osm/blob/157a15ef7ebc99d48f60f93ec5a4a31a11a15f4e/.github/workflows/main.yml#L55-L58 and it will not timeout?

Thanks for addressing them @shalier, this was frustrating when I did it last month :-/

jsturtevant:
That did make it run for 10 mins but looks like it still timed out :-(

.github/workflows/main.yml
```diff
@@ -56,6 +56,7 @@ jobs:
         with:
           version: latest
           skip-pkg-cache: true
+          args: --timeout=20m
```
keithmattix, line 59: Is this required given the .golangci file?

pkg/cli/proxy_get.go
```diff
@@ -72,7 +73,7 @@ func getPortForwarder(clientSet kubernetes.Interface, config *rest.Config, names
 	if err != nil {
 		return nil, fmt.Errorf("error getting pod %s/%s", namespace, podName)
 	}
-	if !mesh.ProxyLabelExists(*pod) {
```
keithmattix, line 75: This looks like a functional change; I'd prefer for functional changes to be in a separate PR from linter appeasement

keithmattix (approved):

shashankram (approved):

PR #5049, opened 2022-08-29
Allow SDS NewResponse with no request object

steeling:
Once we switch to the snapshot cache, we will not be calling SDS NewResponse with a request object.

This PR does not make that switch, but allows the NewResponse function to be called with a nil request object.

This PR also has one other big change, which is to switch from an Inbound Validation secret context for each inbound service, to a single inbound validation context for all services. This makes our secret generation logic much simpler, while simultaneously reducing resource consumption across the fleet.

I've also removed some unnecessary  test cases, particularly those which were exercising already tested paths, and were testing unexported methods. This sticks with the go recommendation to keep unit tests on exported methods, to avoid internal changes breaking tests.


pkg/cli/verifier/testdata/httpbin1_permissive.json
```diff
@@ -1837,12 +1837,12 @@
      }
     },
     {
-     "name": "root-cert-for-mtls-inbound:httpbin/httpbin",
```
keithmattix, line 1840: Is there a reason this was service specific originally? /cc @shashankram

pkg/envoy/ads/response_test.go
```diff
@@ -33,11 +33,11 @@ var _ = Describe("Test ADS response functions", func() {
 	fakeCertManager, err := certificate.FakeCertManager()
 	Expect(err).ToNot(HaveOccurred())
 	proxyUUID := uuid.New()
-	proxySvcAccount := tests.BookstoreServiceAccount
+	proxySvcID := tests.BookstoreServiceIdentity
```
keithmattix, line 36: `proxySvcID` is still functionally a service account right? This is just de-coupling from Kubernetes right?

pkg/envoy/sds/builder.go
```diff
@@ -20,11 +20,12 @@ type sdsBuilder struct {
 
 	// The DiscoveryRequest contains the certs to generate secrets for
 	// requestedCerts is expected to be a list of either "service-cert:namespace/service" or "root-cert:namespace/service"
-	requestedCerts []string
+	requestedCerts map[string]any
```
keithmattix, line 23: I haven't seen `any` before; is this a go1.19 alias for `interface{}`?

keithmattix (approved):

pkg/envoy/ads/response_test.go
steeling: yup, that's right!

pkg/envoy/sds/builder.go
steeling: yes, exactly, it's intended to be a more idiomatic name for interface{}

pkg/cli/verifier/testdata/httpbin1_permissive.json
steeling: My guess is we deemed it more flexible at the time, but with the new builder approach it's pretty trivial to move between the 2, and having a single item is both easier to conceptualize, less code, and less compute resources.

I'll wait to resolve for @shashankram's comment

shashankram: Originally, each inbound secret would also specify the allowed client SANs within the validation context. Later, the SAN matching was delegated to the network RBAC filter in the inbound listener instead of doing it via the secret config. So the optimization makes sense now - it was something that was missed during the refactoring of this code before.

pkg/envoy/sds/builder.go
```diff
@@ -53,62 +57,59 @@ func (b *sdsBuilder) SetTrustDomain(trustDomain string) *sdsBuilder {
...
-	b.serviceIdentitiesForOutboundServices = serviceIdentitiesForServices
+	b.identitiesForSecrets = make(map[string][]identity.ServiceIdentity)
+	for svc, serviceIdentities := range serviceIdentitiesForServices {
+		b.identitiesForSecrets[secrets.NameForUpstreamService(svc)] = serviceIdentities
+	}
 	return b
 }
 
+// TODO(2683): Remove this function once we use the snapshot cache, and remove requestedCerts.
+func (b *sdsBuilder) shouldBuild(secretName string) bool {
```
shashankram, line 68: Could you add a comment stating what this function does.

shashankram (approved):

pkg/envoy/sds/builder.go
```diff
@@ -20,11 +20,12 @@ type sdsBuilder struct {
 
 	// The DiscoveryRequest contains the certs to generate secrets for
 	// requestedCerts is expected to be a list of either "service-cert:namespace/service" or "root-cert:namespace/service"
```
jaellio, line 22:
```suggestion
	// requestedCerts is expected to be a list of either "service-cert:namespace/service" or "root-cert"
```

jaellio (approved):

pkg/envoy/sds/builder.go
steeling: will do!

PR #5050, opened 2022-08-30
feat(certs): add intent, conditions, and componentStatuses to MRC resource

### LINKED ISSUE #5044: Add new fields and statuses/conditions to MRC resource
To enable declarative certificate management and rotation, the MeshRootCertificate resource needs a new `intent` field to describe whether or not it is meant to be "active" or "passive". More states may be possible later (e.g. "decomissioned"), so this should not be expressed with a simple boolean field.

Additionally, the MeshRootCertificate status subresource (represented by the `MeshRootCertificateStatus`) type needs to hold the following fields:

```go
// Status of the certificate component, one of (`True`, `False`, `Unknown`).
type MeshRootCertificateComponentStatus string

type MeshRootCertificateComponentStatuses struct {
	Webhooks        MeshRootCertificateComponentStatus `json:"webhooks"`
	XDSControlPlane MeshRootCertificateComponentStatus `json:"xdsControlPlane"`
	Sidecar         MeshRootCertificateComponentStatus `json:"sidecar"`
	Bootstrap       MeshRootCertificateComponentStatus `json:"bootstrap"`
	Gateway         MeshRootCertificateComponentStatus `json:"gateway"`
}

type MeshRootCertificateCondition struct {
	// Type of the condition, known values are (`Ready`, `Accepted`,
	// `IssuingRollout`, `ValidatingRollout`, `IssuingRollback`, `ValidatingRollout`).
	Type MeshRootCertificateConditionType `json:"type"`

	// Status of the condition, one of (`True`, `False`, `Unknown`).
	Status MeshRootCertificateConditionStatus `json:"status"`

	// LastTransitionTime is the timestamp corresponding to the last status
	// change of this condition.
	// +optional
	LastTransitionTime *metav1.Time `json:"lastTransitionTime,omitempty"`

	// Reason is a brief machine readable explanation for the condition's last
	// transition (should be in camelCase)
	// +optional
	Reason string `json:"reason,omitempty"`

	// Message is a human readable description of the details of the last
	// transition, complementing reason.
	// +optional
	Message string `json:"message,omitempty"`
}

// MeshRootCertificateStatus defines the status of the MeshRootCertificate resource
type MeshRootCertificateStatus struct {
	// State specifies the state of the certificate provider
	// All states are specified in constants.go
	State string `json:"state"`

	// Set of statuses for each certificate component in the cluster (e.g. webhooks, bootstrap, etc.)
	ComponentStatuses MeshRootCertificateComponentStatuses `json:"componentStatuses"`

	// List of status conditions to indicate the status of a MeshRootCertificate.
	// Known condition types are `Ready` and `InvalidRequest`.
	// +optional
	Conditions []MeshRootCertificateCondition `json:"conditions"`
}
```

Note: consider NOT remove the existing constants for the `status.state` fields (e.g. `ValidatingRollout`) yet so that MRCEvent handling doesn't have to be refactored in the PR for this issue. 

P.S  - All names are TBD/suggestions. Feel free to change them in the implementation of this issue.

shalier:
**Description**:
Adds MeshRootCertificateIntent, MeshRootCertificateComponentStatuses, and MeshRootCertificateCondition to MRC resource
and resolves #5044

**Testing done**:

**Affected area**:

2. Is this a breaking change?


cmd/osm-bootstrap/osm-bootstrap.go
```diff
@@ -430,6 +430,39 @@ func (b *bootstrap) createMeshRootCertificate() error {
 
 	createdMRC.Status = configv1alpha2.MeshRootCertificateStatus{
 		State: constants.MRCStateActive,
```
keithmattix, line 432: This initial state for a created MRC should now be `Pending`

pkg/apis/config/v1alpha2/meshrootcertificate.go
```diff
@@ -111,11 +115,69 @@ type TresorCASpec struct {
...
+// one of (`True`, `False`, `Unknown`).
+type MeshRootCertificateConditionStatus string
+
+// MeshRootCertificateConditionType specifies the type of the condition,
+// one of (`Ready`, `Accepted`, `IssuingRollout`, `ValidatingRollout`, `IssuingRollback`, `ValidatingRollback`).
+type MeshRootCertificateConditionType string
+
+// MeshRootCertificateComponentStatuses is the set of statuses for each certificate component in the cluster.
+type MeshRootCertificateComponentStatuses struct {
+	Webhooks        MeshRootCertificateComponentStatus `json:"webhooks"`
```
keithmattix, line 136: @jaellio do you see anything incorrect or missing?

pkg/constants/constants.go
```diff
@@ -210,6 +210,32 @@ const (
...
+	MRCConditionStatusUnknown = "unknown"
+
+	// MRCConditionTypeReady is the ready condition type for certificate condition of the MeshRootCertificate
+	MRCConditionTypeReady = "ready"
+
+	// MRCConditionTypeAccepted is the accepted condition type for certificate condition of the MeshRootCertificate
+	MRCConditionTypeAccepted = "accepted"
+
+	// MRCConditionTypeIssuingRollout is the issuing rollout condition type for certificate condition of the MeshRootCertificate
+	MRCConditionTypeIssuingRollout = "issuingRollout"
```
keithmattix, line 229: Condition types should be CamelCase by [convention](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties). ComponentStatuses should also probably be CamelCase

cmd/osm-bootstrap/crds/config_mesh_root_certificate.yaml
```diff
@@ -152,6 +152,9 @@ spec:
                     - required: ["certManager"]
                     - required: ["vault"]
                     - required: ["tresor"]
+            intent:
+              description: Specifies the intent of the MeshRootCertificate can be (Active, Passive)
```
jaellio, line 156: nit: Not sure if "behavior" is the best word here. What are your thoughts?
```suggestion
              description: Intent specifies the behavior of the MeshRootCertificate, can be Active or Passive
```

pkg/apis/config/v1alpha2/meshrootcertificate.go
```diff
@@ -111,11 +115,69 @@ type TresorCASpec struct {
...
+}
+
+// MeshRootCertificateStatus defines the status of the MeshRootCertificate resource.
 type MeshRootCertificateStatus struct {
-	// State specifies the state of the certificate provider
+	// State specifies the state of the certificate provider.
 	// All states are specified in constants.go
 	State string `json:"state"`
+
+	// Set of statuses for each certificate component in the cluster (e.g. webhooks, bootstrap, etc.).
```
jaellio, line 174: nit: Maybe remove the period from the end of the comment to be consistent with the rest of the file.
```suggestion
	// Set of statuses for each certificate component in the cluster (e.g. webhooks, bootstrap, etc.)
```

pkg/apis/config/v1alpha2/meshrootcertificate.go
jaellio: LGTM!

cmd/osm-bootstrap/osm-bootstrap.go
```diff
@@ -429,7 +429,40 @@ func (b *bootstrap) createMeshRootCertificate() error {
 	}
 
 	createdMRC.Status = configv1alpha2.MeshRootCertificateStatus{
-		State: constants.MRCStateActive,
+		State: constants.MRCStatePending,
```
jaellio, line 432: For the default MRC created on start-up should the state be active?

pkg/apis/config/v1alpha2/meshrootcertificate.go
```diff
@@ -21,6 +21,10 @@ type MeshRootCertificate struct {
 	// +optional
 	Spec MeshRootCertificateSpec `json:"spec,omitempty"`
 
+	// Intent of the MeshRootCertificate resource
+	// +optional
+	Intent MeshRootCertificateIntent `json:"intent,omitempty"`
```
jaellio, line 26: Could you also update the `preset-mesh-root-certificate.yaml` to include the `intent` field? Thanks!

pkg/certificate/providers/config.go
```diff
@@ -68,6 +68,39 @@ func NewCertificateManager(ctx context.Context, kubeClient kubernetes.Interface,
 			},
```
jaellio, line 68: Intent could also be set here.

cmd/osm-bootstrap/osm-bootstrap.go
keithmattix: I'm not 100% sure on that; wouldn't we want the controller to verify that the MRC is active first? What if the issuer can't be contacted?

cmd/osm-bootstrap/crds/config_mesh_root_certificate.yaml
jsturtevant: would `role` make sense here over `intent`?  for example the role of the certificate would be `active` or `passive`.

cmd/osm-bootstrap/osm-bootstrap.go
jsturtevant: would the conditions also be known in this case?

cmd/osm-bootstrap/crds/config_mesh_root_certificate.yaml
keithmattix: I can see that! I was thinking `role` was more aligned with `validating` or `issuing` rather than active or passive, but I can see either way!

cmd/osm-bootstrap/osm-bootstrap.go
keithmattix: For the preset MRC, we may still have to connect to the issuer (a k8s secret, cert-manager, vault, etc.) and that connection may fail; wouldn't the MRC be waiting for a controller to confirm things are working as expected (i.e. pending) rather than assuming everything's working during creation?

charts/osm/templates/preset-mesh-root-certificate.yaml
```diff
@@ -41,6 +41,7 @@ data:
           "port": {{.Values.osm.vault.port | mustToJson}}
         }
         {{- end}}
-      }
+      },
+      "intent": {{.Values.osm.intent}}
```
jaellio, line 45: Thanks for adding this here! I don't think `intent` should be added as a value. If the user is using the default MRC the `intent` should be `active` (the default MRC should only be created if there are no other MRCs in the osm namespace and thus it can be assumed it is the `active` MRC). What are your thoughts on this?

cmd/osm-bootstrap/osm-bootstrap.go
jaellio: That makes sense to me! I think I had this field confused with `intent` in the MRC spec.

charts/osm/templates/preset-mesh-root-certificate.yaml
shalier: Ah gotcha, that makes sense

keithmattix (approved):

pkg/certificate/providers/config.go
```diff
@@ -66,8 +66,42 @@ func NewCertificateManager(ctx context.Context, kubeClient kubernetes.Interface,
 				Provider:    option.AsProviderSpec(),
 				TrustDomain: trustDomain,
 			},
+			Intent: constants.MRCIntentPassive,
```
jaellio, line 69: nit: I think the intent here could be active, but I don't think this value impacts the functionality of the mrcCompatClient.

PR #5051, opened 2022-08-30
remove test that checks if error was returned on grpc server

steeling:
We don't issue the certificate in the grpc server anymore, so the specific error case we were testing was removed.


jaellio (approved):
Good catch, thanks!

PR #5057, opened 2022-08-31
switch to envoy wellknown types for config names

steeling:
The snapshot cache implementation requires a "consistent cache", meaning references to other config objects require the referenced config is present.

It traces these configs through the wellknown names. While the envoy sidecar itself does not care about most of these wellknown names, the snapshot cache implementation does.

I also fix a nil pointer dereference and remove some unused code

### DIFF
--- a/pkg/cli/verifier/envoy_config.go
+++ b/pkg/cli/verifier/envoy_config.go
@@ -11,13 +11,13 @@ import (
 	xds_listener "github.com/envoyproxy/go-control-plane/envoy/config/listener/v3"
 	xds_route "github.com/envoyproxy/go-control-plane/envoy/config/route/v3"
 	xds_secret "github.com/envoyproxy/go-control-plane/envoy/extensions/transport_sockets/tls/v3"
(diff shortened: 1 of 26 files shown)


pkg/cli/verifier/envoy_config.go
```diff
@@ -331,10 +331,10 @@ func findOutboundFilterChainForServicePort(meshSvc service.MeshService, dstIPRan
 func getFilterForProtocol(protocol string) string {
 	switch protocol {
 	case constants.ProtocolHTTP:
-		return envoy.HTTPConnectionManagerFilterName
+		return wellknown.HTTPConnectionManager
```
keithmattix, line 334: @shashankram removed wellknown types in #4834 because they were deprecated; could you link to Envoy docs containing justification for bringing them back?

steeling: if https://github.com/envoyproxy/go-control-plane/pull/589 gets in then i can close this

PR #5058, opened 2022-08-31
decouple envoy proxy from k8s by removing pod metadata

steeling:
This is a step in reducing our code's knowledge of direct kubernetes concepts.

By removing pod metadata, we can remove a call to GetPodForProxy, and remove strictly Kubernetes objects off the envoy.Proxy object. This helps us push down the k8s calls into a lower layer of the stack, and lets our business logic be abstracted form K8s.

It should also be noted that we only stored Pod metadata, in a custom struct, to later convert it into stats headers, of type `map[string]string`. This code change both gets rid of certain possible errors that could be returned, removes 100+ lines of code, and improves our layers of abstraction.

Part of https://github.com/openservicemesh/osm/issues/4863


steeling:
> Are we not replacing pod metadata with anything? Seems like a lot of information being lost

I get it on the fly when needed from the compute.client instead of storing it on the proxy. 

It's only used to parse out stats headers, so it's more of the responsbility of the envoy xDS builders than it is of storing it on the proxy.

keithmattix (commented):
So are we not doing WASM anymore?

steeling:
> So are we not doing WASM anymore?

This has no functional end result change, it maintains the status quo

steeling:
> @steeling please add more details to the commit && PR description describing the motivation for the change.
> 
> Thanks

done!

pkg/envoy/ads/cache_callbacks.go
```diff
@@ -38,12 +38,6 @@ func (s *Server) OnStreamOpen(ctx context.Context, streamID int64, typ string) e
 
 	proxy := envoy.NewProxy(kind, uuid, si, utils.GetIPFromContext(ctx), streamID)
 
-	if err := s.recordPodMetadata(proxy); err == errServiceAccountMismatch {
-		// Service Account mismatch
-		log.Error().Err(err).Str("proxy", proxy.String()).Msg("Mismatched service account for proxy")
```
jsturtevant, line 43: Are we losing anything by not have the runtime check against the pods Service account?  dropping/moving the metadata struct to remove the dependency is good but is there a need to do this verification to ensure the cert matches what we expect?

steeling: Ya I thought about that.. we can bring it back.

FWIW it's indirectly present in the GetPodForProxy call, but this is called on each iteration of building the configs and wouldn't return early as compared to here.

I'll leave the comment open to other's thoughts on if we want to keep it

jsturtevant: >FWIW it's indirectly present in the GetPodForProxy call, but this is called on each iteration of building the configs and wouldn't return early as compared to here.

ah now I remember it being there and wondering why it was in two places :-).  I like the explicit and early failure check here and I think it could still be platform independent.  let's see what other say

steeling: re-added this back!

pkg/compute/kube/client.go
```diff
@@ -284,3 +285,44 @@ func (c *client) GetHostnamesForService(svc service.MeshService, localNamespace
...
+
+	return map[string]string{
+		"osm-stats-pod":       pod.Name,
+		"osm-stats-namespace": pod.Namespace,
+		"osm-stats-kind":      workloadKind,
+		"osm-stats-name":      workloadName,
+	}, nil
+}
+
+// VerifyProxy verifies that the proxy is formed correctly and aligns with what is in the k8s control plane.
```
shashankram, line 323: This comment is quite ambiguous to me. What does `proxy is formed correctly and aligns with what is in the k8s control plane` mean? 

Can this be reworded to `VerifyProxy attempts to lookup a pod that matches the given proxy instance by its service identity, namespace, and UUID`

shashankram (approved):

pkg/compute/types.go
```diff
@@ -37,4 +37,9 @@ type Interface interface {
 
 	// ListServicesForProxy gets the services that map to the given proxy.
 	ListServicesForProxy(p *envoy.Proxy) ([]service.MeshService, error)
+
+	GetProxyStatsHeaders(p *envoy.Proxy) (map[string]string, error)
+
+	// VerifyProxy verifies that the proxy is formed correctly and aligns with what is in the control plane.
```
keithmattix, line 43: If you fix Shashank's comment above, fix it here too

keithmattix (approved):

pkg/compute/kube/client.go
steeling: done!

pkg/compute/types.go
steeling: done!

pkg/envoy/ads/stream.go
```diff
@@ -51,12 +51,6 @@ func (s *Server) StreamAggregatedResources(server xds_discovery.AggregatedDiscov
 	streamID := atomic.AddInt64(&s.nextStreamID, 1)
 	proxy := envoy.NewProxy(kind, uuid, si, utils.GetIPFromContext(server.Context()), streamID)
 
-	if err := s.recordPodMetadata(proxy); err == errServiceAccountMismatch {
```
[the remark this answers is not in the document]
jsturtevant: @steeling double checking if we need a Verify proxy here or just in cache_callbacks.go?

PR #5064, opened 2022-09-01
ref(cds): leverage builder pattern for CDS

SanyaKochhar:
**Description**:

**Testing done**:

**Affected area**:

2. Is this a breaking change?


pkg/envoy/cds/builder.go
```diff
@@ -28,9 +31,139 @@ import (
 	"github.com/openservicemesh/osm/pkg/trafficpolicy"
 )
 
+type clusterBuilder struct {
+	proxyIdentity identity.ServiceIdentity
+
+	outboundMeshTrafficPolicy *trafficpolicy.OutboundMeshTrafficPolicy
```
steeling, line 37: I noticed for all of the *trafficpolicy.*TrafficPolicy objects, we only use the `clusterConfigs`. Can we reduce the stored property to just those fields that are used?

pkg/envoy/cds/builder.go
```diff
@@ -28,9 +31,139 @@ import (
...
+
+	egressTrafficPolicy *trafficpolicy.EgressTrafficPolicy
+
+	envoyOutboundPassthroughCluster string
+
+	egressEnabled bool
+
+	metricsEnabled bool
+
+	tracingEnabled bool
```
steeling, line 51: can we collapse enabled, and address into a single field, such that if address is not the empty string then it's enabled?

pkg/envoy/cds/builder.go
```diff
@@ -28,9 +31,139 @@ import (
...
+
+	outboundMeshTrafficPolicy *trafficpolicy.OutboundMeshTrafficPolicy
+
+	inboundMeshTrafficPolicy *trafficpolicy.InboundMeshTrafficPolicy
+
+	sidecarSpec configv1alpha2.SidecarSpec
+
+	egressTrafficPolicy *trafficpolicy.EgressTrafficPolicy
+
+	envoyOutboundPassthroughCluster string
```
steeling, line 45: nit: I would consider leaving this as a non-field, and continue to use the constant value in its place

pkg/envoy/cds/builder.go
```diff
@@ -28,9 +31,139 @@ import (
...
+	b.egressTrafficPolicy = egressTrafficPolicy
+	return b
+}
+
+func (b *clusterBuilder) SetEnvoyOutBoundPassthroughCluster(envoyOutboundPassthroughCluster string) *clusterBuilder {
+	b.envoyOutboundPassthroughCluster = envoyOutboundPassthroughCluster
+	return b
+}
+
+func (b *clusterBuilder) SetEgressEnabled(egressEnabled bool) *clusterBuilder {
```
steeling, line 97: nit: all of these exported methods will need comments to pass linting

pkg/envoy/cds/builder.go
```diff
@@ -28,9 +31,139 @@ import (
...
+
 // replacer used to configure an Envoy cluster's altStatName
 var replacer = strings.NewReplacer(".", "_", ":", "_")
 
+func (b *clusterBuilder) Build() ([]*xds_cluster.Cluster, error) {
+	var clusters []*xds_cluster.Cluster
+
+	// Build upstream clusters based on allowed outbound traffic policies
+	if b.outboundMeshTrafficPolicy != nil {
+		clusters = append(clusters, upstreamClustersFromClusterConfigs(b.proxyIdentity, b.outboundMeshTrafficPolicy.ClustersConfigs, b.sidecarSpec)...)
```
steeling, line 130: instead of passing these fields in, I'd consider making `upstreamClustersFromClusterConfigs` a method on the builder which has access to those fields. ie:

```go
clusters = append(clusters, b.buildUpstreamClusters()
```

pkg/envoy/cds/builder.go
```diff
@@ -28,9 +31,139 @@ import (
...
+	}
+
+	// Build local clusters based on allowed inbound traffic policies
+	if b.inboundMeshTrafficPolicy != nil {
+		clusters = append(clusters, localClustersFromClusterConfigs(b.inboundMeshTrafficPolicy.ClustersConfigs)...)
```
steeling, line 135: same here, and below

pkg/envoy/cds/builder.go
```diff
@@ -28,9 +31,139 @@ import (
...
+	if b.metricsEnabled {
+		clusters = append(clusters, getPrometheusCluster())
+	}
+
+	// Add an outbound tracing cluster (from localhost to tracing sink)
+	if b.tracingEnabled {
+		clusters = append(clusters, getTracingCluster(b.tracingAddress, b.tracingPort))
+	}
+
+	return removeDups(clusters), nil
```
steeling, line 164: I don't think we need remove dups anymore, that was an edge case from the original multicluster attempt

steeling (commented):
this looks great! A handful of comments, but mostly nits!

pkg/envoy/cds/builder.go
```diff
@@ -28,9 +28,139 @@ import (
...
 )
 
+type clusterBuilder struct {
+	proxyIdentity identity.ServiceIdentity
+
+	outboundMeshTrafficPolicy *trafficpolicy.OutboundMeshTrafficPolicy
+
+	inboundMeshTrafficPolicy *trafficpolicy.InboundMeshTrafficPolicy
+
+	sidecarSpec configv1alpha2.SidecarSpec
```
jsturtevant, line 38: should this be a more generic set of configuration options?  Sidecar spec sounds like k8s object?

steeling: that's a great question! I'd argue that's a bit out of scope, and the sidecarspec is actually defined by us, so if in a world where OSM ran outside of K8s, I think it would be reasonable for us to define our required API's, which could still include the SidecarSpec

But I concede it takes some squinting to get to that conclusion :)

jsturtevant: makes sense. The `Sidecar` terminology through me off a bit but looking at struct it seems like it has some generic configuration for the "proxy". It does have some things that are a bit overly specific to kubernetes but we do own it and could tweak it in the future.  Thanks!

pkg/envoy/cds/builder.go
SanyaKochhar: that would make sense but we do have a default value set for tracing address and port even if the address is nil or observability isn't enabled: https://github.com/openservicemesh/osm/blob/c5967d6df3d61ae0f864324824bdfaa78c58cc78/pkg/utils/mesh_config.go#L46

steeling: That makes sense on the meshconfig, but on the builder we can have it being an empty string still. I'll leave a comment on the code where it's set

pkg/envoy/cds/builder.go
```diff
@@ -204,6 +329,31 @@ func getPrometheusCluster() *xds_cluster.Cluster {
...
+		},
+		LbPolicy: xds_cluster.Cluster_ROUND_ROBIN,
+		LoadAssignment: &xds_endpoint.ClusterLoadAssignment{
+			ClusterName: constants.EnvoyTracingCluster,
+			Endpoints: []*xds_endpoint.LocalityLbEndpoints{
+				{
+					LbEndpoints: []*xds_endpoint.LbEndpoint{{
+						HostIdentifier: &xds_endpoint.LbEndpoint_Endpoint{
+							Endpoint: &xds_endpoint.Endpoint{
+								Address: envoy.GetAddress(address, port),
```
steeling, line 347: super nit: I'd consider just setting the address on the builder as the value of envoy.GetAddress

pkg/envoy/cds/response.go
```diff
@@ -16,74 +15,46 @@ import (
...
 	} else if enabled {
-		clusters = append(clusters, getPrometheusCluster())
+		cb.SetMetricsEnabled(enabled)
 	}
 
 	// Add an outbound tracing cluster (from localhost to tracing sink)
-	if cfg.GetMeshConfig().Spec.Observability.Tracing.Enable {
-		clusters = append(clusters, getTracingCluster(cfg))
+	if enabled := cfg.GetMeshConfig().Spec.Observability.Tracing.Enable; enabled {
+		cb.SetTracingEnabled(enabled)
```
steeling, line 50: here we can just do:

```go
if cfg.GetMeshConfig().Spec.Observability.Tracing.Enable {
   // sets the value, including the port.
   cb.SetTracingAddress(utils.GetEnvoyTracingAddress(cfg.GetMeshConfig())
}
```

Then in the builder you can have a method:

```go
func (b *clusterBuilder) TracingEnabled() bool {
  return b.tracingAddress != ""
}
```

pkg/envoy/cds/response.go
```diff
@@ -16,74 +15,46 @@ import (
...
 		log.Warn().Str("proxy", proxy.String()).Msg("Could not find pod for connecting proxy, no metadata was recorded")
 	} else if enabled {
-		clusters = append(clusters, getPrometheusCluster())
+		cb.SetMetricsEnabled(enabled)
 	}
 
 	// Add an outbound tracing cluster (from localhost to tracing sink)
-	if cfg.GetMeshConfig().Spec.Observability.Tracing.Enable {
-		clusters = append(clusters, getTracingCluster(cfg))
+	if enabled := cfg.GetMeshConfig().Spec.Observability.Tracing.Enable; enabled {
```
steeling, line 49: nit: you don't need the `enabled` var, since you're in a truthy block, you're guaranteed enabled is true, and it's a *small* amount of less cognitive overhead to just use the true value

although I'm proposing getting rid of even storing the enabled var altogether in the following comment

steeling (approved):

pkg/envoy/cds/builder.go
SanyaKochhar: it seems lint doesn't seem to have an issue with this, but I can add them if you feel it's necessary

pkg/envoy/cds/builder.go
SanyaKochhar: seems like this other test case also expects duplicates to exist if an egress traffic cluster config lists the same cluster twice, is this also multicluster related?
https://github.com/openservicemesh/osm/blob/main/pkg/envoy/cds/response_test.go#L437-L460

pkg/envoy/cds/builder.go
```diff
@@ -28,9 +31,117 @@ import (
...
+	if b.inboundMeshTrafficClusterConfigs != nil {
+		clusters = append(clusters, b.buildLocalClusters()...)
+	}
+
+	// Add egress clusters based on applied policies
+	if b.egressTrafficClusterConfigs != nil {
+		clusters = append(clusters, b.getEgressClusters()...)
+	}
+
+	// Add an outbound passthrough cluster for egress if global mesh-wide Egress is enabled
```
SanyaKochhar, line 121: note the change in logic here: https://github.com/openservicemesh/osm/blob/main/pkg/envoy/cds/response.go#L47-L57

Prev:
- get outbound passthru cluster + handle error
- check if egress enabled - if yes, add outboundpassthroughcluster

Now:
- check if egress enabled - if yes, only then try to get outbound passthru cluster and handle error

pkg/envoy/cds/builder.go
```diff
@@ -28,9 +31,117 @@ import (
...
+
+	outboundMeshTrafficClusterConfigs []*trafficpolicy.MeshClusterConfig
+
+	inboundMeshTrafficClusterConfigs []*trafficpolicy.MeshClusterConfig
+
+	egressTrafficClusterConfigs []*trafficpolicy.EgressClusterConfig
+
+	sidecarSpec configv1alpha2.SidecarSpec
+
+	egressEnabled bool
```
keithmattix, line 45: This should be renamed to globalEgressEnabled to distinguish between the meshconfig setting and the egress policy api

keithmattix (approved):
One small nit that's not blocking; use your best judgement. Great work!

PR #5065, opened 2022-09-01
Remove crd conversion webhook

nshankar13:

**Description**: Resolves issue: https://github.com/openservicemesh/osm/issues/4967

As discussed in the issue, the crd conversion webhook currently does not act differently than how k8s would update CRDs with the conversion strategy set to `None`. Typically, the conversionwebhook is necessary for schema changes between CRDs on upgrades (for instance, if the field `foo` was renamed to the field `bar`). However, we can handle these scenarios separately down the road if the need arises. 

**Testing done**: Tested upgrading from v1.0.0 to a v1.2.1-test version built from this branch and the upgrades succeeded. 

`OSM successfully upgraded mesh [osm] in namespace [osm-system]`

**Affected area**:

2. Is this a breaking change?


cmd/osm-bootstrap/osm-bootstrap.go
```diff
@@ -203,48 +195,9 @@ func main() {
...
-				"Error initializing certificate manager of kind %s", certProviderKind)
-		}
-	}
-
-	// Initialize the crd conversion webhook server to support the conversion of OSM's CRDs
-	if err := crdconversion.NewConversionWebhook(ctx, crdClient, certManager, osmNamespace); err != nil {
-		events.GenericEventRecorder().FatalEvent(err, events.InitializationError, fmt.Sprintf("Error creating crd conversion webhook: %s", err))
-	}
-
 	version.SetMetric()
```
steeling, line 201: Can you remove the reconciler as well?

nshankar13: @steeling looks like we also add the reconcile label to the mutatingwebhook and validatingwebhook, should we handle the reconciler removal in a follow up PR? Or will removing the conversionwebhook while keeping the reconciler introduce bugs?

steeling: we should delete it for all. it's broken on upgrades in that an old reconciler will stomp over any changes we try to make

shashankram (changes requested):
Thanks for working on a fix for the upgrade issue.

I want to ensure we address a couple of things before we make a decision on completely removing the CRD conversion framework.
1. Does this actually fix the bug. If yes, kindly provide a repro for the issue and add a e2e test to prove this is the correct fix.
2. Have a proposal on how CRD schema changes are to be handled. Completely removing this will only push this problem down the road for someone else that needs to update the CRD schema to worry about. I don't believe this is the correct approach to solving this problem.
3. Instead of deleting the framework, add a feature flag to optionally turn it on, as it may be desirable to re-integrate this in the future. We do not know of how we can seamlessly handle CRD conversions when the API schema changes, so I am uncomfortable trading 1 bug fix without a concrete proposal to address future needs, which will arise.

Overall, I'd like to be more cautious on making such changes without a definitive way to repro and verify the fix while addressing future extensibility.

steeling:
> Thanks for working on a fix for the upgrade issue.
> 
> I want to ensure we address a couple of things before we make a decision on completely removing the CRD conversion framework.
> 
> 1. Does this actually fix the bug. If yes, kindly provide a repro for the issue and add a e2e test to prove this is the correct fix.

I have a repro for this, happy to demo it.

> 2. Have a proposal on how CRD schema changes are to be handled. Completely removing this will only push this problem down the road for someone else that needs to update the CRD schema to worry about. I don't believe this is the correct approach to solving this problem.

I have a proposal out for that as well. I've sent it to you.

> 3. Instead of deleting the framework, add a feature flag to optionally turn it on, as it may be desirable to re-integrate this in the future. We do not know of how we can seamlessly handle CRD conversions when the API schema changes, so I am uncomfortable trading 1 bug fix without a concrete proposal to address future needs, which will arise.

I disagree with adding a feature flag for this. Version control makes it trivial to bring it back if we desire, by simply reverting this PR. In the proposal I demonstrate how and why we won't need a conversion webhook, and that there are a lot of (unsafe) misconceptions about how the webhook works.

> 
> Overall, I'd like to be more cautious on making such changes without a definitive way to repro and verify the fix while addressing future extensibility.

shashankram:
> > Thanks for working on a fix for the upgrade issue.
> > I want to ensure we address a couple of things before we make a decision on completely removing the CRD conversion framework.
> > 
> > 1. Does this actually fix the bug. If yes, kindly provide a repro for the issue and add a e2e test to prove this is the correct fix.
> 
> I have a repro for this, happy to demo it, although adding a test for this would be difficult (but not impossible). I think we should move forward with the fix now, and can circle back to an E2E test in the future.
> 
> For those wondering the steps are simple:
> 
> 1. Deploy OSM v1.0.0
> 2. Scale the bootstrap pod to 0
> 3. Upgrade to v1.2.0
> 4. The new bootstrap pod fails to initialize
> 5. Attempt to scale the old bootstrap pod back up
> 6. Both the old and new bootstrap pod cannot start, and the upgrade fails.
> 7. Rollout restart the injector and controller, and now all components are down and nothing can turn on.
> 
> > 2. Have a proposal on how CRD schema changes are to be handled. Completely removing this will only push this problem down the road for someone else that needs to update the CRD schema to worry about. I don't believe this is the correct approach to solving this problem.
> 
> [There's a proposal here with the full description](https://docs.google.com/document/d/1BnRHTxwXTcxl8NkvCRSRZJdFIATJNLr4GWUeWHswD_I/edit#)
> 
> > 3. Instead of deleting the framework, add a feature flag to optionally turn it on, as it may be desirable to re-integrate this in the future. We do not know of how we can seamlessly handle CRD conversions when the API schema changes, so I am uncomfortable trading 1 bug fix without a concrete proposal to address future needs, which will arise.
> 
> I disagree with adding a feature flag for this. Version control makes it trivial to bring it back if we desire, by simply reverting this PR. In the proposal I demonstrate how and why we won't need a conversion webhook, and that there are a lot of (unsafe) misconceptions about how the webhook works.
> 
> > Overall, I'd like to be more cautious on making such changes without a definitive way to repro and verify the fix while addressing future extensibility.

@steeling Thanks for sharing your proposal, looking at it now.

Also please share a repro script and link it to the bug if possible.

dockerfiles/Dockerfile.osm-crds
```diff
@@ -2,9 +2,13 @@ FROM --platform=$BUILDPLATFORM busybox:1.33 AS builder
 ARG TARGETPLATFORM
 # Talking to the internet in an arm64 container doesn't seem to work from a
 # amd64 Mac, so download the kubectl binary in a stage running the native arch.
-RUN wget https://dl.k8s.io/release/v1.23.5/bin/$TARGETPLATFORM/kubectl -O /bin/kubectl && \
+RUN wget https://dl.k8s.io/release/v1.22.2/bin/$TARGETPLATFORM/kubectl -O /bin/kubectl && \
     chmod +x /bin/kubectl
 FROM busybox:1.33
 COPY --from=builder /bin/kubectl /bin
+COPY --from=builder /bin/sh /bin/sh
```
jsturtevant, line 9: Coping in a shell doesn't seem like the best long-term solution here.

dockerfiles/Dockerfile.osm-crds
```diff
@@ -2,9 +2,13 @@ FROM --platform=$BUILDPLATFORM busybox:1.33 AS builder
 ARG TARGETPLATFORM
 # Talking to the internet in an arm64 container doesn't seem to work from a
 # amd64 Mac, so download the kubectl binary in a stage running the native arch.
-RUN wget https://dl.k8s.io/release/v1.23.5/bin/$TARGETPLATFORM/kubectl -O /bin/kubectl && \
+RUN wget https://dl.k8s.io/release/v1.22.2/bin/$TARGETPLATFORM/kubectl -O /bin/kubectl && \
```
jsturtevant, line 5: Why dropping the kubectl version here?

charts/osm/templates/osm-deployment.yaml
```diff
@@ -36,10 +36,10 @@ spec:
       {{- if not (.Capabilities.APIVersions.Has "security.openshift.io/v1") }}
       {{- include "restricted.securityContext" . | nindent 6 }}
       {{- end }}
-      initContainers:
-        - name: init-osm-controller
-          image: {{ .Values.osm.curlImage }}
-          command: ["curl", "http://osm-bootstrap.{{ include "osm.namespace" . }}.svc.cluster.local:9091/healthz", "--connect-timeout", "2", "--retry", "50", "--retry-connrefused", "--retry-delay", "5"]
+      # initContainers:
```
steeling, line 39: this init container needs to stay, it's actually the part that checks to ensure the bootstrap pod has run successfully

cmd/osm-bootstrap/crds/patch_crds.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
```
steeling, line 1: nit: delete this script

cmd/osm-bootstrap/osm-bootstrap.go
```diff
@@ -175,6 +172,50 @@ func main() {
 		namespace:    osmNamespace,
 	}
 
+	crd_files, err := filepath.Glob("/osm-crds/*.yaml")
```
steeling, line 175: looks like this file needs to be formatted

cmd/osm-bootstrap/osm-bootstrap.go
```diff
@@ -175,6 +172,50 @@ func main() {
...
+			log.Fatal().Err(err).Msgf("Error reading CRD file %s", file)
+		}
+
+		crd := &apiv1.CustomResourceDefinition{}
+		_, _, err = decode(yaml, nil, crd)
+		if err != nil {
+			log.Fatal().Err(err).Msgf("Error decoding CRD file %s", file)
+		}
+
+		crd.Labels[constants.ReconcileLabel] = fmt.Sprintf("%t", enableReconciler)
```
steeling, line 197: nit: use strconv.FormatBool for parity with the reconciler

cmd/osm-bootstrap/osm-bootstrap.go
```diff
@@ -175,6 +172,50 @@ func main() {
...
+
+		crdExisting, err := crdClient.CustomResourceDefinitions().Get(context.Background(), crd.Name, metav1.GetOptions{})
```
steeling, line 199: need to check if the err is not nil and not an NotFound error

cmd/osm-bootstrap/osm-bootstrap.go
```diff
@@ -175,6 +172,50 @@ func main() {
...
+		if apierrors.IsNotFound(err) {
+			if _, err = crdClient.CustomResourceDefinitions().Create(context.Background(), crd, metav1.CreateOptions{}); err != nil {
```
steeling, line 201: we need to apply the ApplyAnnotation, which enables future apply's, utils.

```
			if err := util.CreateApplyAnnotation(newCRD, unstructured.UnstructuredJSONScheme); err != nil {

```

cmd/osm-bootstrap/osm-bootstrap.go
```diff
@@ -175,6 +172,50 @@ func main() {
...
+				log.Fatal().Err(err).Msgf("Error creating crd : %s", crd.Name)
+			}
+			log.Info().Msgf("Successfully created crd: %s", crd.Name)
+		} else {
+			log.Info().Msgf("Patching conversion webhook configuration for crd: %s, setting to \"None\"", crd.Name)
+
+			crdExisting.Spec = crd.Spec
```
steeling, line 208: nit: let's set the reconciler label here for posterity, and in case it has changed

cmd/osm-bootstrap/osm-bootstrap.go
steeling, line 175: nit: let's put this whole thing into a separate function

cmd/osm-bootstrap/osm-bootstrap.go
```diff
@@ -203,48 +244,9 @@ func main() {
 		metricsstore.DefaultMetricsStore.ErrCodeCounter,
 		metricsstore.DefaultMetricsStore.HTTPResponseTotal,
```
steeling, line 245: As mentioned on the call, can you bring the reconciler to above the httpServer, and log fatal on an error starting that?

keithmattix (approved):

steeling (approved):

PR #5072, opened 2022-09-02
switch back to wellknown http_connection_manager name

steeling:
This is purely to make this PR smaller https://github.com/openservicemesh/osm/pull/5056. Once this is in, I can rebase into the other to reduce the diff size and make review a bit easier

Necessary because of https://github.com/envoyproxy/go-control-plane/pull/589

### DIFF
--- a/pkg/envoy/types.go
+++ b/pkg/envoy/types.go
@@ -3,6 +3,8 @@
 package envoy
 
 import (
+	"github.com/envoyproxy/go-control-plane/pkg/wellknown"
+
 	"github.com/openservicemesh/osm/pkg/logger"
 )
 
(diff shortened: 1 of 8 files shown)


pkg/envoy/types.go
```diff
@@ -90,7 +90,7 @@ const (
 // *Note: HTTP typed filters referenced in RDS require a wellknown name
 const (
 	// HTTP filters
-	HTTPConnectionManagerFilterName = "http_connection_manager"
+	HTTPConnectionManagerFilterName = "envoy.filters.network.http_connection_manager"
```
keithmattix, line 93: Should this reference the wellknown names package?

steeling: done!

jaellio (approved):

keithmattix (approved):

Issue #5040, opened 2022-08-25
Remove MRCCompatClient


jaellio:
Part of #4998 

The MRCCompatClient was created to support the transition to using a MeshRootCertificate custom resource to configure the certificate provider on osm install and facilitate OSM root certificate rotation. As specified in #4718, the MRCCompatClient allowed us to support a single implementation of certificate.Manager regardless of whether or not an MRC resource was used to configure the certificate provider. If the `enableMeshRootCertificate` binary flag was set to false an MRCCompatClient was created.  

In v1.3 a MeshRootCertificate resource will be created on install by default and the MRCCompatClient is no longer needed. This issue tracks the removal of the MRCCompatClient.

Resources:
- MRCCompatClient added in #4718 
-  #4816 
- `enableMeshRootCertificate` flag added in #4871 

The removal of the `enableMeshRootCertificate` flag is tracked in #5041.

jaellio:
Blocked on:
- #4818 
- #4723 
- #4815 
- #4962

ceekay47:
Hi @jaellio , I would to like work on this. Can you please help me get started? Thanks!

jaellio:
HI @ceekay47! Thank you so much for your interest in picking up this issue! This change is currently blocked on several other issues. Let me see if there is another good first issue that is currently unblocked.

ceekay47:
Sure @jaellio. That would be really awesome. Thanks!

PR #5073, opened 2022-09-02
feat(certs): define loops to reconcile MRC status

### LINKED ISSUE #5046: Create pattern for spawning level-triggered reconciliation loop for certs
Many steps in the MRC certificate rotation [process](https://docs.google.com/document/d/1QF7wL1su4CUzwo3Yo2_vFL90XG2d-wUhBU6uifNrkws/edit#heading=h.sy0tyhbqfylp) requires the spawning/management of independent reconciliation loops based on certain combinations of MRC conditions. To reduce the cognitive overhead of re-writing these loops for each step, there should be a general pattern for starting/managing the level-triggered reconciliation loops. 

Implementation is TBD, but consider aptly-named functions containing the infinite loops that can be spawned in separate goroutines as the requisite events come in. Error handling, context termination, and stauses updates will likely look similar between different steps, so those may be able to be abstracted out

jaellio:
**Description**:

Resolves #5046 

Creates an `MRCReconciler` to define how to check and update the status of a specified MRC. There are multiple reconciliation loops that will be created during root certificate rotation. The details for the following loops can be found [here](https://docs.google.com/document/d/1QF7wL1su4CUzwo3Yo2_vFL90XG2d-wUhBU6uifNrkws/edit#heading=h.tix4uddk7ofj).

1. Watch for all `status.CertificateStatus` fields to be **Validating**
2. Watch for all `status.CertificateStatus` fields to be **Issuing**
3. Watch for the conditions `Provisioning`, `IssuingRollout`, and `ValidatingRollout` to be **True**

This PR contains no functional changes and sets up future changes to define the check and update functions required for the above MRC reconciliation loops.

**Testing done**:
- CI
- unit tests

**Affected area**:

5. Is this a breaking change? No

6. Has documentation corresponding to this change been updated in the [osm-docs](https://github.com/openservicemesh/osm-docs) repo (if applicable)? No

### DIFF
--- a/cmd/osm-controller/osm-controller.go
+++ b/cmd/osm-controller/osm-controller.go
@@ -215,7 +215,7 @@ func main() {
 		}
 	} else {
 		certManager, err = providers.NewCertificateManager(ctx, kubeClient, kubeConfig, osmNamespace,
-			certOpts, k8sClient, 5*time.Second, trustDomain)
+			certOpts, computeClient, 5*time.Second, trustDomain)
 		if err != nil {
 			events.GenericEventRecorder().FatalEvent(err, events.InvalidCertificateManager,
 				"Error fetching certificate manager of kind %s", certProviderKind)
--- a/cmd/osm-controller/osm-controller.go
+++ b/cmd/osm-controller/osm-controller.go
@@ -197,6 +197,8 @@ func main() {
 
 	meshSpec := smi.NewSMIClient(informerCollection, osmNamespace, k8sClient, msgBroker)
 
+	computeClient := kube.NewClient(k8sClient)
+
 	certOpts, err := getCertOptions()
 	if err != nil {
 		log.Fatal().Err(err).Msg("Error getting certificate options")
@@ -206,7 +208,7 @@ func main() {
 	var certManager *certificate.Manager
 	if enableMeshRootCertificate {
 		certManager, err = providers.NewCertificateManagerFromMRC(ctx, kubeClient, kubeConfig, osmNamespace,
-			certOpts, k8sClient, informerCollection, 5*time.Second)
+			certOpts, computeClient, informerCollection, 5*time.Second)
 		if err != nil {
 			events.GenericEventRecorder().FatalEvent(err, events.InvalidCertificateManager,
 				"Error fetching certificate manager of kind %s from MRC", certProviderKind)
@@ -219,8 +221,6 @@ func main() {
 				"Error fetching certificate manager of kind %s", certProviderKind)
 		}
 	}
-
-	computeClient := kube.NewClient(k8sClient)
 
 	ingress.Initialize(kubeClient, k8sClient, stop, certManager, msgBroker)
 
--- a/pkg/certificate/fake_manager.go
+++ b/pkg/certificate/fake_manager.go
@@ -9,6 +9,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"github.com/openservicemesh/osm/pkg/apis/config/v1alpha2"
+	fakeConfigClientset "github.com/openservicemesh/osm/pkg/gen/client/config/clientset/versioned/fake"
 
 	"github.com/openservicemesh/osm/pkg/certificate/pem"
 	"github.com/openservicemesh/osm/pkg/constants"
@@ -18,7 +19,9 @@ var (
 	validity = time.Hour
 )
 
-type fakeMRCClient struct{}
+type fakeMRCClient struct {
+	fakeConfigClient *fakeConfigClientset.Clientset
+}
 
 func (c *fakeMRCClient) GetCertIssuerForMRC(mrc *v1alpha2.MeshRootCertificate) (Issuer, pem.RootCertificate, error) {
 	return &fakeIssuer{}, pem.RootCertificate("rootCA"), nil
@@ -32,6 +35,19 @@ func (c *fakeMRCClient) List() ([]*v1alpha2.MeshRootCertificate, error) {
 			TrustDomain: "fake.domain.com",
 		},
 	}}, nil
+}
+
+func (c *fakeMRCClient) Get(name string) *v1alpha2.MeshRootCertificate {
+	mrc, _ := c.fakeConfigClient.ConfigV1alpha2().MeshRootCertificates("osm-system").Get(context.Background(), name, metav1.GetOptions{})
+	return mrc
+}
+
+func (c *fakeMRCClient) Update(obj *v1alpha2.MeshRootCertificate) (*v1alpha2.MeshRootCertificate, error) {
+	return c.fakeConfigClient.ConfigV1alpha2().MeshRootCertificates("osm-system").Update(context.Background(), obj, metav1.UpdateOptions{})
+}
+
+func (c *fakeMRCClient) UpdateStatus(obj *v1alpha2.MeshRootCertificate) (*v1alpha2.MeshRootCertificate, error) {
+	return c.fakeConfigClient.ConfigV1alpha2().MeshRootCertificates("osm-system").UpdateStatus(context.Background(), obj, metav1.UpdateOptions{})
 }
 
 func (c *fakeMRCClient) Watch(ctx context.Context) (<-chan MRCEvent, error) {
--- a/pkg/certificate/fake_manager.go
+++ b/pkg/certificate/fake_manager.go
@@ -9,7 +9,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"github.com/openservicemesh/osm/pkg/apis/config/v1alpha2"
-	fakeConfigClientset "github.com/openservicemesh/osm/pkg/gen/client/config/clientset/versioned/fake"
+	"github.com/openservicemesh/osm/pkg/compute"
 
 	"github.com/openservicemesh/osm/pkg/certificate/pem"
 	"github.com/openservicemesh/osm/pkg/constants"
@@ -20,34 +20,11 @@ var (
 )
 
 type fakeMRCClient struct {
-	fakeConfigClient *fakeConfigClientset.Clientset
+	compute.Interface
 }
 
 func (c *fakeMRCClient) GetCertIssuerForMRC(mrc *v1alpha2.MeshRootCertificate) (Issuer, pem.RootCertificate, error) {
 	return &fakeIssuer{}, pem.RootCertificate("rootCA"), nil
-}
-
-// List returns the single, pre-generated MRC. It is intended to implement the certificate.MRCClient interface.
-func (c *fakeMRCClient) List() ([]*v1alpha2.MeshRootCertificate, error) {
-	// return single empty object in the list.
-	return []*v1alpha2.MeshRootCertificate{{
-		Spec: v1alpha2.MeshRootCertificateSpec{
-			TrustDomain: "fake.domain.com",
-		},
-	}}, nil
-}
-
-func (c *fakeMRCClient) Get(name string) *v1alpha2.MeshRootCertificate {
-	mrc, _ := c.fakeConfigClient.ConfigV1alpha2().MeshRootCertificates("osm-system").Get(context.Background(), name, metav1.GetOptions{})
-	return mrc
-}
-
-func (c *fakeMRCClient) Update(obj *v1alpha2.MeshRootCertificate) (*v1alpha2.MeshRootCertificate, error) {
-	return c.fakeConfigClient.ConfigV1alpha2().MeshRootCertificates("osm-system").Update(context.Background(), obj, metav1.UpdateOptions{})
-}
-
-func (c *fakeMRCClient) UpdateStatus(obj *v1alpha2.MeshRootCertificate) (*v1alpha2.MeshRootCertificate, error) {
-	return c.fakeConfigClient.ConfigV1alpha2().MeshRootCertificates("osm-system").UpdateStatus(context.Background(), obj, metav1.UpdateOptions{})
 }
 
 func (c *fakeMRCClient) Watch(ctx context.Context) (<-chan MRCEvent, error) {
--- a/pkg/certificate/providers/compat.go
+++ b/pkg/certificate/providers/compat.go
@@ -14,21 +14,6 @@ func (c *MRCCompatClient) List() ([]*v1alpha2.MeshRootCertificate, error) {
 	}, nil
 }
 
-// Get returns the pre-generated MRC.
-func (c *MRCCompatClient) Get(name string) *v1alpha2.MeshRootCertificate {
-	return c.mrc
-}
-
-// UpdateStatus returns the updated MRC. Should never be called when using the MRCCompatClient.
-func (c *MRCCompatClient) UpdateStatus(obj *v1alpha2.MeshRootCertificate) (*v1alpha2.MeshRootCertificate, error) {
-	return c.mrc, nil
-}
-
-// Update returns the updated MRC. Should never be called when using the MRCCompatClient.
-func (c *MRCCompatClient) Update(obj *v1alpha2.MeshRootCertificate) (*v1alpha2.MeshRootCertificate, error) {
-	return c.mrc, nil
-}
-
 // Watch is a basic Watch implementation for the MRC attached to the compat client
 func (c *MRCCompatClient) Watch(ctx context.Context) (<-chan certificate.MRCEvent, error) {
 	ch := make(chan certificate.MRCEvent)
--- a/pkg/certificate/providers/compat.go
+++ b/pkg/certificate/providers/compat.go
@@ -14,6 +14,21 @@ func (c *MRCCompatClient) List() ([]*v1alpha2.MeshRootCertificate, error) {
 	}, nil
 }
 
+// Get returns the pre-generated MRC.
+func (c *MRCCompatClient) Get(name string) *v1alpha2.MeshRootCertificate {
+	return c.mrc
+}
+
+// UpdateStatus returns the updated MRC. Should never be called when using the MRCCompatClient.
+func (c *MRCCompatClient) UpdateStatus(obj *v1alpha2.MeshRootCertificate) (*v1alpha2.MeshRootCertificate, error) {
+	return c.mrc, nil
+}
+
+// Update returns the updated MRC. Should never be called when using the MRCCompatClient.
+func (c *MRCCompatClient) Update(obj *v1alpha2.MeshRootCertificate) (*v1alpha2.MeshRootCertificate, error) {
+	return c.mrc, nil
+}
+
 // Watch is a basic Watch implementation for the MRC attached to the compat client
 func (c *MRCCompatClient) Watch(ctx context.Context) (<-chan certificate.MRCEvent, error) {
 	ch := make(chan certificate.MRCEvent)
--- a/pkg/certificate/reconciler.go
+++ b/pkg/certificate/reconciler.go
@@ -1,0 +1,48 @@
+package certificate
+
+import (
+	"context"
+	"time"
+)
+
+// CheckAndUpdate checks for a MRC to reach a desired condition and when the condition is met,
+// it updates the status and terminates the loop
+func (r *MRCReconciler) CheckAndUpdate(ctx context.Context, mrcClient MRCClient, checkInterval time.Duration) {
+	ticker := time.NewTicker(checkInterval)
+	go func() {
+		for {
+			select {
+			case <-ctx.Done():
+				ticker.Stop()
+				log.Info().Msgf("Stopping status reconciliation loop for MRC %s", r.mrcName)
+				return
+			case <-ticker.C:
+				mrc := mrcClient.Get(r.mrcName)
+				if mrc == nil {
+					log.Error().Msgf("failed to get MRC %s in status reconciliation loop", r.mrcName)
+					continue
+				}
+
+				meetsCondition, err := r.checkStatus(ctx, mrc)
+				// checkStatus will return an error if the MRC is no longer in the expected rotation state
+				// i.e. prior to this reconciliation loop, another component has already moved the rotation
+				// to the next state after seeing the valid status condition without this check, the MRC
+				// reconciliation loop might never terminate
+				if err != nil {
+					return
+				}
+				if !meetsCondition {
+					continue
+				}
+
+				// update status does not need to implement its own retry logic. Updating the status will
+				// be retried in the MRC reconciler's next iteration
+				if err := r.updateStatus(ctx, mrc); err != nil {
+					log.Warn().Err(err).Msgf("failed to update status of MRC %s in status reconciliation loop. Condition was met", r.mrcName)
+					continue
+				}
+				return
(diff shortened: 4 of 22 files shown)


cmd/osm-controller/osm-controller.go
```diff
@@ -197,10 +197,12 @@ func main() {
 		events.GenericEventRecorder().FatalEvent(err, events.InitializationError, "Error creating informer collection")
 	}
 
-	k8sClient := k8s.NewClient(osmNamespace, osmMeshConfigName, informerCollection, policyClient, msgBroker)
+	k8sClient := k8s.NewClient(osmNamespace, osmMeshConfigName, informerCollection, policyClient, configClient, msgBroker)
```
keithmattix, line 200: I think for the time being, we're looking to leverage the informer cache and reduce dependencies on raw clients. Originally, I was worried about the informer cache being stale, but I realized that all of our k8s query methods leveraged the informer cache anyway. Does that make sense?

pkg/certificate/providers/compat.go
```diff
@@ -14,6 +14,21 @@ func (c *MRCCompatClient) List() ([]*v1alpha2.MeshRootCertificate, error) {
 	}, nil
 }
 
+// Get returns the pre-generated MRC.
+func (c *MRCCompatClient) Get(name string) *v1alpha2.MeshRootCertificate {
+	return c.mrc
+}
+
+// UpdateStatus returns the updated MRC. Should never be called when using the MRCCompatClient.
```
jaellio, line 22: TODO(#5040)

cmd/osm-controller/osm-controller.go
jaellio: I updated `GetMeshRootCertificate` to use the informer cache. For `UpdateMeshRootCertificate` and `UpdateMeshRootCertificateStatus` I am still utilizing the raw config client.

pkg/certificate/reconciler.go
```diff
@@ -0,0 +1,35 @@
...
+				}
+
+				if !r.checkStatus(ctx, mrc) {
+					continue
+				}
+
+				if err := r.updateStatus(ctx, mrc); err != nil {
+					log.Warn().Err(err).Msgf("failed to update status of MRC %s in status reconciliation loop. Condition was met", r.mrcName)
+				}
+				return
```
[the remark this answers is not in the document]
jaellio: Do you mean the return statement? I believe so. I could use `break` here instead.

pkg/certificate/reconciler.go
```diff
@@ -0,0 +1,35 @@
...
+	go func() {
+		for {
+			select {
+			case <-ctx.Done():
+				ticker.Stop()
+				log.Info().Msgf("Stopping status reconciliation loop for MRC %s", r.mrcName)
+				return
+			case <-ticker.C:
+				mrc := mrcClient.Get(r.mrcName)
+				if mrc != nil {
```
keithmattix, line 20:
```suggestion
				if mrc == nil {
```

Also, we probably want to `continue` after this step....we need to decide how to handle errors that occur at this level. With a kubernetes error like this (while retrieving an MRC), I'm inclined to just keep retrying, but I'm interested in others' opinions

pkg/certificate/reconciler.go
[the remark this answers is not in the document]
keithmattix: Yep; it exits the whole goroutine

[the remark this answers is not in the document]
jsturtevant: oh, I see this is intentional, If `checkStatus()` passes then you want to update and stop the loop.  I was unclear on the usage.  A comment here or on `CheckMRCStatus` that say if this returns true the loop will complete

pkg/certificate/types.go
```diff
@@ -153,3 +156,16 @@ type MRCEventBroker interface {
...
+
+// UpdateMRCStatus updates the status of an MRC to the desired state
+type UpdateMRCStatus func(ctx context.Context, mrc *v1alpha2.MeshRootCertificate) error
+
+// CheckMRCStatus checks if the MRC status is in the desired state
+type CheckMRCStatus func(ctx context.Context, mrc *v1alpha2.MeshRootCertificate) bool
+
+// MRCReconciler defines the actions of a reconciler used to determine when and how to update an MRC's status
+type MRCReconciler struct {
+	mrcName      string
```
keithmattix, line 168: Shouldn't these fields be exported?

pkg/k8s/client.go
```diff
@@ -564,3 +567,42 @@ func (c *Client) GetUpstreamTrafficSetting(namespace *types.NamespacedName) *pol
...
+}
+
+// UpdateMeshRootCertificateStatus updates the status of a MeshRootCertificate.
+func (c *Client) UpdateMeshRootCertificateStatus(obj *configv1alpha2.MeshRootCertificate) (*configv1alpha2.MeshRootCertificate, error) {
+	return c.configClient.ConfigV1alpha2().MeshRootCertificates(c.osmNamespace).UpdateStatus(context.Background(), obj, metav1.UpdateOptions{})
+}
+
+// ListMeshRootCertificates returns the MRCs stored in the informerCollection's store
+func (c *Client) ListMeshRootCertificates() ([]*configv1alpha2.MeshRootCertificate, error) {
+	// informers return slice of pointers so we'll convert them to value types before returning
```
keithmattix, line 593:
```suggestion
	// informers return slice of generic, essentially untyped pointers so we'll convert them to value types before returning
```

pkg/certificate/reconciler.go
```diff
@@ -0,0 +1,35 @@
+package certificate
+
+import (
+	"context"
+	"time"
+)
+
+// CheckAndUpdate checks for a MRC to reach a desired condition and updates the status when the condition is met
+func (r *MRCReconciler) CheckAndUpdate(ctx context.Context, mrcClient MRCClient, checkInterval time.Duration) {
```
keithmattix, line 9: We need a test for this; let me know if you need help setting the test up.

keithmattix (changes requested):
This is really close! Thanks for your work!

pkg/certificate/reconciler.go
```diff
@@ -0,0 +1,35 @@
...
+				if mrc != nil {
+					log.Error().Msgf("failed to get MRC %s in status reconciliation loop", r.mrcName)
+				}
+
+				if !r.checkStatus(ctx, mrc) {
+					continue
+				}
+
+				if err := r.updateStatus(ctx, mrc); err != nil {
+					log.Warn().Err(err).Msgf("failed to update status of MRC %s in status reconciliation loop. Condition was met", r.mrcName)
```
jsturtevant, line 29: What happens in this case?  Are there any safe guards to ensure the loop gets restarted?  Is it left the implementor of UpdateMRCStatus to ensure proper number of retries are completed?

pkg/certificate/reconciler.go
jaellio: Oops, yep, this should be `==` rather than `!=`

pkg/certificate/reconciler.go
jaellio: Added a unit test!

pkg/certificate/reconciler.go
jaellio: Per our offline conversation, I have added an error return value to `checkStatus`. An error will be returned if the mrc reconciler loop should exit. For example, the loop should terminate if another control plane replica executing the same loop, sees the condition is met and updates the status. The other loops waiting for the same condition may never see the condition in the desired state, but they will see the status as having already been updated, signifying the loop can terminate. A false bool and nil error will be returned if the condition has not been met but the check should be retried.

pkg/certificate/types.go
jaellio: I had envisioned only creating the `MRCReconciler` in the certificate package. Does it make more sense for these fields to be exported?

pkg/certificate/reconciler.go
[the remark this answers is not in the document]
jaellio: Added a clarifying comment!

pkg/certificate/fake_manager.go
```diff
@@ -18,7 +19,9 @@ var (
 	validity = time.Hour
 )
 
-type fakeMRCClient struct{}
+type fakeMRCClient struct {
```
steeling, line 22: Can we get rid of this whole thing? There's one place where it's used, which can be easily switched to the `tresorFake.NewFake`

pkg/certificate/providers/compat.go
```diff
@@ -14,6 +14,21 @@ func (c *MRCCompatClient) List() ([]*v1alpha2.MeshRootCertificate, error) {
...
+func (c *MRCCompatClient) UpdateStatus(obj *v1alpha2.MeshRootCertificate) (*v1alpha2.MeshRootCertificate, error) {
```
steeling, line 23: these should return errors, since if we're calling it it's an error. We should not enable the event handler if this is being used

pkg/k8s/client.go
```diff
@@ -564,3 +567,42 @@ func (c *Client) GetUpstreamTrafficSetting(namespace *types.NamespacedName) *pol
...
+	return c.configClient.ConfigV1alpha2().MeshRootCertificates(c.osmNamespace).Update(context.Background(), obj, metav1.UpdateOptions{})
+}
+
+// UpdateMeshRootCertificateStatus updates the status of a MeshRootCertificate.
+func (c *Client) UpdateMeshRootCertificateStatus(obj *configv1alpha2.MeshRootCertificate) (*configv1alpha2.MeshRootCertificate, error) {
+	return c.configClient.ConfigV1alpha2().MeshRootCertificates(c.osmNamespace).UpdateStatus(context.Background(), obj, metav1.UpdateOptions{})
+}
+
+// ListMeshRootCertificates returns the MRCs stored in the informerCollection's store
+func (c *Client) ListMeshRootCertificates() ([]*configv1alpha2.MeshRootCertificate, error) {
```
steeling, line 592: suggestion: Let's add another method: `AddMRCEventsHandler`, and remove the informersCollection from the MRCComposer

pkg/certificate/types.go
```diff
@@ -122,6 +122,9 @@ type Manager struct {
 // MRCClient is an interface that can watch for changes to the MRC. It is typically backed by a k8s informer.
 type MRCClient interface {
 	List() ([]*v1alpha2.MeshRootCertificate, error)
```
steeling, line 124: We could also change these methods to resemble the methods that are on the KubeClient, which would allow us to just use the compute interface directly

pkg/certificate/types.go
```diff
@@ -153,3 +156,16 @@ type MRCEventBroker interface {
 	// an error if the subscription goes awry.
 	Watch(context.Context) (<-chan MRCEvent, error)
 }
+
+// UpdateMRCStatus updates the status of an MRC to the desired state
+type UpdateMRCStatus func(ctx context.Context, mrc *v1alpha2.MeshRootCertificate) error
```
steeling, line 161: Is there a strong reason to type alias these methods?

The CheckMRCStatus doesn't make any API calls, so I imagine there will only be one implementation. I'd recommend we just make that a pointer receiver.

The UpdateMRCStatus is analogous to UpdateMeshRootCertificateStatus, so that can just be embedded

pkg/certificate/types.go
steeling: prefer unexported if it's unused outside the package

pkg/certificate/reconciler.go
```diff
@@ -0,0 +1,48 @@
+package certificate
+
+import (
+	"context"
+	"time"
+)
+
+// CheckAndUpdate checks for a MRC to reach a desired condition and when the condition is met,
+// it updates the status and terminates the loop
+func (r *MRCReconciler) CheckAndUpdate(ctx context.Context, mrcClient MRCClient, checkInterval time.Duration) {
```
steeling, line 10: so it seems this, and CheckStatus are the only 2 new methods.

The certificate package is getting a little complicated with the number of types we have, and the layers of composition required. In general small building blocks are favorable, but we're starting to create a web of building blocks that on first glance is hard to grok.

I wonder if it makes more sense to put these methods on one of the other existing structs for now?

pkg/k8s/client.go
```diff
@@ -564,3 +567,42 @@ func (c *Client) GetUpstreamTrafficSetting(namespace *types.NamespacedName) *pol
 	}
 	return nil
 }
+
+// GetMeshRootCertificate returns a MeshRootCertificate resource with namespaced name
```
steeling, line 571: nit: consider adding the k8s client stuff in a separate PR to reduce the size of this one.

pkg/certificate/fake_manager.go
jaellio: I believe using `tresorFake` would result in a circular dependency.

pkg/certificate/reconciler.go
jaellio: When we get rid of the MRCCompatClient I think the certificate package will become less complex, but I agree that the MRCReconciler adds another layer of complexity.

There are several different MRC reconciliation loops. Each loop is checking for different status conditions and will make different updates to the MRC.  I had envisioned having unique implementations of `updateStatus` and `checkStatus` for each loop. The goal of the `MRCReconciler` was to define a reconciliation pattern applicable to all loops while also providing the ability to uniquely specify how to check and update the MRC. The `MRCReconciler` would be configured in response to an MRC event.

I am not sure there is a struct in `pkg/certificate` where it makes sense to put these methods. I'd say the `MRCComposer` would make sense but that would result in circular dependency if we want to call `checkAndUpdate` in `manager.go`

pkg/k8s/client.go
jaellio: Created #5127

pkg/certificate/providers/compat.go
jaellio: Removed since using the compute.Interface directly

pkg/certificate/types.go
jaellio: Updated!

pkg/k8s/client.go
jaellio: Added!

pkg/certificate/reconciler.go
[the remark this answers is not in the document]
jaellio: Added an errorcode

pkg/certificate/reconciler.go
jaellio: @steeling Let me know if this makes sense or if I should find a way to avoid a new MRCReconciler struct :)

pkg/certificate/types.go
keithmattix: From my understanding, users will pass in an arbitrary `UpdateMRCStatus` function for the reconciler to call later, thus the type alias

pkg/certificate/reconciler.go
```diff
@@ -0,0 +1,51 @@
...
+					log.Error().Msgf("failed to get MRC %s in status reconciliation loop", r.mrcName)
+					continue
+				}
+
+				meetsCondition, err := r.checkStatus(ctx, mrc)
+				// checkStatus will return an error if the MRC is no longer in the expected rotation state
+				// i.e. prior to this reconciliation loop, another component has already moved the rotation
+				// to the next state after seeing the valid status condition without this check, the MRC
+				// reconciliation loop might never terminate
+				if err != nil {
```
keithmattix, line 33: I'd like to see the ability to change behavior (e.g. log, emit a metric, etc.) based on the type of error. For the situation described in the comment (MRC is no longer in the expected rotation state), could you emit a specific error type (e.g. `ErrUnexpectedMRCStatusInReconciler`)?

pkg/certificate/reconciler_test.go
```diff
@@ -0,0 +1,216 @@
...
+	"github.com/openservicemesh/osm/pkg/tests"
+)
+
+func TestCheckAndUpdate(t *testing.T) {
+	a := tassert.New(t)
+
+	mrcClient := &fakeMRCClient{}
+	checkStatus := func(ctx context.Context, mrc *v1alpha2.MeshRootCertificate) (bool, error) {
+		if mrc.Status.State != constants.MRCStatePending {
+			return false, fmt.Errorf("incorrect rotation state")
```
keithmattix, line 29: Here's an example where "ErrUnexpectedMRCStatusInReconciler" would be used

pkg/certificate/reconciler_test.go
```diff
@@ -0,0 +1,216 @@
...
+	"github.com/openservicemesh/osm/pkg/messaging"
+	"github.com/openservicemesh/osm/pkg/tests"
+)
+
+func TestCheckAndUpdate(t *testing.T) {
+	a := tassert.New(t)
+
+	mrcClient := &fakeMRCClient{}
+	checkStatus := func(ctx context.Context, mrc *v1alpha2.MeshRootCertificate) (bool, error) {
+		if mrc.Status.State != constants.MRCStatePending {
```
keithmattix, line 28: I don't know how I feel about depending on the top-level `status.state`. The certificateStatuses are what should be informing the condition updates

pkg/certificate/reconciler_test.go
```diff
@@ -0,0 +1,216 @@
...
+				mrc.Status.ComponentStatuses.Sidecar == constants.MRCComponentStatusValidating &&
+				mrc.Status.ComponentStatuses.XDSControlPlane == constants.MRCComponentStatusValidating &&
+				mrc.Status.ComponentStatuses.Webhooks == constants.MRCComponentStatusValidating),
+			nil
+	}
+	updateStatus := func(ctx context.Context, mrc *v1alpha2.MeshRootCertificate) error {
+		mrc.Status.State = constants.MRCStateValidating
+		mrc.Status.Conditions = append(mrc.Status.Conditions, v1alpha2.MeshRootCertificateCondition{
+			Type:   constants.MRCConditionTypeValidatingRollout,
+			Status: "True",
```
keithmattix, line 42: nit: I lightly prefer a constant here

pkg/certificate/reconciler_test.go
```diff
@@ -0,0 +1,216 @@
...
+			Reason: "CertificatePassivelyInUse",
```
keithmattix, line 43: For `Reasons` we should _definietly_ have those exported as constants somewhere. Our API should have a specific set of Reasons that have very clear meanings documented somewhere

keithmattix (changes requested):
Only one blocking comment; looks good overall though

pkg/certificate/types.go
steeling: so there's 2 methods defined. To me that means it should be an interface. 

["Interfaces are named collections of method signatures."](https://gobyexample.com/interfaces)

That would further remove the need for a type alias. For simplicity, and to avoid duplication, we should define the interface such that the compute.Client/k8s.Controller adheres to that interface

pkg/certificate/reconciler.go
steeling: That sounds pretty complex. To dynamically change the method that is called based on the event type, when that method is a field on the struct, instead of being explicit about `UpdateStatusForNewReadyState`, or something similar.

That hides some "magic", and forces a new reader to dig around a bit more to discover what is happening.

What if we simply put the reconciler logic on the MRCClient?

pkg/certificate/types.go
keithmattix: These aren't "methods" though in the truest sense. These functions don't need receivers, and usually won't have receivers

pkg/certificate/fake_manager.go
steeling: but the usages (there's now 2) of this is outside of the package. Can we get rid of this, and migrate the usage of FakeCertManager to tresor.NewFake?

pkg/certificate/reconciler.go
steeling: It’s really difficult to sign off on this PR as is. The MRCReconciler is never instantiated, nor is there an interface to demonstrate how it is used. 

Instead, can we hook it into whatever calls CheckAndRotate?

A PR should come in before this that demonstrates how this is plumbed in, even if the reconciler loop does nothing. Then we can flesh out the implementation. 

Or make this a full implementation that has update logic. But right now we’ve created a middle piece that doesn’t connect on either end so it’s hard to see how the roads will meet

steeling (changes requested):

keithmattix:
@steeling I don't really agree with the conclusions you've raised here. The issue that spawned this PR (#5046) is about creating a framework for watching MRC statuses and waiting until it reaches a certain state. In general, I don't feel that issues like this demand an implementation in the same PR as the framework; the design doc for the rotation process clearly calls out a pattern of watching statuses. Preemptively creating a type to flatten that repeated logic doesn't feel like something worth blocking, even if that pattern has to change slightly.

steeling:
> @steeling I don't really agree with the conclusions you've raised here. The issue that spawned this PR (#5046) is about creating a framework for watching MRC statuses and waiting until it reaches a certain state. In general, I don't feel that issues like this demand an implementation in the same PR as the framework; the design doc for the rotation process clearly calls out a pattern of watching statuses. Preemptively creating a type to flatten that repeated logic doesn't feel like something worth blocking, even if that pattern has to change slightly.

I simply don't understand what direction this is going in to give this an adequate review. What's going to instantiate the MRCReconciler?

Can we instead have a first PR that creates and populates a reconciler object (maybe this reconciler object has a CheckAndRotate that does nothing).

This PR creates a control loop but neither starts the control loop, nor provides an implementation of any actuation within that loop. Its really difficult (for me) to understand what the intent is here

steeling:
> @steeling I don't really agree with the conclusions you've raised here. The issue that spawned this PR (#5046) is about creating a framework for watching MRC statuses and waiting until it reaches a certain state. In general, I don't feel that issues like this demand an implementation in the same PR as the framework; the design doc for the rotation process clearly calls out a pattern of watching statuses. Preemptively creating a type to flatten that repeated logic doesn't feel like something worth blocking, even if that pattern has to change slightly.

I simply don't understand what direction this is going in to give this an adequate review. What's going to instantiate the MRCReconciler?

Can we instead have a first PR that creates and populates a reconciler object (maybe this reconciler object has a CheckAndRotate that does nothing).

This PR creates a control loop but neither starts the control loop, nor provides an implementation of any actuation within that loop. Its really difficult (for me) to understand what the intent is here

pkg/certificate/fake_manager.go
jaellio: It is used in `manager_test.go` which is in this package.

jsturtevant:
I agree it was a bit hard to understand the intent here initially (for example my comment https://github.com/openservicemesh/osm/pull/5073#discussion_r967383598)  and it would be nice to see how it was being invoked. That said I would like to see these implementations make some progress and feel that refactoring could be followed up.

#5147 and this PR seems like they should be merged TBH.  Neither is fully functional without the other and together they tell a story of how these components interact to achieve the end goal.  By implementing a single slice through the whole path we can understand the various trade-offs and would help inform the overall design.  

I am in favor of the path that makes us get to the point where we have one transition is working e2e.  Which could be merging either of these and then integrating or be integrating up front.

jaellio:
Closing in favor of #5151

PR #5076, opened 2022-09-06
temporarily switch to a manual workflow for helm chart

steeling:
Steps:

1. Manually release helm chart v1.2.0 
2. Revert this PR
3. Re-release v1.2.1


shashankram:
To understand the context of this change better, it would be good to describe the motivation for the change in the PR description along with a Github issue referenced to indicate an existing problem it is attempting to circumvent.

PR #5085, opened 2022-09-07
Update gh-pages with v1.1.2 helm chart because the old version of contour is gone

keithmattix:
Since bitnami no longer hosts the bundled contour version we use, CI [fails](https://github.com/openservicemesh/osm/runs/8218093775?check_suite_focus=true) on v1.1.x of OSM. This PR manually updates the gh-pages helm repo with the v1.1.2 version of OSM


shashankram (approved):

jsturtevant (approved):

steeling (commented):
approved

PR #5086, opened 2022-09-07
change Envoy node ID to the UUID

steeling:
This is considered more stable due to trust domains being able to rotate.

This is in preparation for the switch to the envoy snapshot cache, so we don't have to parse the uuid from the cert, which is considered more fragile

### DIFF
--- a/pkg/envoy/xdsutil.go
+++ b/pkg/envoy/xdsutil.go
@@ -1,9 +1,7 @@
 package envoy
 
 import (
-	"fmt"
 	"net"
-	"strings"
 
 	xds_accesslog_filter "github.com/envoyproxy/go-control-plane/envoy/config/accesslog/v3"
 	xds_core "github.com/envoyproxy/go-control-plane/envoy/config/core/v3"
@@ -17,7 +15,6 @@ import (
 
 	configv1alpha2 "github.com/openservicemesh/osm/pkg/apis/config/v1alpha2"
 
-	"github.com/openservicemesh/osm/pkg/constants"
 	"github.com/openservicemesh/osm/pkg/envoy/secrets"
 	"github.com/openservicemesh/osm/pkg/errcode"
 	"github.com/openservicemesh/osm/pkg/identity"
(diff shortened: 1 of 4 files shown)


pkg/envoy/xdsutil.go
```diff
@@ -214,47 +211,6 @@ func GetADSConfigSource() *xds_core.ConfigSource {
...
-		"$(POD_NAME)",
-		workloadKind,
-		workloadName,
-	}
-
-	return strings.Join(items, constants.EnvoyServiceNodeSeparator)
-}
-
-// ParseEnvoyServiceNodeID parses the given Envoy service node ID and returns the encoded metadata
-func ParseEnvoyServiceNodeID(serviceNodeID string) (*PodMetadata, error) {
```
jsturtevant, line 234: looks like these functions were only used in tests?

pkg/injector/bootstrap.go
```diff
@@ -30,16 +31,16 @@ func (wh *mutatingWebhook) createEnvoyBootstrapFromExisting(newBootstrapSecretNa
...
 
-	return wh.marshalAndSaveBootstrap(newBootstrapSecretName, namespace, config, cert)
+	return wh.marshalAndSaveBootstrap(bootstrapConfigName(proxyUUID), namespace, config, cert)
 }
 
-func (wh *mutatingWebhook) createEnvoyBootstrapConfig(name, namespace, osmNamespace string, cert *certificate.Certificate, originalHealthProbes models.HealthProbes) (*corev1.Secret, error) {
+func (wh *mutatingWebhook) createEnvoyBootstrapConfig(proxyUUID uuid.UUID, namespace string, cert *certificate.Certificate, originalHealthProbes models.HealthProbes) (*corev1.Secret, error) {
 	builder := bootstrap.Builder{
-		NodeID: cert.GetCommonName().String(),
+		NodeID: proxyUUID.String(),
```
jsturtevant, line 41: Do we need to handle anything on the ADS side for the node id?

steeling: nope, we never read from it. The current snapshot cache implementation already writes the UUID, so this fixes what was previously broken on it (although it wasn't actually used anyways)

pkg/envoy/xdsutil.go
steeling: correct, they should be removed regardless :)

jsturtevant (approved):

Issue #5048, opened 2022-08-29
Missing name in tracing


piotrgwiazda:
**Bug description**:

In https://github.com/openservicemesh/osm/blob/main/pkg/injector/envoy_container.go#L43 the `--service-cluster` parameter is set based on ServiceAccount and pod.Namespace.

As a result for a simple deployment all services in a mesh are identified as "default.".
- The name component could be based on some other parameter, e.g. deployment name
- The namespace component of name is always missing, even in example tracing documentation https://release-v0-11.docs.openservicemesh.io/docs/guides/observability/tracing/

**Affected area (please mark with X where applicable)**:
- Install                    [ ]
- SMI Traffic Access Policy  [ ]
- SMI Traffic Specs Policy   [ ]
- SMI Traffic Split Policy   [ ]
- Permissive Traffic Policy  [ ]
- Ingress                    [ ]
- Egress                     [ ]
- Envoy Control Plane        [ ]
- CLI Tool                   [ ]
- Metrics                    [ ]
- Certificate Management     [ ]
- Sidecar Injection          [X ]
- Logging                    [ ]
- Debugging                  [ ]
- Tests                      [ ]
- Demo                       [ ]
- CI System                  [ ]

**Expected behavior**:

- Services are easily distinguishable in Jeager tracing by providing proper argument to `--service-cluster` parameter into Envoy.
- There is a documentation describing how to modify the naming pattern

**Steps to reproduce the bug (as precisely as possible)**:

Install OSM

**How was OSM installed?**:

As AKS add-on

**Anything else we need to know?**:

**Bug report archive**:

**Environment**:
- OSM version (use `osm version`):
- Kubernetes version (use `kubectl version`):
- Size of cluster (number of worker nodes in the cluster):
- Others:

keithmattix:
Thanks for raising these issues @piotrgwiazda! @whitneygriffith I'm assigning this to you to make sure our tracing works as expected. Feel free to use this issue to gather any more context you need

whitneygriffith:
@piotrgwiazda we were able to fix the missing name in tracing and have updated the docs which will be available in our next release which will be out in <2 weeks! If you need this fix before then feel free to build the latest [OSM](https://github.com/openservicemesh/osm)

Issue #440, opened 2020-04-10
Improve Demo : Compare with Istio BookInfo


snehachhabria:
Sophisticate the demo to cover more scenarios so that it can be compared better with Istio's BookInfo

PR #5097, opened 2022-09-07
fix(): Missing name in tracing

whitneygriffith:
Resolved missing namespace component of ClusterID

Part of #5048

Signed-off-by: Whitney Griffith <whitney.griffith16@gmail.com>

**Description**:

[Envoy Sidecar Container Spec](https://github.com/openservicemesh/osm/blob/main/pkg/injector/envoy_container.go#L43) sets the [--service-cluster](https://www.envoyproxy.io/docs/envoy/latest/operations/cli#cmdoption-service-cluster) parameter based on ServiceAccount and pod.Namespace. Namespace component is always missing, even in [example tracing documentation](https://release-v1-2.docs.openservicemesh.io/docs/guides/observability/tracing/).

**Testing done**:

- [x]  Test in AKS environment
- [x] Test in Kind environment 

**Additional Work**:

- [x] Update [example tracing documentation](https://release-v1-2.docs.openservicemesh.io/docs/guides/observability/tracing/) [PR](https://github.com/openservicemesh/osm-docs/pull/440)

**Affected area**:

2. Is this a breaking change?

### DIFF
--- a/pkg/injector/envoy_container.go
+++ b/pkg/injector/envoy_container.go
@@ -38,9 +38,10 @@ func getPlatformSpecificSpecComponents(meshConfig v1alpha2.MeshConfig, podOS str
 	return
 }
 
-func getEnvoySidecarContainerSpec(pod *corev1.Pod, meshConfig v1alpha2.MeshConfig, originalHealthProbes models.HealthProbes, podOS string) corev1.Container {
+func getEnvoySidecarContainerSpec(pod *corev1.Pod, namespace string, meshConfig v1alpha2.MeshConfig, originalHealthProbes models.HealthProbes, podOS string) corev1.Container {
 	// cluster ID will be used as an identifier to the tracing sink
-	clusterID := fmt.Sprintf("%s.%s", pod.Spec.ServiceAccountName, pod.Namespace)
+	// pod.Namespace is unset in the API request to the webhook so namespace is derived from req.Namespace
+	clusterID := fmt.Sprintf("%s.%s", pod.Spec.ServiceAccountName, namespace)
 	securityContext, containerImage := getPlatformSpecificSpecComponents(meshConfig, podOS)
 
 	logLevel := meshConfig.Spec.Sidecar.LogLevel
--- a/pkg/injector/bootstrap_test.go
+++ b/pkg/injector/bootstrap_test.go
@@ -43,6 +43,8 @@ var _ = Describe("Test functions creating Envoy bootstrap configuration", func()
 		},
 	}
 
+	namespace := "namespace"
+
 	meshConfig := v1alpha2.MeshConfig{
 		Spec: v1alpha2.MeshConfigSpec{
 			Sidecar: v1alpha2.SidecarSpec{
@@ -87,7 +89,7 @@ var _ = Describe("Test functions creating Envoy bootstrap configuration", func()
 
 	Context("test unix getEnvoySidecarContainerSpec()", func() {
 		It("creates Envoy sidecar spec", func() {
-			actual := getEnvoySidecarContainerSpec(pod, meshConfig, originalHealthProbes, constants.OSLinux)
+			actual := getEnvoySidecarContainerSpec(pod, namespace, meshConfig, originalHealthProbes, constants.OSLinux)
 
 			expected := corev1.Container{
 				Name:            constants.EnvoyContainerName,
@@ -200,7 +202,7 @@ var _ = Describe("Test functions creating Envoy bootstrap configuration", func()
 
 	Context("test Windows getEnvoySidecarContainerSpec()", func() {
 		It("creates Envoy sidecar spec", func() {
-			actual := getEnvoySidecarContainerSpec(pod, meshConfig, originalHealthProbes, constants.OSWindows)
+			actual := getEnvoySidecarContainerSpec(pod, namespace, meshConfig, originalHealthProbes, constants.OSWindows)
 
 			expected := corev1.Container{
 				Name:            constants.EnvoyContainerName,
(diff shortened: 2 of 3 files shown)


keithmattix (approved):

keithmattix (changes requested):
Can you reproduce the bug with a failing unit test?

steeling:
> Can you reproduce the bug with a failing unit test?

+1, object meta is [embedded](https://eli.thegreenplace.net/2020/embedding-in-go-part-1-structs-in-structs/), so this change looks like a no-op to me

pkg/injector/envoy_container.go
```diff
@@ -40,7 +40,7 @@ func getPlatformSpecificSpecComponents(meshConfig v1alpha2.MeshConfig, podOS str
 
 func getEnvoySidecarContainerSpec(pod *corev1.Pod, meshConfig v1alpha2.MeshConfig, originalHealthProbes models.HealthProbes, podOS string) corev1.Container {
 	// cluster ID will be used as an identifier to the tracing sink
-	clusterID := fmt.Sprintf("%s.%s", pod.Spec.ServiceAccountName, pod.Namespace)
+	clusterID := fmt.Sprintf("%s.%s", pod.Spec.ServiceAccountName, pod.ObjectMeta.Namespace)
```
shashankram, line 43: The problem here is that `pod.Namespace` is unset in the API request to the webhook, so you can't rely on it. You instead need to make use of the `namespace` variable derived from `req.Namespace` in `createPatch()` by passing it to `getEnvoySidecarContainerSpec()`.

shashankram (changes requested):

pkg/injector/envoy_container.go
keithmattix: I never knew this! Do you have a link to some docs or code that shows namespace being removed?

shashankram: This is the reason we rely on the following code:
https://github.com/openservicemesh/osm/blob/a551827f40c4ff858d1ebb938db1f6a680abe950/pkg/injector/patch.go#L28

I strongly suspect that might be the underlying cause here, as it was definitely an issue I ran into when I implemented automatic sidecar injection a couple of years ago.

keithmattix: @whitneygriffith does this make sense?

whitneygriffith: @shashankram @keithmattix this makes sense! Thanks!

whitneygriffith: Why do we unset the namespace? @shashankram

whitneygriffith:
> Can you reproduce the bug with a failing unit test?

The latest version of OSM does not have missing namespace in the name. See pictures attached.

However our [latest docs](https://docs.openservicemesh.io/docs/guides/observability/tracing/) is showing outdated images. 

Jaeger UI search for bookbuyer traces

bookbuyer interactions with bookstore-v1 and bookestore-v2

Directed acyclic graph of bookstore demo application interactions

pkg/injector/envoy_container.go
shashankram: We do not unset the namespace, but rather the namespace is unset in the request body from the API server. This is what I observed a long time ago and the reason why we use `req.Namespace` and not `pod.Namespace` for the namespace value.

whitneygriffith:
> > Can you reproduce the bug with a failing unit test?
> 
> The latest upstream version of OSM does not have missing namespace in the name. See pictures attached.
> 
> However our [latest docs](https://docs.openservicemesh.io/docs/guides/observability/tracing/) is showing outdated images. PR to fix this: [docs/tracing: update jaeger ui examples #440](https://github.com/openservicemesh/osm-docs/pull/440)
> 
> Jaeger UI search for bookbuyer traces
> 
> bookbuyer interactions with bookstore-v1 and bookestore-v2
> 
> Directed acyclic graph of bookstore demo application interactions

The current OSM version, v1.1.2 , in the AKS add on has the missing namespace  
Screen Shot 2022-09-12 at 3 35 13 PM

jsturtevant:
@whitneygriffith do I understand correctly there is not change needed on main?  Do we know what commit fixed this and needs to be backported to 1.1.X branch?  Does 1.2 have the fix?

whitneygriffith:
> @whitneygriffith do I understand correctly there is not change needed on main? Do we know what commit fixed this and needs to be backported to 1.1.X branch? Does 1.2 have the fix?

As of now, no changes are needed on main. 1.2 does not have the fix. I am confirming that main works within the aks environment before advising the customer to expect the fix in our upcoming release. 

Also need to investigate which commit fixed this.

whitneygriffith:
> > @whitneygriffith do I understand correctly there is not change needed on main? Do we know what commit fixed this and needs to be backported to 1.1.X branch? Does 1.2 have the fix?
> 
> As of now, no changes are needed on main. 1.2 does not have the fix. I am confirming that main works within the aks environment before advising the customer to expect the fix in our upcoming release.
> 
> Also need to investigate which commit fixed this.

The namespace value is missing in the aks environmen with osm built from the latest main

pkg/injector/envoy_container.go
whitneygriffith: Relying on `req.namespace` resolved the missing namespace in AKS environment

whitneygriffith:
Missing namespace is resolved with relying on `req.namespace`. PR is ready for review.

shashankram (approved):

whitneygriffith:
> Is it worth adding a test to something like https://github.com/openservicemesh/osm/blob/main/tests/e2e/e2e_metrics_test.go?

Will investigate adding a unit test to ensure this bug doesn't happen again

keithmattix (approved):

PR #5101, opened 2022-09-08
Remove pre-povisioned ADS certificate 

jsturtevant:

**Description**:

This PR:

-  removes the pre-provisioning of the ADS cert that is provisioned.  When it was pre-provisioned it ended up with the `ads.cluster.local` trust domain appended.  When rotating it would end up using this even though it was requested that the provisioned cert already included the `FullCNProvided` which would have given a cert with CN `ads`.  It was an implementation detail that the cert that was pre-provisioned ended up being re-used because our cache mechanism looks up the certs by "prefix" which was `ads`.
- ~Sets up all the internal certificates to have the trust domain as part of the CN.  Prior ADS ended up with trust domain (due to issue explained above) but webhooks didn't. I've chosen to include the Trust Domain as part of the all the certs to be consistent in naming our CN and will be necessary as we add spiffe support which requires a TrustDomain.~  Webhooks in k8s need to be in format of `service.namespace.svc` (https://github.com/openservicemesh/osm/pull/5101#discussion_r966251904)

 

**Testing done**:
Manual, unit, e2e

**Affected area**:

2. Is this a breaking change?

### DIFF
--- a/pkg/webhook/server.go
+++ b/pkg/webhook/server.go
@@ -90,6 +90,7 @@ func (s *Server) setCert() error {
 	// This is a certificate issued for the webhook handler
 	// This cert does not have to be related to the Envoy certs, but it does have to match
 	// the cert provisioned.
+	// Kubernetes requires webhooks to have format of 'servicename.namespace.svc' without trust domain
 	webhookCert, err := s.cm.IssueCertificate(
 		s.certCommonName(),
 		certificate.Internal,
--- a/cmd/osm-controller/osm-controller.go
+++ b/cmd/osm-controller/osm-controller.go
@@ -54,10 +54,6 @@ import (
 	"github.com/openservicemesh/osm/pkg/version"
 )
 
-const (
-	xdsServerCertificateCommonName = "ads"
-)
-
 var (
 	verbosity                  string
 	meshName                   string // An ID that uniquely identifies an OSM instance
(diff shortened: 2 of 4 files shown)


pkg/webhook/server.go
```diff
@@ -92,8 +92,7 @@ func (s *Server) setCert() error {
 	// the cert provisioned.
 	webhookCert, err := s.cm.IssueCertificate(
 		s.certCommonName(),
-		certificate.Internal,
-		certificate.FullCNProvided())
+		certificate.Internal)
```
keithmattix, line 95: Kubernetes webhooks [require](https://kubernetes.io/blog/2019/03/21/a-guide-to-kubernetes-admission-controllers/#tls-certificates) a specific SAN format of `service.namespace.svc` with NO trust domain attached (link mentions common name, but this was [changed](https://cloud.google.com/kubernetes-engine/docs/deprecations/webhookcompatibility) to SAN in k8s v1.23). 

I just want to make sure this change is compatible with that

jsturtevant: interesting, I was working in my local env but I will read up on this

jsturtevant: ok, looks like you are right, the tests failed with.
```
Error from server (InternalError): error when creating "STDIN": Internal error occurred: failed calling webhook "osm-validator.k8s.io": failed to call webhook: Post "https://osm-validator.osm-system.svc:9093/validate?timeout=10s": x509: certificate is valid for osm-validator.osm-system.svc.cluster.local, not osm-validator.osm-system.svc
```

Looks like when I updated the webhook locally it was still using the old cert. Will drop this part of the change

jsturtevant: updated and left a comment for future me :-)

jaellio (approved):
LGTM!

keithmattix (approved):
LGTM - it's interesting to note that the ADS server will start off having no cert initially; that's soon fixed, but something to be aware of (assuming I'm understanding this correctly

shashankram (approved):

jsturtevant:
> LGTM - it's interesting to note that the ADS server will start off having no cert initially; that's soon fixed, but something to be aware of (assuming I'm understanding this correctly

Before it starts the server it sets up certificate watches

https://github.com/openservicemesh/osm/blob/7745be0902858e48681c3ba6ac556a04a9f462d6/pkg/envoy/ads/grpc.go#L157-L164

which provisions a cert 
https://github.com/openservicemesh/osm/blob/7745be0902858e48681c3ba6ac556a04a9f462d6/pkg/envoy/ads/grpc.go#L87-L95

so it starts with a certificate 👍

keithmattix:
Oh nice perfect! Thanks for pointing this out!

Issue #4723, opened 2022-05-04
Validate MeshRootCertificate


schristoff:
Please describe the Improvement and/or Feature Request
Part of https://github.com/openservicemesh/osm/issues/4502

Add validating webhook for MeshRootCertificate that: 

- Allow at most 2 MeshRootCertificate resources in either the pending-up, pending-down, complete, inactive, error or unknown state to be created. MeshRootCertificates that are inactive are not included in the count 
    - If there are already 2 MeshRootCertificate resources in either the pending-up, pending-down, OSM will set the status to error 
- Check for the correct status settings 
- Prevent out of order status changes 
- Don’t allow operators to change certain values in existing _MeshRootCertificates_ 

Areas of Interest:
Add MeshRootCertificate CRD to validatingWebhookServer
https://github.com/openservicemesh/osm/blob/d4853664a5ee34dfaf7367b54822b307bd19d99e/pkg/validator/server.go#L55-L61

Create a validateFunc that satisfies the above requirements
https://github.com/openservicemesh/osm/blob/966405b29161ee650d01a8c0ebaa5fe4ed324b79/pkg/validator/validators.go#L26

Validations [X]

steeling:
When this gets implemented, with the new trust domain we should also validate that no parts in `strings.Split(TrustDomain, ".")` are the empty string

jaellio:
Blocked on #4962

keithmattix:
Note: Make sure the logic from the [rotation design doc](https://docs.google.com/document/d/1QF7wL1su4CUzwo3Yo2_vFL90XG2d-wUhBU6uifNrkws/edit#heading=h.r5r6bqgddlfx) is include (especially the finalizer).

PR #5105, opened 2022-09-08
feat(certs): add basic MRC validation

jaellio:
**Description**:

- adds BASIC validation of the MRC on add and update to the validating webhook
  - on add, checks for valid trust domain
  - on update, verifies trust domain and provider were not updated
- moves `Intent` field to MRC spec
- adds enum on the `Intent` field to limit possible values to `active` and `passive`

Part of #4723 

**Testing done**:
- [x] CI
- [x] manual MRC creation and updates

**Affected area**:

### DIFF
--- a/pkg/validator/patch.go
+++ b/pkg/validator/patch.go
@@ -18,6 +18,9 @@ const (
 	// ValidatingWebhookName is the name of the validating webhook.
 	ValidatingWebhookName = "osm-validator.k8s.io"
 
+	// ControlPlaneValidatingWebhookName is the name of the validating webhook for control plane resource.
+	ControlPlaneValidatingWebhookName = "osm-control-plane-validator.k8s.io"
+
 	// ValidatorWebhookSvc is the name of the validator service.
 	ValidatorWebhookSvc = "osm-validator"
 )
@@ -48,6 +51,17 @@ func createOrUpdateValidatingWebhook(clientSet kubernetes.Interface, cert *certi
 				Resources:   []string{"traffictargets"},
 			},
 		})
+	}
+
+	controlPlaneRules := []admissionregv1.RuleWithOperations{
+		{
+			Operations: []admissionregv1.OperationType{admissionregv1.Create, admissionregv1.Update},
+			Rule: admissionregv1.Rule{
+				APIGroups:   []string{"config.openservicemesh.io"},
+				APIVersions: []string{"v1alpha2"},
+				Resources:   []string{"meshrootcertificates"},
+			},
+		},
 	}
 
 	vwhcLabels := map[string]string{
@@ -104,7 +118,37 @@ func createOrUpdateValidatingWebhook(clientSet kubernetes.Interface, cert *certi
 					sideEffect := admissionregv1.SideEffectClassNoneOnDryRun
 					return &sideEffect
 				}(),
-				AdmissionReviewVersions: []string{"v1"}}},
+				AdmissionReviewVersions: []string{"v1"},
+			},
+			{
+				Name: ControlPlaneValidatingWebhookName,
+				ClientConfig: admissionregv1.WebhookClientConfig{
+					Service: &admissionregv1.ServiceReference{
+						Namespace: osmNamespace,
+						Name:      ValidatorWebhookSvc,
+						Path:      &webhookPath,
+						Port:      &webhookPort,
+					},
+					CABundle: cert.GetTrustedCAs()},
+				FailurePolicy: &failurePolicy,
+				MatchPolicy:   &matchPolicy,
+				NamespaceSelector: &metav1.LabelSelector{
+					MatchExpressions: []metav1.LabelSelectorRequirement{
+						{
+							Key:      "name",
+							Operator: metav1.LabelSelectorOpIn,
+							Values:   []string{osmNamespace},
+						},
+					},
+				},
+				Rules: controlPlaneRules,
+				SideEffects: func() *admissionregv1.SideEffectClass {
+					sideEffect := admissionregv1.SideEffectClassNoneOnDryRun
+					return &sideEffect
+				}(),
+				AdmissionReviewVersions: []string{"v1"},
+			},
+		},
 	}
 
 	if _, err := clientSet.AdmissionregistrationV1().ValidatingWebhookConfigurations().Create(context.Background(), &vwhc, metav1.CreateOptions{}); err != nil {
--- a/pkg/validator/validators.go
+++ b/pkg/validator/validators.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net"
+	"reflect"
 	"strings"
 
 	mapset "github.com/deckarep/golang-set"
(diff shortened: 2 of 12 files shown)


pkg/validator/validators.go
```diff
@@ -266,3 +268,34 @@ func (kc *policyValidator) upstreamTrafficSettingValidator(req *admissionv1.Admi
...
+		}
+
+		if !reflect.DeepEqual(oldMRC.Spec.Provider, newMRC.Spec.Provider) {
+			return nil, fmt.Errorf("cannot update certificate provider settings for MRC %s/%s. Create a new MRC and initiate root certificate rotation to update the provider", newMRC.GetNamespace(), newMRC.GetName())
+		}
+
+		if oldMRC.Spec.TrustDomain != newMRC.Spec.TrustDomain {
+			return nil, fmt.Errorf("cannot update trust domain for MRC %s/%s. Create a new MRC and initiate root certificate rotation to update the trust domain", newMRC.GetNamespace(), newMRC.GetName())
+		}
+	}
```
steeling, line 299: have you tested the semantics of preventing a deletion of an active MRC with uninstalling the cluster?

pkg/validator/validators.go
```diff
@@ -266,3 +268,34 @@ func (kc *policyValidator) upstreamTrafficSettingValidator(req *admissionv1.Admi
 
 	return nil, nil
 }
+
+func meshRootCertificateValidator(req *admissionv1.AdmissionRequest) (*admissionv1.AdmissionResponse, error) {
+	switch req.Operation {
+	case admissionv1.Create:
```
steeling, line 274: do we want to try to verify the state outside of the existing MRC, and that the system can handle the creation of a new one? Maybe add a comment on whichever we choose

pkg/validator/validators.go
```diff
@@ -266,3 +268,34 @@ func (kc *policyValidator) upstreamTrafficSettingValidator(req *admissionv1.Admi
...
+		newMRC := &configv1alpha2.MeshRootCertificate{}
+		if err := json.NewDecoder(bytes.NewBuffer(req.Object.Raw)).Decode(newMRC); err != nil {
+			return nil, err
+		}
+
+		if newMRC.Spec.TrustDomain == "" {
```
steeling, line 280: nit: add a new func validateMRC() error

We will likely want to add new validations, and it can be reused in the update as well.

steeling (approved):

pkg/validator/patch.go
```diff
@@ -104,7 +118,37 @@ func createOrUpdateValidatingWebhook(clientSet kubernetes.Interface, cert *certi
 					sideEffect := admissionregv1.SideEffectClassNoneOnDryRun
 					return &sideEffect
 				}(),
-				AdmissionReviewVersions: []string{"v1"}}},
+				AdmissionReviewVersions: []string{"v1"},
```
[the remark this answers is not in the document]
keithmattix: +1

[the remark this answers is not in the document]
jaellio: The `osm-validator` webhook has a label selector which excludes resources in the control plane namespace. I added the additional webhook for resources we want to validate that are created in the control plane namespace. Is it possible to keep the label selector and validate MRC resources using the existing webhook?

pkg/validator/patch.go
```diff
@@ -50,6 +53,17 @@ func createOrUpdateValidatingWebhook(clientSet kubernetes.Interface, cert *certi
 		})
 	}
 
+	controlPlaneRules := []admissionregv1.RuleWithOperations{
```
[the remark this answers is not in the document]
jaellio: I added the additional webhook for any resources we want to validate that are in the osm control plane namespace. I kept the rule list generic to reflect this. Happy to update this though!

pkg/validator/validators.go
jaellio: I have not, will make a note of this in the related issue if it isn't already there.

pkg/validator/patch.go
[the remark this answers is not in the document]
steeling: Gotcha! Looks like we're only validating the following resources:

```go
	Rule: admissionregv1.Rule{
		APIGroups:   []string{"policy.openservicemesh.io"},
		APIVersions: []string{"v1alpha1"},
		Resources:   []string{"ingressbackends", "egresses"},
	},
```
and
```go
	Rule: admissionregv1.Rule{
		APIGroups:   []string{"access.smi-spec.io"},
		APIVersions: []string{"v1alpha3"},
		Resources:   []string{"traffictargets"},
	},
```

are there any potential pitfalls to opening up the control plane namespace? 

This is not a blocking comment, nor should it really be a focus of this PR. Splitting them up is totally fine just wanted to sanity check. Feel free to resolve this thread :)

pkg/validator/validators.go
jsturtevant: do we also need to check that `Spec.Provider` information is set here?

pkg/validator/patch.go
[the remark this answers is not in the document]
jaellio: Yes! I think it would make sense to include the MeshConfig if there is validation we'd like to perform outside of schema enforcement.

pkg/validator/validators.go
jaellio: Good call out. I could add validation for the required string value fields

pkg/validator/patch.go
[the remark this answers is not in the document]
jaellio: That's a good point. We also have this requirement for the monitored by label:

```go
MatchLabels: map[string]string{
	constants.OSMKubeResourceMonitorAnnotation: meshName,
},
```

I wonder if there is a way to enforce that either this condition is met or another (the namespace is the `control-plane` namespace.

pkg/validator/validators.go
jaellio: Will address this in a follow up PR

pkg/validator/validators.go
jaellio: Added `Spec.Provider` validation

jsturtevant (approved):

pkg/validator/validators.go
jaellio: And more unit tests :)

pkg/validator/validators.go
```diff
@@ -266,3 +268,82 @@ func (kc *policyValidator) upstreamTrafficSettingValidator(req *admissionv1.Admi
...
+			return nil, err
+		}
+
+		err := validateMRCOnUpdate(oldMRC, newMRC)
+		return nil, err
+	}
+	return nil, nil
+}
+
+func validateMRCOnCreate(mrc *configv1alpha2.MeshRootCertificate) error {
```
keithmattix, line 297: Please add a field to prevent creating an `active` MRC if one already exists

keithmattix (commented):
Maybe I missed this elsewhere, but there was just one omission I think we need to add to the ValidatingWebhook

pkg/validator/validators.go
jaellio: I was planning on addressing more validation in a follow-up PR. When I created this PR I was blocked on https://github.com/openservicemesh/osm/pull/5127. Since I am no longer blocked on this change, I can include the validation that required the MRC methods in the k8s client in this PR.

keithmattix: Ah, if you're planning to add it in a followup PR, that's fine; just add a TODO comment and I'll approve

jaellio: @keithmattix I went ahead and added the active MRC check

keithmattix (approved):

PR #5106, opened 2022-09-09
feat(healthprobes): Allow multiple containers to use health probes

### LINKED ISSUE #4948: Support different liveness/readiness probes on more than one container in a pod

Currently the automatic health probe forwarding (described here: https://release-v1-2.docs.openservicemesh.io/docs/guides/health_checks/health_probes/) only handles automatically generating forwarding rules in the injected Envoy container for one container's configured probe within a pod. This should instead be able to handle generating configuration for forwarding liveness/readiness probes for multiple containers within a pod.

**Further Info**
Apologies in advance if I've just missed something, but it seems to be the case that - as the liveness/readiness probe URLs in container config are replaced with just `/osm-liveness-probe` and `/osm-readiness-probe` respectively, and the rewriting within the injected OSM Envoy sidecar only affects those URLs  - it's impossible to have multiple containers with different probes within a pod, as only one URL per probe is available. 

As the current handling of probes will replace every container's liveness/readiness probe URLs in their configuration, this can cause unforeseen problems where probes start talking to entirely different URLs than they're meant to when adopting OSM. I haven't validated but I think the injected OSM Envoy pod just gets configured to forward through the last container's set of probes.

- Sidecar Injection          [X]

**Possible use cases**

This would allow multiple containers within a pod to have separate liveness/readiness checks while still going through OSM.

keithmattix:
Create an Envoy cluster per (container, health probe) tuple and inject the container name into the path prefix so that OSM supports health probes from multiple containers

Fixes #4948

**Description**:
Instead of having a single Envoy cluster for each type of probe (liveness, startup, readiness), we now create an Envoy cluster per container probe. The probe rewrite results in the following containerSpec:
```yaml
- name: nginx
  image: nginx:1.19-alpine
  livenessProbe:
    httpGet:
      path: /osm-liveness-probe/nginx
      port: 15901
      scheme: HTTP
- name: httpbin
  image: kennethreitz/httpbin
  livenessProbe:
    httpGet:
      path: /osm-liveness-probe/httpbin
      port: 15901
      scheme: HTTP
```

The generated Envoy config looks like this:
```json
{
   "clusters":[
      {
         "name":"httpbin_liveness_cluster",
         "type":"STATIC",
         "load_assignment":{
            "cluster_name":"httpbin_liveness_cluster",
            "endpoints":[
               {
                  "lb_endpoints":[
                     {
                        "endpoint":{
                           "address":{
                              "socket_address":{
                                 "address":"127.0.0.1",
                                 "port_value":14001
                              }
                           }
                        }
                     }
                  ]
               }
            ]
         }
      },
      {
         "name":"nginx_liveness_cluster",
         "type":"STATIC",
         "load_assignment":{
            "cluster_name":"nginx_liveness_cluster",
            "endpoints":[
               {
                  "lb_endpoints":[
                     {
                        "endpoint":{
                           "address":{
                              "socket_address":{
                                 "address":"127.0.0.1",
                                 "port_value":80
                              }
                           }
                        }
                     }
                  ]
               }
            ]
         }
      }
   ],
   "listeners":[
      {
         "name":"liveness_listener",
         "address":{
            "socket_address":{
               "address":"0.0.0.0",
               "port_value":15901
            }
         },
         "filter_chains":[
            {
               "filters":[
                  {
                     "name":"envoy.filters.network.http_connection_manager",
                     "typed_config":{
                        "@type":"type.googleapis.com/envoy.extensions.filters.network.http_connection_manager.v3.HttpConnectionManager",
                        "stat_prefix":"health_probes_http",
                        "route_config":{
                           "name":"local_route",
                           "virtual_hosts":[
                              {
                                 "name":"local_service",
                                 "domains":[
                                    "*"
                                 ],
                                 "routes":[
                                    {
                                       "match":{
                                          "prefix":"/osm-liveness-probe/httpbin"
                                       },
                                       "route":{
                                          "cluster":"httpbin_liveness_cluster",
                                          "prefix_rewrite":"/",
                                          "timeout":"1s"
                                       }
                                    },
                                    {
                                       "match":{
                                          "prefix":"/osm-liveness-probe/nginx"
                                       },
                                       "route":{
                                          "cluster":"nginx_liveness_cluster",
                                          "prefix_rewrite":"/",
                                          "timeout":"1s"
                                       }
                                    }
                                 ]
                              }
                           ]
                        }
                     }
                  }
               ]
            }
         ]
      }
   ]
}
```

**Testing done**:
Updated Unit tests
Updated E2E tests
Manual test
<!--

3. Has documentation corresponding to this change been updated in the [osm-docs](https://github.com/openservicemesh/osm-docs) repo (if applicable)? not yet

### DIFF
--- a/pkg/envoy/bootstrap/config.go
+++ b/pkg/envoy/bootstrap/config.go
@@ -353,7 +353,7 @@ func (b *Builder) getProbeResources() ([]*xds_listener.Listener, []*xds_cluster.
 	for containerName, probes := range b.OriginalHealthProbes {
 		// Liveness probe listener + cluster
 		livenessClusterName := fmt.Sprintf("%s_%s", containerName, livenessCluster)
-		livenessListenerBuilder.AddProbe(containerName, livenessClusterName, constants.LivenessProbePath, probes.Liveness, probes.Liveness.IsHTTP)
+		livenessListenerBuilder.AddProbe(containerName, livenessClusterName, constants.LivenessProbePath, probes.Liveness)
 		livenessCluster := buildProbeCluster(livenessClusterName, probes.Liveness)
 		if livenessCluster != nil {
 			clusters = append(clusters, livenessCluster)
@@ -361,7 +361,7 @@ func (b *Builder) getProbeResources() ([]*xds_listener.Listener, []*xds_cluster.
 
 		// Readiness probe listener + cluster
 		readinessClusterName := fmt.Sprintf("%s_%s", containerName, readinessCluster)
-		readinessListenerBuilder.AddProbe(containerName, readinessClusterName, constants.ReadinessProbePath, probes.Readiness, probes.Readiness.IsHTTP)
+		readinessListenerBuilder.AddProbe(containerName, readinessClusterName, constants.ReadinessProbePath, probes.Readiness)
 		readinessCluster := buildProbeCluster(readinessClusterName, probes.Readiness)
 		if readinessCluster != nil {
 			clusters = append(clusters, readinessCluster)
@@ -369,7 +369,7 @@ func (b *Builder) getProbeResources() ([]*xds_listener.Listener, []*xds_cluster.
 
 		// Startup probe listener + cluster
 		startupClusterName := fmt.Sprintf("%s_%s", containerName, startupCluster)
-		startupListenerBuilder.AddProbe(containerName, startupClusterName, constants.StartupProbePath, probes.Startup, probes.Startup.IsHTTP)
+		startupListenerBuilder.AddProbe(containerName, startupClusterName, constants.StartupProbePath, probes.Startup)
 		startupCluster := buildProbeCluster(startupClusterName, probes.Startup)
 		if startupCluster != nil {
 			clusters = append(clusters, startupCluster)
--- a/pkg/envoy/bootstrap/config.go
+++ b/pkg/envoy/bootstrap/config.go
@@ -351,55 +351,28 @@ func (b *Builder) getProbeResources() ([]*xds_listener.Listener, []*xds_cluster.
 	}
 
 	for containerName, probes := range b.OriginalHealthProbes {
-		// Is there a liveness probe in the Pod Spec?
-		if probes.Liveness != nil && !probes.Liveness.IsTCPSocket {
-			clusterName := fmt.Sprintf("%s_%s", containerName, livenessCluster)
-			if probes.Liveness.IsHTTP {
-				livenessListenerBuilder.isHTTP = true
-				livenessListenerBuilder.virtualHostRoutes = append(livenessListenerBuilder.virtualHostRoutes, probeListenerRoute{
-					pathPrefixMatch:   fmt.Sprintf("%s/%s", constants.LivenessProbePath, containerName),
-					clusterName:       clusterName,
-					pathPrefixRewrite: probes.Liveness.Path,
-				})
-			} else {
-				// NOTE: Only 1 HTTPS probe is supported per type (liveness, readiness, startup)
-				// This is due to the fact that we passthrough HTTPS and cannot do any kind of filtering on path
-				// Therefore, the last declared HTTPS probe will be the only one receiving traffic
-				livenessListenerBuilder.httpsClusterName = clusterName
-			}
-			clusters = append(clusters, buildLivenessCluster(clusterName, probes.Liveness))
+		// Liveness probe listener + cluster
+		livenessClusterName := fmt.Sprintf("%s_%s", containerName, livenessCluster)
+		livenessListenerBuilder.AddProbe(containerName, livenessClusterName, constants.LivenessProbePath, probes.Liveness, probes.Liveness.IsHTTP)
+		livenessCluster := buildProbeCluster(livenessClusterName, probes.Liveness)
+		if livenessCluster != nil {
+			clusters = append(clusters, livenessCluster)
 		}
 
-		// Is there a readiness probe in the Pod Spec?
-		if probes.Readiness != nil && !probes.Readiness.IsTCPSocket {
-			clusterName := fmt.Sprintf("%s_%s", containerName, readinessCluster)
-			if probes.Readiness.IsHTTP {
-				readinessListenerBuilder.isHTTP = true
-				readinessListenerBuilder.virtualHostRoutes = append(readinessListenerBuilder.virtualHostRoutes, probeListenerRoute{
-					pathPrefixMatch:   fmt.Sprintf("%s/%s", constants.ReadinessProbePath, containerName),
-					clusterName:       clusterName,
-					pathPrefixRewrite: probes.Readiness.Path,
-				})
-			} else {
-				readinessListenerBuilder.httpsClusterName = clusterName
-			}
-			clusters = append(clusters, buildReadinessCluster(clusterName, probes.Readiness))
+		// Readiness probe listener + cluster
+		readinessClusterName := fmt.Sprintf("%s_%s", containerName, readinessCluster)
+		readinessListenerBuilder.AddProbe(containerName, readinessClusterName, constants.ReadinessProbePath, probes.Readiness, probes.Readiness.IsHTTP)
+		readinessCluster := buildProbeCluster(readinessClusterName, probes.Readiness)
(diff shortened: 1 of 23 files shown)


pkg/envoy/bootstrap/envoy_config_health_probes.go
```diff
@@ -33,25 +33,25 @@ const (
 	startupListener   = "startup_listener"
 )
 
-func getLivenessCluster(originalProbe *models.HealthProbe) *xds_cluster.Cluster {
+func getLivenessCluster(clusterName string, originalProbe *models.HealthProbe) *xds_cluster.Cluster {
```
steeling, line 36: nit: change change get to generate/build

pkg/envoy/bootstrap/envoy_config_health_probes.go
```diff
@@ -89,34 +89,32 @@ func getProbeCluster(clusterName string, port int32) *xds_cluster.Cluster {
...
+	pathPrefixRewrite string
+	timeout           time.Duration
 }
 
-func getReadinessListener(originalProbe *models.HealthProbe) (*xds_listener.Listener, error) {
-	if originalProbe == nil {
-		return nil, nil
-	}
-	return getProbeListener(readinessListener, readinessCluster, constants.ReadinessProbePath, constants.ReadinessProbePort, originalProbe)
+type probeListenerBuilder struct {
```
steeling, line 99: Having both a models.HealthProbe, and a `probeListenerBuilder` seems like an unnecessary layer of indirection. Can we reduce to one or the other?

keithmattix: `models.HealthProbe` represents a set of 3 probes: startup, listener, and readiness. The `probeListenerBuilder` builds an Envoy listener for any generic probe based on the information set on the builder. I don't see the redundancy here; could you expound your thought process?

pkg/envoy/bootstrap/envoy_config_health_probes.go
keithmattix: done!

pkg/envoy/bootstrap/envoy_config_health_probes.go
jsturtevant: It looks like `getProbeListener` was replaced with `probeListenerBuilder`.   `getProbeListener` could be reused reducing the indirection and remove some of the additional mapping above in `getProbeResources()`.  

I think the pro for that approach would be a little less mapping code in `getProbeResources`?

keithmattix: I'm not sure how `getProbeListener` could be reused here seeing as all that logic moved to `Build`?

shashankram (approved):

pkg/envoy/bootstrap/envoy_config_health_probes.go
jsturtevant: moved to build in the `probeListenerBuilder`?  I think the suggestion was to drop that but might be miss-understanding.    I could go either way personally, what is here is pretty readable but I see how it adds some additional mapping

jaellio (approved):

pkg/envoy/bootstrap/config.go
```diff
@@ -330,43 +331,106 @@ func GetValidationContextSDSConfigYAML() ([]byte, error) {
 // startup, readiness and liveness probes.
 // These will not change during the lifetime of the Pod.
 // If the original probe defined a TCPSocket action, listener and cluster objects are not configured
-// to serve that probe.
+// to serve that probe since the osm-healthcheck container that does that job is configured separately.
 func (b *Builder) getProbeResources() ([]*xds_listener.Listener, []*xds_cluster.Cluster, error) {
```
steeling, line 335: nit: this method is a little big, i might suggest separating out the listener and the cluster builders here into 2 methods

pkg/envoy/bootstrap/config.go
```diff
@@ -330,43 +331,106 @@ func GetValidationContextSDSConfigYAML() ([]byte, error) {
...
 
-	// Is there a readiness probe in the Pod Spec?
-	if b.OriginalHealthProbes.Readiness != nil && !b.OriginalHealthProbes.Readiness.IsTCPSocket {
-		listener, err := getReadinessListener(b.OriginalHealthProbes.Readiness)
-		if err != nil {
-			log.Error().Err(err).Msgf("Error getting readiness listener")
-			return nil, nil, err
+	for containerName, probes := range b.OriginalHealthProbes {
+		// Is there a liveness probe in the Pod Spec?
+		if probes.Liveness != nil && !probes.Liveness.IsTCPSocket {
```
steeling, line 355: this indicates somewhat deep knowledge of the probeListenerBuilder internals, which makes it fragile to refactoring. Instead, can we simply say `append(livenessListenerBuilder.probes, probes.Liveness)`

And defer the logic of inspecting the isHTTP to the internals of the builder?

pkg/envoy/bootstrap/config.go
```diff
@@ -330,43 +331,106 @@ func GetValidationContextSDSConfigYAML() ([]byte, error) {
...
+			clusterName := fmt.Sprintf("%s_%s", containerName, livenessCluster)
+			if probes.Liveness.IsHTTP {
+				livenessListenerBuilder.isHTTP = true
```
steeling, line 358: Should this have isHTTP set globally on the livesnessBuilder, even if there are others that are no http?

pkg/envoy/bootstrap/config.go
steeling: I think this would also really re-use some of the logic across liveness, readiness, startup

steeling (approved):

pkg/envoy/bootstrap/config.go
keithmattix: The spirit behind this logic already existed, but the idea is that TCPSocket probes don't go through this process at all since they are transformed into an HTTP probe elsewhere

pkg/envoy/bootstrap/config.go
keithmattix: Not sure what you mean; there are up to 3 different builders representing the different probes that exist in Kubernetes. As such, each builder has an `IsHTTP` field

steeling: is it possible for there to be more than 1 liveness builder in the list of `range b.OriginalHealthProbes`?

If so, the isHTTP is a global field, even if there is one HTTP liveness, and one TCP liveness

pkg/envoy/bootstrap/config.go
steeling: I think there's a misunderstanding. I'm simply saying, add a `builder,AddProbe(probes.Liveness)` to the builder, and move the logic present here (which is duplicated across 3 instances), into the actual builder code.

pkg/envoy/bootstrap/config.go
steeling: to clarify, i'm suggesting the code turns into:


```go
for containerName, probes := range b.OriginalHealthProbes {
  livenessBuilder.AddProbe(probes.Liveness)
  readinessBuilder.AddProbe(probes.Readiness)
  startupBuilder.AddProbe(probes.Startup)
}
```

You can also add the type and the container name as fields to the model.Probe, or simply add extra params to the AddProbe method

pkg/envoy/bootstrap/config.go
steeling: to clarify, i'm suggesting the code turns into:


```go
for containerName, probes := range b.OriginalHealthProbes {
  livenessBuilder.AddProbe(probes.Liveness)
  readinessBuilder.AddProbe(probes.Readiness)
  startupBuilder.AddProbe(probes.Startup)
}
```

You can also add the type and the container name as fields to the model.Probe, or simply add extra params to the AddProbe method

keithmattix: Ah, I see; agreed 👍🏾

pkg/envoy/bootstrap/config.go
keithmattix: Ah, you're absolutely right! Great catch; what I'll do is add a TLS inspector so that Envoy can detect HTTPS traffic

PR #5109, opened 2022-09-12
Abstract control plane logic from envoy config generation and serving

### LINKED ISSUE #5141: Decouple envoy config serving, generation, and control loop
Right now, envoy config generation, serving, and the control loop -- which is the part that watches for k8s updates and schedules proxy updates on the workerpool, are all intermingled in the same package.

By separating these out, we gain finer control over generating the configs, serving the configs (serializing them to the envoy proxies), and the control loop, which could help us integrate with a non-envoy based proxy if desired. These layers of abstraction will help make the code easier to maintain, read, and reason about.

steeling:
This PR adds 3 new domain layers from the existing envoy config generation, and separates out a few more. The tl;dr is that it decouples our control plane loops (notifications, proxy registration, worker pools, etc) from config generation, and from serving those configs.

It does this by introducing a new `osm.ControlPlane` type that has callbacks for registering a newly connected proxy. This adds the proxy to our control plane loop and notifies a new ConfigGenerator type, which simply returns a generic configuration object, and finally feeds that to a new ConfigServer type, which is responsible for serving those configs to the proxies.

It further moves all of the old `NewResponse` functions to methods on the EnvoyConfigGenerator, which is now separated from the builder logic, and further separated from the serving logic. 

This has large implications for how we can support new types of proxies, proxy waypoints, or other config generation mechanisms.

Note that the new ControlPlane code accepts configs of generic types. For our use case, these will be of type `map[string][]types.Resource`. Leveraging generics instead of an interface is what allows us to decouple config generation from proxy updates, without requiring implementations to type cast the config.

There's 2 additional side effects:

1. The health probe code was only used for the envoy server, which *always* returned true. I've gone ahead and removed this code, in favor of the `health.SimpleHandler`, which performs the same function.
2. I move the scenario tests to rds_test.go (which is what its testing), to also prevent the need to export the generation of RDS specific configs

Closes #5141


cmd/osm-controller/osm-controller.go
```diff
@@ -233,10 +236,15 @@ func main() {
 	)
 
 	proxyRegistry := registry.NewProxyRegistry()
-
+	// meshCatalog, proxyRegistry, k8sClient.GetMeshConfig().Spec.Observability.EnableDebugServer, osmNamespace, , k8sClient, msgBroker
```
keithmattix, line 239: Can this comment be removed?

pkg/osm/control_plane.go
```diff
@@ -0,0 +1,53 @@
...
+	"github.com/openservicemesh/osm/pkg/envoy/registry"
+	"github.com/openservicemesh/osm/pkg/messaging"
+	"github.com/openservicemesh/osm/pkg/workerpool"
+)
+
+const (
+	workerPoolSize = 0
+)
+
+type ProxyConfigServer[T any] interface {
```
keithmattix, line 18: Could you add some comments (in the code or PR description) about what these types represent and why they're generic

pkg/envoy/server/cache.go
```diff
@@ -0,0 +1,96 @@
...
+
+// OnFetchRequest is called when a fetch request is received
+func (s *Server) OnFetchRequest(_ context.Context, req *discovery.DiscoveryRequest) error {
+	// Unimplemented
+	return errors.New("Unsupported XDS server connection type")
+}
+
+// OnFetchResponse is called when a fetch request is being responded to
+func (s *Server) OnFetchResponse(req *discovery.DiscoveryRequest, resp *discovery.DiscoveryResponse) {
+	// Unimplemented
```
keithmattix, line 44: I'd prefer an explicit error here as well for consistency's sake

pkg/envoy/server/cache.go
```diff
@@ -0,0 +1,96 @@
...
+}
+
+// --- Delta stream types below. Callback interfaces still requires these to be defined
+
+// OnDeltaStreamOpen is called when a Delta stream is being opened
+func (s *Server) OnDeltaStreamOpen(ctx context.Context, streamID int64, typ string) error {
+	log.Debug().Msgf("OnDeltaStreamOpen id: %d typ: %s", streamID, typ)
+	return s.callbacks.ProxyConnected(ctx, streamID)
```
keithmattix, line 52: Do we support delta xDS? Does the snapshot cache give it to us for free?

pkg/envoy/server/cache.go
```diff
@@ -0,0 +1,96 @@
...
+	log.Debug().Msgf("OnStreamDeltaRequest node: %s, type: %s, nonce: %s, resNames: %s", req.Node.Id, req.TypeUrl, req.ResponseNonce, req.GetResourceNamesSubscribe())
+	return nil
+}
+
+// OnStreamDeltaResponse is called when a Delta request is getting responded to
+func (s *Server) OnStreamDeltaResponse(streamID int64, req *discovery.DeltaDiscoveryRequest, resp *discovery.DeltaDiscoveryResponse) {
+	log.Debug().Msgf("OnStreamDeltaResponse node: %s type: %s, v: %s, nonce: %s, NumResources: %d", req.Node.Id, resp.TypeUrl, resp.SystemVersionInfo, resp.Nonce, len(resp.Resources))
+}
+
+// scLogger implements envoy control plane's log.Logger and delegates calls to the `log` variable defined in
```
keithmattix, line 72: Nice - TIL

pkg/envoy/server/health.go
```diff
@@ -7,7 +7,7 @@ func (s *Server) Liveness() bool {
 
 // Readiness is the Kubernetes readiness probe handler.
 func (s *Server) Readiness() bool {
-	return s.ready
+	return true
```
keithmattix, line 10: This doesn't feel right; why are we effectively removing our readiness probe?

keithmattix (commented):
Impressive! I just need some more info on the new types introduced in pkg/osm

cmd/osm-controller/osm-controller.go
steeling: will do!

pkg/osm/control_plane.go
steeling: Yup, good call!

pkg/envoy/server/cache.go
steeling: this has to fit the interface, so an error can't be returned. Note that this won't ever be called since the OnFetchRequest will preempt this

pkg/envoy/server/cache.go
steeling: yup! we can now switch over to delta xDS. pretty sweet!

pkg/envoy/server/health.go
steeling: There were no cases were s.ready was false prior to the server being started and able to handle requests, and we never switch it to false in any cases, so it was an unnecessary appendage.

keithmattix: We defaulted the field to true? That feels like a bug....would prefer not to propagate it if possible

steeling: essentially there's no not ready case. The server turns on and it's ready. There was never any downstream checks to perform

steeling: i've gone ahead and removed the health check anyways, to just leverage the health.SimpleHandler

pkg/envoy/generator/eds/builder.go
```diff
@@ -19,6 +19,7 @@ const (
 )
 
 // EndpointsBuilder is a helper type to build Envoy endpoints resources
```
jaellio, line 21:
```suggestion
```

pkg/osm/callbacks.go
```diff
@@ -0,0 +1,140 @@
...
+
+func (cp *ControlPlane[T]) update(ctx context.Context, proxy *envoy.Proxy) error {
+	resources, err := cp.configGenerator.GenerateConfig(ctx, proxy)
+	if err != nil {
+		return err
+	}
+	if err := cp.configServer.UpdateProxy(ctx, proxy, resources); err != nil {
+		return err
+	}
+	log.Debug().Msgf("successfully updated resources for proxy %s", proxy.String())
```
jaellio, line 105: nit:
```suggestion
	log.Debug().Str("proxy", proxy.String()).Msg("successfully updated resources for proxy)
```

pkg/osm/callbacks.go
```diff
@@ -0,0 +1,140 @@
...
+	return nil
+}
+
+func (cp *ControlPlane[T]) scheduleUpdate(ctx context.Context, proxy *envoy.Proxy) {
+	var wg sync.WaitGroup
+	wg.Add(1)
+	cp.workqueues.AddJob(
+		func() {
+			t := time.Now()
+			log.Debug().Msgf("Starting update for proxy %s", proxy.String())
```
jaellio, line 86:
```suggestion
			log.Debug().Str("proxy", proxy.String()).Msg("Starting update for proxy")
```

pkg/osm/control_plane_test.go
```diff
@@ -0,0 +1,211 @@
...
+	"github.com/openservicemesh/osm/pkg/envoy"
+	"github.com/openservicemesh/osm/pkg/envoy/registry"
+	"github.com/openservicemesh/osm/pkg/identity"
+	"github.com/openservicemesh/osm/pkg/messaging"
+)
+
+type fakeConfig string
+
+type fakeGenerator struct {
+	// map of
```
jaellio, line 31: Not sure if this comment is intentional
```suggestion
```

jaellio (approved):

pkg/envoy/generator/eds/builder.go
steeling: done!

pkg/osm/callbacks.go
steeling: done!

pkg/osm/callbacks.go
steeling: done!

pkg/osm/control_plane_test.go
steeling: done!

keithmattix (approved):

cmd/osm-controller/osm-controller.go
```diff
@@ -29,14 +30,16 @@ import (
 	"github.com/openservicemesh/osm/pkg/certificate"
 	configClientset "github.com/openservicemesh/osm/pkg/gen/client/config/clientset/versioned"
 	policyClientset "github.com/openservicemesh/osm/pkg/gen/client/policy/clientset/versioned"
+	"github.com/openservicemesh/osm/pkg/osm"
```
shashankram, line 33: nit: leave an empty line after API imports before local pkg imports.

steeling: done!

PR #5110, opened 2022-09-12
ref: merge SMI client into compute client and k8s client

allenlsy:
**Description**:

Refactor: merge SMI client into compute client and k8s client

* Operations for k8s resources are moved to k8s client
* Operations that involves OSM domain knowledge are moved to compute client

**Testing done**:

* Unit + e2e test

**Affected area**:

No

2. Is this a breaking change?

No

No


pkg/catalog/debugger.go
```diff
@@ -10,10 +10,10 @@ import (
 
 // ListSMIPolicies returns all policies OSM is aware of.
 func (mc *MeshCatalog) ListSMIPolicies() ([]*split.TrafficSplit, []identity.K8sServiceAccount, []*spec.HTTPRouteGroup, []*access.TrafficTarget) {
-	trafficSplits := mc.meshSpec.ListTrafficSplits()
-	serviceAccounts := mc.meshSpec.ListServiceAccounts()
-	trafficSpecs := mc.meshSpec.ListHTTPTrafficSpecs()
-	trafficTargets := mc.meshSpec.ListTrafficTargets()
+	trafficSplits := mc.Interface.ListTrafficSplits()
```
steeling, line 13: since interface is embedded, you don't need to call it. I think `mc.ListTrafficSplits()` is a cleaner read, and makes it more easy for us to merge these 2 in the future.

pkg/catalog/debugger.go
```diff
@@ -10,10 +10,10 @@ import (
 
 // ListSMIPolicies returns all policies OSM is aware of.
 func (mc *MeshCatalog) ListSMIPolicies() ([]*split.TrafficSplit, []identity.K8sServiceAccount, []*spec.HTTPRouteGroup, []*access.TrafficTarget) {
```
steeling, line 12: Optional: we could remove the ListSMIPolicies method, and have the single calling site, call each of the below methods individually. Up to you

pkg/catalog/egress.go
```diff
@@ -186,7 +186,7 @@ func (mc *MeshCatalog) buildHTTPRouteConfigs(egressPolicy *policyv1alpha1.Egress
 
 			// A TypedLocalObjectReference (Spec.Matches) is a reference to another object in the same namespace
 			httpRouteName := fmt.Sprintf("%s/%s", egressPolicy.Namespace, match.Name)
-			if httpRouteGroup := mc.meshSpec.GetHTTPRouteGroup(httpRouteName); httpRouteGroup == nil {
+			if httpRouteGroup := mc.Interface.GetHTTPRouteGroup(httpRouteName); httpRouteGroup == nil {
```
steeling, line 189: same: avoid mc.Interface in favor of embedded usage

pkg/compute/kube/client.go
```diff
@@ -549,3 +553,84 @@ func (c *client) ListNamespaces() ([]string, error) {
 	}
 	return names, nil
 }
+
+// ListTrafficSplits implements mesh.MeshSpec by returning the list of traffic splits.
+func (c *client) ListTrafficSplitsByOptions(options ...smi.TrafficSplitListOption) []*smiSplit.TrafficSplit {
+	var trafficSplits []*smiSplit.TrafficSplit
+
+	for _, trafficSplit := range c.kubeController.ListTrafficSplits() {
```
steeling, line 561: a good rule of thumb: if this method can be placed in the mesh catalog with no changes it should.

That is, we can leverage the passthrough interface, and have this method simply on the catalog itself.

Then we can even unexport it if it's not used elsewhere.

steeling: The reason for this, is that the higher level functions, that don't need knowledge about k8s, can be used out of the compute client, and re-used with another client. This makes it more simple to plug in a new interface

pkg/k8s/types.go
```diff
@@ -73,8 +78,35 @@ const (
...
+	// HTTPRouteGroup lookup identifier
+	HTTPRouteGroup InformerKey = "HTTPRouteGroup"
+	// TCPRoute lookup identifier
+	TCPRoute InformerKey = "TCPRoute"
+	// TrafficTarget lookup identifier
+	TrafficTarget InformerKey = "TrafficTarget"
 )
 
+// TrafficTargetListOpt specifies the options used to filter TrafficTarget objects as a part of its lister
+type TrafficTargetListOpt struct {
```
steeling, line 92: I think these are defined in multiple places, and should be removed from k8s package?

pkg/smi/smi.go
```diff
@@ -0,0 +1,95 @@
+package smi
+
+import (
+	smiAccess "github.com/servicemeshinterface/smi-sdk-go/pkg/apis/access/v1alpha3"
+	smiSplit "github.com/servicemeshinterface/smi-sdk-go/pkg/apis/split/v1alpha2"
+)
+
+// FilterTrafficSplit applies the given TrafficSplitListOption filter on the given TrafficSplit object
+func FilterTrafficSplit(trafficSplit *smiSplit.TrafficSplit, options ...TrafficSplitListOption) *smiSplit.TrafficSplit {
```
steeling, line 9: consider creating multiple methods for each List* call instead of having a generic filter.

Up to you

pkg/compute/kube/client.go
steeling: This would further allow us to unexport all the types defined in SMI, bring those types into the catalog package, and remove the SMI package altogether

pkg/smi/types.go
```diff
@@ -94,3 +77,21 @@ func WithKubeController(c k8s.Controller) TrafficSplitListOption {
 		o.KubeController = c
 	}
 }
+
+// GetSmiClientVersionHTTPHandler returns an http handler that returns supported smi version information
+func GetSmiClientVersionHTTPHandler() http.Handler {
```
steeling, line 82: @shashankram do we still need this? Can we just recommend users inspect the CRDs?

pkg/k8s/types.go
```diff
@@ -74,8 +79,22 @@ const (
...
+	// HTTPRouteGroup lookup identifier
+	HTTPRouteGroup InformerKey = "HTTPRouteGroup"
+	// TCPRoute lookup identifier
+	TCPRoute InformerKey = "TCPRoute"
+	// TrafficTarget lookup identifier
+	TrafficTarget InformerKey = "TrafficTarget"
 )
 
+// TrafficTargetListOption is a function type that implements filters on TrafficTarget lister
+type TrafficTargetListOption func(o *smi.TrafficTargetListOpt)
```
steeling, line 93: nit: these types are duplicated/unused

steeling (approved):

keithmattix (approved):

PR #5115, opened 2022-09-12
cleanup code in preparation for PR #5109: control plane separation

steeling:
This is some general code cleanup meant to reduce the diff on PR #5109. It includes:

1. Return an appropriate CA for tresor fake.
2. Rename sdsBuilder, to avoid lint stutter of sds.SDSBuilder
3. Export additional components in preparation for separating out the NewResponse funcs (and corresponding tests) from the builder logic.
4. Colocate some methods so we can get a clean file move
5. Add additional comments to exported methods.


pkg/envoy/cds/builder.go
```diff
@@ -476,7 +476,8 @@ func (b *clusterBuilder) buildLocalClusters() []*xds_cluster.Cluster {
 	return clusters
 }
 
-func getTypedHTTPProtocolOptions(httpProtocolOptions *extensions_upstream_http.HttpProtocolOptions) (map[string]*any.Any, error) {
+// GetTypedHTTPProtocolOptions marshals the provided httpProtocolOptions into the consumable type for the envoy config.
+func GetTypedHTTPProtocolOptions(httpProtocolOptions *extensions_upstream_http.HttpProtocolOptions) (map[string]*any.Any, error) {
```
jsturtevant, line 480: It looks like this is only called from unexported functions? It feels strange to export it here, Is this part of #5109? I follow why  exporting the builders makes sense for future code, but not sure about these functions which seem like implementation details to the builders

pkg/certificate/providers/tresor/fake/fake.go
```diff
@@ -107,7 +107,16 @@ func (c *fakeMRCClient) GetCertIssuerForMRC(mrc *v1alpha2.MeshRootCertificate) (
 		return nil, nil, err
 	}
 	issuer, err := tresor.New(ca, rootCertOrganization, 2048)
-	return issuer, pem.RootCertificate("rootCA"), err
+	if err != nil {
+		return nil, nil, err
+	}
+
+	cert, err := issuer.IssueCertificate("rootCA", 24*time.Hour)
```
jsturtevant, line 114: interesting, was this causing any bugs in tests?

pkg/envoy/ads/profile.go
```diff
@@ -18,6 +19,23 @@ const (
 	MaxXdsLogsPerProxy = 20
 )
 
+// GetXDSLog implements XDSDebugger interface and a log of the XDS responses sent to Envoy proxies.
+func (s *Server) GetXDSLog() map[string]map[envoy.TypeURI][]time.Time {
```
keithmattix, line 23: This map type is gnarly; may be a good candidate for a type alias?

keithmattix (approved):

pkg/certificate/providers/tresor/fake/fake.go
steeling: not currently, but I have some tests in an upcoming PR that will rely on a real CA cert being issued.

pkg/envoy/ads/profile.go
steeling: I agree that would be an improvement, but might pushback on that being out of scope for this PR if that's ok with everyone. Can implement if there's a strong push for it in this pr though

shashankram (approved):

Issue #5095, opened 2022-09-07
Update the IssueCertificate Provider Interface to include more than CN string


jsturtevant:
**Bug description**:

Right now, only CN is passed to the interface. This has all the information needed but has a lot of variability depending on the cert type.  If we make this a struct with the values, it will make creating the spiffe id for the URI SAN easier vs parsing it manually.

**Affected area (please mark with X where applicable)**:
- Install                    [ ]
- SMI Traffic Access Policy  [ ]
- SMI Traffic Specs Policy   [ ]
- SMI Traffic Split Policy   [ ]
- Permissive Traffic Policy  [ ]
- Ingress                    [ ]
- Egress                     [ ]
- Envoy Control Plane        [ ]
- CLI Tool                   [ ]
- Metrics                    [ ]
- Certificate Management     [ ]
- Sidecar Injection          [ ]
- Logging                    [ ]
- Debugging                  [ ]
- Tests                      [ ]
- Demo                       [ ]
- CI System                  [ ]

**Expected behavior**:

**Steps to reproduce the bug (as precisely as possible)**:

**How was OSM installed?**:

**Anything else we need to know?**:

**Bug report archive**:

**Environment**:
- OSM version (use `osm version`):
- Kubernetes version (use `kubectl version`):
- Size of cluster (number of worker nodes in the cluster):
- Others:

Issue #5002, opened 2022-08-16
Improve behavior of certificate issuance and rotation fuctions


shashankram:
**Please describe the Improvement and/or Feature Request**
Some of the code pertaining to cert issuance and rotation seems to do more than it should, specifically `IssueCertificate` and `checkAndRotate` have logic in them that seems out of place.

- `IssueCertificate()` checks if a cert should be rotated. This should not be the case. `IssueCertificate()` should either issue a new cert, or get the cert from the cache if present. `startRotationTicker()` routine already implements logic to periodically rotate expired certs. It's more intuitive to not execute rotation business logic in the issuance API.

- `checkAndRotate()` should only issue a certificate if the cert has expired, and not call IssueCertificate unless the cert has expired. Currently, `checkAndRotate()` calls into `IssueCertificate()` *before* verifying if a cert needs to be rotated. This is also ambiguous to reason.

- Disambiguate `prefix` is in function parameters. From an API perspective, this seems like an internal implementation rather than something the caller should be aware of.

**Scope (please mark with X where applicable)**
- Certificate Management     [X]

**Possible use cases**
Improve API semantics.

steeling:
FYI we'll need to move off the sync.map so we can do an atomic compare and swap

shashankram:
> FYI we'll need to move off the sync.map so we can do an atomic compare and swap

Why is that necessary?

jsturtevant:
> Disambiguate prefix is in function parameters. From an API perspective, this seems like an internal implementation rather than something the caller should be aware of.

I found a need to have more structure to the incoming IssueCertificate request.  Right now we take this prefix and optionally attach the trust domain (if not present).  To be able integrate spiffe, I have to know the internal structure of this a bit more than I would like to get the parts needed to for the Spiffe id in the certificate provider.  For example I need the trust domain, and currently the CN can come in the format of `ads.trustcomain.com`, `sa.ns.trustdomain.com`, and `proxyuid.type.sa.ns.trustdomain`.  Parsing these feels fragile.  

I am thinking we could make the concept of encoding the Identity into the IssueCertificate a bit more first class by providing more than just the pre-encoded CN to the Providers.  This would allow the providers to construct the names as needed.  

Going to explore this a bit more but wanted to get thoughts in case someone has already thought about this in some capacity.

PR #5116, opened 2022-09-13
Update IssueCertificate Provider Interface to be more easily extended

jsturtevant:
**Description**:
This refactors the IssueCertificate API to use the Options pattern throughout the code base. This does a few things:

- make it possible to add additional information in the Providers IssueCertificate. For instance, we re-use the information to generate a SpiffeID, where as prior we needed guess the trust domain and deconstruct the CN to figure it out
- Simplifies the calls to the Managers IssueCertificate API.  
  - In the case of `prefix` it was sometimes the prefix, sometimes not. In all cases it was the cache key and we can hide that detial further.
  - The `certtype` is internal to manager and will be removed in future when the information is stored on the MRC so this is a step in that direction by only using it in cert package and unexporting it.

fixes: #5095, https://github.com/openservicemesh/osm/issues/5002#issuecomment-1225010079

**Testing done**:

**Affected area**:

no
2. Is this a breaking change?
no

4. Has documentation corresponding to this change been updated in the [osm-docs](https://github.com/openservicemesh/osm-docs) repo (if applicable)?
no

### DIFF
--- a/pkg/certificate/options.go
+++ b/pkg/certificate/options.go
@@ -2,26 +2,112 @@ package certificate
 
 import (
 	"fmt"
+	"time"
+
+	"github.com/openservicemesh/osm/pkg/identity"
 )
 
-// IssueOption is an option that can be passed to IssueCertificate.
-type IssueOption func(*issueOptions)
+// IssueOption is an option that can be passed to IssueCertificate on the CertificateManager
+type IssueOption func(*IssueOptions)
 
-type issueOptions struct {
-	fullCNProvided bool
+// IssueOptions is passed to the Certificate Providers when creating certificates
+type IssueOptions struct {
+	fullCNProvided   bool
+	trustDomain      string
+	commonNamePrefix string
+	certType         certType
+	ValidityDuration time.Duration
 }
 
-func (o *issueOptions) formatCN(prefix, trustDomain string) CommonName {
+func (o IssueOptions) cacheKey() string {
+	return o.commonNamePrefix
+}
+
+// CommonName constructs the CommonName for the certificate.
+// If the FullCommonName option is set it will use configured name.
+// Otherwise it uses the name configured and appends the trustdomain
+func (o IssueOptions) CommonName() CommonName {
 	if o.fullCNProvided {
-		return CommonName(prefix)
+		return CommonName(o.commonNamePrefix)
 	}
-	return CommonName(fmt.Sprintf("%s.%s", prefix, trustDomain))
+
+	if o.commonNamePrefix == "" {
+		return CommonName(o.trustDomain)
+	}
+
+	return CommonName(fmt.Sprintf("%s.%s", o.commonNamePrefix, o.trustDomain))
 }
 
-// FullCNProvided tells IssueCertificate that the provided prefix is actually the full trust domain, and not to append
-// the issuer's trust domain.
-func FullCNProvided() IssueOption {
-	return func(opts *issueOptions) {
+func withCommonNamePrefix(prefix string) IssueOption {
+	return func(opts *IssueOptions) {
+		opts.commonNamePrefix = prefix
+	}
+}
+
+func withFullCommonName() IssueOption {
+	return func(opts *IssueOptions) {
 		opts.fullCNProvided = true
 	}
+}
+
+func withCertType(certType certType) IssueOption {
+	return func(opts *IssueOptions) {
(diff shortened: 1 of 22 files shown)


steeling (approved):

jsturtevant:
Example of extending the options for the SpiffeId.  Dropping it here since it shouldn't be included in this PR but I pushed it when I was testing and shows how the Issue provider doesn't need to know how to construct this.

```
func (o IssueOptions) SpiffeUrl() *url.URL {
	if o.fullCNProvided {
		// if the trust domain is already appended remove it.
		path := strings.TrimSuffix(o.commonNamePrefix, "."+o.trustDomain)
		return &url.URL{
			Scheme: "spiffe",
			Path:   path,
			Host:   o.trustDomain,
		}
	}

	return &url.URL{
		Scheme: "spiffe",
		Path:   o.commonNamePrefix,
		Host:   o.trustDomain,
	}
}
```

pkg/certificate/types.go
```diff
@@ -82,13 +84,13 @@ type Certificate struct {
...
 
-	certType CertType
+	certType certType
 }
 
 // Issuer is the interface for a certificate authority that can issue certificates from a given root certificate.
 type Issuer interface {
 	// IssueCertificate issues a new certificate.
-	IssueCertificate(CommonName, time.Duration) (*Certificate, error)
+	IssueCertificate(IssueOptions) (*Certificate, error)
```
steeling, line 93: nit: might have this take an interface with the single method GetCommonName()?

Makes it a bit easier to look at and realize that IssueOptions has no other exported fields/methods

jsturtevant: I considered this. There is one exported field `ValidityDuration` which would need to go to `getValidityDuration` and will eventually add `GetSpiffeId` to the interface.  Not sure the best approach. So open to suggestions

pkg/certificate/options.go
```diff
@@ -2,26 +2,112 @@ package certificate
...
+// ForCommonName creates an internal certificate with a given full common name
+func ForCommonName(fullCommonName string) IssueOption {
+	return func(opts *IssueOptions) {
+		opts.commonNamePrefix = fullCommonName
+		opts.certType = internal
+		opts.fullCNProvided = true
+	}
+}
+
+// NewCertOptions creates the IssueOptions for the issuing a certificate
```
jaellio, line 95: nit: 
```suggestion
// NewCertOptions creates the IssueOptions for issuing a certificate
```

pkg/certificate/options_test.go
```diff
@@ -0,0 +1,89 @@
...
+func TestIssueOptions_CommonName(t *testing.T) {
+	tests := []struct {
+		name         string
+		trustDomain  string
+		issueOption  []IssueOption
+		want         CommonName
+		wantCertType certType
+	}{
+		{
+			name:         "ForService appends trust domain",
```
jaellio, line 16: nit: 
```suggestion
			name:         "ForServiceIdentity appends trust domain",
```

pkg/certificate/options_test.go
```diff
@@ -0,0 +1,89 @@
...
+			wantCertType: service,
+		},
+		{
+			name:         "ForCommonName uses common name passed without appending trust domain",
+			issueOption:  []IssueOption{ForCommonName("sa.ns.cluster.local")},
+			want:         CommonName("sa.ns.cluster.local"),
+			wantCertType: internal,
+		},
+		{
+			name:         "ForCommonName uses common name passed without appending trust domain",
```
jaellio, line 29: Could we update the names of these tests to differentiate them?

jsturtevant, line 29:
```suggestion
			name:         "ForCommonName uses common name passed without appending trust domain when trust domain is present",
```

pkg/certificate/types.go
jsturtevant: I think it was in reference to `IssueOptions`. After thinking some more, I believe I choose a struct since it is really just data not an implementation of a set of functionality but I realize that is kind of fuzzy.  Happy to adjust if you anyone has strong opinion here.

keithmattix (approved):

jaellio (approved):

pkg/certificate/types.go
steeling: Ya that sounds reasonable to me!

Issue #5022, opened 2022-08-24
Add feature flag that turns spiffe id generation on in Certificates


jsturtevant:
This does not implement the feature but creates the feature flag to that can be used to turn the feature on/off.

Issue #5027, opened 2022-08-24
Add support in the OSM debugger to view this addition certificate fields for Spiffe


jsturtevant:

keithmattix:
s/OMS/OSM?

jsturtevant:
completed in #5131

Issue #5024, opened 2022-08-24
Implement spiffe ID generation in x509 certificates


jsturtevant:
This adds the field to the certificate in the SAN URI field but does not turn on validation or use it.

May want to clean up our IssueCertificate Interface, right now it uses prefixes and it could benefit from more structure to make parsing decisions easier later

jsturtevant:
completed in #5131

Issue #5023, opened 2022-08-24
Add feature flag that uses on spiffe id validation in certificates and uses spiffe id for Identity


jsturtevant:
This can't be done at same time in backwards compatible way with out re-issuing all the Service Certificates.

It might be possible to automate this but might be just as easy to just wait given time frame to flip a switch or start a new deployment with this enabled. Another option is creating a cli helper tool that can verify the certs are set through the Mesh

jsturtevant:
Since we can rely on the MRC handing rotation of the certificate for us we don't need an additional flag to insure the Serivce certs are all generated properly. #5160 uses the MRC rotation approach. Closing.

PR #5131, opened 2022-09-15
Create SPIFFE compatible certs

jsturtevant:
**Description**:
Adds support for generating SPIFFE Compatible Certs.  It does **_not_** use the spiffe Identity encoded in the cert which will be addressed in #5023.

fixes: #5022, #5027, #5024

TODO:
- [x] Add e2e test
- [x] add validation (after #5105 merges)

~Requires both #5116 and #5105~ (both have merged now)

**Testing done**:

**Affected area**:

2. Is this a breaking change?


jsturtevant:
This is ready for review but could be slightly smaller after https://github.com/openservicemesh/osm/pull/5135 merges and will have small update after https://github.com/openservicemesh/osm/pull/5105 (which could be done in follow up)

jsturtevant:
e2e tests are passing with certs being provisioned with SPIFFE URI SAN

```
• [SLOW TEST:190.262 seconds]
[Tier 2][Bucket 10][] SPIFFE
/home/runner/work/osm/osm/tests/framework/common.go:90
  with Tressor
  /home/runner/work/osm/osm/tests/e2e/e2e_spiffe_test.go:23
    has a Pod Certificate with a SPIFFE ID
    /home/runner/work/osm/osm/tests/e2e/e2e_spiffe_test.go:32
```

```
• [SLOW TEST:232.645 seconds]
[Tier 2][Bucket 1][] 1 Client pod -> 1 Server pod test using cert-manager
/home/runner/work/osm/osm/tests/framework/common.go:90
  with SPIFFE Enabled
  /home/runner/work/osm/osm/tests/e2e/e2e_certmanager_test.go:24
    Tests HTTP traffic for client pod -> server pod
```

```
• [SLOW TEST:98.213 seconds]
[Tier 2][Bucket 4][] 1 Client pod -> 1 Server pod test using Vault
/home/runner/work/osm/osm/tests/framework/common.go:90
  with SPIFFE Enabled
  /home/runner/work/osm/osm/tests/e2e/e2e_hashivault_test.go:24
    Tests HTTP traffic for client pod -> server pod
```

pkg/certificate/providers/vault/tools.go
```diff
@@ -15,9 +15,18 @@ func getIssueURL(role string) string {
...
+	issuanceData := map[string]interface{}{
+		commonNameField: options.CommonName().String(),
+		ttlField:        getDurationInMinutes(options.ValidityDuration),
 	}
+
+	if options.SpiffeEnabled {
+		log.Trace().Str("cn", options.CommonName().String()).Msg("Generating SPIFFE enabled Certificate")
+
+		// Comma delimited list; For SPIFFE compatibility it should only be one.
+		issuanceData[uriSans] = options.SpiffeURL().String()
```
steeling, line 28: would it be possible to change the `CommonName()` method to generate the SPIFFE URL if SPIFFE is enabled.

And to always populate issuanceData with the uriSans = options.CommonName()?

That way the providers can stay "dumb" and not need to know about spiffe?

jsturtevant: I didn't do this since we are still using common name to do validation, which is required to be able to do an upgrade in backwards compat way, i.e. issue all certs with SPIFFE URI SANs (via a MRC rotation) then turn on validation using that field (#5025 and #5023)

Another aspect of this is, the Ingresses we support only currently do validation via the DNS SAN (nginx and contour) which will be in the format of the common name.

One thing I considered was to do something like `if options.SpiffeId() != "" then urisans = options.spiffeId` which would remove the extra `SpiffeEnabled`. Maybe making it `options.UriSan() != ""` could keep them "dumb"

jsturtevant:
I've rebased ontop of https://github.com/openservicemesh/osm/pull/5135 and https://github.com/openservicemesh/osm/pull/5105

pkg/certificate/providers/vault/tools.go
steeling: gotcha ya that makes sense. Up to you on your last point, i'm relatively indifferent there. Although I do like the idea of abstracting Spiffe away for the UriSan method

steeling (approved):

pkg/certificate/providers/vault/tools.go
jsturtevant: I do too.  I'll make this update

jsturtevant: updated

keithmattix (approved):

charts/osm/values.yaml
```diff
@@ -487,6 +487,8 @@ osm:
     enableRetryPolicy: false
     # -- Enable the MeshRootCertificate to configure the OSM certificate provider
     enableMeshRootCertificate: false
+    # -- Enable the MeshRootCertificate to configure certificates with SPIFFE IDs
```
jaellio, line 490: Nit: 
```suggestion
    # -- Enable adding a SPIFFE ID to certificatess
```

pkg/apis/config/v1alpha2/meshrootcertificate.go
```diff
@@ -36,6 +36,10 @@ type MeshRootCertificateSpec struct {
 
 	// Intent of the MeshRootCertificate resource
 	Intent MeshRootCertificateIntent `json:"intent"`
+
+	// SpiffeEnabled will add a SPIFFE ID to the certificates, creating a SPIFFE compatible x509 SVID document
+	// To use SPIFFE ID for validation and routing,  'enableSPIFFEValidation' must be true in the Mesh Config after the certificate is made 'active'
```
jaellio, line 41: Update certificate to meshRootCertificate for clarity
```suggestion
	// To use SPIFFE ID for validation and routing, 'enableSPIFFE' must be true in the MeshConfig after the MeshRootCertificate is made 'active'
```

pkg/certificate/providers/vault/certificate_manager_test.go
```diff
@@ -147,11 +155,27 @@ func TestIssueCertificate(t *testing.T) {
...
+			tassert.Equal(tc.cn+".cluster.local", xCert.Subject.CommonName)
+
+			if tc.spiffeEnabled {
+				uri := xCert.URIs[0]
+				tassert.Equal("spiffe://cluster.local/"+tc.cn, uri.String())
+
+				// Vaults test mock doesn't create private keys properly?
+				// svid, err := x509svid.Parse(cert.CertChain, cert.PrivateKey)
+				// tassert.NoError(err)
+				// tassert.Equal("spiffe://cluster.local/"+tc.cn, svid.ID.String())
```
jaellio, line 178: Not sure if this is meant to be included here

pkg/validator/validators.go
```diff
@@ -325,6 +325,10 @@ func (kc *validator) validateMRCOnUpdate(oldMRC *configv1alpha2.MeshRootCertific
 		return fmt.Errorf("cannot update trust domain for MRC %s. Create a new MRC and initiate root certificate rotation to update the trust domain", getNamespacedMRC(oldMRC))
 	}
 
+	if oldMRC.Spec.SpiffeEnabled != newMRC.Spec.SpiffeEnabled {
```
jaellio, line 328: Awesome!! Thanks for adding this!

jaellio (approved):

pkg/certificate/providers/vault/certificate_manager_test.go
jsturtevant: yes sorry, missed cleaning this up.   We have coverage in other places so will remove.  Interesting the certs produced here aren't valid.

jaellio: It's a good callout. Do you think it's worth tracking in a bug?

jsturtevant: It's more a bug or issue with Vault's testing api. I don't know that this is a critical check and don't think it is probably worth tracking down the issue in vault.  This was more of a sanity on my part to make sure the certs we produce pass SPIFFE validation.  We have that check in Tressor so I feel like we have it covered if anyone was to change the URI SAN implementation it would be caught.

Issue #5134, opened 2022-09-16
Failure on re-running e2e test on existing Kind cluster


allenlsy:
**Bug description**:

When running e2e test locally with Kind cluster, we want to re-use the created Kind cluster to perform test multiple times. It works for the first run, running with args `-cleanupKindCluster=false -installType=KindCluster`. For the consequent runs, we removed `-installType=KindCluster`. This will use `SelfInstall` as the install type by default, which means no new Kind cluster will be created. Then OSM cannot be installed on the existing Kind cluster correctly. It cannot load the OSM images data from localhost:5000.

I found that loading the OSM image data is only executed when install type is `KindCluster`. If it is set to `KindCluster`, it will also fail since there is an existing Kind cluster.

Therefore, there is no way to re-use existing Kind cluster for e2e test.

**Affected area (please mark with X where applicable)**:
- Tests                      [x]

**Expected behavior**:

Skip creating new Kind cluster if an existing Kind cluster is found. If current context is not a non-Kind k8s cluster, it should still create one.  

**Steps to reproduce the bug (as precisely as possible)**:

* Build osm images and push to local registry: `make docker-build DOCKER_BUILDX_OUTPUT=type=docker`
* Run a e2e test: example `go test ./tests/e2e -test.v -ginkgo.v -ginkgo.progress -kindClusterVersion='' -test.timeout 0 -test.failfast -ginkgo.failFast -ginkgo.focus='Test HTTP traffic with SMI TrafficTarget' -cleanupKindCluster=false -cleanupKindClusterBetweenTests=false -installType=KindCluster`
* Run the command again without `-installType=KindCluster`

Be sure to have a non-empty and same CTR_TAG env var for both commands

**How was OSM installed?**:

On Kind cluster

**Anything else we need to know?**:

PR #5137, opened 2022-09-16
test: skip creating Kind cluster if exists in e2e test

allenlsy:

**Description**:

Skip creating Kind cluster if there is a required one during the e2e test.

Resolves https://github.com/openservicemesh/osm/issues/5134

**Testing done**:

Manually tested two cases:

* When required Kind cluster exists (by default `kind-osm-e2e`), it skips creating a new one and continues with the tests
* When required Kind cluster does not exist, it creates a new one as usual

image

**Affected area**:

No

2. Is this a breaking change?

No

No


jsturtevant:
I ran into this yesterday, nice improvement to local developer workflow 💯

tests/framework/common.go
```diff
@@ -18,6 +18,8 @@ import (
 	"strings"
 	"time"
 
+	"golang.org/x/exp/slices"
```
jsturtevant, line 21: This is the experimental package https://github.com/golang/exp

> In short, code in this subrepository is not subject to the Go 1 compatibility promise. (No subrepo is, but the promise is even more likely to be violated by go.exp than the others.)

This is probably fine for our test setup and this function in particular since its pretty simple but wanted to call it out

jaellio (approved):
Super helpful @allenlsy! Thank you!!

Issue #5007, opened 2022-08-17
'osm version' command returns invalid output with locally compiled versions


shashankram:
**Bug description**:
When using locally compiled binaries, the `osm version` command returns an invalid output:
```
$ osm version
CLI Version: version.Info{Version:"dev", GitCommit:"dddb8fa5a7a60dbb3c03843252807ee7bde2ffc8", BuildDate:""}

MESH NAME   MESH NAMESPACE   VERSION   GIT COMMIT                                 BUILD DATE
osm         osm-system       dev       dddb8fa5a7a60dbb3c03843252807ee7bde2ffc8   
Error: 1 error occurred:
	* Failed to output latest release information: could not parse "dev" as version
```

Since the version defaults to `dev` in the Makefile for local builds, outputting `dev` as a version should suffice to indicate a developer build of the software.

**Affected area (please mark with X where applicable)**:
- CLI Tool                   [X]

**Expected behavior**:
Developer builds should output the version as `dev`

**Steps to reproduce the bug (as precisely as possible)**:
Compile control plane binaries locally and deploy it using the CLI.
Run the `osm version` command.

jaellio:
Assigning to @lam-man

jaellio:
For reference, this is the PR that added the remote osm version to the `osm version` CLI output https://github.com/openservicemesh/osm/pull/4395

lam-man:
Thanks @jaellio . I will work on this.

lam-man:
I was able to reproduce the problem. Working on the fix.

lam-man:
PR for the issue has been created. Thanks!

PR #5143, opened 2022-09-20
fix(cmd/cli/version.go): fix 'osm version' command to make it compatible with local compiled version

lam-man:
Signed-off-by: Wen Lin <linwen1991@gmail.com>

**Description**:
Fix `osm version` command to make it compatible with local compiled version. Fixing is for [this issue](https://github.com/openservicemesh/osm/issues/5007#issuecomment-1242172429). The issue occurs while compare the `dev` version with latest available version. However, the regex only parses numbered version. When we try to parse `dev` version, we will see error: `Failed to output latest release information: could not parse `dev` as version` . The fix here disabled the comparison of `dev` and latest available version. 

**Testing done**:

- Test by running the commands `./bin/osm version` and `osm version` locally. 
- Manually deployed the fixed version to an Azure K8s Service cluster and run the commands `./bin/osm version` and `osm version`.

**Affected area**:
| CLI Tool                   | [ X] |

1. Does this change contain code from or inspired by another project? N

2. Is this a breaking change? N

3. Has documentation corresponding to this change been updated in the [osm-docs] (https://github.com/openservicemesh/osm-docs) repo (if applicable)? N/A

### DIFF
--- a/cmd/cli/version_latest_test.go
+++ b/cmd/cli/version_latest_test.go
@@ -17,11 +17,11 @@ func TestOutputLatestReleaseVersion(t *testing.T) {
 		output  string
 	}{
 		{
-			name:    "invalid current version",
-			current: "v1.0.0.0",
+			name:    "customized current version",
+			current: "dev",
 			latest:  "v1.0.0",
-			err:     fmt.Errorf("illegal version string \"v1.0.0.0\""),
-			output:  "",
+			err:     nil,
+			output:  "\nOSM in dev version. Latest available version is v1.0.0.\n\n",
 		},
 		{
 			name:    "invalid latest version",
--- a/cmd/cli/version_latest.go
+++ b/cmd/cli/version_latest.go
@@ -45,9 +45,8 @@ func outputLatestReleaseVersion(out io.Writer, latestRelease string, currentRele
 	}
 	current, err := version.ParseSemantic(currentRelease)
 	if err != nil {
-		return err
-	}
-	if current.LessThan(latest) {
+		fmt.Fprintf(out, "\nOSM in %s version. Latest available version is %s.\n\n", currentRelease, latestRelease)
+	} else if current.LessThan(latest) {
 		fmt.Fprintf(out, "\nOSM %s is now available. Please see https://github.com/openservicemesh/osm/releases/latest.\nWARNING: upgrading could introduce breaking changes. Please review the release notes.\n\n", latestRelease)
 	}
 	return nil


cmd/cli/version.go
```diff
@@ -17,9 +17,12 @@ import (
 	"github.com/openservicemesh/osm/pkg/version"
 )
 
-const versionHelp = `
+const (
+	versionHelp = `
 This command prints the OSM CLI and remote mesh version information
 `
+	dev = `dev`
```
keithmattix, line 24: I'd prefer `devVersion` as the var name to remove ambiguity.

lam-man: Make sense. It is clearer on this. I will change it. Thanks!

lam-man: Changes made and tested in local + AKS. Thanks

cmd/cli/version.go
```diff
@@ -96,8 +99,10 @@ func newVersionCmd(out io.Writer) *cobra.Command {
 				latestReleaseVersion, err := getLatestReleaseVersion()
 				if err != nil {
 					multiError = multierror.Append(multiError, fmt.Errorf("Failed to get latest release information: %w", err))
-				} else if err := outputLatestReleaseVersion(versionCmd.out, latestReleaseVersion, cliVersionInfo.Version); err != nil {
-					multiError = multierror.Append(multiError, fmt.Errorf("Failed to output latest release information: %w", err))
+				} else if cliVersionInfo.Version != devVersion {
```
shashankram, line 102: The version can be any version string, not necessarily `dev`.

For e.g. the version can be set to `foo` as follows:
```
VERSION=foo make build-osm
```

Why can't we print whatever version is encoded instead of only allowing specific version strings?

shashankram (changes requested):

cmd/cli/version.go
lam-man: Great point. I forgot that the version name is customizable. I will try to modify the regex. Thanks!

shashankram: Hey @lam-man, I am thinking we could simply print the version string instead of displaying an error. Would that work?

lam-man: This definitely will work. Should we also print out the latest available version?

shashankram: Sure, as long as it fixes the existing bug where a custom version string gets incorrectly displayed as reported in #5007

cmd/cli/version_latest_test.go
```diff
@@ -17,11 +17,11 @@ func TestOutputLatestReleaseVersion(t *testing.T) {
 		output  string
 	}{
 		{
-			name:    "invalid current version",
-			current: "v1.0.0.0",
+			name:    "customized current version",
```
shashankram, line 20: Could you add a new test to the table instead of updating the existing scenario.

lam-man: The existing scenario will fail with the change, as we are going to accept all customized version. I can add a new test and change the current test case accordingly. Please let me know if you have any other questions or suggestions. Thanks

shashankram (approved):

keithmattix (approved):

Issue #5136, opened 2022-09-16
Add capability to export and configure Envoy access logs


shashankram:
**Please describe the Improvement and/or Feature Request**

Add capability to configure Envoy access logging per pod(s) and namespace, with regards to the access log format and the ability to export access logs to an OpenTelemetry collector.

Tasks:
- [x] [Design doc](https://docs.google.com/document/d/1CZo79XlSBgwr2dOrbR1YDYzPveuk6lr6cAIkl75rW3g/edit?usp=sharing)
- [x] Add ability to configure access logging format per pod(s) and namespace
- [x] Add ability to export access log to an OpenTelemetry collector (deployed independently by the user): https://github.com/openservicemesh/osm/issues/4999
- [x] Documentation

**Scope (please mark with X where applicable)**
- Telemetry [X]

**Possible use cases**

1. Users want the ability to configure the access logging format and configure which request attributes get logged

2. Users want the ability to export logs to a centralized collector.
As quoted in https://github.com/openservicemesh/osm/issues/4999#issue-1339266824:

*The more useful need for collecting Envoy access logs is to have to ability to correlate traffic events and have long-term availability to the logs for possible forensic analysis.

User Story - As a SecOps user, I would like the ability to configure OSM to send Envoy access logs in the mesh to an external data store (SEIM), so I can replay traffic events in needed for further analysis.*

PR #5144, opened 2022-09-20
apis: add Telemetry and ExtensionService APIs

shashankram:
**Description**:
Adds 2 new APIs based on the [Telemetry design doc](https://docs.google.com/document/d/1CZo79XlSBgwr2dOrbR1YDYzPveuk6lr6cAIkl75rW3g/edit?usp=sharing)

1. Telemetry to configure telemetry related config, starting with Envoy access logs.

2. ExtensionService to specify configuration pertaining to external services that OSM can integrate with, starting with config necessary to connect to an OpenTelemetry collector.

Part of #5136

**Testing done**:
E2e

**Affected area**:


keithmattix (approved):

pkg/apis/config/v1alpha2/extensionservice.go
```diff
@@ -0,0 +1,48 @@
...
+	// +optional
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	// Spec defines the specification of the extension service.
+	// +optional
+	Spec ExtensionServiceSpec `json:"spec,omitempty"`
+}
+
+// ExtensionServiceSpec defines the specification of the extension service.
+type ExtensionServiceSpec struct {
```
steeling, line 26: It might be useful to check out the various settings for auth extension https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/ext_authz/v3/ext_authz.proto

ie: breaking up gRPC vs HTTP, adding channel credentials (if allowing non-meshed traffic), adding a `headersToAdd` or something similar for HTTP. Just a thought

shashankram: I am thinking we could introduce protocol specific fields such as `grpc, http, tcp` etc. on the resource for this purpose. Additionally, we would need a `tls` field for TLS based communication with the extension service. 

I'd prefer to not handle this in this PR as I chose to introduce the min required fields to unblock the Access log + OpenTelemetry integration work - which for the first pass will rely on non-TLS based log exporting.

We could iterate on this API to add protocol specific and TLS fields like you suggested, as they would merely be field additions to the same API version.

steeling (approved):

Issue #5091, opened 2022-09-07
Handle MRCUpdated event when spec.State set to passive


jaellio:
Defined in [Update status Step 2](https://docs.google.com/document/d/1QF7wL1su4CUzwo3Yo2_vFL90XG2d-wUhBU6uifNrkws/edit#heading=h.gmxmn5fw46m9)

- osm received MRCUpdated event with `spec.intent=passive` and all `status.certificateStatues=Issuing`
- issuers don't change
- set `status.condition.IssuingRollback` to `True` and the reason to `PassiveStateAndCertIsBeingIssued`
- set `status.condition.ValidatingRollback` to `True` and the reason to `PassiveStateAndCertIsBeingIssued`
- start mrcReconciliation loop to watch for all `status.certificateStatuses` to be `Validating`. Once the condition is met, set the `status.condition.IssuingRollback` to `False` and the reason to `RollbackComplete`. Also, set the `status.state` to `Validating`

Issue #5092, opened 2022-09-07
Handle MRCUpdated event when spec.State set to active


jaellio:
Defined in [Start issue with new certificate Step 5](https://docs.google.com/document/d/1QF7wL1su4CUzwo3Yo2_vFL90XG2d-wUhBU6uifNrkws/edit#heading=h.e5jt97f5t23g)

- [ ] swap issuers
- [ ] start timer loop to update containerStatus
- [ ] start MRC reconciliation loop for IssuingRollout to True (**Verify this should be initiated in this step) 
- [ ] start MRC reconciliation loop for Provisioning, IssuingRollout, and ValidatingRollout to be True

Issue #5089, opened 2022-09-07
Handle MRCUpdated event when ValidatingRollout condition reason set to Pending


jaellio:
Defined in [Use new cert for validation Step 3](https://docs.google.com/document/d/1QF7wL1su4CUzwo3Yo2_vFL90XG2d-wUhBU6uifNrkws/edit#heading=h.sy0tyhbqfylp)

All control plane components receive MRCUpdated event and check for ValidatingRollout as Pending
  - Each component starts its process of adding the new MRC to the cert chain 
    - Individually time-bounded: either xDuration (specified per certificate type by user in spec) or when one of the component replicas finishes the job
      - **For sidecar data plane certificates in osm-controller, do we just want to wait that duration every time? A fresh replica won’t have any entries in its proxy registry, rushing ahead and setting the `certificateStatus` to `Validating` too early. Unless we have osm-controller components coordinate, using the timer is probably the best option for now (coordination via configmap can come later)
    - When each component is done (or when xDuration is up), update `spec.certificateStatuses` field to `Validating`
    - If an error occurs, update `spec.certificateStatuses` to `Error`
      - We may just need to let SDS fail, restart, and wait for bootstrap certs for that proxy to be updated
      - Eventually, retry logic will use this info to set a configmap (or some other signal) so that the erroring component knows to try again
      - This means/implies that control plane components should ignore errored MRCs on startup
If the component pod restarts and fixes itself, it should be able to update

**Steps:**
- initiate time bounded loop for each cert type
- upon completion, set the respective `spec.certificateStatuses` field to `Validating`. On error, set the respective `spec.certificateStatuses` field to `Error`

PR #5147, opened 2022-09-21
feat(certs): handle MRCUpdated event when intent set passive

jaellio:
**Description**:

Adds logic to update MRC conditions and components. Handles MRCUpdated event when intent set to passive.

Part of #5091 

**Testing done**:

**Affected area**:

2. Is this a breaking change?

### DIFF
--- a/pkg/certificate/components.go
+++ b/pkg/certificate/components.go
@@ -1,0 +1,24 @@
+package certificate
+
+import (
+	"github.com/openservicemesh/osm/pkg/apis/config/v1alpha2"
+	"github.com/openservicemesh/osm/pkg/constants"
+)
+
+// componentsAreIssuing returns true if all certificate component statuses are `issuing`
+func componentsAreIssuing(mrc *v1alpha2.MeshRootCertificate) bool {
+	return mrc.Status.ComponentStatuses.Bootstrap == constants.MRCComponentStatusIssuing &&
+		mrc.Status.ComponentStatuses.Gateway == constants.MRCComponentStatusIssuing &&
+		mrc.Status.ComponentStatuses.Sidecar == constants.MRCComponentStatusIssuing &&
+		mrc.Status.ComponentStatuses.Webhooks == constants.MRCComponentStatusIssuing &&
+		mrc.Status.ComponentStatuses.XDSControlPlane == constants.MRCComponentStatusIssuing
+}
+
+// componentsHaveError returns true if any certificate component statuses are `error`
+func componentsHaveError(mrc *v1alpha2.MeshRootCertificate) bool {
+	return mrc.Status.ComponentStatuses.Bootstrap == constants.MRCComponentStatusIssuing &&
+		mrc.Status.ComponentStatuses.Gateway == constants.MRCComponentStatusIssuing &&
+		mrc.Status.ComponentStatuses.Sidecar == constants.MRCComponentStatusIssuing &&
+		mrc.Status.ComponentStatuses.Webhooks == constants.MRCComponentStatusIssuing &&
+		mrc.Status.ComponentStatuses.XDSControlPlane == constants.MRCComponentStatusIssuing
+}
--- a/pkg/certificate/components.go
+++ b/pkg/certificate/components.go
@@ -16,9 +16,9 @@ func componentsAreIssuing(mrc *v1alpha2.MeshRootCertificate) bool {
 
 // componentsHaveError returns true if any certificate component statuses are `error`
 func componentsHaveError(mrc *v1alpha2.MeshRootCertificate) bool {
-	return mrc.Status.ComponentStatuses.Bootstrap == constants.MRCComponentStatusIssuing &&
-		mrc.Status.ComponentStatuses.Gateway == constants.MRCComponentStatusIssuing &&
-		mrc.Status.ComponentStatuses.Sidecar == constants.MRCComponentStatusIssuing &&
-		mrc.Status.ComponentStatuses.Webhooks == constants.MRCComponentStatusIssuing &&
-		mrc.Status.ComponentStatuses.XDSControlPlane == constants.MRCComponentStatusIssuing
+	return mrc.Status.ComponentStatuses.Bootstrap == constants.MRCComponentStatusError &&
+		mrc.Status.ComponentStatuses.Gateway == constants.MRCComponentStatusError &&
+		mrc.Status.ComponentStatuses.Sidecar == constants.MRCComponentStatusError &&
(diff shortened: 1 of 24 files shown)


pkg/certificate/manager.go
```diff
@@ -152,9 +154,83 @@ func (m *Manager) handleMRCEvent(mrcClient MRCClient, event MRCEvent) error {
 			m.mu.Unlock()
 		}
 	case MRCEventUpdated:
-		// TODO
+		oldMRC := event.OldMRC
+		newMRC := event.NewMRC
+
+		// check if the MRC's intent has been updated from active to passive and the
+		// certificate statuses are all issuing
+		if oldMRC.Intent != newMRC.Intent && newMRC.Intent == constants.MRCIntentPassive &&
```
steeling, line 162: is the idea that there will be multiple if statements here, for each stage change, each one creating a reconciler?

Can you add a comment that might show what all of those checks might be?

pkg/certificate/manager.go
```diff
@@ -152,9 +154,83 @@ func (m *Manager) handleMRCEvent(mrcClient MRCClient, event MRCEvent) error {
...
+			log.Debug().Str("mrc", newMRC.Name).Msg("handling active to passive MRC status change")
+			err := handleActiveToPassiveMRCStatusChange(mrcClient, event)
+			if err != nil {
+				return err
+			}
+
+			// define reconciler
+			mrcReconciler := MRCReconciler{
+				mrcName:      newMRC.Name,
+				updateStatus: nil,
```
steeling, line 174: can you add a comment explaining what the updateStatus might do in this example?

pkg/certificate/manager.go
```diff
@@ -152,9 +154,83 @@ func (m *Manager) handleMRCEvent(mrcClient MRCClient, event MRCEvent) error {
...
+		}
+		if !componentsAreIssuing(mrc) {
+			return fmt.Errorf("mrc components are not issuing")
+		}
+		if mrc.Intent != constants.MRCIntentPassive {
+			return fmt.Errorf("mrc does not have passive intent")
+		}
+
+		// set new conditions
+		setMRCCondition(mrc, newIssuingRollbackCond.Type, newIssuingRollbackCond.Status, mrcConditionReason(newIssuingRollbackCond.Reason), "")
```
steeling: These look to be setting the condition on the same MRC object.. is the idea that it will set one condition on the old and one condition on the new?

pkg/certificate/manager.go
jaellio: The checks for MRCUpdated events are roughly outlined in this doc https://docs.google.com/document/d/1QF7wL1su4CUzwo3Yo2_vFL90XG2d-wUhBU6uifNrkws/edit#, but are more specifically outlined in the following issues:
- #5091 
- #5089
- #5092

I can also add a comment.

pkg/certificate/manager.go
jaellio: Sure! As outlined in #5091:
> Once the condition is met, set the `status.condition.IssuingRollback` to `False` and the reason to `RollbackComplete`. Also, set the `status.state` to `Validating`

pkg/certificate/manager.go
steeling: and i'm guessing it also performs the actual update via the compute client? So it seems the job of the update is split between the cert manager, which has several checks for switching on status case, and the `handleActiveToPassiveMRCStatusChange`, which *actually* also performs the update. So that's a bit confusing to me.

It seems the mrcReconciler would likely want to reconcile both the old and the new mrc, and not just the new one? Or is the idea that there can be multiple reconciler objects running simultaneously?

If that's the case there needs to be extra protections to make sure that one mrc doesn't race ahead while another gets stuck behind.

Finally i go back to my original argument: why is this logic spread across 2 objects? Why not put the entire event loop/handleActiveToPassiveMRCStatusChange on a new reconciler object, and have the reconciler object be long lived? Why does it need to be a temporal object that finishes upon completion? Instead of temporal object, it can be a goroutine on a more permanent object?

steeling: with that, instead of having an updateFunc variable, it can be 

```
func (r *reconciler) reconcile(ctx context.Context, oldName, newName string) {
   for range ticker.C {
    oldMRC, newMRC := get(oldName), get(newName)
    switch{
     case isActive(oldMRC, newMRC):
          r.updateActive(oldMRC, newMRC)
     case isPassive(oldMRC, newMRC):
          r.updatePassive(oldMRC, newMRC)
      .....
    }
}
```


That's a really rough outline, but i think sets the stage.

steeling:
I'd personally prefer to see a PR like this come before https://github.com/openservicemesh/osm/pull/5073. I think there's a lot of caveats like how many can run at once (might infer if we need a mutex, how the MRCReconciler signifies that it is done reconciling, or other implementation details). Curious on your thoughts and how others may feel though

pkg/certificate/components.go
```diff
@@ -0,0 +1,24 @@
...
+	return mrc.Status.ComponentStatuses.Bootstrap == constants.MRCComponentStatusIssuing &&
+		mrc.Status.ComponentStatuses.Gateway == constants.MRCComponentStatusIssuing &&
+		mrc.Status.ComponentStatuses.Sidecar == constants.MRCComponentStatusIssuing &&
+		mrc.Status.ComponentStatuses.Webhooks == constants.MRCComponentStatusIssuing &&
+		mrc.Status.ComponentStatuses.XDSControlPlane == constants.MRCComponentStatusIssuing
+}
+
+// componentsHaveError returns true if any certificate component statuses are `error`
+func componentsHaveError(mrc *v1alpha2.MeshRootCertificate) bool {
+	return mrc.Status.ComponentStatuses.Bootstrap == constants.MRCComponentStatusError &&
```
shalier, line 19: Should these be `||`'s instead of `&&`'s

pkg/certificate/conditions.go
```diff
@@ -0,0 +1,108 @@
...
+	for i, condition := range mrc.Status.Conditions {
+		if condition.Type != newCondition.Type {
+			continue
+		}
+
+		// only update transition time on status change
+		if condition.Status == newCondition.Status {
+			newCondition.LastTransitionTime = condition.LastTransitionTime
+		} else {
+			log.Debug().Str("mrc", mrc.Name).Msgf("updating last transition time for condition type %s", condition.Type)
```
shalier, line 74: Should the log be in the `if` statement?

jaellio:
Closing in favor of #5201

PR #5148, opened 2022-09-22
Certificate status loops

### LINKED ISSUE #5090: Create timer loop for updating certificateStatus
Defined in [OSM MRC Status Design Doc](https://docs.google.com/document/d/1QF7wL1su4CUzwo3Yo2_vFL90XG2d-wUhBU6uifNrkws/edit#heading=h.sy0tyhbqfylp)

Used for updating `certificateStatuses` to `Validating`, `Issuing`, and on delete.

- [ ] update MRC spec to add upper bounds ???
- [ ] implement a looping pattern to be used by each control plane component. Loop terminates at end of time bound or when the task completes

keithmattix:
**Description**:
Fixes #5090 

**Testing done**:

**Affected area**:

2. Is this a breaking change?

### DIFF
--- a/pkg/certificate/manager.go
+++ b/pkg/certificate/manager.go
@@ -3,14 +3,14 @@ package certificate
 import (
 	"context"
 	"errors"
-	"fmt"
 	"math/rand"
-	"strings"
 	"sync"
 	"time"
 
 	"github.com/cskr/pubsub"
+	"k8s.io/client-go/util/retry"
 
+	"github.com/openservicemesh/osm/pkg/apis/config/v1alpha2"
 	"github.com/openservicemesh/osm/pkg/constants"
 	"github.com/openservicemesh/osm/pkg/errcode"
 	"github.com/openservicemesh/osm/pkg/logger"
@@ -21,11 +21,12 @@ var (
 )
 
 // NewManager creates a new CertificateManager with the passed MRCClient and options
-func NewManager(ctx context.Context, mrcClient MRCClient, getServiceCertValidityPeriod func() time.Duration, getIngressCertValidityDuration func() time.Duration, checkInterval time.Duration) (*Manager, error) {
+func NewManager(ctx context.Context, mrcClient MRCClient, getServiceCertValidityPeriod func() time.Duration, getIngressCertValidityDuration func() time.Duration, checkInterval time.Duration, ownedCertTypes []CertificateType) (*Manager, error) {
 	m := &Manager{
 		serviceCertValidityDuration: getServiceCertValidityPeriod,
 		ingressCertValidityDuration: getIngressCertValidityDuration,
 		pubsub:                      pubsub.New(1),
+		ownedCertTypes:              ownedCertTypes,
 	}
 
 	err := m.start(ctx, mrcClient)
@@ -83,7 +84,7 @@ func (m *Manager) start(ctx context.Context, mrcClient MRCClient) error {
 					return
 				}
 
-				err = m.handleMRCEvent(mrcClient, event)
+				err = m.handleMRCEvent(ctx, mrcClient, event)
 				if err != nil {
 					log.Error().Err(err).Msgf("error encountered processing MRCEvent")
 					continue
@@ -116,10 +117,10 @@ func (m *Manager) start(ctx context.Context, mrcClient MRCClient) error {
 	return nil
 }
 
-func (m *Manager) handleMRCEvent(mrcClient MRCClient, event MRCEvent) error {
+func (m *Manager) handleMRCEvent(ctx context.Context, mrcClient MRCClient, event MRCEvent) error {
 	switch event.Type {
 	case MRCEventAdded:
-		mrc := event.MRC
+		mrc := event.NewMRC
 		if mrc.Status.State == constants.MRCStateError {
 			log.Debug().Msgf("skipping MRC with error state %s", mrc.GetName())
 			return nil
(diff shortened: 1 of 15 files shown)


pkg/certificate/transitions.go
```diff
@@ -0,0 +1,143 @@
...
+func newDeferredCertificateStatusUpdate(wait time.Duration, certificateTypes []CertificateType, mrcClient MRCClient, targetMRCName string, targetCertStatus v1alpha2.MeshRootCertificateComponentStatus) *deferredCertificateStatusUpdate {
+	d := &deferredCertificateStatusUpdate{
+		wait:                    wait,
+		certificateTypes:        certificateTypes,
+		mrcClient:               mrcClient,
+		targetMRCName:           targetMRCName,
+		targetCertificateStatus: targetCertStatus,
+	}
+
+	d.certificateStatusUpdates = map[CertificateType]func() error{
```
steeling, line 54: instead of this map, why not make these all receivers on a type, and have a switch case that calls each method by name based on the cert type? removes a layer of indirection, easier code nav, improved readability

pkg/certificate/manager.go
```diff
@@ -21,11 +20,12 @@ var (
 )
 
 // NewManager creates a new CertificateManager with the passed MRCClient and options
-func NewManager(ctx context.Context, mrcClient MRCClient, getServiceCertValidityPeriod func() time.Duration, getIngressCertValidityDuration func() time.Duration, checkInterval time.Duration) (*Manager, error) {
+func NewManager(ctx context.Context, mrcClient MRCClient, getServiceCertValidityPeriod func() time.Duration, getIngressCertValidityDuration func() time.Duration, checkInterval time.Duration, ownedCertTypes []CertificateType) (*Manager, error) {
```
steeling, line 23: isOwnedCertTypes an all or nothing thing? My guess is yet. What if we simplified for now, and just had a bool for enabling the MRCReconciler

pkg/certificate/transitions.go
steeling, line 54: can we push all of these into a switch statement in genUpdateFunc, which can then just become updateFunc instead of returning a func.

Removes a big layer of indirection and improves readability

pkg/certificate/transitions.go
```diff
@@ -0,0 +1,143 @@
...
+		mrc := d.mrcClient.GetMeshRootCertificate(d.targetMRCName)
+		if mrc == nil {
+			return fmt.Errorf("no MRC found with name %s in the osm control plane namespace", d.targetMRCName)
+		}
+
+		if !set(mrc, d.targetCertificateStatus) {
+			return nil
+		}
+
+		_, err := d.mrcClient.UpdateMeshRootCertificateStatus(mrc)
```
steeling, line 121: this issues N update calls in consecutive order with no waits. That's going to trigger a lot of udpates. 

Is the intent that it does this with no waiting, or is the idea to perform a check between each one? Can you add a comment?

pkg/certificate/transitions.go
```diff
@@ -0,0 +1,143 @@
...
+		wait:                    wait,
+		certificateTypes:        certificateTypes,
+		mrcClient:               mrcClient,
+		targetMRCName:           targetMRCName,
+		targetCertificateStatus: targetCertStatus,
+	}
+
+	d.certificateStatusUpdates = map[CertificateType]func() error{
+		CertificateTypeSidecar: d.genUpdateFunc(func(mrc *v1alpha2.MeshRootCertificate, status v1alpha2.MeshRootCertificateComponentStatus) bool {
+			if mrc.Status.ComponentStatuses.Sidecar == status {
```
steeling, line 56: Can a single controller know that the sidecars are done? Should this not have a mechanism to signal which controllers have seen this complete?

pkg/certificate/transitions.go
```diff
@@ -0,0 +1,143 @@
...
+		}),
+		CertificateTypeBootstrap: d.genUpdateFunc(func(mrc *v1alpha2.MeshRootCertificate, status v1alpha2.MeshRootCertificateComponentStatus) bool {
+			if mrc.Status.ComponentStatuses.Bootstrap == status {
+				return false
+			}
+			mrc.Status.ComponentStatuses.Bootstrap = status
+
+			return true
+		}),
+		CertificateTypeMutatingWebhook: d.genUpdateFunc(func(mrc *v1alpha2.MeshRootCertificate, status v1alpha2.MeshRootCertificateComponentStatus) bool {
```
steeling, line 71: Same here, there could be 10 mutating webhooks

pkg/certificate/manager.go
keithmattix: No, ownedCertTypes is how the top-level control plane command is able to claim ownership over one or more types of certificates (e.g sidecar, Gateway, etc). This is so  the cert manager knows which certificate status fields on the MRC to update

pkg/certificate/transitions.go
keithmattix: Ah good point; we should queue up the calls before making a final update call. Good catch!

pkg/certificate/transitions.go
keithmattix: Because we don't have any coordination between control plane components (or even replicas of the same component), this implementation just wants on a timer to complete and whichever controller replica updates the status first wins

pkg/certificate/manager.go
steeling: but is there ever a use case where a control plane component uses some of these but not all?

keithmattix: Yes, the sets are disjoint. The injector only handles the mutating webhook and bootstrap certs, and the controller owns the rest

keithmattix:
Closing in favor of #5151

PR #5151, opened 2022-09-23
Add framework for updating the MRC

steeling:
A very simple set of skeleton code for updating the MRC.

Future PR's will flesh out each function.


pkg/apis/config/v1alpha2/meshrootcertificate.go
```diff
@@ -171,8 +158,19 @@ type MeshRootCertificateStatus struct {
...
+	// - "xdsControlPlane"
+	// - "sidecar"
+	// - "bootstrap"
+	// - "gateway"
+
+	// The values in the map can be one of (`True`, `False`, `Unknown`), True indicating the value is ready.
+	// NOTE: There is a caveat that since these components belong to horizontally scalable pods, it is possible that not
+	// all of these components will be ready. That is, one controller might mark the ADS server as ready, while all other
+	// controllers have yet to rotate their controller cert.
+	ComponentStatuses map[string]string `json:"componentStatuses"`
```
keithmattix, line 173: -1 on using a map instead of strong types. Not only is this Kubernetes/Go convention, it increases the possibility of typos and inconsistent string mappings throughout the code

pkg/certificate/manager.go
```diff
@@ -16,13 +16,25 @@ import (
...
 )
 
+const (
+	// mrcDurationPerStage is the amount of time we leave each MRC in a stage before moving to the next stage. This is
+	// intended to accommodate the rotation of *all* rotations across all injector and controller pods for:
+	// 1. Bootstrap Cert rotation for each live proxy
+	// 2. Service Cert rotation and xDS push for each connected proxy
+	// 3. xDS server cert rotation on the controller
+	// 4. Mutating and Validating Webhook rotation for the servers and for the webhook configuration objects.
+	mrcDurationPerStage = 5 * time.Minute
```
keithmattix, line 26: This should be added as an MRC field in a followup PR

keithmattix (changes requested):

pkg/apis/config/v1alpha2/meshrootcertificate.go
keithmattix: Link to k8s api conventions: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#lists-of-named-subobjects-preferred-over-maps

steeling: good call! reverted that piece.

pkg/apis/config/v1alpha2/meshrootcertificate.go
```diff
@@ -136,11 +132,12 @@ type MeshRootCertificateConditionType string
 
 // MeshRootCertificateComponentStatuses is the set of statuses for each certificate component in the cluster.
 type MeshRootCertificateComponentStatuses struct {
-	Webhooks        MeshRootCertificateComponentStatus `json:"webhooks"`
-	XDSControlPlane MeshRootCertificateComponentStatus `json:"xdsControlPlane"`
-	Sidecar         MeshRootCertificateComponentStatus `json:"sidecar"`
-	Bootstrap       MeshRootCertificateComponentStatus `json:"bootstrap"`
-	Gateway         MeshRootCertificateComponentStatus `json:"gateway"`
+	ValidatingWebhook *bool `json:"validatingWebhook"`
```
jaellio, line 135: Per Keith's [proposal](https://gist.github.com/keithmattix/9ac73abbcb5721b0ce58102ac3ebff29), I thought the component statuses were going to be updated to the following: 

```go
type CertificateStatus struct {
  CurrentState CertificateStatusState
  QueuedTransition Transition
  Error string
}

type QueuedTransition struct {
  NewState CertificateStatusState
  At time.Time // An RFC 3339 formatted time. The certificateManager in `leaderMode` will perform the transition once it observes the MRC at a time after transition.At 
}
```

pkg/apis/config/v1alpha2/meshrootcertificate.go
```diff
@@ -155,7 +152,7 @@ type MeshRootCertificateCondition struct {
 	// LastTransitionTime is the timestamp corresponding to the last status
 	// change of this condition.
 	// +optional
-	LastTransitionTime *metav1.Time `json:"lastTransitionTime,omitempty"`
+	LastTransitionTime metav1.Time `json:"lastTransitionTime,omitempty"`
```
jaellio, line 155: Could you explain why this is no longer going to be a pointer?

pkg/certificate/mrc_reconciler.go
```diff
@@ -0,0 +1,184 @@
...
+	UseCaseGateway UseCase = "gateway"
+)
+
+func (uc UseCase) String() string {
+	return string(uc)
+}
+func (m *Manager) handleMRCEvent(event MRCEvent) error {
+	mrc := event.MRC
+	// TODO(5046): improve logging in this method.
+	if m.shouldUpdateMRCComponentStatus(mrc) {
```
jaellio, line 40: Could we also add `shouldUpdateMRCCondition`?

pkg/certificate/mrc_reconciler.go
```diff
@@ -0,0 +1,184 @@
...
+
+// Component statuses determine if we're ready to transition the mrc to the next state.
+// TODO(5046): add unit tests for this, and other methods in here.
+func (m *Manager) shouldUpdateMRCComponentStatus(mrc *v1alpha2.MeshRootCertificate) bool {
+	if len(mrc.Status.Conditions) == 0 {
+		return false
+	}
+
+	// Has enough time passed to consider the component status as ready?
+	lastUpdate := mrc.Status.Conditions[len(mrc.Status.Conditions)-1].LastTransitionTime.Time
```
jaellio, line 63: Could you explain the condition indexing here? I think we might want to be checking the `queuedTransition` time (from Keith's proposal) for each `componentStatus`.

pkg/certificate/types.go
```diff
@@ -125,6 +128,7 @@ type Manager struct {
 type MRCClient interface {
 	List() ([]*v1alpha2.MeshRootCertificate, error)
 	MRCEventBroker
+	UpdateMeshRootCertificate(mrc *v1alpha2.MeshRootCertificate) error
```
jaellio, line 131: Could we embed the `compute.Interface` here? We could avoid adding the `UpdateMeshRootCertificate` method signature and also have access to other MRC update/get/list methods

pkg/certificate/mrc_reconciler.go
```diff
@@ -0,0 +1,164 @@
...
+	}
+
+	//The truncate is needed so we truncate the monotonic clock.
+	return transitionAfter.Truncate(0).After(time.Now())
+}
+
+// update the MRCComponentStatus
+func (m *Manager) updateMRCComponentStatus(mrc *v1alpha2.MeshRootCertificate) error {
+	var state v1alpha2.MeshRootCertificateComponentStatus
+	switch mrc.Status.State {
```
keithmattix, line 80: You can't use state like this; we talked about how both `state` and conditions shouldn't be set until all component statuses agree.

keithmattix (changes requested):

pkg/certificate/mrc_reconciler.go
steeling: that was just a variable,  i probably shouldn't have named it state. renamed to status

pkg/certificate/types.go
jaellio: Could you explain for my own understanding why we're not adding the `compute.Interface` here? Thanks!

steeling: it would pull in all the other unused methods. interfaces should be thin, and describe only what is necessary. It makes creating another implementation much easier, and creates a clear definition of exactly what is needed.

Take a look at some of the goals and guiding principles in the refactoring doc https://docs.google.com/document/d/1L_u-XlVU8Fog-uFS3Uf0O2nz8IArQy82cd6Npxm038Y/edit#heading=h.cvqx4tbzt6i7

pkg/certificate/mrc_reconciler.go
```diff
@@ -0,0 +1,161 @@
...
+	// TODO(5046): check if the mrc is in a terminal state.
+	return true
+}
+
+func (m *Manager) updateMRCState(mrc *v1alpha2.MeshRootCertificate) error {
+	// TODO(5046): update the MRC state & conditions, and issue the update via the MRCClient.
+	// add the retry loop as well.
+
+	// If it's not in a terminal state, set the next transition time.
+	if isTerminal(mrc) {
```
keithmattix, line 117: I don't know that `transitionAfter` is tied to terminal states like this. Please refer to the [gist](https://gist.github.com/keithmattix/9ac73abbcb5721b0ce58102ac3ebff29)

cmd/osm-bootstrap/osm-bootstrap.go
```diff
@@ -434,13 +434,6 @@ func (b *bootstrap) createMeshRootCertificate() error {
 
 	createdMRC.Status = configv1alpha2.MeshRootCertificateStatus{
```
keithmattix, line 435: Can we add TODOs about creating the MRC in osm-boostrap per the [gist](https://gist.github.com/keithmattix/9ac73abbcb5721b0ce58102ac3ebff29)

keithmattix (commented):
Two more things, but I think this is ready to go after them

pkg/certificate/types.go
jaellio: Okay, I must have misunderstood this comment on my PR https://github.com/openservicemesh/osm/pull/5073#discussion_r970050870

pkg/certificate/manager.go
```diff
@@ -16,13 +16,25 @@ import (
 	"github.com/openservicemesh/osm/pkg/logger"
 )
 
+const (
+	// mrcDurationPerStage is the amount of time we leave each MRC in a stage before moving to the next stage. This is
+	// intended to accommodate the rotation of *all* rotations across all injector and controller pods for:
+	// 1. Bootstrap Cert rotation for each live proxy
+	// 2. Service Cert rotation and xDS push for each connected proxy
+	// 3. xDS server cert rotation on the controller
+	// 4. Mutating and Validating Webhook rotation for the servers and for the webhook configuration objects.
```
jaellio, line 25: Also ingress gateway certs

pkg/certificate/mrc_reconciler.go
steeling: there would be no next transition time if it's in a terminal state though, no? since there's nothing to transition to

cmd/osm-bootstrap/osm-bootstrap.go
steeling: done

pkg/certificate/manager.go
steeling: good catch!

pkg/certificate/mrc_reconciler.go
keithmattix: ut this blindly adds a transition time even if we haven't determined that componentStatus needs to update based on intent and/or conditions. One of updateMRCState or shouldUpdateMRCComponentStatus needs a TODO to add conditional logic based on the gist (probably the latter function)

steeling: why does it do it blindly? that's not called out anywhere so I don't see the need to make that decision. Not setting a transitionTime if we know there's no future transition makes more sense to me

keithmattix: I'm saying the code here blindly sets the transition. As it stands, there's no place in the PR for osm-controller to check that, e.g. `spec.intent == Passive && status.Conditions[type=Accepted].Status == True && status.Conditions[type=ValidatingRollout] => { x.Status == False && x.Reason == Pending }` on the MRC in order to set the transition time

keithmattix (approved):

pkg/apis/config/v1alpha2/meshrootcertificate.go
```diff
@@ -123,9 +123,25 @@ type TresorCASpec struct {
 type MeshRootCertificateIntent string
 
 // MeshRootCertificateComponentStatus specifies the status of the certificate component,
-// can be (`True`, `False`, `Unknown`).
+// can be (`Issuing`, `Validating`, `Unknown`).
 type MeshRootCertificateComponentStatus string
```
shalier, line 127: Should there also be an `error` status for `MeshRootCertificateComponentStatus`?

jaellio (approved):
Mostly nits. Might need to revisit some things, but want to unblock this change

PR #5178, opened 2022-09-29
Changes renewBefore to use cert validity duration

### LINKED ISSUE #5150: Bump RenewBeforeCertExpires above 30s
RenewBeforeCertExpires is hardcoded to 30s + some jitter between 0-5s. This doesn't give much time to rotate certs on larger implementations.

This should likely be set to a function of 1/3 of the original validity duration of the certificate itself, or something similar.

Otherwise, there's a lot of potential for bugs here:

1. Keep alive set to 60s on client side, doesn't recognize a proxy has disconnected, so it takes at least 60s to reconnect. This would miss an update. 
2. In a large enough fleet, it could take a while for the rotation mechanism to churn through all the certs, particularly if using an external PKI, where we need to issue sync RPC's to an external service. If each call takes ~1s to issue a new cert, and there are more than 30 certs, we're very likely to have some expire.

shalier:
**Description**:
change renewBefore to use a third of the certificate's validity duration plus some noise
and resolves #5150 

**Testing done**:

**Affected area**:

2. Is this a breaking change?

### DIFF
--- a/pkg/certificate/manager.go
+++ b/pkg/certificate/manager.go
@@ -139,12 +139,12 @@ func (m *Manager) GetTrustDomain() string {
 // ShouldRotate determines whether a certificate should be rotated.
 func (m *Manager) ShouldRotate(c *Certificate) bool {
 	// The certificate is going to expire at a timestamp T
-	// We want to renew earlier. How much earlier is defined in renewBeforeCertExpires.
+	// We want to renew earlier. How much earlier is defined as a third of the certificate's validity duration.
 	// We add a few seconds noise to the early renew period so that certificates that may have been
 	// created at the same time are not renewed at the exact same time.
 	intNoise := rand.Intn(noiseSeconds) // #nosec G404
 	secondsNoise := time.Duration(intNoise) * time.Second
-	renewBefore := RenewBeforeCertExpires + secondsNoise
+	renewBefore := m.getValidityDurationForCertType(c.certType)/3 + secondsNoise
 	// Round is called to truncate monotonic clock to the nearest second. This is done to avoid environments where the
 	// CPU clock may stop, resulting in a time measurement that differs significantly from the x509 timestamp.
 	// See https://github.com/openservicemesh/osm/issues/5000#issuecomment-1218539412 for more details.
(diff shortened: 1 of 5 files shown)


pkg/certificate/manager.go
```diff
@@ -139,12 +139,12 @@ func (m *Manager) GetTrustDomain() string {
...
 func (m *Manager) ShouldRotate(c *Certificate) bool {
 	// The certificate is going to expire at a timestamp T
-	// We want to renew earlier. How much earlier is defined in renewBeforeCertExpires.
+	// We want to renew earlier. How much earlier is defined as a third of the certificate's validity duration.
 	// We add a few seconds noise to the early renew period so that certificates that may have been
 	// created at the same time are not renewed at the exact same time.
 	intNoise := rand.Intn(noiseSeconds) // #nosec G404
 	secondsNoise := time.Duration(intNoise) * time.Second
-	renewBefore := RenewBeforeCertExpires + secondsNoise
+	renewBefore := m.getValidityDurationForCertType(c.certType)/3 + secondsNoise
```
steeling, line 147: unfortunately this will break if a user has changed their validity duration. Imagine a time of 24 hours, but then a user changes to 1 week.

Now we have expired certs.

A good heuristic might be this + a floor. So we can say 1/3 or 5 mins (whatever we set here, we should add a minimum accepted validity duration of 2x the renewal period, so 10 mins if 5 mins, or set 5 mins and 2.5 mins).

There's other ways to accomodate this as well, but this seems like a good first step to me

keithmattix (approved):

pkg/utils/mesh_config.go
```diff
@@ -116,6 +117,15 @@ func GetIngressGatewayCertValidityPeriod(mc v1alpha2.MeshConfig) time.Duration {
 		return defaultServiceCertValidityDuration
```
steeling, line 117: might as well fix this here, looks like this got copy-pasted incorrectly and should be defaultIngressGatewayCertValidityDuration

steeling (approved):

PR #5183, opened 2022-09-30
feat(certs): retrieve CA for passive MRC

jaellio:

**Description**:

**Testing done**:

**Affected area**:

2. Is this a breaking change?


pkg/certificate/conditions.go
```diff
@@ -0,0 +1,111 @@
...
+	validatingRollback v1alpha2.MeshRootCertificateConditionType = "ValidatingRollback"
+)
+
+const (
+	trueStatus    v1alpha2.MeshRootCertificateConditionStatus = "True"
+	falseStatus   v1alpha2.MeshRootCertificateConditionStatus = "False"
+	unknownStatus v1alpha2.MeshRootCertificateConditionStatus = "Unknown"
+)
+
+const (
```
keithmattix, line 26: I feel like these Reasons should probably live in the same package as the MRC resource itself

pkg/certificate/conditions.go
```diff
@@ -0,0 +1,111 @@
...
+
+	// condition type not found, append it to condition list
+	mrc.Status.Conditions = append(mrc.Status.Conditions, newCondition)
+	log.Debug().Str("mrc", namespacedMRCName(mrc)).Msgf("added new condition of type %s", conditionType)
+}
+
+func getMRCCondition(mrc *v1alpha2.MeshRootCertificate, conditionType v1alpha2.MeshRootCertificateConditionType) *v1alpha2.MeshRootCertificateCondition {
+	for _, condition := range mrc.Status.Conditions {
+		if condition.Type == conditionType {
+			return &condition
```
keithmattix, line 94: I feel like you probably don't want a pointer here since you care about the underlying value that the condition is set to at the moment, but I could be wrong /cc @steeling for his perspective

pkg/certificate/mrc_reconciler.go
```diff
@@ -170,10 +180,42 @@ func (m *Manager) updateMRCState(mrc *v1alpha2.MeshRootCertificate) error {
...
 	} else {
 		mrc.Status.TransitionAfter = &metav1.Time{Time: time.Now().Add(mrcDurationPerStage)}
+		log.Debug().Str("mrc", namespacedMRCName(mrc)).Msgf("updated MRC TransitionAfter to %s", mrc.Status.TransitionAfter)
+	}
+
+	if shouldRetrieveCA(mrc) {
+		_, _, err := m.mrcClient.GetCertIssuerForMRC(mrc)
+		if err != nil {
+			log.Error().Err(err).Str("mrc", namespacedMRCName(mrc)).Msg("failed to retrieve CA on passive MRC creation")
+			if !m.leaderMode {
```
keithmattix, line 190: Only the leader should attempt to retrieve the CA. There's not necessarily a reason for each control plane component to try and retrieve the CA (but I'm willing to be convinced)

pkg/certificate/mrc_reconciler.go
```diff
@@ -170,10 +180,42 @@ func (m *Manager) updateMRCState(mrc *v1alpha2.MeshRootCertificate) error {
...
+				return err
+			}
+
+			// CA not accepted
+			setMRCCondition(mrc, constants.MRCConditionTypeAccepted, falseStatus, errorRetrievingCAReason, "")
```
keithmattix, line 195: These messages should be set. In the case of an error, the error should be the message. Else, the message should be a more human readable description of why the condition is what it is

pkg/certificate/conditions.go
```diff
@@ -0,0 +1,111 @@
...
+		Type:    conditionType,
+		Status:  conditionStatus,
+		Reason:  string(conditionReason),
+		Message: message,
+	}
+
+	now := metav1.NewTime(time.Now())
+	newCondition.LastTransitionTime = &now
+
+	for i, condition := range mrc.Status.Conditions {
```
keithmattix, line 68: @steeling @jaellio what do you think of an optimization here? Could the certificate manager maintain a map of condition names/types -> condition struct pointers for O(1) updates? The condition slice isn't huge, but multiple `setMRCCondition` calls will result in multiple iterations over the slice

pkg/certificate/mrc_reconciler.go
```diff
@@ -212,3 +258,19 @@ func (m *Manager) setIssuers(mrc *v1alpha2.MeshRootCertificate) error {
 	}
 	return nil
 }
+
+// retryMRCUpdateOnConflict updates an MRC using the specified function with RetryOnConflict
+func (m *Manager) retryMRCUpdateOnConflict(mrc *v1alpha2.MeshRootCertificate, updateMRCFunc func(*v1alpha2.MeshRootCertificate) error) error {
+	return retry.RetryOnConflict(retry.DefaultBackoff, func() error {
```
keithmattix, line 264: Consider using `DefaultRetry` instead per [documentation](https://pkg.go.dev/k8s.io/client-go/util/retry#pkg-variables), but nice work here!

keithmattix (commented):
I just added some (unsolicited) feedback, but this looks great!

pkg/certificate/conditions.go
jaellio: Would we recreate this mapping for each MRC event?

keithmattix: No, we could store it on the certificate manager

keithmattix: Yes!

pkg/certificate/conditions.go
jaellio: pkg/apis/config/v1alpha2 contains primarily generated files. It is acceptable to add non-generated resources to this package?

pkg/certificate/mrc_reconciler.go
jaellio: Updated so that only the leader will attempt to retrieve the CA.

pkg/certificate/mrc_reconciler.go
jaellio: Updated. I might define the messages as constants with the other condition variables https://github.com/openservicemesh/osm/pull/5183#discussion_r985827964

pkg/certificate/conditions.go
jaellio: I might hold off on this optimization for a future PR.

pkg/certificate/mrc_reconciler.go
jaellio: Updated.

pkg/certificate/mrc_reconciler.go
```diff
@@ -170,10 +181,36 @@ func (m *Manager) updateMRCState(mrc *v1alpha2.MeshRootCertificate) error {
 		mrc.Status.TransitionAfter = nil
 	} else {
 		mrc.Status.TransitionAfter = &metav1.Time{Time: time.Now().Add(mrcDurationPerStage)}
+		log.Debug().Str("mrc", namespacedMRCName(mrc)).Msgf("updated MRC TransitionAfter to %s", mrc.Status.TransitionAfter)
+	}
+
+	if m.shouldRetrieveCA(mrc) {
```
jsturtevant, line 187: nit: This name was a little misleading, it is more "ensureIssuerforMRC" I think?

pkg/certificate/conditions.go
jsturtevant: Is this function only used in the tests?

pkg/certificate/conditions.go
```diff
@@ -0,0 +1,106 @@
...
+		if condition.Type == conditionType {
+			return &condition
+		}
+	}
+	return nil
+}
+
+// mrcHasCondition returns true if the given MRC has a condition matching the specified condition
+// Only the type, status, and reason are used in comparison
+func mrcHasCondition(mrc *v1alpha2.MeshRootCertificate, condition v1alpha2.MeshRootCertificateCondition) bool {
```
jsturtevant, line 97: same, is it only used in the test?

pkg/certificate/mrc_reconciler.go
```diff
@@ -57,16 +60,19 @@ func (m *Manager) handleMRCEvent(event MRCEvent) error {
 	log.Err(err).Msg("error setting issuers on mrc")
 	// TODO(5046): improve logging in this method.
 	if shouldUpdateMRCComponentStatus(mrc) {
```
jsturtevant, line 62: When the initial certificate is created, it will flow through this event three times?

pkg/certificate/mrc_reconciler.go
jaellio: Good point, I agree

pkg/certificate/conditions.go
```diff
@@ -0,0 +1,106 @@
...
+	for i, condition := range mrc.Status.Conditions {
+		if condition.Type != newCondition.Type {
+			continue
+		}
+
+		// only update transition time on status change
+		if condition.Status == newCondition.Status {
+			newCondition.LastTransitionTime = condition.LastTransitionTime
+		} else {
+			log.Debug().Str("mrc", namespacedMRCName(mrc)).Msgf("updating last transition time for condition type %s", condition.Type)
```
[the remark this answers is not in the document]
jaellio: I think the wording is correct, but it is confusing. Above in this method we set the `newCondition`'s `LastTransitionTime` to now. If the condition already exists on the MRC and the status of the `newCondition` and existing `condition` are the same, then the `LastTransitionTime` should not be updated so the `newCondition`'s `LastTransitionTime` is set to the current `LastTransitionTime`. Otherwise, the `LastTransitionTime` will be updated (even though this was already done above. I've refactored the code a bit so this will make more sense.

pkg/certificate/conditions.go
jaellio: Yes, I'll remove them and we can add them back in the future if/when needed.

cmd/osm-controller/osm-controller.go
```diff
@@ -206,8 +207,10 @@ func main() {
 	// Intitialize certificate manager/provider
 	var certManager *certificate.Manager
 	if enableMeshRootCertificate {
+		// the osm-controller is responsible for updating the condition and state of the MRC resources
+		leaderMode := true
```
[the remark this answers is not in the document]
jaellio: I think the idea is that the osm-controller (all replicas) will be responsible for updating the conditions and the osm-injectors will just receive events. All of the osm-controllers will be racing to update status but I have wrapped each update in a RetryOnConflict. Maybe there is a better name than `leaderMode` to represent this functionality?

pkg/certificate/conditions.go
```diff
@@ -0,0 +1,106 @@
...
+	errorRetrievingCAReason   = "ErrorRetrievingCA"
+
+	// ValidatingRollout type
+	passiveStateValidatingReason      = "PassiveState"
+	passivelyInUseForValidatingReason = "CertificatePassivelyInUseForValidation"
+	notAcceptedValidatingReason       = "NotAccepted"
+
+	// IssuingRollout type
+	passiveStateIssuingReason      = "PassiveState"
+	passivelyInUseForIssuingReason = "CertificatePassivelyInUseForIssuing"
```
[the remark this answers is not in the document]
jaellio: Agreed. I was trying to define a unique "PassiveState" constant for the ValidatingRollout and IssuingRollout Types and was using "Validating" and "Issuing" to distinguish between the two.

pkg/certificate/mrc_reconciler.go
jaellio: Do you mean the initial MRC on start up?

jsturtevant: something like a new mrc: `Kubectl apply -f new-mrc.yaml`

cmd/osm-controller/osm-controller.go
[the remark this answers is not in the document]
jsturtevant: ok, I'll think on it some more but I think leader isn't the right term here as it invokes leader election which is not happening there. @keithmattix thoughts?

pkg/certificate/conditions.go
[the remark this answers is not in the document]
jsturtevant: looks like this is only Used in a test?  Could kick naming down the road till it is used? I feel like I don't have enough context as to how it is used

pkg/certificate/conditions.go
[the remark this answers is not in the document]
jsturtevant: oh, yes sorry I read it wrong the first time.  thanks!

pkg/certificate/conditions.go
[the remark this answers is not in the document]
keithmattix: PassiveIntent is the reason why IssuingRollout is false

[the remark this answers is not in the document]
jsturtevant: > PassiveIntent is the reason why IssuingRollout is false

I am not sure I follow

pkg/certificate/conditions.go
```diff
@@ -0,0 +1,106 @@
...
+	ready              v1alpha2.MeshRootCertificateConditionType = "Ready"
+	accepted           v1alpha2.MeshRootCertificateConditionType = "Accepted"
+	issuingRollout     v1alpha2.MeshRootCertificateConditionType = "IssuingRollout"
+	validatingRollout  v1alpha2.MeshRootCertificateConditionType = "ValidatingRollout"
+	issuingRollback    v1alpha2.MeshRootCertificateConditionType = "IssuingRollback"
+	validatingRollback v1alpha2.MeshRootCertificateConditionType = "ValidatingRollback"
+)
+
+// MRC condition statuses
+const (
```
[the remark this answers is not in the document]
jaellio: Updated! Using corev1.ConditionStatus instead of creating a new type.

cmd/osm-controller/osm-controller.go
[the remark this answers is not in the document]
keithmattix: Yeah, I didn't put a ton of thought into naming. You're right, leader implies that only one replica is doing the writing which isn't true. Maybe `conditionWriter` is a better name

keithmattix (approved):

pkg/certificate/conditions.go
jaellio: Move condition types to v1alpha2.

pkg/apis/config/v1alpha2/meshrootcertificate.go
```diff
@@ -122,6 +122,14 @@ type TresorCASpec struct {
 // can be (Active, Passive).
 type MeshRootCertificateIntent string
 
+const (
+	// Passive is the passive option for the Intent of the MeshRootCertificate
+	Passive MeshRootCertificateIntent = "passive"
```
keithmattix, line 127: I'd prefer adding prefixes to these variables for readability. `v1alpha2.MRCIntentPassive` is a clearer variable name than `v1alpha2.Passive`

pkg/apis/config/v1alpha2/meshrootcertificate.go
```diff
@@ -142,14 +150,30 @@ const (
...
-// one of (`True`, `False`, `Unknown`).
-type MeshRootCertificateConditionStatus string
-
 // MeshRootCertificateConditionType specifies the type of the condition,
 // one of (`Ready`, `Accepted`, `IssuingRollout`, `ValidatingRollout`, `IssuingRollback`, `ValidatingRollback`).
 type MeshRootCertificateConditionType string
 
+const (
+	// Ready specifies if the MRC is in a ready state.
+	Ready MeshRootCertificateConditionType = "ready"
```
keithmattix, line 159: Same as above, perhaps `MRCConditionTypeReady`?

keithmattix (approved):
These last two comments are nits, but LGTM

pkg/certificate/conditions.go
keithmattix: Addressed in other comments

cmd/osm-controller/osm-controller.go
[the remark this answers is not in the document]
jaellio: Updated `leaderMode` to `conditionWriter`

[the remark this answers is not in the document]
keithmattix: This needs to be `conditionWriter`, no?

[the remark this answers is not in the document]
jsturtevant: What happens if one of the controllers that doesn't have a lot of work (maybe not a lot of proxies attached) is the first to report that things are ready to go?

pkg/certificate/mrc_reconciler.go
jsturtevant: If I am reading this correctly, a user creates a new MRC with `intent=passive` then then the componentstatus is set, this triggers a change firing another event, which then sets the state which then fires another event which finally sets the issuers?  Otherwise this if statement won't do anything other than set componentstatus on the first submision by the user since all these if's are nested.

It seems to me we should be able to do all that in go?

pkg/certificate/conditions.go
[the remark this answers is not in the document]
jsturtevant: ok, though `CertificatePassivelyInUseForIssuing` doesn't algin with "MRC shouldn't be used to issue new certs?".

jaellio:
Closing in favor of #5201

PR #5186, opened 2022-10-04
feat(certs): adds check to move componentStatuses to Validating

### LINKED ISSUE #5181: Move each componentStatus to ValidatingRollout
[updateMRCComponentStatus](https://github.com/openservicemesh/osm/blob/main/pkg/certificate/mrc_reconciler.go#L131), should see that the MRC in question has the following attributes:

```spec.intent == Passive && status.Conditions[type=Accepted].Status == True && status.Conditions[type=ValidatingRollout] => { x.Status == False && x.Reason == Pending }```

And then move the componentStatuses to Validating based on ownedUseCases.

### PR DESCRIPTION
**Pull request by shalier**
**Description**:
Adds check that the MRC meets the attribute required to move the component statuses to Validating and resolves #5181

**Testing done**:

**Affected area**:

2. Is this a breaking change?

### DIFF
--- a/pkg/certificate/mrc_reconciler.go
+++ b/pkg/certificate/mrc_reconciler.go
@@ -1,6 +1,7 @@
 package certificate
 
 import (
+	"strings"
 	"time"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -130,7 +131,12 @@ func shouldUpdateState(mrc *v1alpha2.MeshRootCertificate) bool {
 // update the MRCComponentStatus
 func (m *Manager) updateMRCComponentStatus(mrc *v1alpha2.MeshRootCertificate) error {
 	// TODO(5046): determine the status to set the owned components to.
-	var status v1alpha2.MeshRootCertificateComponentStatus
+	// default component status is Unknown
+	var status v1alpha2.MeshRootCertificateComponentStatus = constants.MRCComponentStatusUnknown
+
+	if shouldChangeStatusToValidating(mrc) {
+		status = constants.MRCComponentStatusValidating
+	}
 
 	for _, useCase := range m.ownedUseCases {
 		switch useCase {
@@ -151,6 +157,33 @@ func (m *Manager) updateMRCComponentStatus(mrc *v1alpha2.MeshRootCertificate) er
 	}
 
 	return m.mrcClient.UpdateMeshRootCertificate(mrc)
+}
+
+// checks if the given MRC meets the attributes to move the componentStatuses to Validating
+func shouldChangeStatusToValidating(mrc *v1alpha2.MeshRootCertificate) bool {
+	// case-insensitive check for if strings are equal
+	if strings.EqualFold(string(mrc.Spec.Intent), constants.MRCIntentPassive) {
+		acceptedCond := getCondition(mrc, constants.MRCConditionTypeAccepted)
+
+		if strings.EqualFold(string(acceptedCond.Status), constants.MRCConditionStatusTrue) {
+			validRolloutCond := getCondition(mrc, constants.MRCConditionTypeValidatingRollout)
+
+			if strings.EqualFold(string(validRolloutCond.Status), constants.MRCConditionStatusFalse) && strings.EqualFold(validRolloutCond.Reason, constants.MRCConditionReasonPending) {
+				return true
+			}
+		}
+	}
+	return false
+}
+
+// returns the MeshRootCertificateCondition that matches the condtion type specified
+func getCondition(mrc *v1alpha2.MeshRootCertificate, condType v1alpha2.MeshRootCertificateConditionType) *v1alpha2.MeshRootCertificateCondition {
+	for _, condition := range mrc.Status.Conditions {
+		if condition.Type == condType {
+			return &condition
+		}
+	}
+	return nil
 }
 
 // isTerminal checks if the MRC is in a terminal state.
--- a/pkg/certificate/mrc_reconciler.go
+++ b/pkg/certificate/mrc_reconciler.go
@@ -131,12 +131,7 @@ func shouldUpdateState(mrc *v1alpha2.MeshRootCertificate) bool {
 // update the MRCComponentStatus
 func (m *Manager) updateMRCComponentStatus(mrc *v1alpha2.MeshRootCertificate) error {
 	// TODO(5046): determine the status to set the owned components to.
-	// default component status is Unknown
-	var status v1alpha2.MeshRootCertificateComponentStatus = constants.MRCComponentStatusUnknown
-
-	if shouldChangeStatusToValidating(mrc) {
-		status = constants.MRCComponentStatusValidating
-	}
+	status := getNextStatus(mrc)
 
 	for _, useCase := range m.ownedUseCases {
 		switch useCase {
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: pkg/certificate/mrc_reconciler.go [hunk 1]
```diff
@@ -153,6 +159,33 @@ func (m *Manager) updateMRCComponentStatus(mrc *v1alpha2.MeshRootCertificate) er
 	return m.mrcClient.UpdateMeshRootCertificate(mrc)
 }
 
+// checks if the given MRC meets the attributes to move the componentStatuses to Validating
+func shouldChangeStatusToValidating(mrc *v1alpha2.MeshRootCertificate) bool {
+	// case-insensitive check for if strings are equal
+	if strings.EqualFold(string(mrc.Spec.Intent), constants.MRCIntentPassive) {
+		acceptedCond := getCondition(mrc, constants.MRCConditionTypeAccepted)
+
+		if strings.EqualFold(string(acceptedCond.Status), constants.MRCConditionStatusTrue) {
```
steeling, line 168: this can result in a nil pointer dereference, from getCondition returning nil

pkg/certificate/mrc_reconciler.go
```diff
@@ -153,6 +159,33 @@ func (m *Manager) updateMRCComponentStatus(mrc *v1alpha2.MeshRootCertificate) er
...
+			if strings.EqualFold(string(validRolloutCond.Status), constants.MRCConditionStatusFalse) && strings.EqualFold(validRolloutCond.Reason, constants.MRCConditionReasonPending) {
+				return true
+			}
+		}
+	}
+	return false
+}
+
+// returns the MeshRootCertificateCondition that matches the condtion type specified
+func getCondition(mrc *v1alpha2.MeshRootCertificate, condType v1alpha2.MeshRootCertificateConditionType) *v1alpha2.MeshRootCertificateCondition {
```
steeling, line 180: is there exactly 1 condition at a given condition type? If not, I'd consider renaming this func, or changing its signature to refelect that

pkg/certificate/mrc_reconciler.go
```diff
@@ -153,6 +159,33 @@ func (m *Manager) updateMRCComponentStatus(mrc *v1alpha2.MeshRootCertificate) er
 	return m.mrcClient.UpdateMeshRootCertificate(mrc)
 }
 
+// checks if the given MRC meets the attributes to move the componentStatuses to Validating
+func shouldChangeStatusToValidating(mrc *v1alpha2.MeshRootCertificate) bool {
+	// case-insensitive check for if strings are equal
+	if strings.EqualFold(string(mrc.Spec.Intent), constants.MRCIntentPassive) {
+		acceptedCond := getCondition(mrc, constants.MRCConditionTypeAccepted)
```
steeling, line 166: suggestion: the code should either declare its intent through intuitive function names, or add comments explaining how/why its performing these checks. From reading this I'm not sure what it's supposed to be doing

pkg/certificate/mrc_reconciler.go
```diff
@@ -153,6 +159,33 @@ func (m *Manager) updateMRCComponentStatus(mrc *v1alpha2.MeshRootCertificate) er
 	return m.mrcClient.UpdateMeshRootCertificate(mrc)
 }
 
+// checks if the given MRC meets the attributes to move the componentStatuses to Validating
+func shouldChangeStatusToValidating(mrc *v1alpha2.MeshRootCertificate) bool {
+	// case-insensitive check for if strings are equal
+	if strings.EqualFold(string(mrc.Spec.Intent), constants.MRCIntentPassive) {
```
steeling, line 165: can we prefer early returns. That is, instead of 

```go
if ... {
   if {
      ....
   }
}
return false
```

do this:

```go
if !... {
   return false
}
if ... {
  ...
}
```

pkg/certificate/mrc_reconciler.go
```diff
@@ -130,7 +131,12 @@ func shouldUpdateState(mrc *v1alpha2.MeshRootCertificate) bool {
 // update the MRCComponentStatus
 func (m *Manager) updateMRCComponentStatus(mrc *v1alpha2.MeshRootCertificate) error {
 	// TODO(5046): determine the status to set the owned components to.
-	var status v1alpha2.MeshRootCertificateComponentStatus
+	// default component status is Unknown
+	var status v1alpha2.MeshRootCertificateComponentStatus = constants.MRCComponentStatusUnknown
+
+	if shouldChangeStatusToValidating(mrc) {
```
steeling, line 137: I wonder if we can create a generic implementation for,

`getNextStatus`, which simply returns the next status we should set it too, instead of having 1 per status. Might not be able to, but just a thought

pkg/certificate/mrc_reconciler.go
keithmattix: Yes there should not be multiple statuses of the same type

steeling: gotcha, and is there only 1 condition type with status set to True at any given time?

keithmattix: No there may be multiple types set to True at the same time

pkg/certificate/mrc_reconciler.go
```diff
@@ -130,7 +131,12 @@ func shouldUpdateState(mrc *v1alpha2.MeshRootCertificate) bool {
 // update the MRCComponentStatus
 func (m *Manager) updateMRCComponentStatus(mrc *v1alpha2.MeshRootCertificate) error {
 	// TODO(5046): determine the status to set the owned components to.
-	var status v1alpha2.MeshRootCertificateComponentStatus
+	// default component status is Unknown
+	var status v1alpha2.MeshRootCertificateComponentStatus = constants.MRCComponentStatusUnknown
```
jaellio, line 135: nit: I think this could be simplified to the following:
```suggestion
	status := constants.MRCComponentStatusUnknown
```

pkg/certificate/mrc_reconciler.go
```diff
@@ -153,6 +159,33 @@ func (m *Manager) updateMRCComponentStatus(mrc *v1alpha2.MeshRootCertificate) er
 	return m.mrcClient.UpdateMeshRootCertificate(mrc)
 }
 
+// checks if the given MRC meets the attributes to move the componentStatuses to Validating
+func shouldChangeStatusToValidating(mrc *v1alpha2.MeshRootCertificate) bool {
```
jaellio, line 163: Could we add a unit test for `shouldChangeStatusToValidating`?

pkg/certificate/mrc_reconciler.go
shalier: the `constants.MRCComponentStatusUnknown` is a string so I set it the above way so that it can be type `v1alpha2.MeshRootCertificateComponentStatus`. I think instead I'll change the constants to be their specific type

pkg/certificate/mrc_component_status.go
```diff
@@ -0,0 +1,42 @@
...
+)
+
+type (
+	unknownStatus    v1alpha2.MeshRootCertificateComponentStatus
+	validatingStatus v1alpha2.MeshRootCertificateComponentStatus
+	issuingStatus    v1alpha2.MeshRootCertificateComponentStatus
+	unusedStatus     v1alpha2.MeshRootCertificateComponentStatus
+)
+
+func getNextStatus[R any](status v1alpha2.MeshRootCertificateComponentStatus,
```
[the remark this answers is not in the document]
shalier: simplified it, removed the parameterization of the functions and use of generics

pkg/certificate/mrc_reconciler.go
```diff
@@ -152,6 +153,81 @@ func (m *Manager) updateMRCComponentStatus(mrc *v1alpha2.MeshRootCertificate) er
...
+	if !strings.EqualFold(string(mrc.Spec.Intent), constants.MRCIntentPassive) {
+		return false
+	}
+	// checks if Accepted condition has status true
+	if !isConditionTrue(mrc, constants.MRCConditionTypeAccepted) {
+		return false
+	}
+
+	// checks if ValidatingRollout condition has status false with reason Pending
+	if isConditionFalse(mrc, constants.MRCConditionTypeValidatingRollout) && isReasonPending(mrc, constants.MRCConditionTypeValidatingRollout) {
```
jsturtevant, line 196: Should this be the same reason after https://github.com/openservicemesh/osm/pull/5183/files#diff-ee5985129a778d4127e4cf7da9b939021e6e3e7d78106a950391d3a0b96b793eR208?  (I think this two PR are working in the same area?)

I am not sure I understand why we need to do this in two phases.  After it is added the the cert validated, shouldn't we be able to mark it as Validating in the linked PR?  

I think this adds another event change to https://github.com/openservicemesh/osm/pull/5183/files#r988421071  Off of a single submission to the user? 

Since this is done by _each_ osm-controller pod, the first one done could kick off this set of events and another one could still be processing but we are reporting to the user that things are in a good state that isn't true for all the osm-controllers?

jaellio:
Closing for now in favor of #5201. Will revisit in milestone 1.4

PR #5189, opened 2022-10-04
fix(envoy): add keep-alive time to detect sidecar disconnections

### LINKED ISSUE #5149: Set sidecar keep alive
https://github.com/envoyproxy/envoy/issues/6323

We’ve seen customer issues where a sidecar disconnects from the control plane irregularly but thinks it’s still connected, resulting in certificates expiring and requests failing. 

We should also see if the snapshot cache enables keep Alives so we can detect disconnections on the server side. 

This is not verified as the cause of the issue but my leading theory. 

https://github.com/envoyproxy/envoy/issues/6323

SanyaKochhar:

**Description**: Defines a 60s keep-alive time after which probes will start being sent to check if a sidecar is still connected to the control plane. Previously this defaulted to 2 hours

Resolves #5149

**Testing done**:

**Affected area**:

3. Has documentation corresponding to this change been updated in the [osm-docs](https://github.com/openservicemesh/osm-docs) repo (if applicable)? n/a

### DIFF
--- a/pkg/envoy/bootstrap/config.go
+++ b/pkg/envoy/bootstrap/config.go
@@ -16,6 +16,7 @@ import (
 	xds_discovery "github.com/envoyproxy/go-control-plane/envoy/service/discovery/v3"
 	"github.com/golang/protobuf/ptypes/any"
 	"google.golang.org/protobuf/types/known/anypb"
+	"google.golang.org/protobuf/types/known/wrapperspb"
 
 	"github.com/openservicemesh/osm/pkg/constants"
 	"github.com/openservicemesh/osm/pkg/envoy"
@@ -276,6 +277,13 @@ func (b *Builder) Build() (*xds_bootstrap.Bootstrap, error) {
 									},
 								},
 							},
+						},
+					},
+					UpstreamConnectionOptions: &xds_cluster.UpstreamConnectionOptions{
+						TcpKeepalive: &xds_core.TcpKeepalive{
+							KeepaliveProbes:   wrapperspb.UInt32(5),
+							KeepaliveTime:     wrapperspb.UInt32(60),
+							KeepaliveInterval: wrapperspb.UInt32(5),
 						},
 					},
 				},


pkg/envoy/bootstrap/config.go
```diff
@@ -278,6 +278,11 @@ func (b *Builder) Build() (*xds_bootstrap.Bootstrap, error) {
 							},
 						},
 					},
+					UpstreamConnectionOptions: &xds_cluster.UpstreamConnectionOptions{
+						TcpKeepalive: &xds_core.TcpKeepalive{
+							KeepaliveTime: uint32(60),
```
SanyaKochhar, line 283: @steeling we are additionally able to define the keepalive `probes` and `interval` values, does this seem necessary? We haven't defined these for grpc because go's grpc pkg doesn't have these options. If not they will default as follows:

image

pkg/envoy/bootstrap/config.go
```diff
@@ -278,6 +279,11 @@ func (b *Builder) Build() (*xds_bootstrap.Bootstrap, error) {
 							},
 						},
 					},
+					UpstreamConnectionOptions: &xds_cluster.UpstreamConnectionOptions{
```
steeling, line 282: so reading the defaults for the other values present here we have:

* 60s of no activity before sending a keep-alive
* 9 failed probes before terminating
* 75s between probes

that gives us 660s, or 11 minutes before we disconnect, which feels too long. I'd recommend we bring this down to 5 probes and 5 seconds, which gives us 80s to terminate which is much better.

SanyaKochhar: That sounds reasonable. Did you mean 5 second interval or 5 second before the first probe goes out?

steeling: 5s interval

steeling (approved):

keithmattix (approved):

steeling:
there's a flaky test that was recently fixed, can you rebase against main?

PR #5191, opened 2022-10-05
Fix Namespace selector & ensure correct webhook version

jsturtevant:
The name space selectors must be in a specific format https://github.com/kubernetes/kubernetes/issues/92157#issuecomment-825160958. This was causing the webhook to not run validation.

Once those are turned on there is a bug in the upgrade e2e test that leave a webhook around becuase the version of the webhook label is not upgraded and there for not removed by our tooling.

This was found during #5182

**Description**:

**Testing done**:

**Affected area**:
| CI System                  | [x] |
| CLI Tool                   | [x] |
| Upgrade                    | [x] |

no

2. Is this a breaking change?
no 

4. Has documentation corresponding to this change been updated in the [osm-docs](https://github.com/openservicemesh/osm-docs) repo (if applicable)?
no

### DIFF
--- a/pkg/injector/webhook.go
+++ b/pkg/injector/webhook.go
@@ -306,7 +306,7 @@ func createOrUpdateMutatingWebhook(clientSet kubernetes.Interface, cert *certifi
 							Operator: metav1.LabelSelectorOpDoesNotExist,
 						},
 						{
-							Key:      "name",
+							Key:      "kubernetes.io/metadata.name",
 							Operator: metav1.LabelSelectorOpNotIn,
 							Values:   []string{osmNamespace},
 						},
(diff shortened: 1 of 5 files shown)


keithmattix (approved):

jsturtevant:
hmm go test failure looks unrelated.

jaellio (approved):
I think the test failure is related to https://github.com/openservicemesh/osm/pull/5192

steeling:
We should probably add this `createOrUpdateMutatingWebhook` as well

jsturtevant:
> `createOrUpdateMutatingWebhook`

you are right, I meant to do this....

pkg/injector/webhook.go
```diff
@@ -344,7 +344,12 @@ func createOrUpdateMutatingWebhook(clientSet kubernetes.Interface, cert *certifi
 				return err
 			}
 
-			mwhc.ObjectMeta = existing.ObjectMeta // copy the object meta which includes resource version, required for updates.
+			// copy the object meta which includes resource version, required for updates.
+			// but use the OSM Version, which might be different if this is an upgrade
```
shashankram, line 348: It would be nice to have an example in the comments to make the comment clearer.

jsturtevant: something like ?
```suggestion
			// copy the object meta which includes resource version, required for updates.
			// but use the OSM Version, which might be different if this is an upgrade.  For example, 
			// the webhook might have been initially created by version 1.1.2 and the version osm is 1.2.2
```

shashankram: It was a bit hard for me to understand the following:
`required for updates`: what updates? (This is an existing comment prior to your change, but seems ambiguous)
`but use the OSM Version which might be different if this is an upgrade`: It's not clear what upgrade this is referring to - k8s, control plane, etc.

Just reading the comments, I found it hard to make sense of the relation between object meta and OSM version, in this conext.

jsturtevant: oh yes that makes sense.  Let me see if I can come up with something clearer

shashankram: Thanks. This comment isn't blocking, so let me know if you want to merge this as is.

PR #5193, opened 2022-10-06
Use both trust domains if they are different

### LINKED ISSUE #5036: Set SAN Matcher for multiple Certs during a cert rotation
If an MRC changes the trust domain (or in the future switches to a SPIFEE ID), then the envoy config needs to have the DNS SAN matcher set to include both the new and old principal.

We should validate this is correct with E2E tests.

jsturtevant:
When the certificate manager has more than one
trust domain we need to ensure that the services know about both of them since the either could be used.
This usually occurs during a certificate rotation.

fixes #5036 

**Description**:

**Testing done**:

**Affected area**:

2. Is this a breaking change?


pkg/catalog/inbound_traffic_policies.go
```diff
@@ -195,10 +195,14 @@ func (mc *MeshCatalog) getRoutingRulesFromTrafficTarget(trafficTarget access.Tra
 	}
 
 	// Compute the allowed downstream service identities for the given TrafficTarget object
-	trustDomain := mc.GetTrustDomain()
+	trustDomains := mc.certManager.GetTrustDomains()
 	allowedDownstreamPrincipals := mapset.NewSet()
 	for _, source := range trafficTarget.Spec.Sources {
-		allowedDownstreamPrincipals.Add(trafficTargetIdentityToSvcAccount(source).AsPrincipal(trustDomain))
+		allowedDownstreamPrincipals.Add(trafficTargetIdentityToSvcAccount(source).AsPrincipal(trustDomains.Signing))
```
steeling, line 201: nit: feels a bit more idiomatic to have it return a slice of trustdomains

steeling (approved):

pkg/certificate/manager.go
```diff
@@ -128,12 +128,24 @@ func (m *Manager) start(ctx context.Context, mrcClient MRCClient) error {
...
+// AreDifferent returns true if the signing and validating trust domains are different
+func (td TrustDomain) AreDifferent() bool {
+	return td.Signing != td.Validating
+}
+
+// GetTrustDomains returns the trust domains from the configured issuers.
 // Note that the CRD uses a default, so this value will always be set.
-func (m *Manager) GetTrustDomain() string {
+// It is up to the caller to determine if the signing and validating trust domains are different
+func (m *Manager) GetTrustDomains() TrustDomain {
```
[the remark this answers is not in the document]
jsturtevant: not sure I follow, in what sense?

pkg/catalog/inbound_traffic_policies.go
jsturtevant: not sure I understand how it would be miss used but get the idiomatic point.  my thought was I didn't want to loop through a slice since we knew the values upfront.  I am likely going to modify this for #5160.  Maybe can try something different there?  Do you have any thoughts on how to looping through the slice?

pkg/certificate/manager.go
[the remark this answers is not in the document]
steeling: This gets called a few times in the path for generating an envoy config, so it's order magnitude on # of proxies over frequency of proxy broadcasts.

Technically also impacted by # of traffic targets, so in a fully connected mesh in non-permissive mode it can be substantially higher (O(NM)).

pkg/catalog/inbound_traffic_policies.go
steeling: just a simple double loop, one over principles (sources in this line), and another over the slice of trust domains returned, ie:

```go
for _, source := range trafficTarget.Spec.Sources {
  for _, td := range trustDomains {
     allowedDownstreamPrincipals.Add(trafficTargetIdentityToSvcAccount(source).AsPrincipal(td))
  }
}
```

jsturtevant: hmm, since https://github.com/openservicemesh/osm/pull/5193#discussion_r993584627, I am leaning against this.  Would it be simpler if I got rid of trustdomian struct and just return the two strings and do a compare everywhere?

keithmattix (approved):

PR #5194, opened 2022-10-06
envoy: use access log builder and Telemetry config

shashankram:
**Description**:
This change does the following:

- Adds a builder type to build access log config.

- Allows overriding the default access log format using the Telemetry API. Both text based and JSON formats are supported.

- Updates the API used to fetch the Telemetry config to also return associated OpenTelemetry ExtensionService config. This is used to build the OpenTelemetry access log config. (Note: the OpenTelemetry integration isn't complete yet)

- Updates the health probe config code to leverage the access log builder. Previously, the code was duplicated.

- Adds code to allow Telemetry object changes to trigger proxy updates.

- Removes legacy access log related code.

Part of #5136

**Testing done**:
Tested custom access log configuration such as:

```
apiVersion: policy.openservicemesh.io/v1alpha1
kind: Telemetry
metadata:
  name: httpbin-log2
  namespace: httpbin
spec:
  selector:
    app: httpbin
  accessLog:
    format: '{"authority":"%REQ(:AUTHORITY)%","bytes_received":"%BYTES_RECEIVED%","bytes_sent":"%BYTES_SENT%"}'
```

```
apiVersion: policy.openservicemesh.io/v1alpha1
kind: Telemetry
metadata:
  name: httpbin-log
  namespace: httpbin
spec:
  accessLog:
    format: "[%START_TIME%] \"%REQ(:METHOD)% %REQ(X-ENVOY-ORIGINAL-PATH?:PATH)% %PROTOCOL%\" %RESPONSE_CODE% %RESPONSE_FLAGS%\n"
```

**Affected area**:


pkg/envoy/generator/lds/accesslog.go
```diff
@@ -0,0 +1,231 @@
...
+
+type accessLogBuilder struct {
+	name                    string
+	format                  string
+	openTelemetryCluster    string
+	openTelemetryAttributes map[string]string
+}
+
+// NewAccessLogBuilder returns an accessLogBuilder instance used to build access log configuration
+func NewAccessLogBuilder() *accessLogBuilder { //nolint: revive // unexported-return
```
steeling, line 57: Can we add some tests that given a accessLogBuilder with a set of parameters, tests the `Build` method?

steeling (approved):

PR #5195, opened 2022-10-10
feat(cerst): Adds MRCResyncInterval

### LINKED ISSUE #5176: Add resync interval to MRC informer
The `MeshRootCertificate` informers need to have a resync interval in order to provide a retry mechanism for event handlers. This can be done in `pkg/informers/informers.go` when the MRC informer is created in `WithConfigClient`

shalier:
**Description**:
Adds MRCResyncInterval which is half the MrcDurationPerStage and resolves #5176

**Testing done**:

**Affected area**:

2. Is this a breaking change?

### DIFF
--- a/pkg/k8s/client.go
+++ b/pkg/k8s/client.go
@@ -2,6 +2,7 @@ package k8s
 
 import (
 	"context"
+	"time"
 
 	smiAccess "github.com/servicemeshinterface/smi-sdk-go/pkg/apis/access/v1alpha3"
 	smiSpecs "github.com/servicemeshinterface/smi-sdk-go/pkg/apis/specs/v1alpha4"
@@ -570,6 +571,6 @@ func (c *Client) ListServiceExports() []*mcs.ServiceExport {
 }
 
 // AddMeshRootCertificateEventHandler adds an event handler specific to mesh root certificiates.
-func (c *Client) AddMeshRootCertificateEventHandler(handler cache.ResourceEventHandler) {
-	c.informers[informerKeyMeshRootCertificate].AddEventHandler(handler)
+func (c *Client) AddMeshRootCertificateEventHandler(handler cache.ResourceEventHandler, resyncInterval time.Duration) {
+	c.informers[informerKeyMeshRootCertificate].AddEventHandlerWithResyncPeriod(handler, resyncInterval)
 }
--- a/pkg/k8s/informers.go
+++ b/pkg/k8s/informers.go
@@ -19,7 +19,6 @@ import (
 	mcsClient "sigs.k8s.io/mcs-api/pkg/client/clientset/versioned"
 	mcsInformers "sigs.k8s.io/mcs-api/pkg/client/informers/externalversions"
 
-	"github.com/openservicemesh/osm/pkg/certificate"
 	"github.com/openservicemesh/osm/pkg/constants"
 	configClientset "github.com/openservicemesh/osm/pkg/gen/client/config/clientset/versioned"
 	configInformers "github.com/openservicemesh/osm/pkg/gen/client/config/informers/externalversions"
@@ -81,9 +80,6 @@ const (
 	// This is set to 0 because we do not need resyncs from k8s client.
 	// For the MeshConfig resource, we have our own Ticker to turn on periodic resyncs.
 	DefaultKubeEventResyncInterval = 0 * time.Second
-	// MRCResyncInterval is the resync interval to provide a retry mechanism for MRC event handlers
-	// It is half of the MrcDurationPerStage (the amount of time we leave each MRC in a stage before moving to the next stage)
-	MRCResyncInterval = certificate.MrcDurationPerStage / 2
 )
 
 var (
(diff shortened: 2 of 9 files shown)


keithmattix (approved):

pkg/k8s/informers.go
```diff
@@ -144,7 +148,7 @@ func WithConfigClient(configClient configClientset.Interface) ClientOption {
 			opt.FieldSelector = fields.OneTermEqualSelector(metav1.ObjectNameField, c.meshConfigName).String()
 		})
 		meshConfiginformerFactory := configInformers.NewSharedInformerFactoryWithOptions(configClient, DefaultKubeEventResyncInterval, configInformers.WithNamespace(c.osmNamespace), listOption)
-		mrcInformerFactory := configInformers.NewSharedInformerFactoryWithOptions(configClient, DefaultKubeEventResyncInterval, configInformers.WithNamespace(c.osmNamespace))
+		mrcInformerFactory := configInformers.NewSharedInformerFactoryWithOptions(configClient, MRCResyncInterval, configInformers.WithNamespace(c.osmNamespace))
```
steeling, line 151: @keithmattix  what're your thoughts on adding this to the event handler instead of the underlying informer? That way if anything else wants to use the informer it is consistent with the user across other informers.

This would mean exposing the `AddEventHandlerWithResyncPeriod` instead of the AddEventHandler that is called in the MRCComposer's Watch method

keithmattix: Hmm good question....is there any use of the MRC informer that shouldn't have this resync?

steeling: i think consistency across these is more important, so a future feature doesn't get surprised. surprises are bad

keithmattix: Not sure that constitutes a "surprise"; in my opinion, the bigger surprise would be that the behavior of a similar feature (e.g. MRC) is different than a future feature. While one should generally try to avoid surprises, I feel like "surprises are bad" overly simplifies the nuance that accompanies these kinds of conversations

jsturtevant: >That way if anything else wants to use the informer it is consistent with the user across other informers.

are you suggesting leaving this at zero except if we add a second handler as we are doing here?  

If we are we using our "own ticker" (not sure where this is implemented) is changing the default resync going to conflict?

pkg/k8s/informers.go
```diff
@@ -78,8 +79,11 @@ const (
 const (
 	// DefaultKubeEventResyncInterval is the default resync interval for k8s events
 	// This is set to 0 because we do not need resyncs from k8s client, and have our
-	// own Ticker to turn on periodic resyncs.
+	// own Ticker to turn on periodic resyncs for the MeshConfig resource.
```
jsturtevant, line 82: This default ticker is used for more than the just meshconfig?

pkg/k8s/informers.go
```diff
@@ -78,8 +79,11 @@ const (
...
 	DefaultKubeEventResyncInterval = 0 * time.Second
+	// MRCResyncInterval is the resync interval to provide a retry mechanism for MRC event handlers
+	// It is half of the MrcDurationPerStage (the amount of time we leave each MRC in a stage before moving to the next stage)
+	MRCResyncInterval = certificate.MrcDurationPerStage / 2
```
jsturtevant, line 86: how was 2.5 mins chosen?

pkg/k8s/informers.go
jsturtevant: I found it: https://github.com/openservicemesh/osm/blob/68765c5158ed575f1785f6cee3a07d753be2bd9f/pkg/ticker/ticker.go#L55

shalier: @jsturtevant I was trying to say that the ticker that we made for periodic resyncs was used just for the MeshConfig resource. Not that the DefaultKubeEventResyncInterval is used just for MeshConfig.
Let me know if the below change makes more sense!
```
	// DefaultKubeEventResyncInterval is the default resync interval for k8s events
	// This is set to 0 because we do not need resyncs from k8s client.
	// For the MeshConfig resource, we have our own Ticker to turn on periodic resyncs.
```

pkg/k8s/informers.go
shalier: Talked with @steeling and decided 1/2 of what we recommend for setting the transition time.

pkg/k8s/informers.go
jsturtevant: yes, it took me a moment to understand being new to this area.  The new suggested comment makes it much clearer!

pkg/k8s/informers.go
steeling: I'm suggesting to keep the informers resync at 0 to keep consistency (I would be surprised if I had used 10+ other of our informer uses, all with no resync, but then i have a use case for that one, and it has a resync).

But instead add a method that exposes the AddEventHandlerWithResyncPeriod, so the caller is aware of the resync and explicitly states it when adding the handler.

jsturtevant: makes sense to me

keithmattix: Cool, I don't care too much one way or another.

pkg/k8s/client.go
```diff
@@ -571,5 +576,5 @@ func (c *Client) ListServiceExports() []*mcs.ServiceExport {
 
 // AddMeshRootCertificateEventHandler adds an event handler specific to mesh root certificiates.
 func (c *Client) AddMeshRootCertificateEventHandler(handler cache.ResourceEventHandler) {
-	c.informers[informerKeyMeshRootCertificate].AddEventHandler(handler)
+	c.informers[informerKeyMeshRootCertificate].AddEventHandlerWithResyncPeriod(handler, MRCResyncInterval)
```
steeling, line 579: nit: i might add the resync interval as a param to the method, and let the caller decide.

steeling (approved):

pkg/k8s/informers.go
jsturtevant: @steeling @shalier @keithmattix This ends up resulting in `The specified resyncPeriod 2m30s is invalid because this shared informer doesn't support resyncing`.  

I tested it out and the base informer needs to have a resync enabled (aka a value > 1s) otherwise it resync is disabled. https://github.com/kubernetes/client-go/blob/c81636cd4415f0858a99c2cc18023740df7aa15f/tools/cache/shared_informer.go#L516-L524

We've been running the tests in https://github.com/openservicemesh/osm/pull/5201 without resync so I don't know that we need it?

steeling: yes resync was only necessary when we were doing time based changes

Issue #5158, opened 2022-09-27
Remove pkg/debugger dependency on k8s.Controller


steeling:
The debugger package shouldn't need a dependency on k8s.Controller. Currently it calls GetPodForProxy, and uses the pod info to port-forward.

ie: `PortForwardToProxy(ctx context.Context, proxy *envoy.Proxy, int localPort) (close func(), error)`. This should replace the calls in pkg/debugger, such that we can remove the k8s.Controller for the debug struct.

steeling:
The `PortForwardToProxy` method should exist on the compute client in `pkg/compute/kube`

steeling:
The `PortForwardToProxy` method should exist on the compute client in `pkg/compute/kube`

Issue #4926, opened 2022-07-20
Normalize trafficpolicy objects by decoupling from envoy configs


steeling:
Subtask of #4862. 

Currently traffic policy objects are coupled with envoy configs in that they contain denormalized data, to more closely represent the envoy config. However, do to domain separation, they should have no knowledge of envoy configs, and let the construction responsibility belong with the builders.

An example of this is in `trafficpolicy.InboundTrafficPolicy`, which has a list of `trafficpolicy.Rules`. 

The rules have a set of `AllowedPrincipals`, however the `AllowedPrincipals` are identical for every `rule` within an `InboundTrafficPolicy`. This should be normalized, and the `AllowedPrincipals` should be pulled into the `InboundTrafficPolicy`, and out of the `Rule`.

This makes it more clear to the reader how these permutations may exist within our own constructs, and decouples from envoy.

Note that there may not be too many opportunities for this decoupling -- the opportunities will likely present themselves while doing other refactoring.

PR #5199, opened 2022-10-12
Decouple debugger and k8s.Controller

nshankar13:

**Description**: This PR decouples the k8s controller from the debug config. Instead of implementing GetPodForProxy in the k8s client, we now use ConfigFromProxy in the compute/kube client. 

This, as well as other methods moved from k8s and debugger to compute do not have functional changes, aside from getEnvoyConfig - which now uses the k8s portforward method instead of its own portforward method.

Resolves: https://github.com/openservicemesh/osm/issues/5158

**Testing done**:

**Affected area**:

2. Is this a breaking change?


pkg/compute/kube/client.go
```diff
@@ -442,10 +472,135 @@ func (c *client) GetProxyStatsHeaders(p *models.Proxy) (map[string]string, error
...
+	}
+
+	return policy
+}
+
+func key(name, namespace string) string {
+	return types.NamespacedName{Name: name, Namespace: namespace}.String()
+}
+
+func (c *client) getExtensionService(svc policyv1alpha1.ExtensionServiceRef) *configv1alpha2.ExtensionService {
```
steeling, line 583: Let's move this method back into the k8s client. Remember that K8s client is responsible for simple crud operations like Get (by name/namespace), List, update, and delete. This falls into that category

pkg/compute/kube/client.go
```diff
@@ -442,10 +472,135 @@ func (c *client) GetProxyStatsHeaders(p *models.Proxy) (map[string]string, error
...
+// order of preference: 1. selector match, 2. namespace match, 3. global match
+func (c *client) getTelemetryPolicy(proxy *models.Proxy) *policyv1alpha1.Telemetry {
+	pod, _ := c.GetPodForProxy(proxy)
+	if pod == nil {
+		return nil
+	}
+
+	var policy *policyv1alpha1.Telemetry
+
+	for _, resource := range c.kubeController.List(informerKeyTelemetry) {
```
steeling, line 541: we should not be exporting informer key internals. The kubecontroller may not always use an informer

pkg/debugger/server.go
```diff
@@ -42,14 +42,14 @@ func (ds DebugConfig) GetHandlers() map[string]http.Handler {
...
 	return DebugConfig{
-		certDebugger:   certDebugger,
-		xdsDebugger:    xdsDebugger,
-		meshCatalog:    meshCatalog,
-		proxyRegistry:  proxyRegistry,
-		kubeClient:     kubeClient,
-		kubeController: kubeController,
+		certDebugger:  certDebugger,
+		xdsDebugger:   xdsDebugger,
+		meshCatalog:   meshCatalog,
```
steeling, line 49: can we remove the meshcatalog here as well?

pkg/k8s/client.go
```diff
@@ -463,7 +404,7 @@ func (c *Client) ListTCPTrafficSpecs() []*smiSpecs.TCPRoute {
 // GetTCPRoute returns an SMI TCPRoute resource given its name of the form <namespace>/<name>
 func (c *Client) GetTCPRoute(namespacedName string) *smiSpecs.TCPRoute {
 	// client-go cache uses <namespace>/<name> as key
-	routeIf, exists, err := c.getByKey(informerKeyTCPRoute, namespacedName)
+	routeIf, exists, err := c.GetByKey(informerKeyTCPRoute, namespacedName)
```
steeling, line 407: please keep all of these methods unexported.

pkg/k8s/client.go
```diff
@@ -27,7 +24,7 @@ import (
 func NewClient(osmNamespace, meshConfigName string, broker *messaging.Broker, opts ...ClientOption) (*Client, error) {
 	// Initialize client object
 	c := &Client{
-		informers:      map[informerKey]cache.SharedIndexInformer{},
+		informers:      map[InformerKey]cache.SharedIndexInformer{},
```
steeling, line 27: let's keep the informerkey's unexported as well

pkg/osm/callbacks.go
```diff
@@ -45,7 +45,7 @@ func (cp *ControlPlane[T]) ProxyConnected(ctx context.Context, connectionID int6
 
 	proxy := models.NewProxy(kind, uuid, si, utils.GetIPFromContext(ctx), connectionID)
 
-	if err := cp.catalog.VerifyProxy(proxy); err != nil {
+	if _, err := cp.catalog.GetPodForProxy(proxy); err != nil {
```
steeling, line 48: Let's unexport GetPodForProxy, and keep verifyproxy

pkg/compute/kube/client.go
```diff
@@ -442,10 +472,135 @@ func (c *client) GetProxyStatsHeaders(p *models.Proxy) (map[string]string, error
...
+		log.Error().Str(errcode.Kind, errcode.GetErrCodeWithMetric(errcode.ErrPodBelongsToMultipleServices)).
+			Msgf("Found more than one pod with label %s = %s in namespace %s. There can be only one!",
+				constants.EnvoyUniqueIDLabelName, proxyUUID, svcAccount.Namespace)
+		return nil, errMoreThanOnePodForUUID
+	}
+
+	pod := pods[0]
+	log.Trace().Msgf("Found Pod with UID=%s for proxyID %s", pod.ObjectMeta.UID, proxyUUID)
+
+	if pod.Namespace != svcAccount.Namespace {
```
steeling, line 510: We can move these validations to verify proxy, and out of GetPodForProxy if we want. Either way

pkg/osm/callbacks.go
nshankar13: Debugger needs GetPodForProxy, I think we need to keep this exported.

nshankar13: In which case we probably don't need verifyproxy.

pkg/debugger/server.go
nshankar13: Discussed offline, will create a follow up task for this.

pkg/compute/types.go
```diff
@@ -76,9 +79,17 @@ type Interface interface {
 
 	GetProxyStatsHeaders(p *models.Proxy) (map[string]string, error)
 
-	// VerifyProxy attempts to lookup a pod that matches the given proxy instance by service identity, namespace, and UUID
-	VerifyProxy(proxy *models.Proxy) error
+	// GetPodForProxy returns the pod that the given proxy is attached to, based on the UUID and service identity.
+	GetPodForProxy(proxy *models.Proxy) (*v1.Pod, error)
```
steeling, line 83: let's remove this, the compute client should not leak k8s objects.

pkg/compute/kube/client.go
```diff
@@ -442,10 +464,117 @@ func (c *client) GetProxyStatsHeaders(p *models.Proxy) (map[string]string, error
 	}, nil
 }
 
-// VerifyProxy attempts to lookup a pod that matches the given proxy instance by service identity, namespace, and UUID.
-func (c *client) VerifyProxy(proxy *models.Proxy) error {
-	_, err := c.kubeController.GetPodForProxy(proxy)
-	return err
+// GetPodForProxy returns the pod that the given proxy is attached to, based on the UUID and service identity.
+func (c *client) GetPodForProxy(proxy *models.Proxy) (*v1.Pod, error) {
```
steeling, line 468: let's unexport this

pkg/k8s/client.go
```diff
@@ -321,6 +262,15 @@ func (c *Client) ListIngressBackendPolicies() []*policyv1alpha1.IngressBackend {
 	return backends
 }
 
+// GetExtensionService returns the extension service for the given service ref
+func (c *Client) GetExtensionService(svc policyv1alpha1.ExtensionServiceRef) *configv1alpha2.ExtensionService {
```
steeling, line 266: I'd consider changing this to accept a name, namespace string

pkg/osm/callbacks.go
steeling: My understanding is that the purpose of this PR was to add an abstraction to remove that necessity, through a new method `PortForwardToProxy`

pkg/k8s/client.go
shashankram: I'd keep this as is for convenience of directly passing the ExtensionServiceRef from the Telemetry resource.

pkg/compute/kube/client.go
```diff
@@ -442,10 +464,117 @@ func (c *client) GetProxyStatsHeaders(p *models.Proxy) (map[string]string, error
...
+func (c *client) GetTelemetryConfig(proxy *models.Proxy) models.TelemetryConfig {
+	var config models.TelemetryConfig
+
+	config.Policy = c.getTelemetryPolicy(proxy)
+
+	if config.Policy != nil && config.Policy.Spec.AccessLog != nil && config.Policy.Spec.AccessLog.OpenTelemetry != nil {
+		config.OpenTelemetryService = c.kubeController.GetExtensionService(config.Policy.Spec.AccessLog.OpenTelemetry.ExtensionService)
+	}
+
+	return config
```
shashankram, line 577: It's not clear why these functions are being moved. Could you describe this further?

shashankram (changes requested):

pkg/compute/kube/client.go
steeling: the k8s package is intended to not know anything about OSM objects. It's intent is to operate strictly on K8s objects, while the compute/kube package is aware of how to translate between the 2.

The k8s package should have simple crud-like methods for each object for Get, Update, List, Create, etc.

pkg/osm/callbacks.go
nshankar13: `PortForwardToProxy` would just call and return the unexported `getPodForProxy`, in which case I think it would just make sense to keep GetPodForProxy as an exported method and call that instead of having an additional method for `PortForwardToProxy`.

Regardless of whether we do `PortForwardToProxy` or `GetPodForProxy`, the debugger is still decoupled from the k8s client.

pkg/compute/kube/client.go
nshankar13: GetTelemetryConfig calls getTelemetryPolicy, which calls getPodForProxy. Since GetPodForProxy is moving to the kube client, I think it makes sense for the GetTelemetry methods to move there also - otherwise we would need to call computeClient methods from the k8s client.

pkg/compute/kube/client.go
```diff
@@ -9,22 +9,39 @@ import (
...
+	// errServiceAccountDoesNotMatchProxy is an error for when the service account of a Pod does not match the xDS certificate.
+	errServiceAccountDoesNotMatchProxy = fmt.Errorf("service account does not match proxy")
+
+	// errMoreThanOnePodForUUID is an error for when OSM finds more than one pod for a given xDS certificate. There should always be exactly one Pod for a given xDS certificate.
+	errMoreThanOnePodForUUID = fmt.Errorf("found more than one pod for xDS uuid")
+
+	// errNamespaceDoesNotMatchProxy is an error for when the namespace of the Pod does not match the xDS certificate.
+	errNamespaceDoesNotMatchProxy = fmt.Errorf("namespace does not match proxy")
+)
+
```
shashankram, line 44: When there's no format specifier, use `errors.New()`

pkg/compute/kube/client.go
shashankram: I would argue these Telemetry related methods are unrelated to "Compute". These need to sit in the same pkg as other `Getters` (e.g. GetUpstreamTrafficSetting). If you look at this API, it does an aggregated GET on OSM API resources given some input. 

Also moving the getPodForProxy method should not be a reason to move the Telemetry helpers. Putting this into the "Compute" interface seems highly misleading and hard to reason.

steeling: It's just about walls of abstraction. The plan is to not leak OSM resources into the k8s package, which is for simple crud operations. This allows us to hide complexities of informers, or non-informer based crud operations here, while the compute package is responsible for translating to/from OSM <-> k8s.

The entire purpose of that package is exactly the reason it should be placed there. Maybe compute is a bad name, and we can look to change it?

Note that the `GetUpstreamTrafficSetting` is exported in the Compute package as well.

shashankram: @steeling I get your point. MeshCatalog was meant to translate k8s objects to internal OSM objects. It seems like we have this new Compute interface doing the same. Maybe those can be collapsed into the Catalog?

steeling: Ya I agree the 2 are a bit similar, but let me add some more reasoning and we can decide how we want to proceed from there.

Let's assume the following:
1. Interfaces exist to decouple from implementations, allowing us to leverage other implementations in the future.
2. With that in mind, it makes the most sense to push as much business logic out of specific implementations, to allow that business logic to apply to all instances of an interface. This prevents us from having to reimplement business logic.

With the above assumptions, the intent of each package is:

1. The k8s.Controller (package `k8s`): abstracts *how* we get objects from kubernetes. This may be an informer, a direct query to the k8s master, or, for example in a future where we leverage remote control planes, a local cache of k8s objects.
2. The compute client abstracts the *what*, that is, it abstracts Kubernetes itself. It allows a future where we may leverage running OSM outside of kubernetes
3. The mesh catalog contains just the business logic. With that, there should never be multiple instances, and we should remove the interface itself, and have everything directly rely on the MeshCatalog (as opposed to the MeshCatalogger).

This allows us to not need to re-implement the business logic that exists in the mesh catalog if we were to move off of kubernetes, or query kubernetes in a different manner.

shashankram: The mesh catalog was meant to abstract Kubernetes. So I suggest we simplify these layers of abstraction if the catalog and compute client are meant to do the same. Currently, it's hard to make sense of what goes into the compute client vs the catalog, and it seems a bit arbitrary.

steeling: The meshcatalog contains *a lot* of business logic. That would make the undertaking significantly bigger if we ever wanted to implement a non-k8s client, and we had to implement all the functionality of the mesh catalog. It seems much simpler to keep the layers I described above and *only* require an implementation of the logic specific to the compute client. I disagree that it's arbitrary, in fact there's a simple set of rules to decide:

1. If a method deals with both OSM and K8s objects -> compute client
2. If a method *only* deals with k8s objects -> k8s client
3. If a method *only* deals with OSM objects -> mesh catalog.

I suspect a better approach may be to begin deconstructing some of the methods in the meshcatalog (to eventually remove it), such that the xDS builders are more aware of the underlying fields. For example:

```go
// GetOutboundMeshClusterConfigs returns the cluster configs for the outbound mesh traffic policy for the given downstream identity
func (mc *MeshCatalog) GetOutboundMeshClusterConfigs(downstreamIdentity identity.ServiceIdentity) []*trafficpolicy.MeshClusterConfig {
	var clusterConfigs []*trafficpolicy.MeshClusterConfig

	for _, meshSvc := range mc.ListOutboundServicesForIdentity(downstreamIdentity) {
		meshSvc := meshSvc // To prevent loop variable memory aliasing in for loop

		// ---
		// Create the cluster config for this upstream service
		clusterConfigForServicePort := &trafficpolicy.MeshClusterConfig{
			Name:                          meshSvc.EnvoyClusterName(),
			Service:                       meshSvc,
			EnableEnvoyActiveHealthChecks: mc.GetMeshConfig().Spec.FeatureFlags.EnableEnvoyActiveHealthChecks,
			UpstreamTrafficSetting:        mc.GetUpstreamTrafficSettingByService(&meshSvc),
		}
		clusterConfigs = append(clusterConfigs, clusterConfigForServicePort)
	}

	return clusterConfigs
}
```

The xDS builders/generators could really do the following:

```go
for _, svc := range c.ListOutboundServicesForIdentity(identity) {
   builder.AddOutboundService(svc, c.GetUpstreamTrafficSettingByService(&svc))
}
builder.SetEnableEnvoyHealthChecks(c.GetMeshConfig().Spec.FeatureFlags.EnableEnvoyActiveHealthChecks)
```

This removes the need to have an intermediate `MeshClusterConfig` type, as well as the method in the mesh catalog (and maybe eventually the mesh catalog itself).

I've called this out in #4926 and #4863, but I think time will tell more how much we can remove

pkg/compute/kube/port_forward.go
```diff
@@ -30,12 +34,26 @@ type portForward struct {
 	Ready chan struct{}
 }
 
-func (ds DebugConfig) forwardPort(req portForward) {
+// PortForwardToProxy takes the given proxy and port forwards to the pod from this proxy
+func (c *client) PortForwardToProxy(proxy *models.Proxy, w http.ResponseWriter, url string, kubeConfig *rest.Config) {
```
steeling, line 38: this api is not very flexible; for instance, it can't be used by a CLI since it requires writing to an http.ResponseWriter. How can we make this more flexible?

Also, the name `PortForwardToProxy` doesn't describe the inputs, outputs, but instead describes the internals of *how* it's achieving getting the config. I'd suggest changing that as well.

An alternative would be to either return the envoy config, or simply setup the port forwarding, and let the client call out to localhost.


Also notice that there is a port-forwarder already in `pkg/k8s/port-forward.go`, it likely makes sense to reuse this

pkg/compute/kube/port_forward.go
```diff
@@ -0,0 +1,23 @@
...
+import (
+	"fmt"
+
+	"k8s.io/client-go/rest"
+
+	"github.com/openservicemesh/osm/pkg/models"
+)
+
+// EnvoyConfigFromProxy takes the given proxy, port forwards to the pod from this proxy, and returns the envoy config
+func (c *client) EnvoyConfigFromProxy(proxy *models.Proxy, url string, kubeConfig *rest.Config) (string, error) {
```
steeling, line 12: nit: rename to `ConfigFromProxy`, change `url` to `configType`, and move this to `client.go`

steeling:
@nshankar13 friendly reminder to avoid stomping prior commits, so your reviewers can view just what has changed between reviews

pkg/compute/kube/client.go
```diff
@@ -444,10 +468,137 @@ func (c *client) GetProxyStatsHeaders(p *models.Proxy) (map[string]string, error
...
+// ConfigFromProxy takes the given proxy, port forwards to the pod from this proxy, and returns the envoy config
+func (c *client) ConfigFromProxy(proxy *models.Proxy, configType string, kubeConfig *rest.Config) (string, error) {
+	var envoyConfig string
+
+	pod, err := c.getPodForProxy(proxy)
+	if err != nil {
+		msg := fmt.Sprintf("Error getting Pod from proxy %s", proxy)
+		log.Error().Err(err).Msg(msg)
+		return envoyConfig, err
```
steeling, line 538: nit: it's more clear to your reader if you return the empty string directly, vs. forcing the reader to back-check that it's not set. It's minor, but little things like that help

steeling (approved):

pkg/compute/kube/client.go
```diff
@@ -444,10 +468,135 @@ func (c *client) GetProxyStatsHeaders(p *models.Proxy) (map[string]string, error
...
+			Msgf("Pod with UID=%s belongs to ServiceAccount=%s. The pod's xDS certificate was issued for ServiceAccount=%s",
+				pod.ObjectMeta.UID, pod.Spec.ServiceAccountName, svcAccount)
+		return nil, errServiceAccountDoesNotMatchProxy
+	}
+
+	return &pod, nil
+}
+
+// ConfigFromProxy takes the given proxy, port forwards to the pod from this proxy, and returns the envoy config
+func (c *client) ConfigFromProxy(proxy *models.Proxy, configType string, kubeConfig *rest.Config) (string, error) {
```
shashankram, line 531:
```suggestion
func (c *client) GetProxyConfig(proxy *models.Proxy, configType string, kubeConfig *rest.Config) (string, error) {
```

pkg/k8s/informers.go
```diff
@@ -26,52 +26,52 @@ import (
...
 	// informerKeyTelemetry lookup identifier
 	informerKeyTelemetry informerKey = "Telemetry"
-	// informerKeyExtensionService is the informerKey for an ExtensionService informer
+	// informerKeyExtensionService is the informerKeyfor an ExtensionService informer
 	informerKeyExtensionService informerKey = "ExtensionService"
-	// informerKeyServiceImport is the informerKey for a ServiceImport informer
+	// informerKeyServiceImport is the informerKeyfor a ServiceImport informer
 	informerKeyServiceImport informerKey = "ServiceImport"
-	// informerKeyServiceExport is the informerKey for a ServiceExport informer
+	// informerKeyServiceExport is the informerKeyfor a ServiceExport informer
```
shashankram, line 74: typos: s/informerKeyFor/informer key for/g

shashankram (changes requested):

shashankram:
Could you also indicate in the commit && PR description if there are functional changes to methods moved across files?

shashankram (approved):

Issue #4998, opened 2022-08-15
Root certificate rotation MVP


jaellio:
#Get OSM root certificate rotation in a working state to be tested and receive feedback from customers.

MVP Goals
- Support manual root certificate rotation through status updates using the kubectl --subresource command 
  - Nice to have: CLI commands to navigate rotation stages
- Support happy path rotation and reverting rotation
  - Progression between rotation stages, no MRC deletion
- Metrics for certs in cache rotated (including issuer and validator IDs)
  - Nice to have % distribution metrics
- Certificate rotation documentation
- Happy path certificate rotation demo

Required items:
- [x] #4839
- [x] #4818 
- [ ] #4890 
- [x] #4962 
- [x] #4815 
- [x] #4723 
- [x] #5044
- [x] #5045
- [x] #5046

jsturtevant:
For more details, please review the [Simplified Proposal for Manual Root Cert Rotation MVP](https://docs.google.com/document/d/1bh3Yfn5DXWjmzdKa4XHlk-9KepUNb_EgDDrzd1G_deY/edit#heading=h.86n6n6h1rqmo)

Issue #4835, opened 2022-06-21
Create e2e tests for root certificate rotation


jaellio:
**Please describe the Improvement and/or Feature Request**

Create e2e tests to verify root certificate rotation. 

Scenarios to verify: 
- For each rotation stage verify the following
  - issued certificates are in the expected state (issued by the expected issuer and validated by the expected issuer(s))
  - ADS server is using expected certs
  - webhooks are using expected certs
  - webhook ca bundles contain the expected cert
  - OSM bootstrap secrets are set to the expected certs

Outstanding questions:
- Should the e2e tests also validate the validation performed by the validating webhook?

**Scope (please mark with X where applicable)**
- Tests                      [x]

PR #5201, opened 2022-10-13
feat(cert): use intent to set issuers during root cert rotation

### LINKED ISSUE #5198: Create MVP for manual root certificate rotation

Create a process by which OSM users can manually rotate the root certificate without any downtime. This issue only focuses on creating an MVP and not covering all possible edge cases.

For more details, please review the [Simplified Proposal for Manual Root Cert Rotation MVP](https://docs.google.com/document/d/1bh3Yfn5DXWjmzdKa4XHlk-9KepUNb_EgDDrzd1G_deY/edit#heading=h.86n6n6h1rqmo)

**Possible use cases**

Allow users to rotate root certificates without down time. This can be done using the MRC Intent field and a level triggered approach.

jaellio:
**Description**:

The PR includes the changes required to support an MVP of manual root certificate rotation. With these changes, a user will be able to manually create and update the `intent` of the MRCs and rotate the root certificate without downtime. OSM utilizes a level-triggered approach to bring the state of the mesh to match the MRC intent.

There are 3 possible values for intent. Each describes the desired role of the provider and certificate settings specified in the MRC:
- **active**: the certificate provider issues and signs certificates and the ca_bundle for that provider is included in the validation context
- **passive**: the certificate provider is not used to issue or sign certificates and the ca_bundle for the provider is included in the validation context
- **inactive**: the certificate provider is not used to issue or sign certificates and it is not included in the validation context

To rotate the root certificate or update the certificate provider, the user must perform the following actions:
There is an existing MRC, MRC1 and a new MRC will be created, MRC2
1. create a new MRC with `inactive` intent -> MRC1.intent = active, MRC2.intent = inactive
2. update the intent of the new MRC to `passive`. Wait to allow the certificates to be rotated -> MRC1.intent = active, MRC2.intent = passive
3. update the intent of the new MRC to `active` -> MRC1.intent = active, MRC2.intent = active
4. update the intent of the old MRC to `passive`. Wait to allow the certificates to be rotated -> MRC1.intent = passive, MRC2.intent = active
5. update the intent of the old MRC to `inactive` -> MRC1.intent = inactive, MRC2.intent = active

Note: All combinations of intent are deterministic except for `active` and `active`. The order in which the MRCs are listed determines which will be the `signingIssuer` or the `validatingIssuer`. 

Initial proposal for root certificate rotation simplification: https://docs.google.com/document/d/1bh3Yfn5DXWjmzdKa4XHlk-9KepUNb_EgDDrzd1G_deY/edit?usp=sharing

Resolves #5198 
Part of #4998 

**Testing done**:
- e2es testing rotation with tresor, cert-manager, and vault
- unit tests

**Affected area**:
| New Functionality          | [ x ] |
| Certificate Management     | [ x ] |

7. Is this a breaking change? No

8. Has documentation corresponding to this change been updated in the [osm-docs](https://github.com/openservicemesh/osm-docs) repo (if applicable)? No

### DIFF
--- a/cmd/osm-bootstrap/crds/config_mesh_root_certificate.yaml
+++ b/cmd/osm-bootstrap/crds/config_mesh_root_certificate.yaml
@@ -37,9 +37,9 @@ spec:
       served: true
       storage: true
       additionalPrinterColumns:
-        - description: Current state of the MeshRootCertificate config
-          jsonPath: .status.state
-          name: State
+        - description: Intent for the MeshRootCertificate
+          jsonPath: .spec.intent
+          name: Intent
           type: string
       schema:
         openAPIV3Schema:
@@ -48,6 +48,7 @@ spec:
             spec:
               type: object
               required:
+                - intent
                 - provider
               properties:
                 trustDomain:
@@ -55,12 +56,12 @@ spec:
                   type: string
                   default: cluster.local
                 intent:
-                  description: Intent specifies the role of the MeshRootCertificate, can be active or passive
+                  description: Intent specifies the role of the MeshRootCertificate, can be active, passive
                   type: string
                   enum:
                     - passive
                     - active
-                  default: passive
+                    - inactive
                 spiffeEnabled:
                   description: Adds a SPIFFE ID to the certificates, creating a SPIFFE compatible x509 SVID document
                   type: boolean
--- a/cmd/osm-bootstrap/osm-bootstrap.go
+++ b/cmd/osm-bootstrap/osm-bootstrap.go
@@ -423,20 +423,11 @@ func (b *bootstrap) createMeshRootCertificate() error {
 	if err != nil {
 		return err
 	}
-	createdMRC, err := b.configClient.ConfigV1alpha2().MeshRootCertificates(b.namespace).Create(context.TODO(), defaultMeshRootCertificate, metav1.CreateOptions{})
+	_, err = b.configClient.ConfigV1alpha2().MeshRootCertificates(b.namespace).Create(context.TODO(), defaultMeshRootCertificate, metav1.CreateOptions{})
 	if apierrors.IsAlreadyExists(err) {
 		log.Info().Msgf("MeshRootCertificate already exists in %s. Skip creating.", b.namespace)
 		return nil
 	}
-	if err != nil {
-		return err
-	}
-
-	_, err = b.configClient.ConfigV1alpha2().MeshRootCertificates(b.namespace).UpdateStatus(context.Background(), createdMRC, metav1.UpdateOptions{})
-	if apierrors.IsAlreadyExists(err) {
-		log.Info().Msgf("MeshRootCertificate status already exists in %s. Skip creating.", b.namespace)
-	}
-
 	if err != nil {
 		return err
 	}
--- a/cmd/osm-bootstrap/osm-bootstrap.go
+++ b/cmd/osm-bootstrap/osm-bootstrap.go
@@ -49,7 +49,6 @@ const (
 	meshConfigName                   = "osm-mesh-config"
 	presetMeshConfigName             = "preset-mesh-config"
 	presetMeshConfigJSONKey          = "preset-mesh-config.json"
-	meshRootCertificateName          = "osm-mesh-root-certificate"
 	presetMeshRootCertificateName    = "preset-mesh-root-certificate"
 	presetMeshRootCertificateJSONKey = "preset-mesh-root-certificate.json"
 )
@@ -168,7 +167,7 @@ func main() {
 	if enableMeshRootCertificate {
 		err = bootstrap.ensureMeshRootCertificate()
 		if err != nil {
-			log.Fatal().Err(err).Msgf("Error setting up default MeshRootCertificate %s from ConfigMap %s", meshRootCertificateName, presetMeshRootCertificateName)
+			log.Fatal().Err(err).Msgf("Error setting up default MeshRootCertificate %s from ConfigMap %s", constants.DefaultMeshRootCertificateName, presetMeshRootCertificateName)
 			return
 		}
 	}
@@ -433,36 +432,6 @@ func (b *bootstrap) createMeshRootCertificate() error {
 		return err
 	}
 
-	createdMRC.Status = configv1alpha2.MeshRootCertificateStatus{
-		State: constants.MRCStatePending,
-		Conditions: []configv1alpha2.MeshRootCertificateCondition{
-			{
-				Type:   constants.MRCConditionTypeReady,
-				Status: constants.MRCConditionStatusUnknown,
-			},
-			{
-				Type:   constants.MRCConditionTypeAccepted,
-				Status: constants.MRCConditionStatusUnknown,
-			},
-			{
-				Type:   constants.MRCConditionTypeIssuingRollout,
-				Status: constants.MRCConditionStatusUnknown,
-			},
-			{
-				Type:   constants.MRCConditionTypeValidatingRollout,
-				Status: constants.MRCConditionStatusUnknown,
-			},
-			{
-				Type:   constants.MRCConditionTypeIssuingRollback,
-				Status: constants.MRCConditionStatusUnknown,
-			},
-			{
-				Type:   constants.MRCConditionTypeValidatingRollback,
-				Status: constants.MRCConditionStatusUnknown,
-			},
-		},
-	}
-
 	_, err = b.configClient.ConfigV1alpha2().MeshRootCertificates(b.namespace).UpdateStatus(context.Background(), createdMRC, metav1.UpdateOptions{})
 	if apierrors.IsAlreadyExists(err) {
 		log.Info().Msgf("MeshRootCertificate status already exists in %s. Skip creating.", b.namespace)
@@ -472,7 +441,7 @@ func (b *bootstrap) createMeshRootCertificate() error {
 		return err
 	}
 
-	log.Info().Msgf("Successfully created MeshRootCertificate %s in %s.", meshRootCertificateName, b.namespace)
+	log.Info().Msgf("Successfully created MeshRootCertificate %s in %s.", constants.DefaultMeshRootCertificateName, b.namespace)
 	return nil
 }
 
@@ -490,7 +459,7 @@ func buildMeshRootCertificate(presetMeshRootCertificateConfigMap *corev1.ConfigM
 			APIVersion: "config.openservicemesh.io/configv1alpha2",
 		},
 		ObjectMeta: metav1.ObjectMeta{
-			Name: meshRootCertificateName,
+			Name: constants.DefaultMeshRootCertificateName,
 		},
 		Spec: presetMeshRootCertificateSpec,
 	}
--- a/pkg/apis/config/v1alpha2/meshrootcertificate.go
+++ b/pkg/apis/config/v1alpha2/meshrootcertificate.go
@@ -122,6 +122,19 @@ type TresorCASpec struct {
 // can be (Active, Passive).
 type MeshRootCertificateIntent string
 
+const (
+	// ActiveIntent means the settings and certificate provider in this MRC are used for signing and
+	// validating certificates.
+	ActiveIntent MeshRootCertificateIntent = "active"
+
+	// PassiveIntent means the settings and certificate provider in this MRC are used for validating
+	// certificates.
+	PassiveIntent MeshRootCertificateIntent = "passive"
+
+	// InactiveIntent means the settings and certificate provider in this MRC no longer in use.
+	InactiveIntent MeshRootCertificateIntent = "inactive"
+)
+
 // MeshRootCertificateComponentStatus specifies the status of the certificate component,
 // can be (`Issuing`, `Validating`, `Unknown`).
 type MeshRootCertificateComponentStatus string
--- a/pkg/catalog/inbound_traffic_policies_test.go
+++ b/pkg/catalog/inbound_traffic_policies_test.go
@@ -1,6 +1,7 @@
 package catalog
 
 import (
+	"context"
 	"fmt"
 	"reflect"
 	"testing"
@@ -14,15 +15,16 @@ import (
 	tassert "github.com/stretchr/testify/assert"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
 
 	"github.com/openservicemesh/osm/pkg/apis/config/v1alpha2"
 	policyv1alpha1 "github.com/openservicemesh/osm/pkg/apis/policy/v1alpha1"
 	tresorFake "github.com/openservicemesh/osm/pkg/certificate/providers/tresor/fake"
 	"github.com/openservicemesh/osm/pkg/compute/kube"
 	"github.com/openservicemesh/osm/pkg/constants"
-	"github.com/openservicemesh/osm/pkg/k8s"
-
+	configFake "github.com/openservicemesh/osm/pkg/gen/client/config/clientset/versioned/fake"
 	"github.com/openservicemesh/osm/pkg/identity"
+	"github.com/openservicemesh/osm/pkg/k8s"
 	"github.com/openservicemesh/osm/pkg/service"
 	"github.com/openservicemesh/osm/pkg/tests"
 	"github.com/openservicemesh/osm/pkg/trafficpolicy"
@@ -2388,10 +2390,24 @@ func TestGetInboundMeshTrafficPolicy(t *testing.T) {
 			mockCtrl := gomock.NewController(t)
 			defer mockCtrl.Finish()
 
-			mrc := tresorFake.NewFakeMRC()
-			fakeCertManager := tresorFake.NewFakeWithMRC(mrc, 1*time.Hour)
+			mrc1 := &v1alpha2.MeshRootCertificate{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "osm-mesh-root-certificate",
+					Namespace: "osm-system",
+				},
+				Spec: v1alpha2.MeshRootCertificateSpec{
+					TrustDomain: "cluster.local",
+					Intent:      v1alpha2.ActiveIntent,
+				},
+			}
+
+			configClient := configFake.NewSimpleClientset([]runtime.Object{mrc1}...)
+			mrcClient := tresorFake.NewFakeMRCWithConfig(configClient)
+			fakeCertManager := tresorFake.NewFakeWithMRCClient(mrcClient, 1*time.Hour)
 			mockK8s := k8s.NewMockController(mockCtrl)
 			computeClient := kube.NewClient(mockK8s)
+
+			mrcClient.NewCertEvent(mrc1.Name)
 
 			mc := MeshCatalog{
 				certManager: fakeCertManager,
@@ -2412,8 +2428,22 @@ func TestGetInboundMeshTrafficPolicy(t *testing.T) {
 			tc.prepare(mockK8s, tc.trafficSplits, tc.trafficTargets, tc.upstreamTrafficSettings)
 
 			if tc.newTrustDomain != "" {
-				// rotate a cert in and give it a moment to settle
-				mrc.NewCertEvent("new-trust-domain-cert", constants.MRCStateIssuingRollout, tc.newTrustDomain)
+				// create a new MRC with the newTrustDomain
+				mrc2 := &v1alpha2.MeshRootCertificate{
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      "osm-mesh-root-certificate-2",
+						Namespace: "osm-system",
+					},
+					Spec: v1alpha2.MeshRootCertificateSpec{
+						TrustDomain: tc.newTrustDomain,
+						Intent:      v1alpha2.ActiveIntent,
+					},
+				}
+				_, err := configClient.ConfigV1alpha2().MeshRootCertificates("osm-system").Create(context.Background(), mrc2, metav1.CreateOptions{})
+				assert.NoError(err)
+
+				// generate an MRCEvent for the new MRC to update the issuers and trigger a rotation
+				mrcClient.NewCertEvent(mrc2.Name)
 				assert.Eventually(func() bool {
 					return fakeCertManager.GetTrustDomains().AreDifferent()
 				}, 2*time.Second, 100*time.Millisecond)
--- a/pkg/certificate/fake_manager.go
+++ b/pkg/certificate/fake_manager.go
@@ -5,18 +5,16 @@ import (
 	"fmt"
 	"time"
 
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
 	"github.com/openservicemesh/osm/pkg/apis/config/v1alpha2"
-
 	"github.com/openservicemesh/osm/pkg/certificate/pem"
-)
-
-var (
-	validity = time.Hour
+	osmConfigClient "github.com/openservicemesh/osm/pkg/gen/client/config/clientset/versioned"
 )
 
 // fakeMRCClient implements the MRCClient interface
 type fakeMRCClient struct {
-	mrcList []*v1alpha2.MeshRootCertificate
+	configClient osmConfigClient.Interface
 }
 
 // GetCertIssuerForMRC returns a fakeIssuer and pre-generated RootCertificate. It is intended to implement the certificate.MRCClient interface.
@@ -27,17 +25,26 @@ func (c *fakeMRCClient) GetCertIssuerForMRC(mrc *v1alpha2.MeshRootCertificate) (
 // ListMeshRootCertificates returns the single, pre-generated MRC. It is intended to implement the certificate.MRCClient interface.
 func (c *fakeMRCClient) ListMeshRootCertificates() ([]*v1alpha2.MeshRootCertificate, error) {
 	// return single empty object in the list.
-	return c.mrcList, nil
+	mrcList, err := c.configClient.ConfigV1alpha2().MeshRootCertificates("osm-system").List(context.Background(), v1.ListOptions{})
+	if err != nil {
+		return nil, err
+	}
+	var mrcListPointers []*v1alpha2.MeshRootCertificate
+	for i := range mrcList.Items {
+		mrcListPointers = append(mrcListPointers, &mrcList.Items[i])
+	}
+	return mrcListPointers, nil
 }
 
 // UpdateMeshRootCertificate updates the given mesh root certificate.
 func (c *fakeMRCClient) UpdateMeshRootCertificate(mrc *v1alpha2.MeshRootCertificate) error {
-	// TODO(5046): implement this.
-	return nil
+	_, err := c.configClient.ConfigV1alpha2().MeshRootCertificates("osm-system").Update(context.Background(), mrc, v1.UpdateOptions{})
+	return err
 }
 
 // Watch returns a channel that has one MRCEventAdded. It is intended to implement the certificate.MRCClient interface.
 func (c *fakeMRCClient) Watch(ctx context.Context) (<-chan MRCEvent, error) {
+	// send event for first CA created
 	ch := make(chan MRCEvent)
 	go func() {
 		ch <- MRCEvent{
@@ -67,21 +74,4 @@ func (i *fakeIssuer) IssueCertificate(options IssueOptions) (*Certificate, error
 		TrustedCAs: pem.RootCertificate(i.id),
 		PrivateKey: pem.PrivateKey(i.id),
 	}, nil
-}
-
-// FakeCertManager is a testing helper that returns a *certificate.Manager
-func FakeCertManager() (*Manager, error) {
-	getCertValidityDuration := func() time.Duration { return validity }
-	cm, err := NewManager(
-		context.Background(),
-		&fakeMRCClient{},
-		getCertValidityDuration,
-		getCertValidityDuration,
-		1*time.Hour,
-	)
-	if err != nil {
-		return nil, fmt.Errorf("error creating fakeCertManager, err: %w", err)
-	}
-
-	return cm, nil
 }
--- a/pkg/certificate/fake_manager.go
+++ b/pkg/certificate/fake_manager.go
@@ -14,11 +14,13 @@ import (
 
 // fakeMRCClient implements the MRCClient interface
 type fakeMRCClient struct {
-	configClient osmConfigClient.Interface
+	configClient             osmConfigClient.Interface
+	countGetCertIssuerForMRC int
 }
 
 // GetCertIssuerForMRC returns a fakeIssuer and pre-generated RootCertificate. It is intended to implement the certificate.MRCClient interface.
 func (c *fakeMRCClient) GetCertIssuerForMRC(mrc *v1alpha2.MeshRootCertificate) (Issuer, pem.RootCertificate, error) {
+	c.countGetCertIssuerForMRC++
 	return &fakeIssuer{id: mrc.Name}, pem.RootCertificate("rootCA"), nil
 }
 
--- a/pkg/certificate/fake_manager.go
+++ b/pkg/certificate/fake_manager.go
@@ -5,13 +5,9 @@ import (
 	"fmt"
 	"time"
 
-	v1 "k8s.io/api/core/v1"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-
 	"github.com/openservicemesh/osm/pkg/apis/config/v1alpha2"
 
 	"github.com/openservicemesh/osm/pkg/certificate/pem"
-	"github.com/openservicemesh/osm/pkg/constants"
 )
 
 var (
@@ -19,21 +15,19 @@ var (
 )
 
 // fakeMRCClient implements the MRCClient interface
-type fakeMRCClient struct{}
+type fakeMRCClient struct {
+	mrcList []*v1alpha2.MeshRootCertificate
+}
 
 // GetCertIssuerForMRC returns a fakeIssuer and pre-generated RootCertificate. It is intended to implement the certificate.MRCClient interface.
 func (c *fakeMRCClient) GetCertIssuerForMRC(mrc *v1alpha2.MeshRootCertificate) (Issuer, pem.RootCertificate, error) {
-	return &fakeIssuer{}, pem.RootCertificate("rootCA"), nil
+	return &fakeIssuer{id: mrc.Name}, pem.RootCertificate("rootCA"), nil
 }
 
 // ListMeshRootCertificates returns the single, pre-generated MRC. It is intended to implement the certificate.MRCClient interface.
 func (c *fakeMRCClient) ListMeshRootCertificates() ([]*v1alpha2.MeshRootCertificate, error) {
 	// return single empty object in the list.
-	return []*v1alpha2.MeshRootCertificate{{
-		Spec: v1alpha2.MeshRootCertificateSpec{
-			TrustDomain: "fake.domain.com",
-		},
-	}}, nil
+	return c.mrcList, nil
 }
 
 // UpdateMeshRootCertificate updates the given mesh root certificate.
@@ -47,55 +41,7 @@ func (c *fakeMRCClient) Watch(ctx context.Context) (<-chan MRCEvent, error) {
 	ch := make(chan MRCEvent)
 	go func() {
 		ch <- MRCEvent{
-			Type: MRCEventAdded,
-			MRC: &v1alpha2.MeshRootCertificate{
-				ObjectMeta: metav1.ObjectMeta{
-					Name:      "osm-mesh-root-certificate",
-					Namespace: "osm-system",
-				},
-				Spec: v1alpha2.MeshRootCertificateSpec{
-					TrustDomain: "fake.domain.com",
-					Provider: v1alpha2.ProviderSpec{
-						Tresor: &v1alpha2.TresorProviderSpec{
-							CA: v1alpha2.TresorCASpec{
-								SecretRef: v1.SecretReference{
-									Name:      "osm-ca-bundle",
-									Namespace: "osm-system",
-								},
-							},
-						},
-					},
-				},
-				Status: v1alpha2.MeshRootCertificateStatus{
-					State: constants.MRCStateActive,
-					Conditions: []v1alpha2.MeshRootCertificateCondition{
-						{
-							Type:   constants.MRCConditionTypeReady,
-							Status: constants.MRCConditionStatusUnknown,
-						},
-						{
-							Type:   constants.MRCConditionTypeAccepted,
-							Status: constants.MRCConditionStatusUnknown,
-						},
-						{
-							Type:   constants.MRCConditionTypeIssuingRollout,
-							Status: constants.MRCConditionStatusUnknown,
-						},
-						{
-							Type:   constants.MRCConditionTypeValidatingRollout,
-							Status: constants.MRCConditionStatusUnknown,
-						},
-						{
-							Type:   constants.MRCConditionTypeIssuingRollback,
-							Status: constants.MRCConditionStatusUnknown,
-						},
-						{
-							Type:   constants.MRCConditionTypeValidatingRollback,
-							Status: constants.MRCConditionStatusUnknown,
-						},
-					},
-				},
-			},
+			MRCName: "osm-mesh-root-certificate",
 		}
 		close(ch)
 	}()
--- a/pkg/certificate/manager_test.go
+++ b/pkg/certificate/manager_test.go
@@ -10,6 +10,7 @@ import (
 	tassert "github.com/stretchr/testify/assert"
 	trequire "github.com/stretchr/testify/require"
 
+	"github.com/openservicemesh/osm/pkg/apis/config/v1alpha2"
 	"github.com/openservicemesh/osm/pkg/certificate/pem"
 	"github.com/openservicemesh/osm/pkg/constants"
 	"github.com/openservicemesh/osm/pkg/identity"
@@ -84,7 +85,8 @@ func TestRotor(t *testing.T) {
 
 	stop := make(chan struct{})
 	defer close(stop)
-	certManager, err := NewManager(context.Background(), &fakeMRCClient{}, getServiceCertValidityPeriod, getIngressGatewayCertValidityPeriod, 5*time.Second)
+	certManager, err := NewManager(context.Background(), &fakeMRCClient{mrcList: []*v1alpha2.MeshRootCertificate{activeMRC1}},
+		getServiceCertValidityPeriod, getIngressGatewayCertValidityPeriod, 5*time.Second)
 	require.NoError(err)
 
 	certA, err := certManager.IssueCertificate(ForServiceIdentity(identity.ServiceIdentity(cnPrefix)))
--- a/pkg/certificate/manager_test.go
+++ b/pkg/certificate/manager_test.go
@@ -9,10 +9,11 @@ import (
 	"github.com/cskr/pubsub"
 	tassert "github.com/stretchr/testify/assert"
 	trequire "github.com/stretchr/testify/require"
+	"k8s.io/apimachinery/pkg/runtime"
 
-	"github.com/openservicemesh/osm/pkg/apis/config/v1alpha2"
 	"github.com/openservicemesh/osm/pkg/certificate/pem"
 	"github.com/openservicemesh/osm/pkg/constants"
+	configFake "github.com/openservicemesh/osm/pkg/gen/client/config/clientset/versioned/fake"
 	"github.com/openservicemesh/osm/pkg/identity"
 )
 
@@ -85,7 +86,8 @@ func TestRotor(t *testing.T) {
 
 	stop := make(chan struct{})
 	defer close(stop)
-	certManager, err := NewManager(context.Background(), &fakeMRCClient{mrcList: []*v1alpha2.MeshRootCertificate{activeMRC1}},
+	configClient := configFake.NewSimpleClientset([]runtime.Object{activeMRC1}...)
+	certManager, err := NewManager(context.Background(), &fakeMRCClient{configClient: configClient},
 		getServiceCertValidityPeriod, getIngressGatewayCertValidityPeriod, 5*time.Second)
 	require.NoError(err)
 
--- a/pkg/certificate/manager_test.go
+++ b/pkg/certificate/manager_test.go
@@ -9,9 +9,7 @@ import (
 	"github.com/cskr/pubsub"
 	tassert "github.com/stretchr/testify/assert"
 	trequire "github.com/stretchr/testify/require"
-	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
-	"github.com/openservicemesh/osm/pkg/apis/config/v1alpha2"
 	"github.com/openservicemesh/osm/pkg/certificate/pem"
 	"github.com/openservicemesh/osm/pkg/constants"
 	"github.com/openservicemesh/osm/pkg/identity"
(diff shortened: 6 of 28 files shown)


pkg/certificate/mrc_reconciler.go
```diff
@@ -1,214 +1,212 @@
...
-	transitionAfter := mrc.Status.TransitionAfter
-	if transitionAfter == nil {
-		return false
-	}
+	return m.updateIssuers(issuer, issuer)
+}
 
-	//The truncate is needed so we truncate the monotonic clock.
-	return transitionAfter.Truncate(0).After(time.Now())
+var validMRCIntentCombinations = map[v1alpha2.MeshRootCertificateIntent][]v1alpha2.MeshRootCertificateIntent{
```
jsturtevant, line 62: note: need to all these values to MRC CRD enum

pkg/certificate/mrc_reconciler.go
```diff
@@ -1,214 +1,212 @@
...
+	}
 
-	// UseCaseBootstrap maps to the certificate use case for the  Bootstrap secrets
-	UseCaseBootstrap UseCase = "bootstrap"
+	mrcListLen := len(mrcList)
+	if mrcListLen == 0 {
+		log.Error().Err(ErrMRCNotFound).Msgf("when handling MRC event for MRC %s, found no MRCs in OSM control plane namespace", event.MRCName)
+		return ErrMRCNotFound
+	}
+	if mrcListLen > 2 {
```
jsturtevant, line 22: I think we potentially could have more than 2, if some are in the `inactive` state

jsturtevant: if we filter inactive out then we don't need the additional checks for inactive in the `setissuers` below

pkg/certificate/mrc_reconciler.go
```diff
@@ -1,214 +1,212 @@
...
+	intent2 := mrc2.Spec.Intent
+
+	var signingIssuer, validatingIssuer *issuer
+	switch intent1 {
+	case v1alpha2.ActiveIntent:
+		switch intent2 {
+		case v1alpha2.PassiveIntent:
+			signingIssuer = issuer1
+			validatingIssuer = issuer2
+		case v1alpha2.InactiveIntent:
```
jsturtevant, line 138: could we do a check earlier to filter out inactive certs we don't need this here.

pkg/certificate/mrc_reconciler.go
```diff
@@ -1,214 +1,212 @@
...
+			signingIssuer = issuer1
+			validatingIssuer = issuer1
+		case v1alpha2.DeactiveIntent:
```
jsturtevant, line 141: After going through this as an end user, having a `Deactivated` intent but it is still being used felt a bit funny, would it be possible to mark as passive here?   I think we got rid of `inactive` case and `deactivate` case then the switch becomes flat instead of nested which would make this section easier to understand (though this isn't too bad)

I think the challenge would be that we wouldn't know which one to set as Validator an signer in a deterministic way? Is that important or could we use oldest cert or some other field?

pkg/apis/config/v1alpha2/meshrootcertificate.go
```diff
@@ -122,6 +122,20 @@ type TresorCASpec struct {
...
 
+const (
+	// ActiveIntent
+	ActiveIntent MeshRootCertificateIntent = "active"
+
+	// PassiveIntent
+	PassiveIntent MeshRootCertificateIntent = "passive"
+
+	// DeactiveIntent
+	DeactiveIntent MeshRootCertificateIntent = "deactive"
```
jsturtevant, line 133:
```suggestion
	DeactiveIntent MeshRootCertificateIntent = "deactivate"
```

jsturtevant:
I was able to rotate a new certificate in and remove the old one 🚀

jsturtevant:
one note: 

```
k get meshrootcertificates.config.openservicemesh.io    
NAME                                          STATE
osm-mesh-root-certificate-2  
```

maybe we can make that 

```
k get meshrootcertificates.config.openservicemesh.io
NAME                                                          Intent
osm-mesh-root-certificate-2                      active (or passive or etc)
```

pkg/apis/config/v1alpha2/meshrootcertificate.go
```diff
@@ -122,6 +122,20 @@ type TresorCASpec struct {
...
+
+	// InactiveIntent
+	InactiveIntent MeshRootCertificateIntent = "inactive"
```
keithmattix, line 136: Why is there an intent for both Inactive and Deactivate? Why would the user set one instead of the other?

jsturtevant: This makes the transition from singing -> validation for the Active cert and from validating -> signing for the Passive cert deterministic.  I made a comment here as well: https://github.com/openservicemesh/osm/pull/5201#discussion_r995157614
where I think we might be able to remove this and still make it deterministic most of the time which would simplify somethings. Be curious what you think.

keithmattix: Yeah, if the user is setting these intents, having both (at least the way they're currently named) feels confusing.Maybe `Idle` is a better name for "Inactive" assuming I understand the intended usage correctly.

pkg/certificate/mrc_reconciler.go
jaellio: Added method to filter out inactive certs. Need to add tests for it

pkg/certificate/mrc_reconciler.go
```diff
@@ -1,214 +1,181 @@
...
-			if mrc.Status.ComponentStatuses.Bootstrap != status {
-				return false
-			}
-		case UseCaseGateway:
-			if mrc.Status.ComponentStatuses.Gateway != status {
-				return false
-			}
-		}
+// ValidateMRCIntents validates the intent combination of MRCs
+func ValidateMRCIntents(mrc1, mrc2 *v1alpha2.MeshRootCertificate) error {
```
jaellio, line 75: This could be simplified further with the removal of the `deactivate` intent. The only invalid combination of intents is `active` and `active`.

jsturtevant:
After playing with it we landed on the following pattern:
| Step  | MRC1 old   | MRC2 new     | signer  | validator |
| ----- | ---------- | ------------ | ------- | --------- |
| 1     | active     |              | mcr1    | mcr1      |
| 2     | active     | passive      | mcr1    | mcr2      |
| 3     | active | active | mcr2 or mrc1    | mcr1 or mrc2      |
| 4     | passive    | active       | mcr2    | mcr1      |
| 5     | inactive   | active       | mcr2    | mcr2      |
| 5     |            | active       | mcr2    | mcr2      |

pkg/certificate/mrc_reconciler.go
```diff
@@ -1,214 +1,182 @@
...
-	UseCaseValidatingWebhook UseCase = "validatingWebhook"
-
-	// UseCaseMutatingWebhook maps to the certificate use case for the Mutating Webhook
-	UseCaseMutatingWebhook UseCase = "mutatingWebhook"
-
-	// UseCaseXDSControlPlane maps to the certificate use case for the XDS Control Plane
-	UseCaseXDSControlPlane UseCase = "xdsControlPlane"
+func (m *Manager) handleMRCEvent(event MRCEvent) error {
+	log.Debug().Msgf("handling MRC event for MRC %s", event.MRCName)
+	mrcList, err := m.mrcClient.ListMeshRootCertificates()
```
steeling, line 12: this now becomes an n^2 operation. Can we instead keep track of the 2 mrc's we care about on the manager struct?

pkg/certificate/mrc_reconciler.go
```diff
@@ -1,214 +1,182 @@
...
+	intent1 := mrc1.Spec.Intent
+	intent2 := mrc2.Spec.Intent
+
+	validIntents, ok := validMRCIntentCombinations[intent1]
+	if !ok {
+		log.Error().Err(ErrUnknownMRCIntent).Msgf("unable to find %s intent in set of valid intents. Invalid combination of %s intent and %s intent", intent1, intent1, intent2)
+		return ErrUnknownMRCIntent
+	}
+
+	for _, intent := range validIntents {
```
steeling, line 91: should this be a map to check for access?

pkg/certificate/mrc_reconciler.go
```diff
@@ -1,214 +1,182 @@
...
+	}
 
-// state and condition get updated together.
-func (m *Manager) updateMRCState(mrc *v1alpha2.MeshRootCertificate) error {
-	// TODO(5046): update the MRC state & conditions, and issue the update via the MRCClient.
+	issuer1, err := m.getCertIssuer(mrc1)
+	if err != nil {
+		return err
+	}
+	issuer2, err := m.getCertIssuer(mrc2)
```
steeling, line 112: we can avoid calling this if mrc1 and mrc2 are the same

pkg/certificate/mrc_reconciler.go
```diff
@@ -1,214 +1,182 @@
...
+	return ErrInvalidMRCIntentCombination
 }
 
-// isTerminal checks if the MRC is in a terminal state.
-func isTerminal(mrc *v1alpha2.MeshRootCertificate) bool {
-	// TODO(5046): check if the mrc is in a terminal state.
-	return true
-}
+func (m *Manager) setIssuers(mrc1, mrc2 *v1alpha2.MeshRootCertificate) error {
+	if mrc1 == nil || mrc2 == nil {
```
steeling, line 104: What's the usage here when there is 1 mrc? The caller passes in a single mrc for both params, or passes in nil?

If they're the same, can we break out early, and just set the validating and issuing to the single issuer?

pkg/certificate/mrc_reconciler.go
```diff
@@ -1,214 +1,182 @@
...
-			return err
-		}
-	} else if shouldUpdateState(mrc) {
-		if err := m.updateMRCState(mrc); err != nil {
-			return err
-		}
-	} else if m.shouldSetIssuers(mrc) {
-		log.Info().Msgf("setting new certificate issuers on the Certificate Manager for MRC %s in stage %s", mrc.GetName(), mrc.Status.State)
-		return m.setIssuers(mrc)
+	if err = m.setIssuers(mrc1, mrc2); err != nil {
```
steeling, line 42: can we add a check to see if we should be setting the issuers? ie: if the mrc hasn't changed since the previously set mrc's, then we shouldn't set them.

Calling `getCertIssuer` calls `GetCertIssuerForMRC`, which makes an API call, so could be expensive if we do it repeatedly. And indeed, might cost the customer money depending on their PKI infrastructure billing plan.

pkg/certificate/mrc_reconciler.go
jsturtevant: There is check before this where this is handled via `handleSingleMRC` above.  Might be worth adding that here so it doesn't get mis-used?

pkg/certificate/mrc_reconciler.go
jsturtevant: I am missing where this is getting the n^2?  I am reading it as o(n).

If I missed something and it is n^2 it probably be worth optimizing. 

How many events and MRC's are we expecting in the system?  There will be 2 active and maybe a few inactive, but the expected volume should be pretty low as the old inactive MRCs will likely be cleaned up.  If this doesn't add to much complexity to the code to store and look up then I definitely worth it to optimize otherwise given volume I guess I'm not too concerned at this point

cmd/osm-bootstrap/crds/config_mesh_root_certificate.yaml
```diff
@@ -37,9 +37,9 @@ spec:
       served: true
       storage: true
       additionalPrinterColumns:
-        - description: Current state of the MeshRootCertificate config
-          jsonPath: .status.state
-          name: State
+        - description: Intent for the MeshRootCertificate
+          jsonPath: .spec.intent
+          name: Intent
```
jsturtevant, line 42: As I have been playing around with the MRC for the e2e tests and been thinking this field name might be better as `Role`. for example:

"Moving MRC to Passive intent in cluster\n" compared to "Moving MRC [%s] -> Passive role cluster"

"The MRC is in the Active role in the cluster" compared to "The MRC is in the Active intent in the cluster"

Thoughts?  @jaellio @keithmattix @steeling

steeling: ya that sounds good to me.

pkg/certificate/mrc_reconciler.go
jaellio: Discussed offline. Removed `Deactive`/`Deactivated` intent.

pkg/certificate/mrc_reconciler.go
jaellio: Previously mrc1 and mrc2 were always different. I have updated the logic such that they can be the same when there is only 1 mrc in the mesh.

pkg/certificate/mrc_reconciler.go
jaellio: Updated to handle the single MRC case in each method rather than in its own method.

pkg/certificate/mrc_reconciler.go
jaellio: Added `shouldSetIssuer` that we be called before `getCertIssuer` and `GetCertIssuerForMRC`.

cmd/osm-bootstrap/crds/config_mesh_root_certificate.yaml
jaellio: Okay! I will make the update

pkg/certificate/mrc_reconciler.go
steeling: Sorry I should have clarified what I meant

it gets called for handle event, so for each event we list.

At least on startup, where each existing MRC gets created, it O(N^2), in other circumstances its O(NM), with N being the number of events and M being the number of MRC's

jaellio: Is it possible to ignore MRC events for inactive MRCs and not keep them in the informer cache?

steeling: yes we can add a field selector. seems like a good idea!

jaellio: Thought about it more and we still need to receive inactive events since the change from passive to inactive signals the removal of the related issuer as the validatingIssuer. We would eventually pick up on the change on a resync, but the update could be delayed which would make testing more difficult/slower.

pkg/certificate/mrc_reconciler.go
jaellio: Are you suggesting using a nested map for `validMRCIntentCombinations`?

jsturtevant:
The cert rotation tests are passing: 

```
• [SLOW TEST:306.823 seconds]
[Tier 2][Bucket 11][] MeshRootCertificate
/home/runner/work/osm/osm/tests/framework/common.go:92
  with Tressor
  /home/runner/work/osm/osm/tests/e2e/e2e_meshrootcertificate_test.go:29
    rotates certificates
• [SLOW TEST:220.601 seconds]
[Tier 2][Bucket 11][] MeshRootCertificate
/home/runner/work/osm/osm/tests/framework/common.go:92
  with CertManager
  /home/runner/work/osm/osm/tests/e2e/e2e_meshrootcertificate_test.go:35
    rotates certificates
• [SLOW TEST:207.898 seconds]
[Tier 2][Bucket 11][] MeshRootCertificate
/home/runner/work/osm/osm/tests/framework/common.go:92
  with Vault
  /home/runner/work/osm/osm/tests/e2e/e2e_meshrootcertificate_test.go:41
    rotates certificates
Ran 3 of 67 Specs in 737.334 seconds
SUCCESS! -- 3 Passed | 0 Failed | 0 Pending | 64 Skipped
```

pkg/certificate/mrc_reconciler.go
steeling: looked into it a bit, it's actually not even possible to use field selectors for every field https://github.com/kubernetes/kubernetes/issues/53459, and CRD's is just name/namespace

jaellio: Just found that too!

jsturtevant: I think we could accomplish this via a custom informer index:

https://github.com/kubernetes/client-go/blob/c81636cd4415f0858a99c2cc18023740df7aa15f/tools/cache/shared_informer.go#L222-L223

jsturtevant: I am on board for addressing in a follow up, as this unblocks a few PRs we need for the release.

jaellio: @steeling, is this blocking?

pkg/certificate/mrc_reconciler.go
steeling: yup

pkg/certificate/mrc_reconciler.go
steeling: no but let's add an issue and a TODO if not addressed here

pkg/certificate/mrc_reconciler.go
```diff
@@ -1,214 +1,217 @@
...
-var (
-	allUseCases = []UseCase{
-		UseCaseValidatingWebhook,
-		UseCaseMutatingWebhook,
-		UseCaseXDSControlPlane,
-		UseCaseSidecar,
-		UseCaseBootstrap,
-		UseCaseGateway,
+// ValidateMRCIntents validates the intent combination of MRCs
+func ValidateMRCIntents(mrc1, mrc2 *v1alpha2.MeshRootCertificate) error {
```
steeling, line 51: super nit: by having the params be just the intent field, it makes it a bit more clear what this is evaluating without needing to read the code. More of an API hygiene thing. non-blocking.

jsturtevant:
adds some tests for #4835

cmd/osm-bootstrap/osm-bootstrap.go
```diff
@@ -433,36 +432,6 @@ func (b *bootstrap) createMeshRootCertificate() error {
...
-				Status: constants.MRCConditionStatusUnknown,
-			},
-			{
-				Type:   constants.MRCConditionTypeValidatingRollback,
-				Status: constants.MRCConditionStatusUnknown,
-			},
-		},
-	}
-
 	_, err = b.configClient.ConfigV1alpha2().MeshRootCertificates(b.namespace).UpdateStatus(context.Background(), createdMRC, metav1.UpdateOptions{})
```
keithmattix, line 435: Can we remove this `UpdateStatus` call now that we're not setting status on create?

keithmattix (approved):
LGTM! Thanks for all your work here; I'm proud of how far we've come on this!

Sidenote: I don't see any changes to the preset MRC; just double checking that that's intentional

cmd/osm-bootstrap/osm-bootstrap.go
jaellio: Yes, good catch. It is a NOOP

cmd/osm-bootstrap/crds/config_mesh_root_certificate.yaml
jsturtevant: #5222

pkg/apis/config/v1alpha2/meshrootcertificate.go
```diff
@@ -122,6 +122,19 @@ type TresorCASpec struct {
 // can be (Active, Passive).
 type MeshRootCertificateIntent string
 
+const (
+	// ActiveIntent means the settings and certificate provider in this MRC are used for signing and
+	// validating certificates.
+	ActiveIntent MeshRootCertificateIntent = "active"
```
jsturtevant, line 128: I like these constants here but we should clean up the ones not used in https://github.com/openservicemesh/osm/blob/main/pkg/constants/constants.go#L203

pkg/certificate/mrc_reconciler.go
jsturtevant: did we create an issue to track this?

pkg/certificate/mrc_reconciler.go
```diff
@@ -1,214 +1,245 @@
...
-		case UseCaseGateway:
-			if mrc.Status.ComponentStatuses.Gateway != status {
-				return false
-			}
+	log.Debug().Msg("validating MRC intent combination")
+	if err := validateMRCIntents(mrc1, mrc2); err != nil {
+		return nil, nil, err
+	}
+
+	switch intent1 {
```
jsturtevant, line 116: could you add a comment here, that this ordering the mrc into the validating and singing based on the current intents?

tests/e2e/e2e_meshrootcertificate_test.go
```diff
@@ -226,3 +300,155 @@ func createVaultMRC(name string, intent v1alpha2.MeshRootCertificateIntent) (*v1
 			},
 		}, metav1.CreateOptions{})
 }
+
+/*func verifiyUpdatedPodCert(pod *v1.Pod) {
```
jsturtevant, line 304: can drop this for now

tests/e2e/e2e_meshrootcertificate_test.go
```diff
@@ -226,3 +300,155 @@ func createVaultMRC(name string, intent v1alpha2.MeshRootCertificateIntent) (*v1
...
+	clientSecretName := fmt.Sprintf("envoy-bootstrap-config-%s", clientPodUUID)
+
+	// TODO(jaellio): add a time.wait instead of waiting so long in call
+	err = Td.WaitForBootstrapSecretUpdate(serverPodDef.Namespace, srvSecretName, signingCertName, validatingCertName, time.Second*30)
+	Expect(err).NotTo(HaveOccurred())
+	err = Td.WaitForBootstrapSecretUpdate(clientPodDef.Namespace, clientSecretName, signingCertName, validatingCertName, time.Second*30)
+	Expect(err).NotTo(HaveOccurred())
+}
+
+// func getCertificateModulus(pod *v1.Pod) string {
```
jsturtevant, line 367: can drop this for now

tests/e2e/e2e_meshrootcertificate_test.go
```diff
@@ -49,38 +45,106 @@ var _ = OSMDescribe("MeshRootCertificate",
 		})
 	})
 
+// basicCerRotationScenario rotates a new cert in through the follow pattern:
+// | Step  | MRC1 old   | MRC2 new     | signer    | validator |
+// | ----- | ---------- | ------------ | --------- | --------- |
+// | 1     | active     |              | mcr1      | mcr1      |
```
jsturtevant, line 51:
```suggestion
// | 1     | active     |              | mrc1      | mrc1      |
```

pkg/certificate/mrc_reconciler.go
jaellio: Created #5226

cmd/osm-bootstrap/crds/config_mesh_root_certificate.yaml
jaellio: #5222

cmd/osm-bootstrap/crds/config_mesh_root_certificate.yaml
```diff
@@ -48,19 +48,20 @@ spec:
...
                   default: cluster.local
                 intent:
-                  description: Intent specifies the role of the MeshRootCertificate, can be active or passive
+                  description: Intent specifies the role of the MeshRootCertificate, can be active, passive
                   type: string
                   enum:
                     - passive
                     - active
-                  default: passive
+                    - inactive
```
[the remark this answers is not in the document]
jaellio: I agree, `inactive` makes sense as the default. Can we add a default for the intent field even though in the last OSM release we included an MRC without the intent field? We aren't updating the resource version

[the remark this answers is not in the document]
jsturtevant: good question, I don't think we have logic to support conversion webhooks.  

reading https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#defaulting

It looks like 

```
Defaulting happens on the object

in the request to the API server using the request version defaults,
when reading from etcd using the storage version defaults,
after mutating admission plugins with non-empty patches using the admission webhook object version defaults.
``` 

Which means if someone was using it they might end up with inactive certificate I think but something like this might happen anyways, since we wouldn't be able to read the value properly.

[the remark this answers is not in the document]
jaellio: Can `default` and `required` be used interchangeably here since we specify an enum and therefore limit the optional values?

pkg/certificate/mrc_reconciler.go
jaellio: I'm not sure what you mean. Could you please clarify?

tests/e2e/e2e_meshrootcertificate_test.go
jaellio: I don't see a difference in the suggested change. Let me know if I am missing something!

jsturtevant: m_**c**_r -> m_**r**_c

jaellio: Got it! Thank you!

pkg/certificate/mrc_reconciler.go
jsturtevant: it isn't clear why we have this switch here and why sometimes we return mrc1 first and other times mrc2 first.  A comment could help.

jsturtevant: (looking closer now and maybe a switch is no longer needed?)

cmd/osm-bootstrap/crds/config_mesh_root_certificate.yaml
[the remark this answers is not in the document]
jsturtevant: yea maybe don't need default since it is required field.  Works for me to leave it as is

pkg/certificate/mrc_reconciler.go
jaellio: I converted the switch statement to 2 conditionals and added comments. Let me know if you think this improves readability!

I also realized that the `validateMRCIntents` method wasn't really doing much after we reduced the number of expected intents to passive and active. I modified the method to check that an intent is either active or passive. This should always be the case since we filtered out inactive MRCs and we limited the possible intent values with the MRC open api

pkg/apis/config/v1alpha2/meshrootcertificate.go
jaellio: Removed

cmd/osm-bootstrap/crds/config_mesh_root_certificate.yaml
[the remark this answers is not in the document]
jaellio: Will think about it more, but leaving it as is for now.

pkg/certificate/mrc_reconciler.go
jaellio: Updated!

pkg/certificate/mrc_reconciler.go
jaellio: Updated!

pkg/certificate/mrc_reconciler.go
jaellio: Added a TODO

jsturtevant (approved):
lgtm, there maybe a bit more simplification we can accomplish but we can address it with #5226 which will change the details anyways.

pkg/certificate/fake_manager.go
```diff
@@ -5,97 +5,50 @@ import (
...
-
-var (
-	validity = time.Hour
+	osmConfigClient "github.com/openservicemesh/osm/pkg/gen/client/config/clientset/versioned"
 )
 
 // fakeMRCClient implements the MRCClient interface
-type fakeMRCClient struct{}
+type fakeMRCClient struct {
+	configClient osmConfigClient.Interface
```
steeling, line 17: if we're adding this here, we may as well remove this altogether and use the `MRCComposer`, which has an interface for the compute client. We can leverage a fake or mock compute client for that.

pkg/certificate/mrc_reconciler.go
```diff
@@ -1,214 +1,228 @@
...
-	"github.com/openservicemesh/osm/pkg/constants"
+	"github.com/openservicemesh/osm/pkg/errcode"
 )
 
-// UseCase is how the certificate is used. Each use case will have its own rotation mechanism, that is necessary to
-// determine if we can propagate the rotation forward.
-type UseCase string
+func (m *Manager) handleMRCEvent(event MRCEvent) error {
+	log.Debug().Msgf("handling MRC event for MRC %s", event.MRCName)
+	// TODO(#5226): optimize event handling to reduce cost of listing all MRCs for each event
```
steeling, line 10: Note that we could simply keep the MRC on the issuer struct, and then we have the 2 mrc's we care about. This seems like a simple enough change?

pkg/webhook/server.go
```diff
@@ -63,6 +63,7 @@ func NewServer(name, namespace string, port int, cm *certificate.Manager, handle
 // Run actually starts the server.
 func (s *Server) Run(ctx context.Context) error {
 	if err := s.configureCertificateRotation(ctx); err != nil {
+		log.Error().Err(err).Msg("failed to configure cert rotation")
```
steeling, line 66: prefer to log or return errors, not both (we don't do a good job of this in the code base, but can avoid adding new ones).

cmd/osm-bootstrap/osm-bootstrap.go
```diff
@@ -490,7 +450,7 @@ func buildMeshRootCertificate(presetMeshRootCertificateConfigMap *corev1.ConfigM
 			APIVersion: "config.openservicemesh.io/configv1alpha2",
 		},
 		ObjectMeta: metav1.ObjectMeta{
-			Name: meshRootCertificateName,
+			Name: constants.DefaultMeshRootCertificateName,
```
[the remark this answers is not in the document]
jsturtevant: This isn't a change in behavior for this PR, this was a local constant (see line 52 above) and I moved it to the constant so it would be in one place to be used across tests as well.

pkg/apis/config/v1alpha2/meshrootcertificate.go
```diff
@@ -122,6 +122,19 @@ type TresorCASpec struct {
...
+
+	// PassiveIntent means the settings and certificate provider in this MRC are used for validating
+	// certificates.
+	PassiveIntent MeshRootCertificateIntent = "passive"
```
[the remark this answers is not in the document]
jsturtevant: I personally like passive but if there is strong agreement here, we should switch `active` to something like `signing`.

pkg/certificate/mrc_reconciler.go
jsturtevant: When I chatted with @jaellio last week, it got a bit complicated on the edge cases I believe.  We've got https://github.com/openservicemesh/osm/issues/5226 to track.  If we want this in before release we can work can follow up there, but I would be in favor of doing that separately so we can unblock the other PRs that are waiting on this.

cmd/osm-bootstrap/osm-bootstrap.go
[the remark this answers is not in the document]
steeling: we can add a TODO at least?

pkg/certificate/mrc_reconciler.go
```diff
@@ -1,214 +1,228 @@
...
+	desiredSigningMRC, desiredValidatingMRC, err := getSigningAndValidatingMRCs(filteredMRCList)
+	if err != nil {
+		return err
+	}
 
-	// UseCaseXDSControlPlane maps to the certificate use case for the XDS Control Plane
-	UseCaseXDSControlPlane UseCase = "xdsControlPlane"
+	shouldUpdate, err := m.shouldUpdateIssuers(desiredSigningMRC, desiredValidatingMRC)
+	if err != nil {
+		return nil
```
steeling, line 25: return err

pkg/certificate/mrc_reconciler.go
```diff
@@ -1,214 +1,228 @@
...
+		return nil, nil, ErrUnexpectedNilMRC
 	}
 
-	//The truncate is needed so we truncate the monotonic clock.
-	return transitionAfter.Truncate(0).After(time.Now())
-}
+	intent1 := mrc1.Spec.Intent
+	intent2 := mrc2.Spec.Intent
+
+	if err := validateMRCIntent(intent1); err != nil {
```
steeling, line 86: should these be in a validating webhook?

pkg/certificate/mrc_reconciler.go
```diff
@@ -1,214 +1,228 @@
...
-			}
-		case UseCaseBootstrap:
-			if mrc.Status.ComponentStatuses.Bootstrap != status {
-				return false
-			}
-		case UseCaseGateway:
-			if mrc.Status.ComponentStatuses.Gateway != status {
-				return false
-			}
+	log.Debug().Msgf("validating intent combination of %s and %s", intent1, intent2)
```
steeling, line 93: debug line seems like it should occur before validateMRCIntent call. or consider removing

pkg/certificate/mrc_reconciler.go
```diff
@@ -1,214 +1,228 @@
...
+	} else {
+		log.Trace().Msgf("found single MRC in the mesh when handling MRC event for MRC %s", mrc1.Name)
+		// if there is only one MRC, set mrc2 equal to mrc1
+		mrc2 = mrc1
 	}
 
-	transitionAfter := mrc.Status.TransitionAfter
-	if transitionAfter == nil {
-		return false
+	if mrc1 == nil || mrc2 == nil {
```
steeling, line 78: note: i don't think this possible, but defense in depth is not necessarily a bad thing.

I'd consider a validateMRC function, which has this check, and others like ValidateIntent. This seems like it would be useful in general, and would allow greater re-use in ie: a validating webhook

pkg/certificate/mrc_reconciler.go
steeling: Ok, I won't block, but I'll point it that it looks like we could do a drop in replacement into this code:

```
...
   desiredSigning, desiredValidating := m.getSigningAndValidating(event.MRC)
   // note: no need to filter out inactive anymore, or call ListMeshRootCertificates
...

func getSigningAndValidating(newMRC MRC) {
  mrcs := make(map[string]MRC)
  mrcs[m.signingIssuer.ID] = m.signingIssuer.mrc
  mrcs[m.validatingIssuer.ID] = m.validatingIssuer.mrc
  mrcs[newMRC.Name] = newMRC
  if len(mrcs) > 2 {.../*err*/ }
 
   ... // continue as normal
}
```

I think using a map like this would also clean up some of the logic in that method. But  again, non-blocking

pkg/certificate/mrc_reconciler.go
```diff
@@ -1,214 +1,228 @@
...
+
+	// if the issuers are already set to the desired value, return
+	if signingIssuer != nil && signingIssuer.ID == desiredSigningMRC.Name && validatingIssuer != nil && validatingIssuer.ID == desiredValidatingMRC.Name {
+		log.Debug().Msgf("issuers already set to the desired value. Will not update issuers: validating[%s] and signing[%s]", validatingIssuer.ID, signingIssuer.ID)
+		return false, nil
+	}
+
+	// if desiredSigningMRC != desiredValidatingMRC and both MRCs have active intents, their state is non
+	// deterministic. To avoid continuously resetting the issuers, only update
+	// if the issuers have not already been set
```
steeling, line 143: This comment applies to the logical NOT of this if statement. Consider rewording so that the comment applies to this if statement.

steeling: the comment looks to also not be valid anymore?

pkg/certificate/mrc_reconciler.go
```diff
@@ -1,214 +1,228 @@
...
+	m.validatingIssuer = validatingIssuer
+	log.Trace().Msgf("setting issuers for validating[%s] and signing[%s]", validatingIssuer.ID, signingIssuer.ID)
+	return nil
 }
 
-// state and condition get updated together.
-func (m *Manager) updateMRCState(mrc *v1alpha2.MeshRootCertificate) error {
-	// TODO(5046): update the MRC state & conditions, and issue the update via the MRCClient.
+func (m *Manager) getCertIssuers(desiredSigningMRC, desiredValidatingMRC *v1alpha2.MeshRootCertificate) (*issuer, *issuer, error) {
+	if desiredSigningMRC == nil || desiredValidatingMRC == nil {
```
steeling, line 172: There's a lot of duplicative nil checks. These are internal methods, so no need to be as defensive about nil calls, which should be applied at the API level (exported funcs)

pkg/certificate/mrc_reconciler.go
```diff
@@ -1,214 +1,228 @@
...
-}
+func (m *Manager) getCertIssuer(mrc *v1alpha2.MeshRootCertificate) (*issuer, error) {
+	m.mu.Lock()
+	signingIssuer := m.signingIssuer
+	validatingIssuer := m.validatingIssuer
+	m.mu.Unlock()
+
+	// if the issuer has already been created for the specified MRC,
+	// return the existing issuer
+	if signingIssuer != nil && mrc.Name == signingIssuer.ID {
```
steeling, line 203: if we have this check here, then we can be less defensive about the shouldUpdateIssuer, and remove the checks on whether the desired is already set or not

pkg/certificate/mrc_reconciler.go
```diff
@@ -1,214 +1,228 @@
...
+		return signingIssuer, nil
+	}
+	if validatingIssuer != nil && mrc.Name == validatingIssuer.ID {
+		return validatingIssuer, nil
+	}
```
steeling, line 208: you may also want to check if we are promoting/demoting an MRC, to avoid an extra call to `GetCertIssuerForMRC`, if we're swapping the issuers

pkg/certificate/fake_manager.go
```diff
@@ -5,97 +5,50 @@ import (
...
 
 	"github.com/openservicemesh/osm/pkg/apis/config/v1alpha2"
-
 	"github.com/openservicemesh/osm/pkg/certificate/pem"
-	"github.com/openservicemesh/osm/pkg/constants"
-)
-
-var (
-	validity = time.Hour
+	osmConfigClient "github.com/openservicemesh/osm/pkg/gen/client/config/clientset/versioned"
```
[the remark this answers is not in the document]
jaellio: Thanks! Is this in the uber style guide?

pkg/certificate/mrc_reconciler.go
jaellio: `validateMRCIntent` validates the individual intent. This log line refers to validating the intent combination.

pkg/certificate/mrc_reconciler.go
jaellio: I think this case is already handled by the existing conditionals. For a given MRC we check if the name matches the ID of either issuer and return the issuer on a match.

pkg/certificate/mrc_reconciler.go
jaellio: Rewording the comment and refactoring this section to make it more readable.

pkg/certificate/mrc_reconciler.go
jaellio: Yes, #5205 will add this validation to a validating webhook.

cmd/osm-bootstrap/crds/config_mesh_root_certificate.yaml
ksubrmnn: I think Role makes sense since we're ultimately representing what the certificate can and cannot do.

pkg/certificate/mrc_reconciler.go
jsturtevant: An example of a case where this gets tricky is when a controller restarts. In that case the existing manager won't have any sign/validating issuers set.  By approaching it this way we are handling the events in a trigger level fashion.

This means that on the first event that comes through the issuer/validator would be the same but this might not be the case, there could be more than one Active/Passive.  In this case there are some conditions where the controller could start to re-issue certificates since the `ShouldRotate` function would see this as a change. Once more events came through it would be normalize but this could be a spot where we have strange behavior intermittently.

Using a custom indexer (https://github.com/openservicemesh/osm/issues/5226) that only contains active MRCs we could be more level triggered and  solve this problem of having to filter and list all the MRCS.  

Maybe we could combine the two options as well, which would allow us to exit earlier in the case things are set?

pkg/apis/config/v1alpha2/meshrootcertificate.go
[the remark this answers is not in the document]
ksubrmnn: I think ValidateOnly and Signing make understanding the code a bit more straightforward for newcomers

pkg/catalog/inbound_traffic_policies_test.go
```diff
@@ -2388,11 +2390,25 @@ func TestGetInboundMeshTrafficPolicy(t *testing.T) {
 			mockCtrl := gomock.NewController(t)
 			defer mockCtrl.Finish()
 
-			mrc := tresorFake.NewFakeMRC()
-			fakeCertManager := tresorFake.NewFakeWithMRC(mrc, 1*time.Hour)
+			mrc1 := &v1alpha2.MeshRootCertificate{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "osm-mesh-root-certificate",
```
ksubrmnn, line 2395: Should this be `constants.DefaultMeshRootCertificateName`

pkg/certificate/manager_test.go
```diff
@@ -9,11 +9,11 @@ import (
 	"github.com/cskr/pubsub"
 	tassert "github.com/stretchr/testify/assert"
 	trequire "github.com/stretchr/testify/require"
-	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
 
-	"github.com/openservicemesh/osm/pkg/apis/config/v1alpha2"
 	"github.com/openservicemesh/osm/pkg/certificate/pem"
 	"github.com/openservicemesh/osm/pkg/constants"
+	configFake "github.com/openservicemesh/osm/pkg/gen/client/config/clientset/versioned/fake"
```
ksubrmnn, line 16: uppercase

cmd/osm-bootstrap/osm-bootstrap.go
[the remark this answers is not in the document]
ksubrmnn: @steeling Could you explain more?

pkg/certificate/mrc_reconciler.go
jaellio: I think `shouldUpdateIssuer` should keep the check if the issuers are already set to their desired values. If this condition is true, the issuers should not be updated and therefore `shouldUpdateIssuer` should return false.

pkg/certificate/fake_manager.go
jaellio: The fakeMRCClient in this pkg exists for use in tests to avoid a circular dependency with pkg/certificate/providers. Since the MRCComposer is also defined in pkg/certificate/providers this would also result in a circular dependency.

pkg/certificate/mrc_reconciler.go
jaellio: I agree that this logic could be combined in a validateMRC method. This would help reduce duplicate code between the validating webhook for the mrc and the mrc reconciler. I think it makes sense to update this in #5220 when it more clear what the commonalities are for mrc validation in the webhook and the reconciler.

pkg/certificate/fake_manager.go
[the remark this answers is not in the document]
jsturtevant: I couldn't find a linter that enforced this but I found this which has a little dialog https://stackoverflow.com/a/20864492/697126

pkg/certificate/mrc_reconciler.go
jsturtevant: I am not sure if this handles the scenario `desiredSigningMRC != desiredValidatingMRC and both MRCs have active intents, their state is non-deterministic.` ?

Even if it does, I kind of like being explicit about though and exiting the reconciliation early.  It isn't immediately obvious that there are no changes otherwise.

pkg/certificate/mrc_reconciler_test.go
```diff
@@ -0,0 +1,456 @@
...
+
+			m := &Manager{mrcClient: &fakeMRCClient{}}
+			if tt.currentSigningIssuerID != "" {
+				m.signingIssuer = &issuer{ID: tt.currentSigningIssuerID}
+			}
+			if tt.currentValidatingIssuerID != "" {
+				m.validatingIssuer = &issuer{ID: tt.currentValidatingIssuerID}
+			}
+
+			signingIssuer, validatingIssuer, err := m.getCertIssuers(tt.signingMRC, tt.validatingMRC)
```
jsturtevant, line 310: consider adding a check to see if underlying `m.mrcClient.GetCertIssuerForMRC(mrc)` is called by using the `fakemrcclient` to increment a number each time it is called.  A check like this would ensure that the cached versions are used when expected and we aren't making the additional calls when not expected.

jaellio: Great suggestion. Updated!

pkg/certificate/manager_test.go
jaellio: If this is non-blocking, I'll resolve this in a follow-up PR. There are multiple instances of this casing in aliased package names that could be cleaned up in a single PR. If we want to enforce this pattern we could add it to our linter.

cmd/osm-bootstrap/osm-bootstrap.go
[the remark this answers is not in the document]
jsturtevant: Deleting isn't blocked but upon bootstrapper restart we will see issues.  I created https://github.com/openservicemesh/osm/issues/5237 to track. @ksubrmnn hopefully the issue helps explains and proposes a few options for solving.

pkg/apis/config/v1alpha2/meshrootcertificate.go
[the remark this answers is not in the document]
jsturtevant: I was initially leaning this way as well, but in that case does signing mean "signingAndValidating" since its doing both? The ambiguity and verbosity felt to much.  For this reason, after sitting on it for awhile and trying out several rotations I began to like `passive` and `active`

@jaellio you presented this at the servicemeshcon did you hear any feedback in either direction?

jsturtevant (approved):

jsturtevant:
@keithmattix a few changes have been made since your last look, would you take another look?

keithmattix:
Thanks for the ping! Will take a look!

pkg/certificate/fake_manager.go
steeling: that's likely indicative of a separate issue then. maybe we should bring MRCComposer into the certificate package, or into it's own subpackage so it can be used in both packages independently

pkg/certificate/mrc_reconciler.go
steeling: non-blocking at this point

pkg/certificate/mrc_reconciler.go
```diff
@@ -1,214 +1,176 @@
...
+func filterOutInactiveMRCs(mrcList []*v1alpha2.MeshRootCertificate) []*v1alpha2.MeshRootCertificate {
+	n := 0
+	for _, mrc := range mrcList {
+		if mrc.Spec.Intent != v1alpha2.InactiveIntent {
+			mrcList[n] = mrc
+			n++
+		}
 	}
-	return nil
+	return mrcList[:n]
```
steeling, line 175: is this protecting against potentially 3 active MRC's? That seems like a critical issue and we should return an error?

pkg/certificate/mrc_reconciler.go
steeling: ah gotcha, sounds like a good thing for a test case :)

pkg/certificate/mrc_reconciler.go
```diff
@@ -1,214 +1,176 @@
...
+	}
+
+	return true, nil
 }
 
-// isTerminal checks if the MRC is in a terminal state.
-func isTerminal(mrc *v1alpha2.MeshRootCertificate) bool {
-	// TODO(5046): check if the mrc is in a terminal state.
-	return true
+func (m *Manager) updateIssuers(signingIssuer, validatingIssuer *issuer) error {
```
steeling, line 114: this function doesn't need to return an error

steeling (approved):

keithmattix (approved):
One nit, but nothing worth blocking this PR on. I'd prefer to bias towards action and get this merged and iterate via separate PRs. Nice work!

PR #5204, opened 2022-10-14
feat(certs): use MeshConfig feature flag to enable MRC

### LINKED ISSUE #5041: Deprecate enableMeshRootCertificate binary flag

jaellio:
**Description**:

Replace binary flag with MeshConfig feature flag to enable the use
of the MRC. If OSM has been installed with
EnableMeshRootCertificate disabled, if the feature flag is then
enabled in the MeshConfig the control plane components must be
restarted in order to pick up on the enabled flag, and begin using
the MRC for certificate management.

Adds an e2e to check enabling the MRC after osm install for each
cert provider.

Dependent on #5201 
Resolves #5041 
Part of #4835 

**Testing done**:
- e2e
- Manual testing with demo

**Affected area**:

3. Has documentation corresponding to this change been updated in the [osm-docs](https://github.com/openservicemesh/osm-docs) repo (if applicable)? no

### DIFF
--- a/charts/osm/templates/osm-deployment.yaml
+++ b/charts/osm/templates/osm-deployment.yaml
@@ -63,12 +63,12 @@ spec:
             "--certificate-manager", "{{.Values.osm.certificateProvider.kind}}",
             "--trust-domain", "{{.Values.osm.trustDomain}}",
             {{ if eq .Values.osm.certificateProvider.kind "vault" }}
-            "--vault-host", "{{ required "osm.vault.host is required when osm.certificateProvider.kind==vault" .Values.osm.vault.host }}",
+            "--vault-host", "{{.Values.osm.vault.host}}",
             "--vault-port", "{{.Values.osm.vault.port}}",
             "--vault-protocol", "{{.Values.osm.vault.protocol}}",
-            "--vault-token", "{{ required "osm.vault.token is required when osm.certificateProvider.kind==vault and osm.vault.secret.name and osm.vault.secret.key are empty" .Values.osm.vault.token }}",
-            "--vault-token-secret-name",  "{{ required "osm.vault.secret.name is required when osm.certificateProvider.kind==vault and osm.vault.token is empty" .Values.osm.vault.secret.name }}",
-            "--vault-token-secret-key",  "{{ required "osm.vault.secret.key is required when osm.certificateProvider.kind==vault and osm.vault.token is empty" .Values.osm.vault.secret.key }}",
+            "--vault-token", "{{.Values.osm.vault.token}}",
+            "--vault-token-secret-name",  "{{ .Values.osm.vault.secret.name }}",
+            "--vault-token-secret-key",  "{{ .Values.osm.vault.secret.key }}",
             {{- end }}
             "--cert-manager-issuer-name", "{{.Values.osm.certmanager.issuerName}}",
             "--cert-manager-issuer-kind", "{{.Values.osm.certmanager.issuerKind}}",
--- a/charts/osm/templates/osm-deployment.yaml
+++ b/charts/osm/templates/osm-deployment.yaml
@@ -62,7 +62,6 @@ spec:
             "--ca-bundle-secret-name", "{{.Values.osm.caBundleSecretName}}",
             "--certificate-manager", "{{.Values.osm.certificateProvider.kind}}",
             "--trust-domain", "{{.Values.osm.trustDomain}}",
-            "--enable-mesh-root-certificate={{.Values.osm.featureFlags.enableMeshRootCertificate}}",
             {{ if eq .Values.osm.certificateProvider.kind "vault" }}
             "--vault-host", "{{ required "osm.vault.host is required when osm.certificateProvider.kind==vault" .Values.osm.vault.host }}",
             "--vault-port", "{{.Values.osm.vault.port}}",
--- a/charts/osm/templates/osm-deployment.yaml
+++ b/charts/osm/templates/osm-deployment.yaml
@@ -66,13 +66,9 @@ spec:
             "--vault-host", "{{ required "osm.vault.host is required when osm.certificateProvider.kind==vault" .Values.osm.vault.host }}",
             "--vault-port", "{{.Values.osm.vault.port}}",
             "--vault-protocol", "{{.Values.osm.vault.protocol}}",
-            {{ if and (empty .Values.osm.vault.secret.name) (empty .Values.osm.vault.secret.key) }}
             "--vault-token", "{{ required "osm.vault.token is required when osm.certificateProvider.kind==vault and osm.vault.secret.name and osm.vault.secret.key are empty" .Values.osm.vault.token }}",
-            {{- end }}
-            {{ if empty .Values.osm.vault.token }}
             "--vault-token-secret-name",  "{{ required "osm.vault.secret.name is required when osm.certificateProvider.kind==vault and osm.vault.token is empty" .Values.osm.vault.secret.name }}",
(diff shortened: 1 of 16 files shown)


charts/osm/templates/preset-mesh-root-certificate.yaml
```diff
@@ -1,4 +1,3 @@
-{{- if .Values.osm.featureFlags.enableMeshRootCertificate }}
```
keithmattix, line 1: We still need this right?

pkg/apis/config/v1alpha2/mesh_config.go
```diff
@@ -234,4 +234,7 @@ type FeatureFlags struct {
 
 	// EnableRetryPolicy defines if retry policy is enabled.
 	EnableRetryPolicy bool `json:"enableRetryPolicy"`
+
+	// EnableMeshRootCertificate defines if MRCs are used for certificate management.
```
keithmattix, line 238: Can we add a sentence saying that this feature flag requires a control plane restart?

pkg/validator/patch.go
```diff
@@ -130,7 +131,7 @@ func createOrUpdateValidatingWebhook(clientSet kubernetes.Interface, cert *certi
 						Port:      &webhookPort,
 					},
 					CABundle: cert.GetTrustedCAs()},
-				FailurePolicy: &failurePolicy,
+				FailurePolicy: &ignoreConnectionFailurePolicy,
```
keithmattix, line 134: Why this change and why here?

charts/osm/templates/preset-mesh-root-certificate.yaml
jaellio: We need the preset-mesh-root-certificate applied regardless of whether or not the used enables mesh root certificate rotation. If enableMeshRootCertificate was false initially false and the user then updated the flag to true, on restart the osm-bootstrap would need the preset-mesh-root-certificate. Our system is not currently designed to retroactively apply the preset-mesh-root-certificate configMap.

pkg/apis/config/v1alpha2/mesh_config.go
jaellio: Good idea. We'll also call this out and explain it in more depth in the docs.

pkg/validator/patch.go
jaellio: If enableMeshRootCertificate is set to true on OSM install an MRC will be created by the osm-bootstrap. The osm-controller and osm-injector will not have started up since they are waiting on the osm-bootstrap to be ready. If the validating webhook's failure policy is fail on failed connection to validating webhook server, MRC creation will fail in the osm-bootstrap since the osm-controller hasn't started up and therefore the validating webhook server hasn't been created. By ignoring connection failures specifically for the validation of the MRC, the osm-bootstrap can successfully create an MRC regardless of whether or not the osm-controller is ready.

charts/osm/templates/osm-deployment.yaml
```diff
@@ -62,18 +62,13 @@ spec:
             "--ca-bundle-secret-name", "{{.Values.osm.caBundleSecretName}}",
             "--certificate-manager", "{{.Values.osm.certificateProvider.kind}}",
             "--trust-domain", "{{.Values.osm.trustDomain}}",
-            "--enable-mesh-root-certificate={{.Values.osm.featureFlags.enableMeshRootCertificate}}",
             {{ if eq .Values.osm.certificateProvider.kind "vault" }}
-            "--vault-host", "{{ required "osm.vault.host is required when osm.certificateProvider.kind==vault" .Values.osm.vault.host }}",
```
jaellio, line 67: The host requirement could be added back. We do not make the same requirement in osm-injector deployment.

pkg/validator/patch.go
jsturtevant: This is part of the challenge of the way these are handled in the bootstrap.  How is this done for the meshconfiguraiton crd?

The issue with the ignore failure policy means that if for some reason the webhook is down later on down the road then we could allow multiple active MRCs into the system.

jaellio: We currently don't validate the MeshConfig.

jsturtevant: Is there a reason to use the bootstrapper to deploy the initial MRC?  

Could we include the yaml for MRC instead? I think it is pretty static?

jaellio: > Is there a reason to use the bootstrapper to deploy the initial MRC?
> 
> Could we include the yaml for MRC instead? I think it is pretty static?

Since we don't apply the CRDs using helm I can't create any CRs before applying the CRDs in the osm-bootstrap.

jaellio: As long as the ValidatingWebhookConfiguration has not been applied with the control plane validating webhook or the osm-controller is running, MRC creation will succeed. Since the osm-controller is responsible for creating the ValidatingWebhookConfiguration and the osm-controller hasn't started up, the MRC can be created by the osm-bootstrap during installation.

jaellio: I think this is less of an issue than I thought it was. I was concerned about the fresh install scenario, but I think this scenario is most applicable to enabling the `enableMeshRootCertificate` feature flag after install, and restarting all control plane components. 

Speaking with @jsturtevant, one possible solution to avoiding this timing issue is to only add the control plane webhook if `enableMeshRootCertificate` is set to true.

jsturtevant: Looks like this was dropped and we are still in good shape? The restart scenario was passing?  We also have tests covering the other scenario, where we do a fresh install and those are passing.

pkg/apis/config/v1alpha2/mesh_config.go
jaellio: Updated!

pkg/validator/patch.go
jaellio: Yep, I reverted the policy back to FAIL. The e2e that does a rolling restart is passing. I think conditionally applying the control plane webhook would be a nice feature to have, but I don't think the additional complexity is worth it in this case.

jsturtevant (approved):

cmd/cli/install_test.go
```diff
@@ -373,35 +373,6 @@ var _ = Describe("Running the install command", func() {
 			err := installCmd.run(config)
 			Expect(err.Error()).To(ContainSubstring("osm.vault.host is required"))
 		})
-
-		It("should error when token and token secret key are not set", func() {
```
keithmattix, line 377: Why were these removed?

keithmattix (approved):

cmd/cli/install_test.go
keithmattix: Spoke with Jackie about this offline; we're good to go

cmd/osm-bootstrap/osm-bootstrap.go
```diff
@@ -328,13 +324,24 @@ func (b *bootstrap) ensureMeshConfig() error {
 	return nil
 }
 
-// initiatilizeKubernetesEventsRecorder initializes the generic Kubernetes event recorder and associates it with
+// shouldCreateMeshRootCertificate gets the MeshConfig and returns the values of the enableMeshRootCertificate
+// feature flag
+func (b *bootstrap) shouldCreateMeshRootCertificate() bool {
+	config, err := b.configClient.ConfigV1alpha2().MeshConfigs(b.namespace).Get(context.TODO(), meshConfigName, metav1.GetOptions{})
```
steeling, line 330: can we have this also check if a meshRootCertificate already exists? If there are any we should not create a new one.

steeling (approved):

PR #5207, opened 2022-10-17
Support for root certificate rotation via CLI

### LINKED ISSUE #4835: Create e2e tests for root certificate rotation

jsturtevant:
**Description**:
This build on #5201 (required before this) to provide alpha support for rotation in the CLI. This is a proposal for adding a command to ease the burden of managing the MRCs and adds several tests for trust domain and switching between providers.

example usage would be:

```
./bin/osm alpha certificate rotate -d -y
```

Resolves #4835 

**Testing done**:

**Affected area**:

2. Is this a breaking change?

### DIFF
--- a/cmd/cli/alpha.go
+++ b/cmd/cli/alpha.go
@@ -1,0 +1,26 @@
+package main
+
+import (
+	"fmt"
+	"io"
+
+	"github.com/spf13/cobra"
+)
+
+const alphaDescription = `
+This command consists of multiple subcommands related that are in alpha.
+`
+
+func newAlphaCmd(out io.Writer) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "alpha",
+		Short: "commands that are in alpha",
+		Long:  alphaDescription,
+		Args:  cobra.NoArgs,
+		PersistentPreRun: func(cmd *cobra.Command, args []string) {
+			fmt.Fprintf(out, "*** This command is in Preview.  Only run in dev/test environments ***\n")
+		},
+	}
+	cmd.AddCommand(newCertificateCmd(out))
+	return cmd
+}
--- a/cmd/cli/certificate_rotate.go
+++ b/cmd/cli/certificate_rotate.go
@@ -181,7 +181,7 @@ func (r *rotateCmd) run() error {
 		}
 
 		if oldMrc.Spec.Provider.Tresor != nil {
-			r.deleteTresorSecret(oldMrc.Spec.Provider.Tresor.CA, newMrc.Spec.Provider.Tresor.CA)
+			err := r.deleteTresorSecret(oldMrc.Spec.Provider.Tresor.CA, newMrc.Spec.Provider.Tresor.CA)
 			if err != nil {
 				fmt.Printf("warning: unable to delete secret [%s]", oldMrc.Spec.Provider.Tresor.CA.SecretRef.Name)
 			}
--- a/cmd/cli/certificate_rotate.go
+++ b/cmd/cli/certificate_rotate.go
@@ -1,0 +1,352 @@
+package main
+
+import (
+	"context"
+	"crypto/rand"
+	"fmt"
+	"io"
+	"math/big"
+	"os"
+	"time"
+
+	"github.com/spf13/cobra"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+
+	"github.com/openservicemesh/osm/pkg/apis/config/v1alpha2"
+	"github.com/openservicemesh/osm/pkg/constants"
+	configClientset "github.com/openservicemesh/osm/pkg/gen/client/config/clientset/versioned"
+	osmConfigClient "github.com/openservicemesh/osm/pkg/gen/client/config/clientset/versioned"
+	"github.com/openservicemesh/osm/pkg/gen/client/config/clientset/versioned/scheme"
+)
+
+const rotateDesc = `
+This command rotates the OSM Root Certificate
+`
+
+const meshRotateExample = `
+# Rotate the mesh root certificate that is Active in the osm-system namespace
+osm alpha certificate rotate -d -y
+`
+
+type rotateCmd struct {
+	out io.Writer
+
+	meshName        string
+	newTrustDomain  string
+	clientSet       kubernetes.Interface
+	configClient    osmConfigClient.Interface
+	prompt          bool
+	deleteOld       bool
+	waitForRotation time.Duration
+	mrcFilePath     string
+	mrcName         string
+	prompter        func(a ...any) (int, error)
+}
+
+func newCertificateRotateCmd(out io.Writer) *cobra.Command {
+	rotate := &rotateCmd{
+		out: out,
+	}
+
+	cmd := &cobra.Command{
+		Use:     "rotate",
+		Short:   "rotate the MeshRootCertificate",
+		Long:    rotateDesc,
+		Example: meshRotateExample,
+		RunE: func(_ *cobra.Command, _ []string) error {
+			config, err := settings.RESTClientGetter().ToRESTConfig()
+			if err != nil {
+				return fmt.Errorf("error fetching kubeconfig: %w", err)
+			}
+
+			configClient, err := configClientset.NewForConfig(config)
+			if err != nil {
+				return fmt.Errorf("could not access Kubernetes cluster, check kubeconfig: %w", err)
+			}
+			rotate.configClient = configClient
+
+			clientset, err := kubernetes.NewForConfig(config)
+			if err != nil {
+				return fmt.Errorf("could not access Kubernetes cluster, check kubeconfig: %w", err)
+			}
+			rotate.clientSet = clientset
+			rotate.prompter = fmt.Scanln
+
+			return rotate.run()
+		},
+	}
+
+	f := cmd.Flags()
+
+	f.StringVar(&rotate.meshName, "mesh-name", defaultMeshName, "Name of the service mesh")
+	f.StringVarP(&rotate.newTrustDomain, "trust-domain", "t", "", "If specified the new cert will use this trust domain. Works with certificate provider tresor")
+	f.BoolVarP(&rotate.prompt, "accept", "y", false, "when specified it will not prompt for user input")
+	f.BoolVarP(&rotate.deleteOld, "delete", "d", false, "when specified it will delete the old MeshRootCertificate, otherwise the MeshRootCertificate will remain as inactive")
+	f.DurationVarP(&rotate.waitForRotation, "wait", "w", 60*time.Second, "Time to wait for certificate to propagate to all components at each step")
+	f.StringVarP(&rotate.mrcFilePath, "file", "f", "", "File to use for the new MRC. When not supplied the settings from the existing MRC are copied")
+	f.StringVarP(&rotate.mrcName, "meshrootcertificate", "c", "", "Existing MeshRootCertificate to rotate to active role using the name in currently configured namespace")
+	return cmd
+}
+
+func (r *rotateCmd) promptForContinue(msg string) bool {
+	if !r.prompt {
+		var rotate string
+		fmt.Fprintf(r.out, msg+" y/n:")
+		_, err := r.prompter(rotate)
+		if err != nil {
+			fmt.Printf("warning: error reading value from user: %s\n", err)
+			return true
+		}
+
+		if rotate != "y" {
+			return false
+		}
+	}
+
+	return true
+}
+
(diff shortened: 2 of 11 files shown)


jsturtevant:
This also partially addresses https://github.com/openservicemesh/osm/pull/5207, which was needed to be able to test rotation to vault from other providers.

cmd/cli/certificate_rotate.go
```diff
@@ -0,0 +1,352 @@
...
+		if rotate != "y" {
+			return false
+		}
+	}
+
+	return true
+}
+
+func (r *rotateCmd) run() error {
+	// TODO check if MRC feature is enabled (isMRCEnabled)
```
jaellio, line 112: https://github.com/openservicemesh/osm/pull/5204

cmd/cli/certificate_rotate.go
```diff
@@ -0,0 +1,352 @@
...
+		if err != nil {
+			return nil, err
+		}
+
+		newMRC, ok := obj.(*v1alpha2.MeshRootCertificate)
+		if !ok {
+			return nil, fmt.Errorf("file provided is not recognized as MeshRootCertificate")
+		}
+		if newMRC.Spec.Intent != v1alpha2.InactiveIntent {
+			return nil, fmt.Errorf("file provided must be in Inactive role")
```
jaellio, line 271:
```suggestion
			return nil, fmt.Errorf("file provided must be a MeshRootCertificate in Inactive role")
```

cmd/cli/certificate_rotate.go
```diff
@@ -0,0 +1,352 @@
...
+		return r.configClient.ConfigV1alpha2().MeshRootCertificates(settings.Namespace()).Create(context.Background(), newMRC, metav1.CreateOptions{})
+	}
+
+	return r.createFromExisting(mrc)
+}
+
+func (r *rotateCmd) createFromExisting(mrc *v1alpha2.MeshRootCertificate) (*v1alpha2.MeshRootCertificate, error) {
+	switch {
+	case mrc.Spec.Provider.Tresor != nil:
+		return r.createTressorFromExisting(mrc)
```
jaellio, line 291:
```suggestion
		return r.createTresorFromExisting(mrc)
```

cmd/cli/certificate_rotate.go
```diff
@@ -0,0 +1,352 @@
...
+	case mrc.Spec.Provider.Vault != nil:
+		// since we can't create the secrets in vault we require them to create a file
+		return nil, fmt.Errorf("please use -file option to provide a MeshRootCertificate Vault provider")
```
jaellio, line 294: Could the mrc name also be provided instead of the mrc file?

cmd/cli/certificate_rotate.go
```diff
@@ -0,0 +1,352 @@
...
+	case mrc.Spec.Provider.CertManager != nil:
+		// since we require them to set up cert-manager.io ahead of time we can create it for them
+		return nil, fmt.Errorf("please use -file option to provide a MeshRootCertificate for CertManager provider")
+	default:
+		return nil, fmt.Errorf("unknown certificate provider: %+v", mrc.Spec.Provider)
+	}
+}
+
+func generateName() string {
```
jaellio, line 303: Do you think this should be used by the osm-bootstrap when creating the default MRC?

cmd/cli/certificate_rotate.go
```diff
@@ -0,0 +1,352 @@
...
+
+	s := make([]rune, 8)
+	for i := range s {
+		num, _ := rand.Int(rand.Reader, big.NewInt(int64(len(chars))))
+		s[i] = chars[num.Int64()]
+	}
+	return fmt.Sprintf("%s-%x", constants.DefaultMeshRootCertificateName, string(s))
+}
+
+func (r *rotateCmd) createTressorFromExisting(mrc *v1alpha2.MeshRootCertificate) (*v1alpha2.MeshRootCertificate, error) {
```
jaellio, line 314:
```suggestion
func (r *rotateCmd) createTresorFromExisting(mrc *v1alpha2.MeshRootCertificate) (*v1alpha2.MeshRootCertificate, error) {
```

cmd/cli/certificate_rotate.go
```diff
@@ -0,0 +1,352 @@
...
+				Namespace: settings.Namespace(),
+			},
+			Spec: v1alpha2.MeshRootCertificateSpec{
+				TrustDomain: trustDomain,
+				Intent:      v1alpha2.InactiveIntent,
+				Provider: v1alpha2.ProviderSpec{
+					Tresor: &v1alpha2.TresorProviderSpec{
+						CA: v1alpha2.TresorCASpec{
+							SecretRef: v1.SecretReference{
+								Name:      newName,
```
jaellio, line 333: Maybe the secrets name could be osm-ca-bundle-\<random numbers matching the owning MRC\>?

pkg/certificate/providers/config.go
```diff
@@ -186,20 +186,20 @@ func (c *MRCProviderGenerator) getHashiVaultOSMCertificateManager(mrc *v1alpha2.
 	// A Vault address would have the following shape: "http://vault.default.svc.cluster.local:8200"
 	vaultAddr := fmt.Sprintf("%s://%s:%d", provider.Protocol, provider.Host, provider.Port)
 
-	// If the DefaultVaultToken is empty, query Vault token secret
-	var err error
-	vaultToken := c.DefaultVaultToken
-	if vaultToken == "" {
+	// If provider has secret ref filled, query Vault token secret
+	token := c.DefaultVaultToken
+	if provider.Token.SecretKeyRef.Name != "" && provider.Token.SecretKeyRef.Namespace != "" {
```
jaellio, line 191: Should we also check if the key != ""?
```suggestion
	if provider.Token.SecretKeyRef.Name != "" && provider.Token.SecretKeyRef.Namespace != "" && provider.Token.SecretKeyRef.Key != "" {
```

cmd/cli/alpha.go
```diff
@@ -0,0 +1,26 @@
...
+	"io"
+
+	"github.com/spf13/cobra"
+)
+
+const alphaDescription = `
+This command consists of multiple subcommands related that are in alpha.
+`
+
+func newAlphaCmd(out io.Writer) *cobra.Command {
```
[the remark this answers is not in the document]
jsturtevant: good question, I am up for either and easy change to make. I modeled it after `kubectl alpha` which I like because it is very clear that when using it.

cmd/cli/certificate_rotate.go
jsturtevant: I don't think so, it's nice to have a well-known name for the default, I think.

pkg/certificate/providers/config.go
jsturtevant: is the key required?

cmd/cli/certificate_rotate.go
jsturtevant: makes sense

cmd/cli/certificate_rotate.go
jsturtevant: yes!

cmd/cli/certificate_rotate.go
jsturtevant: added the issue number here, as well as the next line

pkg/certificate/providers/config.go
jaellio: I believe so and I don't think we provide a default. I will confirm this.

cmd/cli/alpha.go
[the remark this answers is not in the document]
jaellio: I like this approach, but I am curious to see what other maintainers think.

pkg/certificate/providers/config.go
jaellio: I think this would fail if the key wasn't set https://github.com/openservicemesh/osm/blob/0d36e289fcf3272e09d712286f5e21bde3ee60dd/pkg/certificate/providers/config.go#L219. I think I can confirm this by modifying the e2e tests or testing it manually.

keithmattix (approved):

cmd/cli/certificate_rotate.go
```diff
@@ -0,0 +1,352 @@
...
+
+	mrc.Spec.Intent = intent
+	_, err = r.configClient.ConfigV1alpha2().MeshRootCertificates(settings.Namespace()).Update(context.Background(), mrc, metav1.UpdateOptions{})
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func (r *rotateCmd) createOrFindNewMRC(mrc *v1alpha2.MeshRootCertificate) (*v1alpha2.MeshRootCertificate, error) {
```
steeling, line 238: nit: I'd consider having separate methods for create, createFromFile, and Find, and bring out the control logic outside of those methods.

This took me a sec to understand why it was structured this way.

cmd/cli/certificate_rotate.go
```diff
@@ -0,0 +1,352 @@
...
+
+	s := make([]rune, 8)
+	for i := range s {
+		num, _ := rand.Int(rand.Reader, big.NewInt(int64(len(chars))))
+		s[i] = chars[num.Int64()]
+	}
+	return fmt.Sprintf("%s-%x", constants.DefaultMeshRootCertificateName, string(s)), string(s)
+}
+
+func (r *rotateCmd) createTresorFromExisting(mrc *v1alpha2.MeshRootCertificate) (*v1alpha2.MeshRootCertificate, error) {
```
steeling, line 314: this method doesn't necessarily need to know that the mrc passed in is pre-existing, so I would consider changing the name to just `createTresorMRC`

cmd/cli/certificate_rotate.go
```diff
@@ -0,0 +1,352 @@
...
+	if err != nil {
+		return err
+	}
+	fmt.Fprintf(r.out, "Found MeshRootCertificate [%s] for mesh [%s] in namespace [%s]\n\n", oldMrc.Name, r.meshName, settings.Namespace())
+
+	newMrc, err := r.createOrFindNewMRC(oldMrc)
+	if err != nil {
+		return fmt.Errorf("unable to create or find new MeshRootCertificate: %s", err)
+	}
+	fmt.Fprintf(r.out, "Created new MeshRootCertificate [%s] in Inactive role\n", newMrc.Name)
```
steeling, line 128: this may have not created an MRC if it found an existing one instead.

cmd/cli/certificate_rotate.go
```diff
@@ -0,0 +1,352 @@
...
+
+	fmt.Fprintf(r.out, "Moving MeshRootCertificate [%s] to Passive role\n", newMrc.Name)
+	if !r.promptForContinue("Are you sure you want to initiate rotation?") {
+		return nil
+	}
+	err = r.updateCertificate(newMrc.Name, v1alpha2.PassiveIntent)
+	if err != nil {
+		return fmt.Errorf("unable to update MeshRootCertificate [%s] with role [%s]: %s", newMrc.Name, v1alpha2.PassiveIntent, err)
+	}
+	fmt.Fprintf(r.out, "waiting for propagation...\n\n")
```
steeling, line 138: On all of the sleeps I would add "waiting %s for propoagation", r.waitForRotation so they know how long we're waiting

cmd/cli/certificate_rotate.go
```diff
@@ -0,0 +1,352 @@
...
+
+	if r.deleteOld {
+		if !r.promptForContinue("Are you sure you want to delete the old MeshRootCertificate?") {
+			return nil
+		}
+		err = r.configClient.ConfigV1alpha2().MeshRootCertificates(settings.Namespace()).Delete(context.Background(), oldMrc.Name, metav1.DeleteOptions{})
+		if err != nil {
+			fmt.Printf("warning: unable to delete MeshRootCertificate [%s]", oldMrc.Name)
+		}
+		if oldMrc.Spec.Provider.Tresor != nil {
```
steeling, line 183: do we have a protection in place that the old mrc's tresor secret ref isn't the same as the new one?

steeling (approved):

cmd/cli/certificate_rotate.go
jsturtevant: added a check and warning

pkg/certificate/providers/config.go
jsturtevant: makes sense, I've added the check you suggested

jaellio (approved):

Issue #4930, opened 2022-07-20
Remove all* instances of functions that accept a client


steeling:
Ok, not quite all, but a good start would be in the `pkg/` subdirectory. We can omit anything for `cmd/`, and `pkg/cli`.

There's a lot of functions that take in a client object, like `GetKubernetesServerVersionNumber` that provide functionality specific to k8s, while taking in a k8s specific client. Instead, it should just be on the client object, as that allows us to abstract the underlying implementation (it should also probably be renamed to `GetInfrastructureServerVersion`, or something of the like

The exception to the rule is when passing in a client, to store it on a struct (dependency injection), such as in a `New*` func

PR #5208, opened 2022-10-17
wip: Remove pkg/certificate dependencies on k8s

shalier:
**Description**:

--will update--

**Testing done**:

**Affected area**:

2. Is this a breaking change?


pkg/certificate/providers/types.go
```diff
@@ -81,7 +81,8 @@ type MRCCompatClient struct {
 
 // MRCProviderGenerator knows how to convert a given MRC to its appropriate provider.
 type MRCProviderGenerator struct {
-	kubeClient kubernetes.Interface
+	compute.Interface
```
steeling, line 84: There's no need to embed this interface, since we don't have callers accessing it directly (we're not leveraging the passthrough features of embedded structs).

pkg/certificate/providers/config.go
```diff
@@ -275,6 +274,81 @@ func (c *MRCProviderGenerator) getCertManagerOSMCertificateManager(mrc *v1alpha2
 	if err != nil {
 		return nil, fmt.Errorf("error instantiating Jetstack cert-manager client: %w", err)
 	}
-
 	return cmClient, nil
 }
+
+// getCertFromKubernetes is a helper function that loads a certificate from a models.Secret
+func getCertFromKubernetes(ns string, secretName string, computeClient compute.Interface) (*certificate.Certificate, error) {
```
steeling, line 281: can this just be a utility method that doesn't accept a compute client [#4930](https://github.com/openservicemesh/osm/issues/4930), and instead extracts the certificate from a models.Secret object?

pkg/certificate/providers/config.go
```diff
@@ -275,6 +274,81 @@ func (c *MRCProviderGenerator) getCertManagerOSMCertificateManager(mrc *v1alpha2
...
+		log.Error().Err(err).Msg("Failed to create new Certificate from PEM")
+		return nil, err
+	}
+
+	return cert, nil
+}
+
+// getCertificateFromSecret is a helper function that ensures creation and synchronization of a certificate
+// using Kubernetes Secrets backend and API atomicity.
+func getCertificateFromSecret(ns string, secretName string, cert *certificate.Certificate, computeClient compute.Interface) (*certificate.Certificate, error) {
```
steeling, line 316: let's rename this to something more indicative of what it does. how about `ensureSecretExists`

let's also make this a pointer receiver on the MRCProviderGenerator, so we don't have to pass a compute client interface around

pkg/k8s/client.go
```diff
@@ -120,6 +122,9 @@ func (c *Client) ListSecrets() []*models.Secret {
 
 // UpdateSecret updates the given secret
 func (c *Client) UpdateSecret(ctx context.Context, secret *models.Secret) error {
```
steeling, line 124: The k8s package shouldn've have any knowledge about OSM domain objects (models.Secret). Instead, it should accept a corev1.Secret.

The compute/kube/client.go.Client object is responsible for doing the translation between the 2

steeling: Let's fix Get/List secret here as well.

jaellio:
Spoke offline with @shalier. Closing this PR for now

PR #5216, opened 2022-10-19
chore(release) bump Chart.yaml version to 1.1.3

shalier:
**Description**:
Bumping the Chart.yaml version per the release guide

**Affected area**:

2. Is this a breaking change?

### DIFF
--- a/docs/release_notes.md
+++ b/docs/release_notes.md
@@ -1,10 +1,19 @@
 # Release Notes
 
+## Release v1.1.3
+
+### Notable changes
+- Certificate rotation now ignores monotonic clock readings when checking expiration
+- Add keep-alive time to detect sidecar disconnections
+
+## Release v1.1.2
+
+### Notable changes
+- Remove CRD conversion webhook (the webhook does not act differently than how if the CRD conversion strategy was set to `None`)
+
 ## Release v1.1.0
 
 ### Notable changes
-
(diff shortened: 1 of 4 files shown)


docs/release_notes.md
```diff
@@ -3,7 +3,8 @@
 ## Release v1.1.0
 
 ### Notable changes
-
+- Certificate rotation now ignores monotonic clock readings when checking expiration
```
jaellio, line 6: Should we add a separate release note for patch releases?

shalier: Like a different section from notable changes or like a header under notable changes?
I added the previous patch release change under Notable changes though, so then maybe we'd want to implement this for releases going forward?

shashankram: You should have a separate note for each release, including patch releases. This file typically is used to populate the final release notes. Regardless, this should accurately convey the information for releases and not update the release notes of an already released version.

shashankram (changes requested):
Thanks for the change. Could you create a separate release note section for the patch release?

docs/release_notes.md
shalier: Added sections for v1.1.2 and v1.1.3 and included the notable changes for each

jaellio (approved):

keithmattix (approved):

PR #5224, opened 2022-10-20
k8s: update app protocol detection for k8s service port

### LINKED ISSUE #5223: Make app protocol detection of service port less error prone
Currently, OSM implicitly uses http as the app protocol for a k8s service port if it can't derive the port's protocol using the `appProtocol` field or from the port's name. 
https://github.com/openservicemesh/osm/blob/ee7e846a2ff0005f962e8bd867944c3f2a46b6f7/pkg/compute/kube/client.go#L511-L525

It's possible for a k8s service to use the same port for multiple protocols, e.g. port 53 for DNS can be TCP or UDP. When a port's protocol is unset based on the [app protocol selection behavior](https://release-v1-2.docs.openservicemesh.io/docs/guides/app_onboarding/app_protocol_selection/), the configuration can result in duplicate Envoy filter_chain_match configuration. This happens because OSM implicitly sets the appProtocol to `http` for unknown port protocols. So a configuration such as the below can result in duplicate filter_match_configuration:
```
ports:
  - name: myPort
    port: 8080
    protocol: udp
  - name: myPort2
    port: 8080
    protocol: tcp
    appProtocol: http
```

> Note: the protocol is implicitly set to http when it's unspecified for backward compatibility reasons. We do not want to break existing users relying on this behavior, as http apps are more common. In the future, we can consider making the protocol specification explicit at all times.

This issue proposes to do the following:
1. When `port.protocol` is UDP, to not set MeshService.Protocol to `http`, but to leave it as is and let it result in an error log when the code that builds the config sees an unsupported protocol
2. Add a warn log when MeshService.Protocol is implicitly set to `http`. This log will be useful to identify scenarios where there is a protocol related error with traffic.

shashankram:
**Description**:
This change updates the app protocol detection for a service port to eliminate a misleading error scenario that results from using the same port for multiple protocols.

It does the following:

1. Only uses http implicitly for TCP based ports that do not have appProtocol or a valid protocol prefix in the port name.

2. Adds a warn log when implicitly using http as the app protocol for TCP based ports.

The result is that duplicate filter chain matches
are eliminated with the same port is used for different protocols, eg. port 8080 being used for TCP and UDP.

Resolves #5223

**Testing done**:
Unit tests

**Affected area**:

3. Has documentation corresponding to this change been updated in the [osm-docs](https://github.com/openservicemesh/osm-docs) repo (if applicable)? `TBD`

### DIFF
--- a/pkg/compute/kube/client.go
+++ b/pkg/compute/kube/client.go
@@ -661,7 +661,7 @@ func (c *client) serviceToMeshServices(svc corev1.Service) []service.MeshService
 		// Order of Preference is:
 		// 1. port.appProtocol field
 		// 2. protocol prefixed to port name (e.g. tcp-my-port)
-		// 3. default to http for for TCP ports
+		// 3. default to http for TCP ports
 		var protocol string
 		for _, p := range constants.SupportedProtocolsInMesh {
 			if strings.HasPrefix(portSpec.Name, p+"-") {
--- a/pkg/compute/kube/client.go
+++ b/pkg/compute/kube/client.go
@@ -661,8 +661,8 @@ func (c *client) serviceToMeshServices(svc corev1.Service) []service.MeshService
 		// Order of Preference is:
 		// 1. port.appProtocol field
 		// 2. protocol prefixed to port name (e.g. tcp-my-port)
-		// 3. default to http
-		protocol := constants.ProtocolHTTP
+		// 3. default to http for for TCP ports
+		var protocol string
 		for _, p := range constants.SupportedProtocolsInMesh {
 			if strings.HasPrefix(portSpec.Name, p+"-") {
 				protocol = p
@@ -672,6 +672,22 @@ func (c *client) serviceToMeshServices(svc corev1.Service) []service.MeshService
 
 		// use port.appProtocol if specified, else use port protocol
 		meshSvc.Protocol = pointer.StringDeref(portSpec.AppProtocol, protocol)
+
+		// If app protocol was not detected and the protocol is TCP, use HTTP
+		// as the default app protocol.
+		// This is required for backward compatibility where users may be relying
+		// on implicit HTTP protocol selection.
+		//
+		// Note: it's possible for MeshService.Protocol to be unset for
+		// non TCP ports (eg. UDP) if we could not infer the corresponding
+		// app protocol via the appProtocol field or the port name.
+		// This will result in an error log for the specific port when the
+		// Envoy config is built for that port.
+		if meshSvc.Protocol == "" && portSpec.Protocol == corev1.ProtocolTCP {
+			log.Warn().Msgf("Did not detect app protocol for TCP port name=%s number=%d for service %s/%s, defaulting to http",
+				portSpec.Name, portSpec.Port, svc.Namespace, svc.Name)
+			meshSvc.Protocol = constants.ProtocolHTTP
+		}
 
 		// The endpoints for the kubernetes service carry information that allows
 		// us to retrieve the TargetPort for the MeshService.
--- a/pkg/compute/kube/client_test.go
+++ b/pkg/compute/kube/client_test.go
@@ -2220,7 +2220,7 @@ func TestServiceToMeshServices(t *testing.T) {
 			name: "k8s service with single port and endpoint, no appProtocol set",
 			// Single port on the service maps to a single MeshService.
 			// Since no appProtocol is specified, MeshService.Protocol should default
-			// to http.
+			// to http since Port.Protocol=TCP.
 			svc: corev1.Service{
 				ObjectMeta: metav1.ObjectMeta{
 					Namespace: "ns1",
@@ -2319,7 +2319,7 @@ func TestServiceToMeshServices(t *testing.T) {
 			name: "k8s headless service with single port and endpoint, no appProtocol set",
 			// Single port on the service maps to a single MeshService.
 			// Since no appProtocol is specified, MeshService.Protocol should default
-			// to http.
+			// to http because Port.Protocol=TCP
 			svc: corev1.Service{
 				ObjectMeta: metav1.ObjectMeta{
 					Namespace: "ns1",
--- a/pkg/compute/kube/client_test.go
+++ b/pkg/compute/kube/client_test.go
@@ -438,7 +438,11 @@ func TestGetServicesForServiceIdentity(t *testing.T) {
 						Selector: map[string]string{
 							"k1": "v1", // matches labels on pod ns1/p1
 						},
-						Ports: []corev1.ServicePort{{}},
+						Ports: []corev1.ServicePort{
+							{
+								Protocol: corev1.ProtocolTCP,
+							},
+						},
 					},
 				},
 				{
@@ -960,7 +964,11 @@ func TestListServicesForProxy(t *testing.T) {
 						Selector: map[string]string{
 							"k1": "v1",
 						},
-						Ports: []corev1.ServicePort{{}},
+						Ports: []corev1.ServicePort{
+							{
+								Protocol: corev1.ProtocolTCP,
+							},
+						},
 					},
 				},
 				{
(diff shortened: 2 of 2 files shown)


pkg/compute/kube/client.go
```diff
@@ -673,6 +673,16 @@ func (c *client) serviceToMeshServices(svc corev1.Service) []service.MeshService
 		// use port.appProtocol if specified, else use port protocol
 		meshSvc.Protocol = pointer.StringDeref(portSpec.AppProtocol, protocol)
 
+		// If app protocol was not detected and the protocol is TCP, use HTTP
+		// as the default app protocol.
+		// This is required for backward compatibility where users may be relying
+		// on implicit HTTP protocol selection.
+		if meshSvc.Protocol == "" && portSpec.Protocol == corev1.ProtocolTCP {
```
keithmattix, line 680: I think this conditional should probably be relaxed. As it stands, we still have error/warning log containing the original protocol in the `protocol: udp` case.

pkg/compute/kube/client_test.go
```diff
@@ -2500,6 +2514,71 @@ func TestK8sServicesToMeshServices(t *testing.T) {
...
+					Port:       80,
+					TargetPort: 8080,
+					Protocol:   "http",
+				},
+				{
+					Namespace:  "ns1",
+					Name:       "s1",
+					Port:       80,
+					TargetPort: 8080,
+					Protocol:   "",
```
keithmattix, line 2578: I don't know if we want a blank protocol to be the end-result here; don't we want to preserve the original protocol and make Envoy error later down the stack?

pkg/compute/kube/client.go
shashankram: Sorry, I didn't follow your comment. Could you clarify what scenario results in which error?

pkg/compute/kube/client_test.go
shashankram: We should leave this blank as the app protocol is unknown. The app protocol is not to be confused with the protocol field. We want to preserve the app protocol, not the protocol. If it's unknown, we should leave it as is for non TCP case.

keithmattix: Ah, I see what you're saying; we're just not going to make OSM aware of anything non-TCP. Got it 👍🏾

pkg/compute/kube/client.go
shashankram: Your assessment is correct. An error in LDS will indicate unsupported/unknown app protocols. I'll add a comment regarding the empty protocol case.

keithmattix (approved):

jsturtevant:
changes look sane. I don't know if understand why we want the value to result in an error when building the config, If we know its invalid shouldn't we error early?

from the linked issue: `When port.protocol is UDP, to not set MeshService.Protocol to http, but to leave it as is and let it result in an error log when the code that builds the config sees an unsupported protocol`

shashankram:
> changes look sane. I don't know if understand why we want the value to result in an error when building the config, If we know its invalid shouldn't we error early?
> 
> from the linked issue: `When port.protocol is UDP, to not set MeshService.Protocol to http, but to leave it as is and let it result in an error log when the code that builds the config sees an unsupported protocol`

It's not an invalid config per-se. The problem currently is that we see mysterious errors (duplicate filter_chain_match) that don't make the issue obvious when it occurs under certain scenarios. With this change, we'd get that error log to correctly root cause this potential scenario. Other than that, since we don't support UDP proxying, it also seems wrong to implicitly set the app protocol to `http` for UDP ports, which is addressed in this change. 

Moreover, given this arises from the k8s service spec, it's hard to enforce how users author the config. Currently, the only thing we expect is to correctly configure [app protocol selection](https://release-v1-2.docs.openservicemesh.io/docs/guides/app_onboarding/app_protocol_selection/).

pkg/compute/kube/client.go
```diff
@@ -661,8 +661,8 @@ func (c *client) serviceToMeshServices(svc corev1.Service) []service.MeshService
 		// Order of Preference is:
 		// 1. port.appProtocol field
 		// 2. protocol prefixed to port name (e.g. tcp-my-port)
-		// 3. default to http
-		protocol := constants.ProtocolHTTP
+		// 3. default to http for for TCP ports
```
SanyaKochhar, line 664: nit:
```suggestion
		// 3. default to http for TCP ports
```

SanyaKochhar (approved):

pkg/compute/kube/client.go
shashankram: Done, thanks

pkg/compute/kube/client.go
```diff
@@ -673,6 +673,22 @@ func (c *client) serviceToMeshServices(svc corev1.Service) []service.MeshService
...
+		// non TCP ports (eg. UDP) if we could not infer the corresponding
+		// app protocol via the appProtocol field or the port name.
+		// This will result in an error log for the specific port when the
+		// Envoy config is built for that port.
+		if meshSvc.Protocol == "" && portSpec.Protocol == corev1.ProtocolTCP {
+			log.Warn().Msgf("Did not detect app protocol for TCP port name=%s number=%d for service %s/%s, defaulting to http",
+				portSpec.Name, portSpec.Port, svc.Namespace, svc.Name)
+			meshSvc.Protocol = constants.ProtocolHTTP
+		}
+
```
jaellio, line 691: Should we log if the `meshSvc.Protocol` was never set or will this failure be seen later?

pkg/compute/kube/client_test.go
```diff
@@ -2315,8 +2328,9 @@ func TestK8sServicesToMeshServices(t *testing.T) {
 				Spec: corev1.ServiceSpec{
 					Ports: []corev1.ServicePort{
 						{
-							Name: "p1",
-							Port: 80,
+							Name:     "p1",
```
jaellio, line 2331: nit: In the above comment for this test, we could clarify the default is only set to http if the port's protocol is TCP

pkg/compute/kube/client.go
shashankram: This failure will be handled later where the protocol is actually used, so we can log it with enough context. Further, when we do add support for the currently unsupported protocols, we wouldn't need to update this code but rather the place where it logs an error.

jaellio: Okay, sounds good. Thanks for clarifying!

jaellio (approved):

pkg/compute/kube/client_test.go
shashankram: Done, thanks

jsturtevant:
>It's not an invalid config per-se.

I guess I don't understand how `we don't support UDP proxying,` and it not being a valid config works together?  Do we support UDP in some scenarios and not others?

shashankram:
> > It's not an invalid config per-se.
> 
> I guess I don't understand how `we don't support UDP proxying,` and it not being a valid config works together? Do we support UDP in some scenarios and not others?

To clarify, it's a valid k8s config, but UDP proxying won't work in OSM. This would result in mysterious errors as I noted earlier, which would now be obvious with the code logging an error when it sees an unsupported protocol.

jsturtevant:
I guess my earlier comments stands. If we don't support UDP, Why let the config get written and an update pushed to envoy with bad configuration, then go searching for logs when something doesn't work as expected.  Wouldn't it be better user experience to fail earlier in the process?

shashankram:
> I guess my earlier comments stands. If we don't support UDP, Why let the config get written and an update pushed to envoy with bad configuration, then go searching for logs when something doesn't work as expected. Wouldn't it be better user experience to fail earlier in the process?

It doesn't get pushed to Envoy with a bad config. We push whatever is correct. Whatever isn't currently capable is logged as an error. So in this case, TCP proxying would work for that port. This is consistent with how unsupported configs are handled, with an error log. We can't prevent users from creating k8s services. So in this case, we would suggest them to use a port exclusion rule if they want UDP proxying, so it bypasses the sidecar completely. But that doesn't mean that port will magically not exist in the k8s service config. The point of letting this fall through to the place where it logs an error is to not silently drop the config.

jsturtevant:
got it, Thanks for helping me understand!

Issue #5219, opened 2022-10-20
Remove pkg/debugger dependency on meshcatalog


nshankar13:
The debugConfig should no longer be coupled to a meshcatalog object. This will require invoking methods such as `ListTrafficSplits()`, `ListServiceAccountsFromTrafficTargets()`, etc from the compute client. 

**Please describe the Improvement and/or Feature Request**

**Scope (please mark with X where applicable)**
- New Functionality          [ ]
- Install                    [ ]
- SMI Traffic Access Policy  [ ]
- SMI Traffic Specs Policy   [ ]
- SMI Traffic Split Policy   [ ]
- Permissive Traffic Policy  [ ]
- Ingress                    [ ]
- Egress                     [ ]
- Envoy Control Plane        [ ]
- CLI Tool                   [ ]
- Metrics                    [ ]
- Certificate Management     [ ]
- Sidecar Injection          [ ]
- Logging                    [ ]
- Debugging                  [ ]
- Tests                      [ ]
- CI System                  [ ]
- Demo                       [ ]
- Project Release            [ ]

**Possible use cases**

PR #5225, opened 2022-10-20
Remove meshcatalog, getSmiPoliciesHandler endpoint, and getOSMConfigHandler endpoint from debugger

nshankar13:

**Description**: Remove the meshcatalog from the debug config. This entails calling the meshcatalog methods from the compute Client now. Also, ListServiceAccountsFromTrafficTargets is moved from the catalog pkg to compute client. No functional changes. 

Update: After an offline discussion, it was decided that the getSmiPoliciesHandler and getOSMConfigHandler endpoints can be removed altogether. 

Resolves: https://github.com/openservicemesh/osm/issues/5219

**Testing done**:

**Affected area**:

2. Is this a breaking change?


pkg/debugger/policy_test.go
```diff
@@ -4,49 +4,46 @@ import (
 	"net/http/httptest"
 	"testing"
 
-	"github.com/golang/mock/gomock"
+	gomock "github.com/golang/mock/gomock"
+	"github.com/openservicemesh/osm/pkg/compute/kube"
+	"github.com/openservicemesh/osm/pkg/k8s"
```
jsturtevant, line 9: this linter error is annoying 😃 fyi you can configure your IDE to do this automatically which has solved having to update this everytime

jsturtevant:
lgtm

pkg/debugger/policy_test.go
nshankar13: Actually probably going to get rid of this file altogether due to upcoming change I'm about to make, but will keep this in mind!

pkg/debugger/namespace_test.go
```diff
@@ -4,30 +4,46 @@ import (
...
-
-	mock := catalog.NewMockMeshCataloger(gomock.NewController(t))
-
-	ds := DebugConfig{
-		meshCatalog: mock,
-	}
-	monitoredNamespacesHandler := ds.getMonitoredNamespacesHandler()
+	mockCtrl := gomock.NewController(t)
+	mockK8s := k8s.NewMockController(mockCtrl)
+	computeClient := kube.NewClient(mockK8s)
```
steeling, line 23: you can just mock out the compute client itself

pkg/debugger/policy.go
```diff
@@ -1,25 +1,12 @@
...
 )
 
-type policies struct {
-	TrafficSplits   []*split.TrafficSplit        `json:"traffic_splits"`
-	ServiceAccounts []identity.K8sServiceAccount `json:"service_accounts"`
-	RouteGroups     []*spec.HTTPRouteGroup       `json:"route_groups"`
-	TrafficTargets  []*access.TrafficTarget      `json:"traffic_targets"`
-}
-
 func (ds DebugConfig) getOSMConfigHandler() http.Handler {
```
steeling, line 10: i don't see any reason to keep `getOSMConfigHandler` around either

steeling (approved):
Can you change the description and the title to reflect the new changes?

jsturtevant:
>Update: After an offline discussion, it was decided that the getSmiPoliciesHandler and getOSMConfigHandler endpoints can be removed altogether.

For posterity, what was the reasoning behind removing these?  Is there any tooling like the bug check that uses those endpoints?

nshankar13:
@jsturtevant we decided to get rid of these endpoints because a user can use kubectl to get the same info. I searched through the repo and these endpoints don't get invoked anywhere else.

jsturtevant (approved):

keithmattix (approved):

PR #5244, opened 2022-11-08
ref(catalog): remove meshCatalogger interface and move compute.Interface to catalog pkg

### LINKED ISSUE #5156: Remove the mock mesh catalog, and the meshcatalogger interface
With the new split of MeshCatalog, compute.Interface, and k8s.Client, there is no need to have an interface for the MeshCatalog. The MeshCatalog strictly contains business logic, with no actual API calls being made within the catalog itself.

As such there's no need to have an interface for it. The biggest hurdle will be to migrate all of the `NewMockMeshCataloger` calls to instead mock the `compute.Interface` via `compute.NewMockInterface`.

Finally, interfaces should be defined where they are consumed, not where their implementations are defined. So let's move the compute.Interface (and it's mock) to the catalog package.

This should probably be in 3-4 PR's, e.g:

1. Migrate NewMockMeshCataloger to NewMockInterface
2. Remove the MeshCatalogger interface
3. Move the compute.Interface and NewMockInterface to the catalog package

SanyaKochhar:
**Description**:
* Replaces mock meschatalogger with real mesh catalog
* Moves compute.Interface and its mock to catalog pkg
* Resolves #5156 

**Testing done**:

**Affected area**:

3. Has documentation corresponding to this change been updated in the [osm-docs](https://github.com/openservicemesh/osm-docs) repo (if applicable)? n/a


pkg/catalog/catalog.go
```diff
@@ -4,17 +4,22 @@ import (
...
 
 	"github.com/openservicemesh/osm/pkg/certificate"
-	"github.com/openservicemesh/osm/pkg/compute"
 	"github.com/openservicemesh/osm/pkg/messaging"
 	"github.com/openservicemesh/osm/pkg/ticker"
 )
 
+// MeshCatalog is the struct for the service catalog
+type MeshCatalog struct {
+	Interface
```
steeling, line 13: Let's rename this to `computeInterface` (unexported)

pkg/catalog/catalog.go
```diff
@@ -4,17 +4,22 @@ import (
...
+	certManager *certificate.Manager
+}
+
 // NewMeshCatalog creates a new service catalog
-func NewMeshCatalog(computeInterface compute.Interface, certManager *certificate.Manager,
+func NewMeshCatalog(catalogInterface Interface, certManager *certificate.Manager,
```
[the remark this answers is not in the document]
steeling: I'd call it the compute interface still

pkg/catalog/fake/fake.go
```diff
@@ -3,15 +3,13 @@ package fake
...
-
 	"github.com/openservicemesh/osm/pkg/catalog"
 	tresorFake "github.com/openservicemesh/osm/pkg/certificate/providers/tresor/fake"
 	"github.com/openservicemesh/osm/pkg/messaging"
 )
 
-// NewFakeMeshCatalog creates a new struct implementing catalog.MeshCataloger interface used for testing.
-func NewFakeMeshCatalog(provider compute.Interface) *catalog.MeshCatalog {
+// NewFakeMeshCatalog creates a new catalog.MeshCatalog struct used for testing.
+func NewFakeMeshCatalog(provider catalog.Interface) *catalog.MeshCatalog {
```
steeling, line 12: let's actually delete this whole file. Should be simple to remove all instances of it

pkg/catalog/types.go
```diff
@@ -1,92 +1,112 @@
...
-	ListOutboundServicesForIdentity(identity.ServiceIdentity) []service.MeshService
+// Interface is an interface to be implemented by components abstracting Kubernetes, and other compute/cluster providers
+type Interface interface {
+	k8s.PassthroughInterface
+	// GetSecret returns the secret for a given namespace and secret name
+	GetSecret(string, string) *models.Secret
 
-	// ListInboundServiceIdentities lists the downstream service identities that are allowed to connect to the given service identity
-	ListInboundServiceIdentities(identity.ServiceIdentity) []identity.ServiceIdentity
+	// ListSecrets returns a list of secrets
```
steeling, line 40: Let's remove all methods that aren't actually used within the catalog package.

pkg/catalog/types.go
```diff
@@ -1,92 +1,112 @@
...
+const (
+	inbound  trafficDirection = "inbound"
+	outbound trafficDirection = "outbound"
+)
 
-	// ListOutboundServicesForIdentity list the services the given service identity is allowed to initiate outbound connections to
-	ListOutboundServicesForIdentity(identity.ServiceIdentity) []service.MeshService
+// Interface is an interface to be implemented by components abstracting Kubernetes, and other compute/cluster providers
+type Interface interface {
+	k8s.PassthroughInterface
```
steeling, line 36: Similarly, let's expand out the k8s.PassthroughInterface, and delete those interfaces (move them into where they are used).

Can be done later though, but maybe add a TODO

pkg/certificate/providers/mrc.go
```diff
@@ -6,15 +6,15 @@ import (
...
 	"github.com/openservicemesh/osm/pkg/certificate"
-	"github.com/openservicemesh/osm/pkg/compute"
 )
 
 // MRCComposer is a composer object that allows consumers
 // to observe MRCs (via List() and Watch()) as well as generate
 // `certificate.Provider`s from those MRCs
 type MRCComposer struct {
-	compute.Interface
+	catalog.Interface
```
steeling, line 17: "interfaces should be defined where they are consumed"

let's create a new interface here with *just* the methods required.

pkg/debugger/server.go
```diff
@@ -39,7 +39,7 @@ func (ds DebugConfig) GetHandlers() map[string]http.Handler {
 // NewDebugConfig returns an implementation of DebugConfig interface.
 func NewDebugConfig(certDebugger *certificate.Manager, xdsDebugger XDSDebugger,
 	proxyRegistry *registry.ProxyRegistry, kubeConfig *rest.Config, kubeClient kubernetes.Interface,
-	computeClient compute.Interface, msgBroker *messaging.Broker) DebugConfig {
+	computeClient catalog.Interface, msgBroker *messaging.Broker) DebugConfig {
```
steeling, line 42: same as above comment

pkg/catalog/fake/fake.go
SanyaKochhar: just curious on the reasoning behind deleting this? It's used in ~7 different tests so we'd just end up repeating this in those

pkg/certificate/providers/mrc.go
SanyaKochhar: I believe GetMeshConfig() is the only method used by pkg/certificate/providers -- does this warrant a whole new interface?

steeling: yup. It keeps packages isolated and decoupled from each other. It allows allows making very trivial implementations for testing purposes if desired.

pkg/catalog/fake/fake.go
steeling: all it does is take in a compute interface, which is all we need to provide the real thing. so we can replace all instances with calls to catalog.NewMeshCatalog instead of this

mockspec/rules
```diff
@@ -7,6 +7,16 @@ k8s; pkg/k8s/mock_controller_generated.go; github.com/openservicemesh/osm/pkg/k8
 
 # pkg/debugger
 debugger; pkg/debugger/mock_debugger_generated.go; github.com/openservicemesh/osm/pkg/debugger; XDSDebugger
+debugger; pkg/debugger/mock_debugger_infra_client_generated.go; github.com/openservicemesh/osm/pkg/debugger; DebuggerInfraClient
 
-# pkg/compute
-compute; pkg/compute/mock_compute_client_generated.go; github.com/openservicemesh/osm/pkg/compute; Interface
+# pkg/catalog
+catalog; pkg/catalog/mock_compute_interface_generated.go; github.com/openservicemesh/osm/pkg/catalog; Interface
```
steeling, line 13: I'd choose a different name than interface

mockspec/rules
```diff
@@ -7,6 +7,16 @@ k8s; pkg/k8s/mock_controller_generated.go; github.com/openservicemesh/osm/pkg/k8
...
+
+# pkg/injector
+injector; pkg/injector/mock_injector_infra_client_interface_generated.go; github.com/openservicemesh/osm/pkg/injector; InjectorInfraClient
```
steeling, line 16: injector.InjectorInfraClient stutters, and it's not used outside of the package. So you can unexport it, and just call it infraClient.

Same for the otheres

mockspec/rules
```diff
@@ -7,6 +7,16 @@ k8s; pkg/k8s/mock_controller_generated.go; github.com/openservicemesh/osm/pkg/k8
 
 # pkg/debugger
 debugger; pkg/debugger/mock_debugger_generated.go; github.com/openservicemesh/osm/pkg/debugger; XDSDebugger
+debugger; pkg/debugger/mock_debugger_infra_client_generated.go; github.com/openservicemesh/osm/pkg/debugger; DebuggerInfraClient
```
steeling, line 10: Instead of creating generated mocks for all of these, can we manually create a struct that adheres to the interface to support what we want for tests? This is likely to be much more idiomatic for the smaller interfaces. Debugger is a good one to start with

pkg/catalog/types.go
```diff
@@ -1,92 +1,89 @@
...
 
-	// GetInboundMeshHTTPRouteConfigsPerPort returns a map of the given inbound traffic policy per port for the given upstream identity and services
-	GetInboundMeshHTTPRouteConfigsPerPort(identity.ServiceIdentity, []service.MeshService) map[int][]*trafficpolicy.InboundTrafficPolicy
+	// GetResolvableEndpointsForService returns the expected endpoints that are to be reached when the service FQDN is resolved under
+	// the scope of the provider
+	GetResolvableEndpointsForService(service.MeshService) []endpoint.Endpoint
 
-	// GetOutboundMeshClusterConfigs returns the cluster configs for the outbound mesh traffic policy for the given downstream identity
-	GetOutboundMeshClusterConfigs(identity.ServiceIdentity) []*trafficpolicy.MeshClusterConfig
+	IsMetricsEnabled(*models.Proxy) (bool, error)
```
steeling, line 58: This method isn't required on this interface

pkg/certificate/providers/config.go
```diff
@@ -44,7 +43,7 @@ var getCA = func(i certificate.Issuer) (pem.RootCertificate, error) {
 // NewCertificateManager returns a new certificate manager with a MRC compat client.
 // TODO(4713): Remove and use NewCertificateManagerFromMRC
 func NewCertificateManager(ctx context.Context, kubeClient kubernetes.Interface, kubeConfig *rest.Config,
-	providerNamespace string, option Options, computeClient compute.Interface, checkInterval time.Duration, trustDomain string) (*certificate.Manager, error) {
+	providerNamespace string, option Options, computeClient ProvidersInfraClient, checkInterval time.Duration, trustDomain string) (*certificate.Manager, error) {
```
steeling, line 46: providers isn't a very descriptive name. Provider of what? I think InfraClient is good enough on its own?

pkg/certificate/providers/config_test.go
```diff
@@ -30,8 +29,8 @@ import (
 
 func TestGetCertificateManager(t *testing.T) {
 	mockCtrl := gomock.NewController(t)
-	computeMock := compute.NewMockInterface(mockCtrl)
-	computeMock.EXPECT().GetMeshConfig().AnyTimes()
+	MockInfraClient := NewMockProvidersInfraClient(mockCtrl)
```
steeling, line 32: local vars shouldn't be capitalized

steeling, line 32: this seems simple enough to avoid mock generation as well. Can we create a simple struct here in its place?

steeling (approved):

Issue #5226, opened 2022-10-20
Optimize MRC event handling on startup


jaellio:
In response to each MRCEvent the MRC reconciler lists all MRCs in the mesh. Listing all MRCs for each event is done to achieve a level-triggered design. On startup or restart, each existing MRC triggers a create event resulting in O(NM) operations, where N is the number events and M is the number of MRCs in the mesh. This could potentially have an impact on the mesh's performance.

One approach to optimizing event handling is to use a custom informer index: 
- example creation of a custom informer index: https://github.com/kubernetes-sigs/controller-runtime/blob/f46919744bee01060c9084a285e049afffd38c9d/pkg/cache/informer_cache.go#L180
- example caller to create index: https://github.com/kubernetes-sigs/controller-runtime/blob/d887b2f1c68fee51b20378ff2a0bd6a8a77674dd/pkg/client/example_test.go
- example of listing using a custom informer index: https://github.com/kubernetes-sigs/controller-runtime/blob/8ad090e0663e046c5a9dad68a572af0e30626c85/pkg/cache/internal/cache_reader.go#L126

Related discussion: https://github.com/openservicemesh/osm/pull/5201#discussion_r996326422

Issue #5139, opened 2022-09-16
E2E tests are flaking


jsturtevant:
**Bug description**:

I noticed some flakes where random things failing. I am not sure if these are all related and after looking at logs and commits in the last few days I don't see any thing jumping out as the issue.  

- 9/16/2022 - https://github.com/openservicemesh/osm/actions/runs/3069568739/jobs/4958431162 (kafka doesn't reach concensus)

```
Using config: /opt/bitnami/zookeeper/bin/../conf/zoo.cfg
> (kafka-zookeeper-0) ZK status check failed: expected nil err, got command terminated with exit code 1
> (kafka-zookeeper-1) ZK status check succeeded!
> (kafka-zookeeper-2) ZK status check succeeded!
> (kafka-zookeeper-0) Stdout /opt/bitnami/java/bin/java
Client port found: 2181. Client address: localhost. Client SSL: false.
```

- 9/16/2022 - https://github.com/openservicemesh/osm/actions/runs/3069634707/jobs/4958564544#step:5:399 (nginx doesn't start)

```
Deployment is not ready: ingress-ns/ingress-nginx-controller. 0 out of 1 expected pods are ready
```

- 9/14/2022 - https://github.com/openservicemesh/osm/actions/runs/3056022379/jobs/4929818398 (tcp server first)

```
  TCP server-first traffic [It]
  /home/runner/work/osm/osm/tests/e2e/e2e_tcp_server_first_test.go:27

  Timed out after 5.002s.
  Didn't get expected response from server
  Expected
      <string>: 
  to contain substring
      <string>: 
      y
```

**Affected area (please mark with X where applicable)**:
- Install                    [ ]
- SMI Traffic Access Policy  [ ]
- SMI Traffic Specs Policy   [ ]
- SMI Traffic Split Policy   [ ]
- Permissive Traffic Policy  [ ]
- Ingress                    [ ]
- Egress                     [ ]
- Envoy Control Plane        [ ]
- CLI Tool                   [ ]
- Metrics                    [ ]
- Certificate Management     [ ]
- Sidecar Injection          [ ]
- Logging                    [ ]
- Debugging                  [ ]
- Tests                      [ ]
- Demo                       [ ]
- CI System                  [ ]

**Expected behavior**:

**Steps to reproduce the bug (as precisely as possible)**:

**How was OSM installed?**:

**Anything else we need to know?**:

**Bug report archive**:

**Environment**:
- OSM version (use `osm version`):
[...]

PR #5249, opened 2022-11-10
nit(certificates): Address PR 5201 feedback

ksubrmnn:

**Description**:
Addresses PR feedback for #5201 

Related to issue: #5198 

**Testing done**:
Re-ran certificate package unit tests

**Affected area**:

N/A
2. Is this a breaking change?
N/A
4. N/A

### DIFF
--- a/cmd/osm-bootstrap/osm-bootstrap_test.go
+++ b/cmd/osm-bootstrap/osm-bootstrap_test.go
@@ -140,7 +140,7 @@ var testPresetMeshRootCertificate = &corev1.ConfigMap{
 	},
 	Data: map[string]string{
 		presetMeshRootCertificateJSONKey: `{
-"intent": "Active",
+"intent": "active",
 "provider": {
 	"tresor": {
 	 "ca": {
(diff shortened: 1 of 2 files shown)


pkg/certificate/mrc_reconciler.go
```diff
@@ -164,13 +167,19 @@ func (m *Manager) getCertIssuer(mrc *v1alpha2.MeshRootCertificate) (*issuer, err
...
 	for _, mrc := range mrcList {
 		if mrc.Spec.Intent != v1alpha2.InactiveIntent {
 			mrcList[n] = mrc
 			n++
 		}
 	}
-	return mrcList[:n]
+
+	//Check if there more than 2 Active MRCs
+	if n > 2 {
```
jsturtevant, line 180: This check is already done later section. Is there a need to do it here?

cmd/osm-bootstrap/osm-bootstrap_test.go
```diff
@@ -12,6 +12,7 @@ import (
 	"k8s.io/client-go/kubernetes"
 	fakeKube "k8s.io/client-go/kubernetes/fake"
 
+	"github.com/openservicemesh/osm/pkg/apis/config/v1alpha2"
```
jaellio, line 15: The pkg is already imported with the alias configv1alpha2.

pkg/certificate/mrc_reconciler.go
ksubrmnn: Sean suggested to add it in the last PR review. I can leave it as is or take it out. It does make some sense if we wanted to re-use this function and be safe.

jsturtevant: Reading Sean's comment, https://github.com/openservicemesh/osm/pull/5201/files#r1015697995, I don't think this code is protecting against 2 active MRCs and instead this is simply removing ones we don't care about.  Maybe @jaellio can comment?

Personally, I would lean towards avoiding it since this is an internal function and not part of an external api, my concern would be having the logic in two places from a maintenance perspective. This function will likely change or be removed in  https://github.com/openservicemesh/osm/issues/5226

jaellio: Yep, I agree. I don't think the check is necessary here.

keithmattix (approved):

jaellio (approved):

jsturtevant:
looks like kafka flake: https://github.com/openservicemesh/osm/issues/5139

PR #5250, opened 2022-11-11
Allow all headless services, not just those backed by Statefulsets with subdomains

keithmattix:
The existing logic requires `hostname` to be set before a `MeshService` is created from a headless service's endpoint. However, `hostname` is only populated on the endpoint when that endpoint is generated from a StatefulSet. ~This PR creates a MeshService for every service so that it's still resolvable via DNS (instead of kube-proxy like ClusterIP services).~

Creating a `MeshService` for every headless service creates an issue if there is indeed a subdomain statefulset behind the headless service. Therefore, we only created a separate MeshService if the headless service it came from doesn't have a hostName in at least one of its endpoints


steeling (approved):

jsturtevant (approved):
/lgtm

PR #5300, opened 2023-03-27
Add more robust CRD conversion patching

keithmattix:
Change the reoncile label to refer to the version of OSM that is in charge of reconciling the CRDs. This ensures that on upgrade, the new version is the only version managing the CRDs

**Description**:
Fixes the upgrade bug from pre-v1.x.3 patches

**Testing done**:
E2E tests

**Affected area**:
| Install                    | [x] |
| Upgrade                    | [x] |

### DIFF
--- a/cmd/osm-bootstrap/osm-bootstrap.go
+++ b/cmd/osm-bootstrap/osm-bootstrap.go
@@ -12,7 +12,6 @@ import (
 	"net/http"
 	"os"
 	"path/filepath"
-	"strconv"
 
 	"github.com/spf13/pflag"
 	admissionv1 "k8s.io/api/admissionregistration/v1"
@@ -243,7 +242,7 @@ func applyOrUpdateCRDs(crdClient *apiclient.ApiextensionsV1Client) {
 			log.Fatal().Err(err).Msgf("Error decoding CRD file %s", file)
 		}
 
-		crd.Labels[constants.ReconcileLabel] = strconv.FormatBool(enableReconciler)
+		crd.Labels[constants.ReconcileLabel] = osmVersion
 
 		crdExisting, err := crdClient.CustomResourceDefinitions().Get(context.Background(), crd.Name, metav1.GetOptions{})
 		if err != nil && !apierrors.IsNotFound(err) {
@@ -262,7 +261,7 @@ func applyOrUpdateCRDs(crdClient *apiclient.ApiextensionsV1Client) {
 		} else {
 			log.Info().Msgf("Patching conversion webhook configuration for crd: %s, setting to \"None\"", crd.Name)
 
-			crdExisting.Labels[constants.ReconcileLabel] = strconv.FormatBool(enableReconciler)
+			crdExisting.Labels[constants.ReconcileLabel] = osmVersion
 			crdExisting.Spec = crd.Spec
 			crdExisting.Spec.Conversion = &apiv1.CustomResourceConversion{
 				Strategy: apiv1.NoneConverter,
--- a/tests/e2e/e2e_reconciler_test.go
+++ b/tests/e2e/e2e_reconciler_test.go
@@ -2,13 +2,24 @@ package e2e
 
 import (
 	"context"
+	"os"
+	"path/filepath"
+	"strconv"
 	"time"
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
-
+	apiv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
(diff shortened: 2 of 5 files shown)


steeling:
Can we also set rollout strategy on the bootstrap deployment to replace?

keithmattix:
> Can we also set rollout strategy on the bootstrap deployment to replace?

I'm not a huge fan of that change honestly. Replace effectively guarantees downtime in the gap between the old replicaset going down and the new one becoming Ready

steeling:
> > Can we also set rollout strategy on the bootstrap deployment to replace?
> 
> I'm not a huge fan of that change honestly. Replace effectively guarantees downtime in the gap between the old replicaset going down and the new one becoming Ready

The *only* thing the bootstrap pod does after it starts up (after it creates/updates CRD's on startup), is the reconciler. The bootstrap pod should really be a job anyways. IMO. Not blocking on this though

Technically it has a webhook for:

1. metrics
2. version reporting
3. health checks

cmd/osm-bootstrap/osm-bootstrap.go
```diff
@@ -260,19 +275,74 @@ func applyOrUpdateCRDs(crdClient *apiclient.ApiextensionsV1Client) {
...
+			patchCRDConversion(originalCRD, clusterCRD, crdClient)
+		} else {
+			log.Info().Msgf("Attempt #%d for CRD %s has succeeded!", attempts, clusterCRD.Name)
+
+		}
+	}
+	return success
+}
+
+func patchCRDConversion(originalCRD, clusterCRD *apiv1.CustomResourceDefinition, crdClient *apiclient.ApiextensionsV1Client) {
```
steeling, line 332: I may be missing something but I'm not sure I see any mechanism that prevents the old bootstrap pod from reconciling a patched CRD

tests/e2e/e2e_reconciler_test.go
```diff
@@ -116,7 +127,100 @@ var _ = OSMDescribe("Test OSM Reconciler",
...
+					}
+					Expect(err).NotTo(HaveOccurred())
+				}
+				// Add existing reconciler to watch the cluster
+				_, cancel := context.WithCancel(context.Background())
+				defer cancel()
+				stop := make(chan struct{})
+				Expect(err).NotTo(HaveOccurred())
+
+				err = reconciler.NewReconcilerClient(Td.Client, Td.APIServerClient, "osm", "test", stop, reconciler.CrdInformerKey)
```
steeling, line 189: You could have the reconciler client remove the webhook conversion as well

steeling:
> @steeling it's those webhooks I'm worried about, especially during upgrades. Any client or operator of an OSM resource that's using an older CRD version will end up triggering a conversion request to OSM bootstrap which, for a time, wouldn't have any replicas able to serve traffic

I'm not sure a couple seconds of downtime on a small subset of operations is necessarily a bad thing during an upgrade, given the issues we've had here. Adding further complexity to the code, and banking on allowing an old reconciler to flip/flop back to prior states is undesirable IMO

cmd/osm-bootstrap/osm-bootstrap.go
keithmattix: The idea behind this PR was that ~nothing could prevent the old bootstrap pod from reconciling since we're on an upgrade path. I'm just noticing that the reconciler informer has a label selector in the lister, _not_ the event handler, so this fix becomes much simpler. Rewriting

keithmattix:
> I'm not sure a couple seconds of downtime on a small subset of operations is necessarily a bad thing during an upgrade, given the issues we've had here. Adding further complexity to the code, and banking on allowing an old reconciler to flip/flop back to prior states is undesirable IMO

If anyone builds on top of OSM, we've now introduced errors/downtime in that upgrade path. Downtime during upgrade can be ok if all parties are aware and plan for it. To me, making that kind of change in a patch release is unexpected and error prone

jaellio:
Hey @keithmattix, could you update the PR description to describe the osm version label approach? Thanks!

keithmattix:
Good catch @jaellio! Changed it

jaellio (approved):
LGTM

cmd/osm-bootstrap/osm-bootstrap.go
```diff
@@ -243,7 +242,7 @@ func applyOrUpdateCRDs(crdClient *apiclient.ApiextensionsV1Client) {
 			log.Fatal().Err(err).Msgf("Error decoding CRD file %s", file)
 		}
 
-		crd.Labels[constants.ReconcileLabel] = strconv.FormatBool(enableReconciler)
+		crd.Labels[constants.ReconcileLabel] = osmVersion
```
steeling, line 245: maybe conditionally add the label based on enableReconciler

steeling (approved):

cmd/osm-bootstrap/osm-bootstrap.go
keithmattix: Makes sense in general, but I'm wary of changing more behavior than I need to. The label wouldn't actually be used until the reconciler is turned on, so I think it's safe to keep this logic as is

PR #5306, opened 2023-04-03
fix(): resolve lint error from unchecked AddEventHandler return values

jaellio:
**Description**:
Fixes lint error introduced in #5283

When upgrading helm to 3.11.1, the k8.io client-go package was also updated to v0.26.3. In v0.26 the API for [`AddEventHandler`](https://pkg.go.dev/k8s.io/client-go@v0.26.3/tools/cache#SharedInformer) was updated to return an error and registration handle to enable individual handler cancelation. 

### DIFF
--- a/pkg/certificate/providers/mrc.go
+++ b/pkg/certificate/providers/mrc.go
@@ -24,7 +24,7 @@ type MRCComposer struct {
 // to be ordered for any particular resources, but NOT across different resources.
 func (m *MRCComposer) Watch(ctx context.Context) (<-chan certificate.MRCEvent, error) {
 	eventChan := make(chan certificate.MRCEvent)
-	m.AddMeshRootCertificateEventHandler(cache.ResourceEventHandlerFuncs{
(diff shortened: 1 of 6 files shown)


keithmattix (approved):

steeling:
i'm surprised we don't have the linter enabled to ensure errors are checked. We should check and do something with those errors most likely

keithmattix:
In general yes, but for a backport, we should probably persist existing behavior right?

steeling (approved):

steeling:
> In general yes, but for a backport, we should probably persist existing behavior right?

I'll just point out this is more than a backport if its making it into main

keithmattix:
> > In general yes, but for a backport, we should probably persist existing behavior right?
> 
> 
> 
> I'll just point out this is more than a backport if its making it into main

Oh I misread which PR this is. That's a fair point...in that case, I don't have a strong preference for which direction we go this PR

